PR #161, opened 2014-11-29
Fix duplicate file name extension

nattress:
If a filename extension and a format are specified with the same string,
we create a file with a double file extension. Ie:

``` javascript
py.image.save_as(figure, 'filename2.png', 'png')
```

This will produce a file named filename2.png.png. When ext and format
are both present, the fix will only append format to the filename if the
format is different from ext. Ie:

``` javascript
py.image.save_as(figure, 'filename2.png', 'pdf')
```

This will create filename2.png.pdf.

### DIFF
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -660,7 +660,7 @@ class image:
             format = ext[1:]
         elif not ext and format:
             filename += '.'+format
-        else:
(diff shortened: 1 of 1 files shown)


theengineear:
Apologies that this never got reviewed. It's pretty old at this point and I'm going to close down. If you want to rebase (or merge) master and reopen, I'll take another look.

PR #170, opened 2014-12-22
Revamp tls.get_subplots

etpinard:
@theengineear @chriddyp 

`tls.get_subplots` is in need of some revamping.

More info in this Asana [task](https://app.asana.com/0/11724913287040/12301257260418).

This [NB](http://nbviewer.ipython.org/github/etpinard/plotly-misc-nbs/blob/master/get-subplots/get-subplots.ipynb) shows off the new features.

Features wishlist:
- [x] Generate more complicated subplot arrangements e.g https://plot.ly/~etpinard/245
- [x] Assign shared axes
- [x] Set irregular horizontal and vertical spacing in-between subplots
- [x] Generate subscenes - subaxes subplot grid.
- [x] Generate insets with ease

API issue to resolve:

- [x] `rowspan` > 1 and `is_empty` https://github.com/plotly/python-api/pull/170#discussion_r22334996

**Solution**: Enforce specs to have dimensions (rows, columns). Use `None` to fill in cells over row / column spans.

- [x] Ordering of the `specs` keyword argument https://github.com/plotly/python-api/pull/170#issuecomment-68561185


theengineear:
@etpinard , this is awesome! I definitely have some comments and I want to tinker around in IPython before I let it dance. (Nice job with the tests, btw!)

etpinard:
@theengineear Thanks! Looking forward to your comments.

plotly/tools.py
```diff
@@ -398,60 +399,327 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
...
+          cells.
+
+        - Each item in the 'specs' is a dictionary.
+            The available keys are:
+
+            * isEmpty (boolean, default=False): flag for empty grid cells
+            * is3D (boolean, default=False): flag for 3d scenes
+            * colspan (int, default=1): span across grid columns
+                                        from left to right
+            * rowspan (int, default=1): span across grid rows
```
theengineear: `row_span` instead of `rowspan` ? (i see this is an HTML thing... but I think it doesn't look Pythonic)

plotly/tools.py
```diff
@@ -398,60 +399,327 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
...
+
+        - Each item in the 'specs' list corresponds to one subplot
+          in a subplot grid. The subplot grid has 'rows' times 'columns'
+          cells.
+
+        - Each item in the 'specs' is a dictionary.
+            The available keys are:
+
+            * isEmpty (boolean, default=False): flag for empty grid cells
+            * is3D (boolean, default=False): flag for 3d scenes
```
theengineear: I don't mind the `row_span` as `rowspan` so much, i think it's used commonly maybe. However let's go with `is_empty` and `is_3d` or something? What do you think?

etpinard: Good call about `is_empty` and `is_3d`  (I've been written a lot of js lately, sorry)

Matplotlib's gridspec uses `rowspan` and `colspan`, so I would vote for keeping them that way.

theengineear: yup, sounds good!

theengineear:
Some initial feedback about using `specs`. When making a spec, you design the figure bottom-to-top and left-to-right. However, in the code, it's top-to-bottom and left-to-right. Can we reverse the order of the spec list in the get_suplots function to fix this?

Example:

```
# spec to be used
specs = [[{}, {'colspan': 2}],
         [{}, {}, {}],
         [{}, {}, {}]]

# resulting grid
This is the format of your plot grid!
[x6,y6]  [x7,y7]  [x8,y8]
[x3,y3]  [x4,y4]  [x5,y5]
[x1,y1]  [x2,y2         ]
```

I *don't* think it would be good to flip it the other way since plotly uses left-to-right and bottom-to-top to organize axis spans.

etpinard:
Good point. Not sure what's best. Either way, it wouldn't be too hard to change to order.

My intuition was the following:

plotly works left-right / bottom-to-top, so the first element in the specs argument should correspond to the bottom-left cell.

theengineear:
Ooh, another thing. You had a little `bind` function in that NB. That's clever. We should figure out a function to go in tools or a method on a figure, or on a trace that will make this simpler. I'm hesitant to add *too* much stuff onto the trace PlotyDict objects, but we might be able to do some fancy (non-confusing) things to the Figure objects that might make this process a little easier.

theengineear:
Re: order. I guess there are two lines of thought. You make the spec visually, or you make it in order. And i agree, I'm not sure what's best, but so far, I've wished it was more visually right for what I'm doing.

etpinard:
@theengineear I'd vote for something like `tools.bind_trace_to_axis` or `tools.bind_data_to_axis`. 

I agree. Lightweight PlotlyDict objects are the way to go.

theengineear:
@etpinard , First, I think trying to have the way you write the spec mirror the printout is short-sighted (that's my bad!). In fact, that got me thinking that the current `specs` isn't flexible enough! Here's wayy too much thought I put into it:

tl;dr:
could the spec be a `dict

```python
"""
The spec is a dictionary of subplot dictionaries.

Each dictionary's key can be used to reference the dictionary from
other dictionaries (used for sharing axes).

    The available keys are:

    * cols (int|sequence): left-most and right-most occupied cols (or left-most if int)
    * rows (int|sequence): bottom-most and top-most occupied rows (or bottom-most if int)
    * share_y (str, default=None): reference to plot y value is shared between
    * share_x (str, default=None): reference to plot x value is shared between
    * layer (int, default=1): consider overlapping plots to have ordered layering.
                              higher layers appear as overlaying (ontop)
    * l (float, default=0.0): padding left of cell
    * r (float, default=0.0): padding right of cell
    * t (float, default=0.0): padding right of cell
    * b (float, default=0.0): padding bottom of cell
    * is_3d (boolean, default=False): flag for 3d scenes

"""

# simple (no overlap, all assume layer 1)
spec = {
    'a': {'rows': [1, 1], 'cols': [1, 3]},
    'b': {'rows': [2, 3], 'cols': [1, 1]},
    'c': {'rows': [2, 3], 'cols': [2, 3]}
}
"""
This is the format of your plot grid!
  
Layer 1:

   ********************************************
   *[            ][                          ]*
   *[            ][                          ]*
   *[b -> x2, y2 ][c -> x3, y3               ]*
   *[a -> x1, y1                             ]*
   ********************************************
"""

# complex (overlap, overlapping object must resolve!)
spec = {
    'a': {'rows': [1, 1], 'cols': [1, 3]},
    'b': {'rows': [2, 3], 'cols': [1, 1]},
    'c': {'rows': [2, 3], 'cols': [2, 3]},
    'd': {'rows': [3, 3], 'cols': [3, 3], 'layer': 2}  # error if unspecified
}

# out
"""
This is the format of your plot grid!
    
Layer 1:

   ********************************************
   *[            ][                          ]*
   *[            ][                          ]*
   *[b -> x2, y2 ][c -> x3, y3               ]*
   *[a -> x1, y1                             ]*
   ********************************************

Layer 2:

   ********************************************
   *                             [d -> x4, y4]*
   *                                          *
   *                                          *
   *                                          *
   ********************************************
                             
"""

# complex shared axes in the spec
spec = {
    'a': {'rows': [1, 1], 'cols': [1, 3]},
    'b': {'rows': [2, 3], 'cols': [1, 1]},
    'c': {'rows': [2, 3], 'cols': [2, 3], 'share_y': 'b'}
}
"""
This is the format of your plot grid!
  
Layer 1:

   ********************************************
   *[            ][                          ]*
   *[            ][                          ]*
   *[b -> x2, y2 ][c -> x3, y2               ]*
   *[a -> x1, y1                             ]*
   ********************************************
"""
```

theengineear:
I'm also wondering *whether* we should entertain inset plots here at all! Either way, I'm still wondering if a different structure might be useful for the `specs`.

etpinard:
Oh man. We could debate about for days. Just like I did in my head before pushing. 

A few remarks:

Apart from being able to stack layers (e.g. to make insets) and having a more declarative syntax --- but longer to type --- I don't see how your API is more flexible (prove me wrong my man!)

I like  having shared_x/yaxes outside of specs. Shared axis plot can be easily made with the default layout (i.e. without setting specs) that way and I wouldn't like having shared_xaxes as an argument and a key in specs (it would be a little too much to throw at the users in my mind).

Mapping a spec {} to a subplot cell using indices seems intutive to me and allows users to make a custom layout while setting only a few keys (i.e no rows cols keys to type in).

To me, get_subplots should yes be flexible, but, also easy and fast to use. If not, what's the point? Users can always set axis domains manually.

I'm almost done coding up the logic for an insets_specs argument. Separating it from specs seems logical: insets are layers on top of the specs and behave differently (e.g. insets will not have shared axes).

-- Just my two cents. Looking forward to your feedback.

theengineear:
Alright, after a bit more thought, I think I'm coming around to the way you've got it :)

Here are the major things that I was tripping up on for the current approach though. I *think* you'll dig the suggestions :smiley_cat:.

* ~~insets~~ solved
* ~~general plot overlay~~ i can't think of a worthwhile use case for this === who cares!
* ordering, this is what set me off in the beginning, i still think it's a bit unintuitive.
* underspecification, I don't know that we should allow underspecification at all.
* overspecification, I don't think we should let overspecification pass either.
* 'is_empty', can we use `None` instead of a dict with `'is_empty': True`?
* symmetry between colspan and rowspan wrt 'is_empty'

More:

**underspecification**:

If you've got a 2x2 subplot grid, i don't think this is OK:

```python
specs = [
    [{'colspan': 2}]
]
```

Rather:

specs = [
    [{'colspan': 2}],
    [{}, {}]
]

**overspecification**:

If you've got a 2x2 subplot grid, i don't think this is OK:

```python
specs = [
    [{'rowspan': 2}, {}],
    [{}, {}]
]
```

Rather:

```python
specs = [
    [{'rowspan': 2}, {}],
    [{'is_empty': True}, {}]
]
```

**ordering**:

maybe it's because i keep writing out the specs separately, but i think it's intuitive to do so

```python
specs = [
    [{'rowspan': 2}, {}],  # this is the bottom row
    [{'is_empty': True}, {}]  # this is the top row
]
```

Rather:

```python
specs = [
    [{'rowspan': 2}, {}],  # this is the top row
    [{'is_empty': True}, {}]  # this is the bottom row
]
```

**is_empty**:

Can we just use `None`?

```python
specs = [
    [{'is_empty': True}, {}]
    [{'rowspan': 2}, {}],
]
```

Rather:

```python
specs = [
    [None, {}]
    [{'rowspan': 2}, {}],
]
```

**col/rowspan symmetry**:

It's not totally obvious (at first) how you account for grids that overlay (i.e., we need 'is_empty' for rows and *no entry* for cols). So, this is two different methods of dealing with the same thing. For clarity (suffering a bit of verboseness), should we make this symmetric?

Imagine a 3x3 grid:

```python
specs = [
    [{'colspan': 2}, {}],
    [{'rowspan': 2}, {}, {}],
    [{'is_empty': True}, {'colspan': 2}]
]
```

Rather:

```python
specs = [
    [{'colspan': 2}, {'is_empty': True}, {}],
    [{'rowspan': 2}, {}, {}],
    [{'is_empty': True}, {'colspan': 2}, {'is_empty': True}]
]
```

#### An elaborate example

So, with the proposed changes, to make the following subplot:

https://plot.ly/~etpinard/723

The *only* way to do it is this (assuming a 3x3 grid):

```python
specs = [
    [{'colspan': 3}, None, None],
    [{'colspan': 2}, None, {'rowspan': 2}],
    [{}, {}, None]
]
```

Where columns expand to the *right* and rows expand *down* **necessarily** into `None` entries. No other options. What do you think?

#### Loose ends

Finally, this might seem verbose, but it's only when things get complex. i.e., you'd never do this:

```python
specs = [
    [{}, {}, {}],
    [{}, {}, {}],
    [{}, {}, {}]
]
```

Clearly, since that's a plain 3x3.

etpinard:
You guessed right. Those are very good points.

I agree with all of them, except for the point on ordering. I still think specs[0][0] should point to the bottom-left corner

etpinard:
Maybe we should somebody else's opinion?

etpinard:
@theengineear 

The Jan 4 commits solve all the issues you mentioned last Friday, except for the ordering one (see above comment).

I haven't updated the NB though, I'll wait for your feedback to do so.

chriddyp:
you guys are the best.

I think we should throw a `bind` method in this PR just to make sure everything feels super solid to us.

Some options


```
tls.get_subplots(rows=3, columns=3, specs = [
    [{'colspan': 3}, None, None],
    [{'colspan': 2}, None, {'rowspan': 2}],
    [{}, {}, None]
], print_grid=True)


This is the format of your plot grid!
[x4,y4 (3)]  [x5,y5 (4)]      ^   
[x2,y2 (1)        ]  [x3,y3 (2)]
[x1,y1 (0)                          ]

fig.insert_subplot_trace(Scatter(name='trace 0'), index=0)
fig.insert_subplot_trace(Scatter(name='trace 1'), index=1)
fig.insert_subplot_trace(Scatter(name='trace 2'), index=2)
fig.insert_subplot_trace(Scatter(name='trace 3'), index=3)
fig.insert_subplot_trace(Scatter(name='trace 4'), index=4)
```

we might need something like a `fig.print_arrangement` method that would print out the subplot grid and indices

Other arguments:
```
fig.insert_subplot_trace(Scatter(name='trace 0'), xaxis=1, yaxis=1)
fig.insert_subplot_trace(Scatter(name='trace 1'), xaxis=2, yaxis=2)
fig.insert_subplot_trace(Scatter(name='trace 2'), xaxis=3, yaxis=3)
fig.insert_subplot_trace(Scatter(name='trace 3'), xaxis=4, yaxis=4)
fig.insert_subplot_trace(Scatter(name='trace 4'), xaxis=5, yaxis=5)
```

plotly/tools.py
```diff
@@ -398,60 +400,464 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
         fig['data'] += [Scatter(x=[1,2,3], y=[2,1,2], xaxis='x2', yaxis='y2')]
 
     Example 2:
+        # subplots with shared x axes
+        fig = tools.get_subplots(rows=2, shared_xaxes=True)
```
chriddyp: What do you think about adding `print_grid=True` to each of these examples?

chriddyp:
I expected this to give me a 3 full columns:

`py.image.ishow(tls.get_subplots(rows=3, columns=3, shared_xaxes=True))`

image

chriddyp:
similarly here:
`py.image.ishow(tls.get_subplots(rows=3, columns=3, shared_yaxes=True, shared_xaxes=True))`
image

plotly/tools.py
etpinard: :+1:

chriddyp:
looks like you crushed this issue, nice! https://github.com/plotly/python-api/issues/152

plotly/tools.py
```diff
@@ -398,60 +400,464 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
...
+
+            * is_3d (boolean, default=False): flag for 3d scenes
+            * colspan (int, default=1): span across grid columns
+                                        from left to right
+            * rowspan (int, default=1): span across grid rows
+                                        from bottom to top
+            * l (float, default=0.0): padding left of cell
+            * r (float, default=0.0): padding right of cell
+            * t (float, default=0.0): padding right of cell
+            * b (float, default=0.0): padding bottom of cell
```
chriddyp: an example or two in here would be fantastic

plotly/tools.py
```diff
@@ -398,60 +400,464 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
...
+        - Note that specs[0][0] has the specs for the bottom-left subplot
+
+        - Each item in 'specs' is a dictionary.
+            The available keys are:
+
+            * is_3d (boolean, default=False): flag for 3d scenes
+            * colspan (int, default=1): span across grid columns
+                                        from left to right
+            * rowspan (int, default=1): span across grid rows
+                                        from bottom to top
```
chriddyp: I found this span language confusing, maybe something like:

> number of subplot columns for this subplot to span

plotly/tools.py
```diff
@@ -398,60 +400,464 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
...
 
-    columns (int, default=1):
-        Number of columns, evenly spaced horizontally on the figure.
+    vertical_spacing (kwarg, float in [0,1], default=0.3 / rows):
+        Space between subplot rows.
+        Applies to all rows (use 'specs' subplot-dependents spacing)
 
-    horizontal_spacing (float in [0,1], default=0.1):
-        Space between subplot columns. Applied to all columns.
+    specs (kwarg, list (of lists) of dictionaries):
```
chriddyp: this is always a `list of lists` right? maybe remove parentheses?

plotly/tools.py
```diff
@@ -398,60 +400,464 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
...
 
-    rows (int, default=1):
-        Number of rows, evenly spaced vertically on the figure.
+        Applies to all columns (use 'specs' subplot-dependents spacing)
 
-    columns (int, default=1):
-        Number of columns, evenly spaced horizontally on the figure.
+    vertical_spacing (kwarg, float in [0,1], default=0.3 / rows):
+        Space between subplot rows.
+        Applies to all rows (use 'specs' subplot-dependents spacing)
```
chriddyp: After reading the 'specs' doc, it wasn't clear to me how to do this

plotly/tools.py
etpinard: I accept a flat `list` if `rows == 1`.

etpinard: I am not opposed to enforcing `specs` to be a `list of lists`.

plotly/tools.py
etpinard: :+1:  We should mention `vertical_spacing` and `horizontal_spacing` in the `specs` doc string.

plotly/tools.py
etpinard: :+1:

plotly/tools.py
etpinard: :+1:

chriddyp:
A nice outcome of the `specs` argument is that you can create non-grid geometries with consistent demnsions pretty easily. A couple examples:

```
tls.get_subplots(rows=6, columns=1, shared_xaxes=True, shared_yaxes=True, print_grid=True, specs=
                 [[{}],
                  [{}],
                  [{}],
                  [{'rowspan': 3}],
                  [None],
                  [None]])
```

image

```
py.image.ishow(tls.get_subplots(rows=4, columns=4, shared_xaxes=True, shared_yaxes=True, print_grid=True, specs=
                 [[{'colspan': 3, 'rowspan': 3}, None, None, {'rowspan': 3}],
                  [None,                         None, None, None],
                  [None,                         None, None, None],
                  [{'colspan': 3},               None, None, {}]]))
```
image

chriddyp:
What do you guys think about an `update` or `style` keyword argument?

It's pretty tedious to loop through all of the axis keys and frequently all of the axes have the same style (usually no gridlines). 
Some examples:








```
help(tls.get_subplots)
...
update (dict or function returning a dict): Additional keys to pass into the constructed axis objects.
```

```
tls.get_subplots(rows=2, update={'showgrid': False, 'zeroline': False, 'showline': False})
```

```
def small_multiple_styles(axis, index):
    axis_style = {}
    if isinstance(axis, XAxis):
        axis_style['showgrid'] = False
    else:
        pass

    if axis['anchor'] == 'x3':
        axis_style['gridcolor'] = 'blue'

    return axis_style

tls.get_subplots(rows=2, update=small_multiple_style)
```

Alternatively, we could have a method that mirrors the indexing arguments of whatever we decide on for `bind`, e.g. if we have

`fig.insert_subplot_trace(Scatter(name='trace 0'), xaxis=1, yaxis=1)`

we could have

`fig.update_axis({'showgrid': False}, xaxis=1, yaxis=1)`

and

`fig.insert_subplot_trace(Scatter(name='trace 0'), index=3)`
`fig.update_axis({'showgrid': False}, index=3)`

(not sure how you would update all of the axes with the `update_axis` method)

etpinard:
@chriddyp keep the suggestions coming!

- I'd vote for calling your `update` argument `style` instead as `tls.get_subplots` creates a new `Figure` object.

- Instead of assigning numbers to `index` in `insert_subplot_trace` and `update_axis`, I'd vote for assigning (row, column) tuples.

chriddyp:
Concerning creating other types of geometries, I wonder if we should add a `width` parameter to `spec` which would compute `domain` with the appropriate spacing. For example:



```
tls.get_subplots(rows=2, columns=2, shared_xaxis=True, shared_yaxis=True,
                 specs=[
                    [{}, {}],
                    [{'width': 0.75, 'height': 0.75}, {}]
                 ])
```

In this case, the `domain`s would be something like `[0, 0.73]`, `[0.77, 1]` and specifying `{'width': 0.75, 'horizontal_spacing': 0}` would set the `domain`s to `[0, 0.75]`, `[0.75, 1]`.

chriddyp:
FWIW, I think that the print-out should match the spec arrangement - it's pretty cool for the spec to look just like your plot

```
# spec to be used
specs = [[{}, {}, {}],
               [{}, {}, {}],
               [{}, {'colspan': 2}, None]]

# resulting grid
This is the format of your plot grid!
[x6,y6]  [x7,y7]  [x8,y8]
[x3,y3]  [x4,y4]  [x5,y5]
[x1,y1]  [x2,y2         ]
```

chriddyp:
Consider this case:

```
spec = [[{}],
        [{'rowspan': 2}],
        [None]]

spec = [[{}],
        [None],
        [{'rowspan': 2}]]
```

Do we allow the latter spec?

plotly/tools.py
```diff
@@ -398,60 +400,464 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
...
+            * t (float, default=0.0): padding right of cell
+            * b (float, default=0.0): padding bottom of cell
+
+    insets (kwarg, list of dictionaries):
+        Inset specifications.
+
+        - Each item in 'insets' is a dictionary.
+            The available keys are:
+
+            * cell (tuple, default=(0,0)) subplot cell indices
```
chriddyp: I found this confusing. Maybe:

> (row, column) index of the subplot cell to overlay inset axes onto

chriddyp:
the inset spec is great!

plotly/tools.py
etpinard: :+1:

etpinard: My engrish isn't the best. Thanks for your suggestions.

etpinard:
Re `width` / `height` https://github.com/plotly/python-api/pull/170#issuecomment-68748389

- Good idea (matplotlib's gridspec allows it). But it would require some work to code it up. It also somewhat breaks my initial philosophy: the width and height of each subplot depend only on the number of row and columns. Col and row span > 1 are nothing more than to subplot merged together.

- with `{'width': , 'height': }` the width and height of all subplots would have to be recomputed in order to preserve `horizontal_spacing` and `vertical_spacing`.

Re ordering https://github.com/plotly/python-api/pull/170#issuecomment-68749866

- You're on @theengineear's side on this one. 

- To me, it's a little weird that the first spec item (i.e. `specs[0][0]`) does not correspond to `xaxis1` / `yaxis1`. 

- What about this solution https://github.com/plotly/python-api/pull/170#commitcomment-9148029 ?

Re `None` layout https://github.com/plotly/python-api/pull/170#issuecomment-68750001

- With `rows == 4`, we allow
```
spec = [[{}],
        [None],
        [{'rowspan': 2}],
        [None]]
```

- With `rows == 3`, we allow 

```
spec = [[{}],
        [{'rowspan': 2}],
        [None]]
```

- Should we make `None` specs opitonal at the end of rows and columns ?

theengineear:
whoa, @chriddyp , demolishing this thread.

Some thoughts on your thoughts:

* `update` key in get_subplots
This seems like it's adding too much behavior to `get_subplots`. Can we just add an 'axes' property to `Figure` so that users can do:

```python
for axis in Figure.axes:
  axis['showgrid'] = False
```

* I think order should matter:

```python
spec = [[{}],
        [{'rowspan': 2}],
        [None]]

spec = [[{}],
        [None],
        [{'rowspan': 2}]]
```

One of these should throw an exception that mentions the direction we assume rowspan and colspan to work in. I think a good exception will make this a simple fix for users.

* still in agreement on spec ordering matching the plot layout. i.e., row order not being the reverse of what's input in specs. However, again, a good exception will help in either case :)

* `width` and `height` specs. Not sure how i feel on this:
  * it adds more complexity to the spec dicts
  * it offers two ways of doing the same (almost the same) thing (you *could* just make the initial grid larger, i.e. more rows/cols)

I'd vote to hold off on this for now and open an issue related to it.

* optional `None` in the spec. I don't see how making this optional would greatly help users, so, I'd vote to keep these required (we can always change it in a non-breaking way)

* `bind` type method
Plotly doesn't really have a concept of subplots, right? Just axes. So I think we won't be able to offer a call signature like `(index=0)`, but will need to go with `(xaxis=1, yaxis=1)`. To do the former, we'd need a function that goes through the axes of a figure and computes the index of each subplot. Maybe i'm being naive, but this seems like a bit of a tough problem.

I think a good solution is to offer (1) the `print_arrangement` suggested--without indices--and (2) a bind method, which simply adds arguments to the dict and can be leveraged by the suggested `insert_*` Figure method. I use the `*` there to reaffirm that inserting at subplots might be a bit tricky.

etpinard:
@theengineear @chriddyp 

What if we allow negative `rowspan` and `colspan` ?

etpinard:
@theengineear [Re going from index to `x{}` and `y{}`. That function is already [coded](https://github.com/plotly/python-api/pull/170/files#diff-2431a3f5483919319eff181a09618fc7R632), so I wouldn't be too hard.

theengineear:
re: negative rowspan. Seems like another way to do the same thing. I'm not necessarily opposed to it though. So you'd have something like:

`[None, None, {'colspan': -3}]`

or 

`[{'colspan': 3}, None, None}]`

Not sure if you guys feel the same way. But `{'colspan': -3}` reads `this column and three back` while `{'colspan': 3}` reads `three total columns starting here`. Maybe that's just me though. Again, if you think this is helpful, not opposed as long as it's symmetric as used above.

etpinard:
@theengineear [Re negative colspan]. Good call. We shouldn't add any *useless* complexity just for the sake of it.

theengineear:
@etpinard , yeah, if the index stuff is already coded, that's definitely a nice user experience.

* print grid for figure
* insert trace at desired subplot index from printed grid

etpinard:
@theengineear @chriddyp 

Let me throw a few more ideas about the grid ordering (before I give up):

- Ideally (your favorite), the `specs` list of lists would line up visually with the printed grid.

- Ideally (my fav), `specs[0][0]` would correspond to the axes `x1` and `y1`.

Both conditions would be fulfilled if `x1` and `y1` is placed at the top-left corner of the grid.

**Problem**: The old `tls.get_subplots()` has `x1` and `y1` on the bottom-left corner. The above solution would break backward compatibility.

So, I can think of two solutions:

1. Have `x1`/`y1` on the bottom-left (top-left) corner if `specs` is undefined (set).

2. Add a `start_cell` argument. Set `start_cell='top-left'` or `start_cell='bottom-left'` depending on your tastes.

etpinard:
Also, do we agree on `index` being a (row, cell) tuple ?

chriddyp:
This is pretty fun, this is going to be great

#### re indexing and coordinates

@etpinard - i'm not too worried about backwards incompatibility on this one - I'm OK with putting in that optional argument to preserve compatibility for folks (we should maybe put this inside the docstring - the main reason for that argument is for backwards compatibility)

All good with this?



```
tls.get_subplots(rows=3, columns=3, specs = [
    [{}, {}, {'colspan': 2}],
    [{'rowspan': 2}, None, None],
    [{'rowspan': 3}, None, None]
], print_grid=True)

This is the format of your plot grid
[x1,y1 (1, 1)]  [x2,y2 (1, 2)]  [x3, y3 (1, 3)]   
[x4,y4 (2, 1)                ]  [             ]
[x5,y5 (3, 1)                                 ]

```

```
fig.append_trace_to_axes(Scatter(), index=(1, 1)) # top left
# in this case, equivalent to:
fig['data'].append(Scatter(xaxis='x1', yaxis='y1'))

# this raises an exception - right?
fig.append_trace_to_axes(Scatter(), index=(2, 2))
```

Note that (row, column) indexing allows the user to set `shared_axes=True` and not have to re-order their graph:

```
tls.get_subplots(rows=3, columns=3, specs = [
    [{}, {}, {'colspan': 2}],
    [{'rowspan': 2}, None, None],
    [{'rowspan': 3}, None, None]
], shared_yaxis=True,
   print_grid=True)

This is the format of your plot grid
[x1,y1 (1, 1)]  [x2,y1 (1, 2)]  [x3, y1 (1, 3)]   
[x4,y2 (2, 1)                ]  [             ]
[x5,y5 (3, 1)                                 ]
```

#### Order
Into it - this raises an exception:
```
spec = [[{}],
        [None],
        [{'rowspan': 2}]]
```

#### `axes` property in `Figure` or `Layout`
into it, great idea.


#### `width` and `height`
alright, let's hold off on this

#### new method: `fig.print_arrangement`


#### display axes grid as an HTML table in ipython notebooks

that'd be kinda sweet, right? just a random idea

theengineear:
@chriddyp and @etpinard , i think we've got enough users that we shouldn't break backwards compatibility here. What about just making a new function and throwing a deprecation warning on the old function?

new:

`make_subplots`

OR JUST

`subplots`

deprecation warning on:

`get_subplots`

Also, beyond the backwards compatibility argument, it's also the opposite of the *natural plotly ordering*. The `[0,1]` domains for axes go left-to-right, and bottom-to-top. However, I think that we're probably okay with going against the grain on this one.

Also, also, I don't think we should include a `start_cell` argument.

Also, also, also, an html grid is a good call, but can happen whenever. Along these lines though, should we default to printing the grid for folks? New users who will most need it would have to search for that functionality.

Finally, if we're using row/col, there's no extra typing saved by using a new index argument. Why not just use row/col?

`fig.append_trace_to_axes(Scatter(), index=(1, 1))`
`fig.append_trace_to_axes(Scatter(), row=1, col=1)`

etpinard:
Starting from the top-left may not be *natural*, but it is what matplotlib [does](http://matplotlib.org/users/gridspec.html).

Re `row` / `col` instead of `index`. I like it. My only concern is that `get_subplots` has `columns` as argument not `cols`.

theengineear:
To clarify my negativity on the `start_cell`:

I'm assuming with that statement what we *would* use a new function and throw a deprecation warning on the old. With this in mind, I don't think the additional complexity is required.

However, if we *do* use the same function, I'd still be in favor of maintaining backwards compatibility. So I'd vote for it being defaulted to the behavior `get_subplots` has now. If this is the case, then **yes** this solution seems reasonable.

Last thought, could we fire a deprecation warning by *default*, would that be annoying?

```
Looks like your code might be using the old ordering for get_subplots. To make sure your code won't break in the future, you should use "start_cell='top-left'" in your function call. This will make the first subplot in the top-left of the figure as opposed to making the first subplot in the bottom-left of the figure.
```

theengineear:
@etpinard , re: row/col vs index. yeah, you're totally right, forgot that it's `rows`/`columns`. Up to you, i might still be in favor of `row` `column` though :). We should try to be consistent though.

So:

`fig.append_trace_to_axes(Scatter(), row=1, column=1)`

OR

`fig.append_trace_to_axes(Scatter(), 1, 1)`

etpinard:
I'm voting for adding `start_cell` in `get_subplots` is this case with

- `start_cell='bottom-left'` by default.
-  fire a deprecation warning if the default is used.

Also, now that we're talking about backward compatibility, is [this](http://nbviewer.ipython.org/github/etpinard/plotly-misc-nbs/blob/master/get-subplots/get-subplots.ipynb#Backward-compatibility-break) considered a bug fix, or a backward compatibility break (that we should remove).

theengineear:
@etpinard , sounds like a plan!

And, I think that's a bug fix, not a compatibility break. It also reduces the complexity, so it's a great change, imo.

theengineear:
Thanks for being so thorough *and* patient :)

etpinard:
No problem.  Thanks for your feedback !

chriddyp:
I vote:

- New function: `make_subplots`, remove our examples of `get_subplots`, [issue depreciation warning](https://docs.python.org/2/library/exceptions.html#exceptions.DeprecationWarning) on use of `get_subplots`, include a note about the new function at the top of the `get_subplot` docstring. Keep `get_subplots` the same as the old `get_subplots` (except with the better spacing). 
  
  > Warning: `get_subplots` is depreciated. Use `make_subplots` instead.
- Default of `make_subplots` is top-left
- `row`, `column` argument as the first 2 keyword arguments
- `fig.append_trace` method (instead of `fig.append_trace_to_axes`, `tls.gind_trace_to_axis`, `fig.insert_subplot_trace`

etpinard:
@chriddyp I love it.

Minor detail: I'd vote for `make_subplots(rows= , cols=)` both 4-letter words. Leave the singular (row, col) for pointing to specific cell in the subplot grid.

chriddyp:
@etpinard - :+1:

theengineear:
@etpinard , @chriddyp , double :+1:

plotly/tools.py
```diff
@@ -386,23 +386,20 @@ def mpl_to_plotly(fig, resize=False, strip_style=False, verbose=False):
 
 ### graph_objs related tools ###
 
-# TODO: Scale spacing based on number of plots and figure size
-def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
-                 vertical_spacing=0.15, print_grid=False):
+def get_subplots(rows=1, columns=1, print_grid=False, **kwargs):
```
etpinard: good old `get_subplots` is back !

plotly/tools.py
```diff
@@ -417,9 +414,46 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
...
+    warnings.warn(
+        "tools.get_subplots is depreciated. "
+        "Please use tools.make_subplots instead."
+    )
+
+    # Throw exception for non-integer rows and columns
+    if not isinstance(rows, int):
+        raise Exception("Keyword argument 'rows' must be an int")
+    if not isinstance(columns, int):
+        raise Exception("Keyword argument 'columns' must be an int")
```
etpinard: with some added features

plotly/tools.py
```diff
@@ -417,9 +414,46 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
...
+
+    # Set 'horizontal_spacing' / 'vertical_spacing' w.r.t. rows / columns
+    try:
+        horizontal_spacing = float(kwargs['horizontal_spacing'])
+    except KeyError:
+        horizontal_spacing = 0.2 / columns
+    try:
+        vertical_spacing = float(kwargs['vertical_spacing'])
+    except KeyError:
+        vertical_spacing = 0.3 / rows
```
etpinard: ... horizontal and vertical spacing that depend on 'columns' and 'rows'

plotly/tools.py
```diff
@@ -452,9 +487,599 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
...
+    # Sanitize 'specs' (must be a list of lists)
+    exception_msg = "Keyword argument 'specs' must be a list of lists"
+    try:
+        specs = kwargs['specs']
+        if not isinstance(specs, list):
+            raise Exception(exception_msg)
+        else:
+            for spec_row in specs:
+                if not isinstance(spec_row, list):
+                    raise Exception(exception_msg)
```
etpinard: 'specs' must now be a list of lists

plotly/tools.py
```diff
@@ -452,9 +487,599 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
...
+    START_CELL_all = {
+        'bottom-left': {
+            # 'natural' setup where x & y domains increase monotonically
+            'col_dir': 1,
+            'row_dir': 1
+        },
+        'top-left': {
+            # 'default' setup visually matching the 'specs' list of lists
+            'col_dir': 1,
+            'row_dir': -1
```
etpinard: this is where the magic happens.

plotly/tools.py
```diff
@@ -452,9 +487,599 @@ def get_subplots(rows=1, columns=1, horizontal_spacing=0.1,
...
+
+            c_spanned = c + spec['colspan'] - 1  # get spanned c
+            r_spanned = r + spec['rowspan'] - 1  # get spanned r
+
+            # Throw exception if 'colspan' | 'rowspan' is too large for grid
+            if c_spanned >= cols:
+                raise Exception("Some 'colspan' value is too large for "
+                                "this subplot grid.")
+            if r_spanned >= rows:
+                raise Exception("Some 'rowspan' value is too large for "
```
etpinard: new exceptions for situations like:

```
spec = [[{}],
        [None],
        [{'rowspan': 2}]]
```

etpinard:
@theengineear @chriddyp 

The logic part of `make_subplots` is done (and tested).

**Important questions**

- [ ] 1. If `shared_xaxes=True` (and `start_cell='top-left'` our new default), should the shared xaxes be on the top or bottom row ? For more, see the now updated [NB](http://nbviewer.ipython.org/github/etpinard/plotly-misc-nbs/blob/master/get-subplots/get-subplots.ipynb#Important-question) .

- [ ] 2. In order to implement `Figure.append_trace` and `Figure.print_arrangement` (that I think should be called `Figure.print_grid` just as in `make_subplots` by the way), I'm thinking about storing `grid_ref` ( which maps row, col -> x, y labels) and `grid_str` (the string repr of the subplot grid) in the `Figure` instance generated by `make_subplots`. These would be *hidden* properties of `fig` that would bypass validation (and hence bypass graph reference).

**Questions about details**

- [ ] 3. Should we make `print_grid=True` the default in `make_subplots` ?
- [ ] 4. Should we rename the keyword arguments `horizontal_spacing` and `vertical_spacing` arguments `hspace` and `vspace` ?

chriddyp:
My votes:
1. shared axes frame the bottom-left

2. i think that's cool, but it should probably still work even if the `fig` was created without `make_subplots` right?

3. +1

4. i vote more descriptive names as they probably won't be used that often. if it was something that would be used in every call, then i think we should shorten. but otherwise, i like the descriptive names

etpinard:
2. Won't work (at least the way I imagined) unless fig knows about how rows and cols the subplots grid has.

theengineear:
@chriddyp , @etpinard 
1. agreed
2. this was the trickiness I was thinking about earlier. i don't think we should include it since we can't promise that it will be in sync with changes to the figure. but this also will bog up appending traces to subplots right? One option is to check if axes have changed (in specific ways) and not offer the functionality when things have changed too much. (maybe that makes sense, or maybe not)
3. agreed
4. agreed

etpinard:
@theengineear @chriddyp 

About point 2.: I'm voting for leaving `fig.print_grid` and `fig.append_trace` for a future PR once we have a better idea of how things will shape up. 

+1 ?

chriddyp:
how about we do 2 in this PR the way that you propose but just raise an exception if someone changes the domains in the axes? this seems like the main difficulty of creating subplots: `'x{}'.format(i)` is too gnarly for folks I think! it'll prolly cover 99% of the cases

etpinard:
Sure. I'm down.

theengineear:
sounds good to me!

etpinard:
@theengineear good call. `_grid_ref` and `_grid_str` make more sense as attributes.

etpinard:
@theengineear @chriddyp version bump and merge?

theengineear:
:hamburger: from me, but I think we should @chriddyp a little time to respond as well!

plotly/graph_objs/graph_objs.py
```diff
@@ -919,6 +919,53 @@ def __init__(self, *args, **kwargs):
...
+        try:
+            grid_str = self._grid_str
+        except KeyError:
+            raise Exception("Use tools.make_subplots "
+                            "to create a subplot grid.")
+        print(grid_str)
+    Figure.print_grid = print_grid
+
+    def append_trace(self, trace, row, col):
+        try:
```
chriddyp: docstring!

plotly/graph_objs/graph_objs.py
```diff
@@ -919,6 +919,53 @@ def __init__(self, *args, **kwargs):
                 kwargs['layout'] = Layout()
         super(Figure, self).__init__(*args, **kwargs)
     Figure.__init__ = __init__  # override method!
+
+    def print_grid(self):
+        try:
```
chriddyp: docstring!

> Print a visual layout of the figure's axes arrangement. This is only valid for figures that are created with `plotly.tools.make_subplots`.

chriddyp:
I'm gonna give this a quick spin :car:

chriddyp:
(ongoing)

- [ ] Missing trace, rightmost xaxis should prolly be on the bottom
image

plotly/tools.py
```diff
@@ -386,23 +386,20 @@ def mpl_to_plotly(fig, resize=False, strip_style=False, verbose=False):
...
 
     Example 1:
-        # stack two subplots vertically
-        fig = tools.get_subplots(rows=2)
-        fig['data'] += [Scatter(x=[1,2,3], y=[2,1,2], xaxis='x1', yaxis='y1')]
-        fig['data'] += [Scatter(x=[1,2,3], y=[2,1,2], xaxis='x2', yaxis='y2')]
+    # stack two subplots vertically
+    fig = tools.get_subplots(rows=2)
+    fig['data'] += [Scatter(x=[1,2,3], y=[2,1,2], xaxis='x1', yaxis='y1')]
+    fig['data'] += [Scatter(x=[1,2,3], y=[2,1,2], xaxis='x2', yaxis='y2')]
```
chriddyp: Maybe changed to:

>    # stack two subplots vertically
>    fig = tools.get_subplots(rows=2)
>    fig.append_trace(Scatter(x=[1,2,3], y=[2,1,2]), row=1, col=1)
>  # alternatively: fig['data'] += [Scatter(x=[1,2,3], y=[2,1,2], xaxis='x1', yaxis='y1')]
>    fig.append_trace(Scatter(x=[1,2,3], y=[2,1,2]), row=2, col=1) 
>  # alternatively: fig['data'] += [Scatter(x=[1,2,3], y=[2,1,2], xaxis='x2', yaxis='y2')]

Not sure about the "alternatively" text - thoughts? @theengineear , @etpinard

chriddyp:
besides this comment (https://github.com/plotly/python-api/pull/170#discussion_r22746521), :dancer: from me

plotly/tools.py
theengineear: yeah, let's give one example and link to documentation, we can use that thing to link to the appropriate domain too: https://github.com/plotly/python-api/blob/80fcc59309d59c227a91075070882904e29fb006/plotly/utils.py#L286-292

with the `@utils.template_doc` wrapper, like in https://github.com/plotly/python-api/blob/a0dad3cc0b70a60e195d4a6ea7eaca06937d1bf3/plotly/plotly/plotly.py#L391

theengineear:
should we make a note about subtle changes (don't have to list them, but just let the user know) here:

https://github.com/plotly/python-api/blob/revamp-get_subplots/plotly/tools.py#L430-433

PR #175, opened 2015-01-09
Add strict requirements for python 2.6. (more)

theengineear:
Though these requirements are unecessary for Python versions after 2.6,
they’re small, harmless libs to import. They ended up being included
in stdlib anyways.


theengineear:
@chriddyp and @etpinard , python 2.6 users need to find this stuff on their own. I don't *think* including it for *all* users will cause any problems. If tests pass, any reason not to include these?

etpinard:
:dancer: ship it !

PR #181, opened 2015-01-24
Widgets save graph

theengineear:
Some annoyingly-implemented save functionality.

You can just call `graph.save(filename='whatevs')`. We might want the save call signature to match py.plot, not sure.


theengineear:
@chriddyp , wanna give it a look? when this PR is merged, it'll work: https://github.com/plotly/streambed/pull/1192

theengineear:
@chriddyp , I was making the tutorial for widgets, when i realized we never merged this guy in. should i fix it up and we can finally let users save graphs made in widgets : )

chriddyp:
yeah, that sounds good

theengineear:
cool, i'll try and fix it up tonight.

theengineear:
@chriddyp , want to take a peek? I added the fadeTo functionality interface on top of the message protocol. A little sloppy maybe? I'm less familiar with how the messaging works though, so there may be a cleaner way.

theengineear:
Note, there seems to be an issue in Python 2.6 with requests where requests[security] needs to be installed (so, beyond just requests). Perhaps a change in circle caused this to pop up? Should be fixed by the last two commits.

chriddyp:
:dancer:

PR #182, opened 2015-01-24
4x iplot load-time performance

chriddyp:
cc @theengineear


theengineear:
These tests are failing: https://github.com/plotly/python-api/blob/master/plotly/tests/test_optional/test_ipython/test_embed.py#L30-62

You can probably just remove them. We can go back in the commit history and snag them again if we reimplement.

I opened up an issue in ipython about this: https://github.com/ipython/ipython/issues/7573

PR #183, opened 2015-01-24
Cached image iplot

chriddyp:
On calling `iplot`, render an image first then replace it with the interactive version once it loads. If offline, the image of the graph will still be viewable. Re-opening notebooks will be fast with the cached images.

Example: http://nbviewer.ipython.org/gist/chriddyp/e567a9d1dd659a6b397b

### DIFF
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -97,8 +97,13 @@ def get_credentials():
 
 ### plot stuff ###
 
-def iplot(figure_or_data, **plot_options):
+def iplot(figure_or_data, cached=False, **plot_options):
     """Create a unique url for this plot in Plotly and open in IPython.
+
+    cached (bool) -- Show a static image of the graph while the interactive
+                     version loads. Note: the interactive version will
+                     not display on notebooks if Javascript can't be run,
+                     e.g. notebooks shared with nbviewer.
 
     plot_options keyword agruments:
     filename (string) -- the name that will be associated with this figure


plotly/plotly/plotly.py
```diff
@@ -97,9 +97,14 @@ def get_credentials():
...
 ### plot stuff ###
 
-def iplot(figure_or_data, **plot_options):
+def iplot(figure_or_data, cached=False, **plot_options):
     """Create a unique url for this plot in Plotly and open in IPython.
 
+    cached (bool) -- Show a static image of the graph while the interactive
+                     version loads. Note: the interactive version will
+                     not display on notebooks if Javascript can't be run,
+                     e.g. notebooks shared with nbviewer.
```
theengineear: so, is this not true then? it will work?

chriddyp:
Alright, here is an example of the proposed behavior: http://nbviewer.ipython.org/gist/chriddyp/a5585876433d42eb0844

- make 2 threaded requests to plotly: 1 for image, 1 for static plot. show the first one that comes back
- save the static images to the notebook
- on page reload, don't load the interactive graphs on page load
- on page scroll, load the interactive graphs as you scroll near them

theengineear:
@chriddyp, sorry if I dropped the :8ball: on getting back here. Do you need/want me to look this over and are we still trying to go ahead on this one?

chriddyp:
Nah, we can close it. Needs a few more days to make it really nice, and I don't have the bandwidth now.

PR #184, opened 2015-01-26
skip image tests (we’re removing these methods for now)

theengineear:
@chriddyp , just removed the tests for this.


chriddyp:
:dancer:

PR #186, opened 2015-02-02
Session module

theengineear:
This factors out the common _config/_credentials/_plot_options from plotly.py into a new session.py module which is meant to *not* rely on other plotly modules such that all modules can safely import it to know what the current session-status is.

### DIFF
--- a/plotly/__init__.py
+++ b/plotly/__init__.py
@@ -28,5 +28,5 @@ Modules:
 
 from __future__ import absolute_import
 
-from plotly import plotly, graph_objs, grid_objs, tools, utils
+from plotly import plotly, graph_objs, grid_objs, tools, utils, session
 from plotly.version import __version__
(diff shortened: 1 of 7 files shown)


theengineear:
@chriddyp and @etpinard , this also fixes a problem that occurs when you sign-in to a different plotly domain... Since tools.embed uses the file-config, what gets embedded was not matching what was iplotted! yikes!

etpinard:
Inspired by (Lightning-viz](http://nbviewer.ipython.org/github/lightning-viz/lightning-example-notebooks/blob/master/plots/scatter.ipynb) ?

theengineear:
nah, this is something i've been meaning to refactor for a while now. it should be a better way to stay organized.

theengineear:
thx! i'll wait for the tests and merge it in.

etpinard:
Cool. :dancer: on my end.

PR #194, opened 2015-02-18
Widget box

theengineear:
This is an *extremely* rough crack at adding a little Plotly container for instantiating, changing, saving, and linking to Plotly plots.


theengineear:
@chriddyp , I don't really have the time to work on this, so I'm just throwing this idea up.

Issue #202, opened 2015-03-10
json encoder should handle NaN and Infinity


chriddyp:
python's json encoder does this really annoying thing where it uses "extended" JSON, which includes NaN and Infinity. This means that `JSON.parse()` (javascript) of `NaN` fails! We should probably just replace with `None`

reference: http://stackoverflow.com/questions/17503981/is-there-a-way-to-override-pythons-json-handler

cc @theengineear

theengineear:
Yup, good call.

theengineear:
This is *fairly* simple to implement based on the link you provided.

One caveat. What about this case:

(1) user has `Inf` and `NaN` from some calculations
(2) user uploads that data to plotly
(3) collaborator snags data from plotly
(4) collaborator *decodes* and get's `None` now!

If they're using numpy, they can get back 'NaN' pretty easily:

http://stackoverflow.com/questions/19456239/convert-python-list-with-none-values-to-numpy-array-with-nan-values

But since it's a many-to-one mapping, there's obviously no way to get both `NaN` and `Inf` back.

I think I'm okay with this if we document the link I provided above.

PR #203, opened 2015-03-10
Override `JSONEncoder.iterencode`

theengineear:
We do this to handle NaN, Inf, and -Inf

Fixes: #202

Todo:

- [x] convert these types to 'null'
- [x] update broken tests (that have NaN, Inf, etc)
- [x] write new *core* test to specify this functionality


theengineear:
@chriddyp , this one looked like a quickie, so I just grabbed it.

plotly/utils.py
```diff
@@ -119,8 +119,72 @@ class PlotlyJSONEncoder(json.JSONEncoder):
...
+    Additionally, this encoder overrides nan functionality so that 'Inf',
+    'NaN' and '-Inf' encode to 'null'. Which is stricter JSON than the Python
+    version.
+
     """
 
+    # we want stricter JSON, so convert NaN, Inf, -Inf --> 'null'
+    nan_str = inf_str = neg_inf_str = 'null'
+
+    # uses code from official python json.encoder module. Same licence applies.
```
chriddyp: woah gnarly

chriddyp: this may be naive, but can't we just extend our current encoder with a check for is `NaN`, `Inf`? like how we do with datetimes https://github.com/plotly/python-api/blob/master/plotly/utils.py#L224

theengineear: situation kinda stinks, I actually documented it because this same thing was confusing to me as well:

https://github.com/plotly/python-api/blob/master/plotly/utils.py#L135-150

We only get a crack at stuff if the JSONEncoder *can't* encode it. What's more, as is mentioned in the SO post, when `allow_nan=False`, we *also* don't get a crack at it in our encoder. bummer!

Here's an example:

```python
import json
import numpy

class Foo(object):
    pass

class NullEncoder(json.JSONEncoder):
    def encode_as_true(self, obj):
        return None
    
    def default(self, obj):
        return self.encode_as_true(obj)

json.dumps(Foo(), cls=NullEncoder)  # here, it works
# 'null'
json.dumps(numpy.ma.core.masked, cls=NullEncoder)  # and again here
# 'null'
json.dumps(float('NaN'), cls=NullEncoder)  # no shot! this doesn't work as expected!
# 'NaN'
```

chriddyp: wow, that's pretty wild. well, that looks good then, nice diggin!

theengineear:
@chriddyp , i'm ready to roll this out when tests pass. nice catch on this!

chriddyp:
boom!

PR #204, opened 2015-03-13
Circle pyenv

theengineear:
Just some fun with circle:

* circle didn't have the pyenv shims ahead of the system python binary's path therefore, even though we were setting up pyenv right, it was still using python 2.7 to run the majority of our test suite
* we *thought* it was working because the variables were set right and we even got python 2.6- and python 3.x-specific errors because we *we're* still catching the general module import test.


theengineear:
@chriddyp , look ok to you?

chriddyp:
wow, good catch! :dancer:

PR #205, opened 2015-03-16
Ipython 3 widgets

theengineear:
This doesn't cover *everything* listed in the [widget compatibility doc](http://ipython.org/ipython-doc/3/whatsnew/version3_widget_migration.html), but it gets our base functionality up and running again.

I don't fully understand why we're not supposed to use `display` anymore, but things seem to work as-is, and unless we run into an issue, I'm not sure it's worth taking the time now to change it.


theengineear:
@chriddyp , IPython updated what the require call imports, we just had to move a few things around.

Also, there was a curious bug related to the parent div of the iframe. It's style was getting wiped (even though it's clearly set in the JS render function) by the time it shows up in the output cell. Not sure where it's happening... I added a hacky override of this behavior in a timeout function : (

Not sure if you have any ideas.

//cc @jackparmer

theengineear:
@chriddyp , @etpinard , poke.

etpinard:
looks good to me. :dancer:

PR #220, opened 2015-05-25
Manifest file

### LINKED ISSUE #22: should pkg_resources be a dependency?
trying to run plotly on a barebones, pip-less python distribution (the Arduino Yun) and ran into an `ImportError` for `pkg_resources` in `graph_objs`. Should it be an official setup.py dependency?

theengineear:
Fixes #22

This uses a MANIFEST.in file to ensure data gets into the sdist tar ball. It also removes the need to use the `pkg_resources` import, which seemed to be a problem for some folks.


theengineear:
@chriddyp , we've wanted to ditch the `pkg_resources` stuff for a while, this ditches it.

chriddyp:
nice! :dancer:

PR #222, opened 2015-05-25
Add Quiver and Streamline Functions

cldougl:
Quiver function plots velocity vectors and Streamline function plots
streamlines with directional arrows


chriddyp:
cc @theengineear @etpinard @aneda

plotly/tools.py
```diff
@@ -15,6 +15,8 @@
 import six
 import requests
 
+import numpy as np
```
chriddyp: we can't make `numpy` a dependency. check out how we do "protected" imports of `pandas` to get around this

plotly/tools.py
```diff
@@ -1269,3 +1271,512 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
             )
 
     return figure
+
+### Quiver Plots ###
```
chriddyp: rm this line for consistent style

plotly/tools.py
```diff
@@ -1269,3 +1271,512 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+def Quiver(X, Y, U, V, **kwargs):
+    """ Return data and layout objects for quiver plot
+    Example 1:
+    # X and Y are the coordinates of the arrow locations (created from numpy meshgrid),
+    # and U and V give the X and Y components of the arrow vectors. 
+    X,Y = np.meshgrid(np.arange(0,np.pi,.2),np.arange(0,np.pi,.2) )
+    U = np.cos(X)
+    V = np.sin(Y)
```
chriddyp: match the docstring pattern of the other functions, like `make_subplots`. the args `X`, `Y`, `U`, `V` should be described at the top, and the examples should be enclosed in ```

plotly/tools.py
```diff
@@ -1269,3 +1271,512 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+    data, layout = Quiver(X,Y,U,V)
+    figure = Figure(data = data, layout = layout)
+    plot_url = py.plot(figure, filename = 'QuiverEx1')
+    Example 2:
```
chriddyp: and include more spacing between example 1 and example 2

plotly/tools.py
```diff
@@ -1269,3 +1271,512 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+    barb_width (kwarg, int greater than or equal to 1, default = width):
+        Change width of lines for barbs.
+    arrow_width (kwarg, int greater than or equal to 1, default = width):
+        Change width of lines for arrows.
+    title (kwarg, string, default=none):
+        Title of plot.
+    xlabel (kwarg, string, default=none):
+        Label of x-axis
+    ylabel (kwarg, string, default=none):
+        Label of y-axis
```
chriddyp: not sure how i feel about adding `title`, `xlabel`, and `ylabel` attributes into `Quiver`. If folks see these labels in `Quiver` then might expect these attributes in other graph objects, or other tools that we make. this makes teaching users about labeling their graphs, through `Layout(title=title)` and `Layout(xaxis=XAxis(title=title))` inconsistent

plotly/tools.py
```diff
@@ -1269,3 +1271,512 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+    Keywords arguments with constant defaults:
+    angle (kwarg, angle in radians, default = np.pi/9):
+        Angle of arrowhead.
+    scale (kwarg, float in [0,1], default = .1):
+        Scales size of the arrows (avoid overlap).
+    arrow_length (kwarg, float in [0,1], default = 0.3):
+        Value multiplied to length of barb to get length of arrowhead.
+    color (kwarg, rgb vale, default = rgb(114, 132, 304)):
+        Set color of both barb and arrow.
+    barb_color (kwarg, rgb vale, default = color):
```
chriddyp: "value" typo

plotly/tools.py
```diff
@@ -1269,3 +1271,512 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        Set color of barbs.
+    arrow_color (kwarg, rgb vale, default = color):
```
chriddyp: "value"
also, this can be any [valid html color string](http://www.w3.org/TR/css3-color/#colorunits), so we might want call the type "color string" instead of "rgb value"

plotly/tools.py
```diff
@@ -1269,3 +1271,512 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+    plot_url = py.plot(figure, filename = 'QuiverEx1')
+    Example 2:
+    X,Y = np.meshgrid(np.arange(0,2,.2),np.arange(0,2,.2))
+    U = np.cos(X)*Y;
+    V = np.sin(X)*Y;
+    data, layout = Quiver(X, Y, U, V)
+    figure = Figure(data = data, layout = layout)
+    plot_url = py.plot(figure, filename = 'QuiverEx2')
+
+    Keywords arguments with constant defaults:
```
chriddyp: in the event that we eventually sometime make vector plots an actual plotly.js chart type, we should probably use the same keyword attributes.. so @etpinard you should check these out

plotly/tools.py
```diff
@@ -1269,3 +1271,512 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+    color = kwargs.pop('color', 'rgb(114, 132, 314)')
+    barb_col = kwargs.pop('barb_color', color)
+    arrow_col = kwargs.pop('arrow_color', color)
+    width = kwargs.pop('width', 1)
+    arrow_width = kwargs.pop('arrow_width', width)
+    barb_width = kwargs.pop('barb_width', width)
+    title = kwargs.pop('title','')
+    xlab = kwargs.pop('xlabel','')
+    ylab = kwargs.pop('ylabel','')
+    hover = kwargs.pop('hover','closest')
```
chriddyp: pep8! configure your editor (sublime instructions here: https://github.com/SublimeLinter/SublimeLinter-pep8) to use pep8 syntax style. this includes having a space after commas :)

plotly/tools.py
chriddyp: alternatively, rewrite the numpy functionality in pure python!

plotly/tools.py
```diff
@@ -1269,3 +1271,512 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+                     mode = 'lines', name = 'Arrow', 
+                     line = Line(width = arrow_width, color = arrow_col))
+        
+    # Data
+    data = Data([trace1, trace2])
+    # Layout
+    layout = Layout(title = title, xaxis = XAxis(title = xlab), 
+                    yaxis = YAxis(title = ylab), hovermode = hover)
+            
+    return data, layout
```
chriddyp: why not just return a Figure?

plotly/tools.py
cldougl: oops-thanks!! got it

plotly/tools.py
etpinard: +1 for returning a `Figure` obj.

plotly/tools.py
```diff
@@ -1269,3 +1271,512 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+    # Combine arrays into matrix
+    YArrows = np.matrix([YPoint1, Yend, YPoint2, nnn])
+    # Make matrix into array readable by plotly
+    YArrows = np.array(YArrows)
+    YArrows = YArrows.flatten('F')
+        
+    # Make trace2: the arrows
+    trace2 = Scatter(x = XArrows, y = YArrows, 
+                     mode = 'lines', name = 'Arrow', 
+                     line = Line(width = arrow_width, color = arrow_col))
```
etpinard: `hoverinfo: none` to remove hover-text on the arrows would be nice (ref [here](https://github.com/plotly/streambed/blob/master/shelly/plotlyjs/static/plotlyjs/src/graph_obj.js#L1190))

same for streamline arrows

plotly/tools.py
etpinard: or alternatively, return only `data` and let the users fill in

```py
layout = Layout(title = title, xaxis = XAxis(title = xlab), 
                         yaxis = YAxis(title = ylab), hovermode = hover)
```

chriddyp: +1

cldougl: Yep, I like returning only data. (I originally just returned data and left layout to have users fill in, but the default hover looks a bit strange with these plots so I added layout as well and returned both separately so it would be easier for the user to add to one or the other if they needed.)

etpinard: Logic around for setting the *correct* default hover mode should be handled in js land.

@cldougl I'd love to show you around plotlyjs someday if you like.

cldougl: ok- all hover info removed from the python script. @etpinard yes! plotlyjs visit soon

chriddyp: @etpinard - is the case where non-monotonic x-data would have `hovermode="closest"`?

etpinard: more like *clustered* data as well as bubble charts (i.e. `Array.isArray(marker.size)`) would get `hovermode: 'closest'`.

plotly/tests/test_core/test_tools/test_Quiver.py
```diff
@@ -0,0 +1,223 @@
+import plotly
+from plotly.graph_objs import *
```
theengineear: :cow2: You shouldn't use the import `*` notation within our own packages.

Instead just import the things you need directly.

plotly/tests/test_core/test_tools/test_Quiver.py
```diff
@@ -0,0 +1,223 @@
...
+import plotly.tools as tls
+from nose.tools import raises
+import numpy as np
+
+
+@raises(Exception)
+def unequal_xy_length():
```
theengineear: does this test run? I think that you need to have `test` in the definition for it to register? might be wrong about that though.

plotly/tests/test_core/test_tools/test_Quiver.py
```diff
@@ -0,0 +1,223 @@
...
+    data = tls.Quiver(x=[1, 2], y=[1], u=[1, 2], v=[1, 2])
+
+
+@raises(Exception)
+def unequal_uv_length():
```
theengineear: ditto, see https://github.com/plotly/python-api/pull/222/files#r31415670

plotly/tests/test_core/test_tools/test_Quiver.py
```diff
@@ -0,0 +1,223 @@
...
+    data = tls.Quiver(x=[1, 2], y=[1, 3], u=[1], v=[1, 2])
+
+
+@raises(Exception)
+def test_wrong_kwarg():
+    data = tls.Quiver(stuff='not gonna work')
+
+
+def test_one_arrow():
+    nan = np.nan
```
theengineear: you can't have a numpy dependency in our core tests, can you switch these from `plotly/tests/test_core` to `plotly/tests/test_optional`?

plotly/tests/test_core/test_tools/test_Quiver.py
```diff
@@ -0,0 +1,223 @@
...
+    trace1 = Scatter(
+            x=[0., 1., nan],
```
theengineear: I believe `None` will suffice in lieu of `np.nan`

plotly/tests/test_core/test_tools/test_Quiver.py
theengineear: even better, can you just remove the numpy dependency from here?

plotly/tests/test_core/test_tools/test_Quiver.py
```diff
@@ -0,0 +1,223 @@
...
+    data = tls.Quiver(x=[1, 2], y=[1], u=[1, 2], v=[1, 2])
+
+
+@raises(Exception)
+def unequal_uv_length():
+    data = tls.Quiver(x=[1, 2], y=[1, 3], u=[1], v=[1, 2])
+
+
+@raises(Exception)
+def test_wrong_kwarg():
```
theengineear: Also, it's usually more powerful to test via `TestCase`, which you get from `from unittest import TestCase`.

Here's an example in this repo:
https://github.com/plotly/python-api/blob/4a19ce2d382495692affd47dd107baab1755f5a7/plotly/tests/test_core/test_plotly/test_credentials.py#L33-73

plotly/tests/test_core/test_tools/test_Quiver.py
```diff
@@ -0,0 +1,223 @@
...
+            y=[0.61548617,  1.,  0.82069826, nan],
+            mode='lines',
+            name='Arrow',
+            line=Line(color='rgb(114, 132, 314)', width=1)
+            )
+    expected = Data([trace1, trace2])
+
+    data = tls.Quiver(x=[0], y=[0], u=[1], v=[1], scale=1)
+
+    np.testing.assert_almost_equal(data[0]['y'], expected[0]['y'])
```
theengineear: again, if you're using TestCase, there's a `self.assertAlmostEqual`

plotly/tests/test_core/test_tools/test_Streamline.py
```diff
@@ -0,0 +1,10 @@
+import plotly
+from plotly.graph_objs import *
+import plotly.tools as tls
+from nose.tools import raises
+import numpy as np
+
+
+@raises(Exception)
+def test_wrong_kwarg():
+    fig = tls.make_subplots(stuff='not gonna work')
```
theengineear: huh?

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
             )
 
     return figure
+
+
+def Quiver(x, y, u, v,
```
theengineear: pep:8ball: you shouldn't Capitalize functions. https://www.python.org/dev/peps/pep-0008/#function-names

theengineear: do you know how to check pep8 stuff in your development environment?

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+    ```
+    x,y = np.meshgrid(np.arange(0,2,.2),np.arange(0,2,.2))
+    u = np.cos(x)*y;
+    v = np.sin(x)*y;
+    data = Quiver(x, y, u, v)
+    ```
+
+    Keywords arguments with constant defaults:
+
+    scale (kwarg, float in [0,1], default = .1):
```
theengineear: They're not fully-fleshed out, but ther eare some good python guidelines that we try to follow here for docstrings:

https://github.com/plotly/dev-docs/blob/master/reference/styleguides/python.md#function-docstrings

Please try to follow them as much as possible and feel free to ask!

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        Set color of arrow.
+
+    barb_width (kwarg, int greater than or equal to 1, default = width):
+        Change width of lines for barbs.
+
+    arrow_width (kwarg, int greater than or equal to 1, default = width):
+        Change width of lines for arrows.
+    """
+
+    if _numpy_imported is False:
```
theengineear: :cow2: for booleans, it's often clearer to do `if not _numpy_imported` here.

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        raise Exception("To use Quiver() please import numpy as np")
```
theengineear: Non standard errors that we raise should either be PlotlyError or a subclass of them: https://github.com/plotly/python-api/blob/master/plotly/exceptions.py#L28

This is sort of a standard error though, perhaps you can just raise an `ImportError`?

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+
+    if len(x) != len(y):
+        raise Exception("x, y, u, and v should be ndarrays of equal dimmension"
```
theengineear: This should be a PlotlyError of a subclass thereof.

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+                        "or lists where x and y are the same length"
+                        "and u and v are the same length")
+
+    if len(u) != len(v):
+        raise Exception("x, y, u, and v should be ndarrays of equal dimmension"
+                        "or lists where x and y are the same length"
+                        "and u and v are the same length")
+
+    VALID_KWARGS = ['angle', 'scale',
```
theengineear: :cow2: Constants like this should be declared at the module level. Perhaps VALID_QUIVER_KWARGS?

:cow2: Also, you can use a `tuple` here instead of a `list` since it's immutable anyways. It's a slight performance increase (although it's more about convention, imo)

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+                    'arrow_scale', 'barb_color',
+                    'arrow_color', 'barb_width',
+                    'arrow_width']
+    for key in kwargs.keys():
+        if key not in VALID_KWARGS:
+            raise Exception("Invalid keyword argument: '{0}'".format(key))
```
theengineear: hey! you remembered to use the `0` for our Python 2.6 support, nice!

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+
+    # Make x y u v into np.array in case user entered list
+    x = np.array(x)
```
theengineear: is there anyway to create quivers *without* requiring a dependency on numpy?

@chriddyp , or @etpinard , would that be desirable?

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+    y = np.array(y)
+    u = np.array(u)
+    v = np.array(v)
+
+    # Make arrays of x & y start values
+    Xstart = x.flatten('F')
```
theengineear: pep:8ball: variable names should be in `snake_case`, so `x_start`

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+    Ystart = y.flatten('F')
```
theengineear: also, the alternative to flatten is to use something like this:

http://stackoverflow.com/questions/952914/making-a-flat-list-out-of-list-of-lists-in-python

```
[item for sublist in l for item in sublist]
```

It's a fairly confusing list comprehension, but it *does* work.

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+    # Make arrays of x & y end values
+    Xend = x + (u * scale)
+    Xend = Xend.flatten('F')
+    Yend = y + (v * scale)
+    Yend = Yend.flatten('F')
+
+    # Make array of nans
+    # These are used as spaces so the separate
+    # arrows can be included in 1 trace
+    nnn = np.empty((len(Xstart)))
```
theengineear: :cow2: try and pick descriptive variable names.

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+    nnn[:] = np.NAN
+
+    # Combine arrays into matrix
+    Xvals = np.matrix([Xstart, Xend, nnn])
```
theengineear: it looks like you rely *heavily* on numpy, which is fine, this can move to *optional* functionality for now, we can always change it in a non-backwards-incompatibility-breaking way later : D

plotly/tools.py
```diff
@@ -1269,3 +1276,547 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+
+    # Make trace2: the arrows
+    trace2 = Scatter(x=XArrows, y=YArrows,
+                     mode='lines', name='Arrow',
+                     line=Line(width=arrow_width, color=arrow_color))
+
+    # Data
+    data = Data([trace1, trace2])
+
+    return data
```
theengineear: :cow2: you might consider making this into a class? this function has > 100 lines of code in it. As an example, it can then have a single method that handles cleaning up the input arguments, which currently takes up ~20 lines.

theengineear:
@cldougl , looking great! I sort of threw a lot of information in there. Please feel free to ping me on anything that's unclear.

There are a couple major things that I think we need to address:

* does this need to rely on numpy? unless we can show it's an enormous performance increase, it seems nicer to *not* depend on it. It might be interesting to see if `math` would suffice?
* class vs function organization. I think both of these should be changed to class definitions over function definitions. Was there a reason we wanted them to be functions and not classes in the first place?

cldougl:
@theengineear thanks for the review! No problem for making them classes. I'll work on removing numpy today and I'll let you know if I have any problems or questions.

plotly/tools.py
cldougl: @theengineear sorry the link to the guidelines doesn't seem to be working (and I think those could be pretty useful for me)

theengineear: You might not have permission to view it?

Can you access the repo at all? https://github.com/plotly/dev-docs/

If you can't, you'll need to ping someone, perhaps Chris about getting access.

theengineear:
@cldougl , i think you've still got a dependency on numpy in your tests that you'll need to remove so we can test this.

cldougl:
@theengineear I can remove the numpy dependency in the test or I can move the tests to optional tests (instead of core). I thought the numpy depend should stay in one of the test cases because most users will probably be using numpy to set their x, y, u, v values

theengineear:
@cldougl , if `Quiver` is intended to be core functionality (which it is), it needs to have it's tests in core, *not* optional. If you want to test it will numpy stuff, you'll need to *also* have some tests in test_optional.

Sorry for the inconvenience :(

The reason is that we test the core stuff on Circle **without** those optional dependencies installed to be sure that users with a similar setup are guaranteed to have a working plotly install.

cldougl:
cc @theengineear @chriddyp

chriddyp:
@cldougl - we're just waiting on consensus of this comment: https://github.com/plotly/python-api/pull/222#commitcomment-11615755. @theengineear is going to ruminate on it later today

cldougl:
@chriddyp sounds good. 
I'm especially agree/am concerned with this point:  "I do think it's confusing to try to remember which Capital Letter chart types require you to call get_trace and which ones "you can just plot"."

theengineear:
re: https://github.com/plotly/python-api/pull/222#commitcomment-11615755

I'm pretty sure we'll want to *not* mix functions/dicts into our graph_objs namespace. Keeping that homogenous is going to be a good idea, I think.

I'm still thinking of the quiver and streamline plots as beta functionality that the Python library supports for *now*, which may make it's way into our graphing library eventually. From this perspective, I'd suggest we roll with a new namespace, like you suggested. Well, I think it should just go in `tools.py`.

```
from plotly.tools import TraceFactory

quiver = TraceFactory.create_quiver(**some_kwargs_here)
fig = go.Figure()
fig['data'].append(quiver)
```

Once you've *created* the quiver object, it's clear that you're getting back a trace that you can play with like any other trace.

Notes:

* there is *no* `Quiver` (with a capital Q) this way, so we'll dodge that confusion
* you add the resulting trace like any other trace and it *is* a trace
* you call `help(TraceFactory)` and `help(TraceFactory.create_quiver)`, this is different, but i think OK?
* when we *do* have a real `Quiver` object later, the user can just do `Quiver(**some_kwargs_here)` without conflict/confusion. We could even *return* a proper `Quiver` object from `create_quiver` (although this might not be backwards compatible)

plotly/tests/test_core/test_tools/test_TraceFactory.py
```diff
@@ -0,0 +1,112 @@
...
+from plotly.graph_objs import graph_objs, Scatter, Data, Marker, Line, Trace
+from plotly.exceptions import PlotlyError
+
+import plotly.plotly as py
+import plotly.tools as tls
+import math
+from nose.tools import raises
+
+
+@raises(PlotlyError)
```
theengineear: :cow2: you can just stick these in your test case:

```
def TestQuiver(TestCase):

    def test_unequal_xy_length():
        kwargs = {'x': [1], 'y': [1], 'u': [1, 2], 'v': [1, 2]}
        self.assertRaises(PlotlyError, tls.TraceFactory.create_quiver, **kwargs)
```

plotly/tests/test_core/test_tools/test_TraceFactory.py
```diff
@@ -0,0 +1,112 @@
...
+@raises(ValueError)
+def test_wrong_arrow_scale():
+    data = tls.TraceFactory.create_quiver(x=[1], y=[1], u=[1], v=[1],
+                                          arrow_scale=-1)
+
+
+class TestQuiver(TestCase):
+
+    def test_one_arrow(self):
+        self.assertAlmostEqual(tls.TraceFactory.create_quiver(x=[1], y=[1],
```
theengineear: :cow2: This seems like more work that it's worth, but trust me : D

*Try and make tests as clear as possible* this is one of the places that we'll look at with a lot of scrutiny for code review and this is what keeps stuff working as expected! It's trivial to you know, but this is super helpful to future debuggers (which could be you!).

How about:

```
    def test_one_arrow(self):
    
        # we should be able to create a single arrow using create_quiver
    
        quiver = tls.TraceFactory.create_quiver(x=[1], y=[1], u=[1], v=[1], scale=1)
        expected_quiver = {
            'y': [1, 2, None, 1.615486170766527, 2, 1.820698256761928, None],
            'x': [1, 2, None, 1.820698256761928, 2, 1.615486170766527, None],
            'type': 'scatter',
            'mode': 'lines'
        }
        self.assertEqual(quiver, expected_quiver)
```

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+
+
+class TraceFactory(dict):
+    @staticmethod
+    def create_quiver(x, y, u, v, scale=.1, arrow_scale=.3,
+                      angle=math.pi/9, **kwargs):
+        """
+        Return data for a quiver plot.
+
+        :param (list or array) x: x coordinates of the arrow locations
```
theengineear: :cow2: the stuff in the `()` in should be real types. do you mean `(list or ndarray)`? This can also be written `(list|ndarray)`. Docstring parsers can then give you type hinting :)

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+
+        # Create quiver
+        quiver = TraceFactory.create_quiver(x, y, u, v, scale=.2,
+                                            arrow_scale=.3, angle=math.pi/6,
+                                            line=Line(color='purple', width=1))
+        # Plot
+        fig=Figure()
+        fig['data'].append(quiver)
+        py.iplot(fig, filename='quiver')
+        ```
```
theengineear: nice job on all the examples \o/

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        """
+        barb_x, barb_y = Quiver(x, y, u, v, scale,
```
theengineear: I'm OK with leaving Quiver as the separate way of making Quiver objects. Two things though:

* if you had all the `Quiver` and `Streamline` methods under TraceFactory, you *might* be able to make things DRYer (the way you've got it now is totally fine though, perhaps even better since we may add a lot of these) hard to say
* you *should* give users one and only one clear way of using this. currently I find it confusing that you can use `TraceFactory.create_quiver` and `Quiver`

As a quick solution, can we change `Quiver` --> `_Quiver` so that it's clear to users that it's meant to be private to the module?

This is especially nice in environments like IPython where prefixing with a `_` or `__` will force the names to be hidden from the namespace by default.

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        self.u *= len(self.x)
+        self.v *= len(self.y)
+        self.validate()
+        self.st_x = []
+        self.st_y = []
+        self.get_streamlines()
+        streamline_x, streamline_y = self.sum_streamlines()
+        arrows_x, arrows_y = self.get_streamline_arrows()
+
+    def validate(self):
```
theengineear: ah, here's an example. If you had this *validate* method in `TraceFactory`, both the `create_quiver` and `create_streamline` could leverage it. Not a big deal for this PR, but perhaps worth a look once you start adding new trace types?

(re: https://github.com/plotly/python-api/pull/222/files#r33422409)

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        Uses list comprehension to flatten array
+
+        :param (array): An iterable to flatten
+        :raises (PlotlyError): If iterable is not nested.
+        :rtype (list): The flattened list.
+        """
+        try:
+            return [item for sublist in array for item in sublist]
+        except TypeError:
+            raise exceptions.PlotlyError("Your data array could not be"
```
theengineear: This will get you a bunch (still gets me). You need to have a space after `be` above. The printout for this is actually:

```
Your data array could not beflattened! Make sure x, y, u, and vare lists or numpy ndarrays!
```

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+                                         "flattened! Make sure x, y, u, and v"
+                                         "are lists or numpy ndarrays!")
+
+    def validate(self):
+        """
+        Validates that args and kwargs meet criteria,
```
theengineear: :cow2: There are generally two ways to do docstrings.

* One-liners https://www.python.org/dev/peps/pep-0257/#one-line-docstrings
* Multi-liners https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

For a multi-line docstring, docstring parsers will return the first line expecting it to be a summary. So, this would be best written as:

```
"""
Validates the user-input arguments.

Specifically, this checks that scale and arrow_scale are positive
and that x, y, u, and v are the same length.

:raises: (ValueError) If scale or arrow_scale is < 1.
:raises: (PlotlyError) If x, y, u, and v are not the same length.

"""
```

Nbd on this one, mostly just a heads up about the *first* line convention.

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        side of the arrowhead. Next uses arrow_scale to scale the length of
+        arrowhead and creates x and y values for arrowhead point1 and point2.
+        Finally x and y values for point1, endpoint and point2s for each
+        arrowhead are separated by a None and zipped to create lists of x and
+        y values for the arrows.
+
+        :rtype: (list) arrow_x: list of point1, endpoint, point2 x_values
+            separated by a None to create the arrowhead.
+        :rtype: (list) arrow_y: list of point1, endpoint, point2 y_values
+            separated by a None to create the barb of the arrow.
```
theengineear: :cow2: Hmm, not positive on this one, but I'm *pretty* sure only one rtype can be given as you *must* return one and only one value from a function.

So, the convention would be:

```
"""
:rtype: (list, list)
"""
```

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        self.u = np.array(u)
+        self.v = np.array(v)
+        self.angle = angle
+        self.arrow_scale = arrow_scale
+        self.density = int(30*density)
+        self.delta_x = self.x[1] - self.x[0]
+        self.delta_y = self.y[1] - self.y[0]
+        self.val_x = self.x
+        self.val_y = self.y
+        # Set up spacing
```
theengineear: :cow2: you can let this breathe a little by adding blank lines before comments : D

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+                 density, angle,
+                 arrow_scale, **kwargs):
+        self.x = np.array(x)
+        self.y = np.array(y)
+        self.u = np.array(u)
+        self.v = np.array(v)
+        self.angle = angle
+        self.arrow_scale = arrow_scale
+        self.density = int(30*density)
+        self.delta_x = self.x[1] - self.x[0]
```
theengineear: This looks like it assumes some things about the `x` array before doing any validation on it. Might help to wrap some of this in a validation method which can return helpful info to users.

Also, are we just blindly trusting that `x[n+1] - x[n]` ∀ `n`. Is this OK?

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        self.delta_y = self.y[1] - self.y[0]
+        self.val_x = self.x
+        self.val_y = self.y
+        # Set up spacing
+        self.blank = np.zeros((self.density, self.density))
+        self.spacing_x = len(self.x)/float(self.density-1)
+        self.spacing_y = len(self.y)/float(self.density-1)
+        self.trajectories = []
+        # Rescale speed onto axes-coordinates
+        self.u = self.u/(self.x[-1]-self.x[0])
```
theengineear: pep:8ball: https://www.python.org/dev/peps/pep-0008/#other-recommendations

`self.u / (self.x[-1] - self.x[0])`

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        self.arrow_scale = arrow_scale
+        self.density = int(30*density)
+        self.delta_x = self.x[1] - self.x[0]
+        self.delta_y = self.y[1] - self.y[0]
+        self.val_x = self.x
+        self.val_y = self.y
+        # Set up spacing
+        self.blank = np.zeros((self.density, self.density))
+        self.spacing_x = len(self.x)/float(self.density-1)
+        self.spacing_y = len(self.y)/float(self.density-1)
```
theengineear: pep:8ball: space around `/`

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        self.trajectories = []
+        # Rescale speed onto axes-coordinates
+        self.u = self.u/(self.x[-1]-self.x[0])
+        self.v = self.v/(self.y[-1]-self.y[0])
+        self.speed = np.sqrt(self.u*self.u + self.v*self.v)
```
theengineear: note, your spacing here is perfecto, though : D

theengineear: :cow2: fwiw, you can also use the `**` operator. `self.u**2 + self.v**2` nbd

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        if self.arrow_scale <= 0:
+            raise ValueError("arrow_scale must be > 0")
+        if self.density <= 0:
+            raise ValueError("density must be > 0")
+        for index in range(len(self.x)-1):
+            if (self.x[index + 1]-self.x[index])-(self.x[1]-self.x[0]) > .0001:
+                raise exceptions.PlotlyError("x must be a 1 dimmensional"
+                                             "evenly spaced array")
+        for index in range(len(self.y)-1):
+            if (self.y[index + 1]-self.y[index])-(self.y[1]-self.y[0]) > .0001:
```
theengineear: is it worth factoring out the *almost equality* check here, its hard to parse and I think it's supposed to be reused a couple places in your code.

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+                raise exceptions.PlotlyError("y must be a 1 dimmensional"
```
theengineear: :cow2: needs a `' '` character after `dimmensional`, also :goat: 'dimmensional' :)

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+                                             "evenly spaced array")
+        if self.u.shape != self.v.shape:
+                raise exceptions.PlotlyError("u and v should both be 2d arrays"
+                                             "with the same dimmensions")
```
theengineear: :goat: `dimensions`

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+
+    def blank_pos(self, xi, yi):
+        """
+        Set up postitions for trajectories to be used with rk4 function.
```
theengineear: :goat: `positions`

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        """
+        return int((xi/self.spacing_x)+0.5), int((yi/self.spacing_y)+0.5)
+
+    def value_at(self, a, xi, yi):
+        """
+        Set up for RK4 function, based on Bokeh's streamline code
+        """
+        if type(xi) == np.ndarray:
```
theengineear: :cow2: `isinstance` instead of `type` --> `if isinstance(xi, np.ndarray):`

Also, is `value_at` required? This seems like it forces users to have `numpy` installed, is this OK/intentional?

theengineear: totally fine if it is, btw : ), just checking to make sure it's what we're expecting.

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+            return x_traj, y_traj
+        else:
+            for xb, yb in zip(xb_changes, yb_changes):
+                self.blank[yb, xb] = 0
+            return None
+
+    def traj(self, xb, yb):
+        """
+        Integrate trajectories, used in get_streamlines()
+        """
```
theengineear: :dolls: (docstring dolls) please document `xb` and `yb`

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+
+        :param (array) st_x: x-values for all streamlines
+        :param (array) st_y: y-values for all streamlines
+        :param (angle in radians) angle: angle of arrowhead. Default = pi/9
+        :param (float in [0,1]) arrow_scale: value to scale length of arrowhead
+            Default = .09
+        :rtype (list) arrows_x: x-values to create arrowhead
+        :rtype (list) arrows_y: y-values to create arrowhead
+        """
+        ArrowEnd_x = np.empty((len(self.st_x)))
```
theengineear: pep:8ball: naming conventions ~~ArrowEnd_x~~ --> `arrow_end_x`

Caps are typically reserved for Classes (like `TraceFactory`)

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        # Combine arrays into matrix
+        arrows_y = np.matrix([point1_y, ArrowEnd_y, point2_y, space])
+        arrows_y = np.array(arrows_y)
+        arrows_y = arrows_y.flatten('F')
+        arrows_y = arrows_y.tolist()
+
+        return arrows_x, arrows_y
+
+    def sum_streamlines(self):
+        """
```
theengineear: :dolls: explanation?

plotly/tests/test_core/test_tools/test_TraceFactory.py
```diff
@@ -0,0 +1,112 @@
...
+                                                                [-1, -5]],
+                                                             v=[[1, 1],
+                                                                [-3, -3]],
+                                                             density=2,
+                                                             arrow_scale=.4,
+                                                             angle=math.pi/6,
+                                                             line=Line(
+                                                              color='purple',
+                                                              width=3))).keys(),
+                         (['y', 'x', 'line', 'type', 'mode']))
```
theengineear: :tiger2: this doesn't really test that `create_streamline` works. Mind adding a few more tests? These should really include bad param tests to make sure we're not blindly returning garbage out when there's garbage in.

theengineear:
@cldougl Got some more comments for ya! The highest-priority stuff is related to test coverage.

Note, I'm pretty sure this *isn't* being run by our tests on Circle. Maybe because it's a PR from your fork? That seems strange though... At any rate, once it gets merged in, you might need to update with another PR if master is failing.

plotly/tools.py
cldougl: @theengineear I totally agree: I started putting everything under TraceFactory and then worried it would be too massive/complicated with all of the chart types so I went back to organizing under separate classes.  `_Quiver` seems like a good solution.

plotly/tools.py
cldougl: @theengineear re: `x[n+1] - x[n]` ∀ `n`  for streamlines, x and y are supposed to be one dimensional evenly spaced arrays so  `x[n+1] - x[n]` ∀ `n` is true. This was the case for every streamline example (python and matlab) I looked at so I think most users making these plots will be familiar with that point, but for new users I mentioned in the docstring `:param (list or array) x: 1 dimensional, evenly spaced list or array` . It's not ok to blindly trust that though so this PlotlyError (line 1671) is supposed to check that the spacing is even:
'for index in range(len(self.x)-1):
+            if (self.x[index + 1]-self.x[index])-(self.x[1]-self.x[0]) > .0001:
+                raise exceptions.PlotlyError("x must be a 1 dimensional"
+                                             " evenly spaced array")'

Does that make sense? 
Or maybe my validation method is not in the right place?

theengineear: yeah i think that validation should be factored out and would be good. i don't *think* that computation should ever take a *relatively long time* compared to the draw time for large plots, so it might be nice to do upfront in validation? I'm mostly just pointing it out to make sure that the way you've got it is intentional though. =)

plotly/tools.py
chriddyp: @cldougl - if you're using sublime, i can show you how to set-up pep8 validation on your editor - I've set mine up to be extremely annoying if I don't adhere to the style.

plotly/tools.py
chriddyp: grammar goat?

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        #Create quiver
+        quiver = TraceFactory.create_quiver(x, y, u, v)
+
+        # Plot
+        fig=Figure()
+        fig['data'].append(quiver)
+        py.iplot(fig, filename='quiver')
+        ```
+
+        Example 3:
```
chriddyp: docstring :dog: says: sometimes it's helpful to give little "titles" to these examples: https://github.com/plotly/python-api/blob/master/plotly/grid_objs/grid_objs.py#L29-L40

So, for yours, it might be:

- `Example 1: Trivial quiver`
- `Example 2: Quiver plot using meshgrid`
- `Example 3: Styling the quiver plot`

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        # Add data
+        x, y = np.meshgrid(np.arange(-np.pi, math.pi, .5),
+                           np.arange(-math.pi, math.pi, .5))
+        u = np.cos(x)*y
+        v = np.sin(x)*y
+
+        # Create quiver
+        quiver = TraceFactory.create_quiver(x, y, u, v, scale=.2,
+                                            arrow_scale=.3, angle=math.pi/6,
+                                            line=Line(color='purple', width=1))
```
chriddyp: the beautiful :bear: says: every example of a chart should be pretty: maybe change this optional argument from `line=Line(color='purple', width=1)` to `name='Wind Velocity', line=Line(width=1)`. the default CSS colors (e.g. purple) are notoriously hard on the eyes! (http://clrs.cc/)

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        :param (list or array) x: x coordinates of the arrow locations
+        :param (list or array) y: y coordinates of the arrow locations
+        :param (list or array) u: x components of the arrow vectors
+        :param (list or array) v: y components of the arrow vectors
+        :param (float in [0,1]) scale: scales size of the arrows(ideally to
+            avoid overlap). Default = .1
+        :param (float in [0,1]) arrow_scale: value multiplied to length of barb
+            to get length of arrowhead. Default = .3
+        :param (angle in radians) angle: angle of arrowhead. Default = pi/9
+
```
chriddyp: docstring :dog:: should also mention what `kwargs` are, in that they are passed through to `plotly.graph_objs.Scatter` and that the user can learn more about accepted kwargs by calling `help(plotly.graph_objs.Scatter)`

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        # Plot
+        fig=Figure()
+        fig['data'].append(quiver)
+        py.iplot(fig, filename='quiver')
+        ```
+
+        Example 3:
+        ```
+        # Add data
```
chriddyp: docstring :dog: - add `import numpy as np` and `import math` so that the user can easily just copy and paste this example.

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        :param (angle in radians) angle: angle of arrowhead. Default = pi/9
+        :param (float in [0,1]) arrow_scale: value to scale length of arrowhead
+            Default = .09
+
+        :rtype: (trace) returns streamline data
+
+        :raises: (ImportError) will happen if numpy is not installed
+        :raises: (PlotlyError) will happen if x or y are not evenly spaced
+            1 dimmensional lists or arrays
+        :raises: (PlotlyError) will happen if u and v are not the same shape.
```
chriddyp: similar comments as above

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        streamline = TraceFactory.create_streamline(x, y, u, v, arrow_scale= 1)
+
+        # Plot
+        fig=Figure()
+        fig['data'].append(streamline)
+        py.iplot(fig, filename='streamline')
+        ```
+
+        Example 2:
+        # from http://nbviewer.ipython.org/github/barbagroup/AeroPython
```
chriddyp: move this line up to the title of the example (the `#` is unnecessary)

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+            raise ValueError("arrow_scale must be > 0")
+        if (len(self.x) != len(self.y) or len(self.u) != len(self.v) or
+           len(self.x) != len(self.u)):
+            raise exceptions.PlotlyError("x, y, u, and v should all be the"
+                                         " same length (or size if ndarray)")
+
+    def scale_uv(self):
+        """
+        Scales u and v. u and v are added to x and y to get the
+        endpoints of the arros so a smaller scale value will
```
chriddyp: :goat: *arros

plotly/tools.py
```diff
@@ -1269,3 +1277,628 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
...
+        :raises: (ImportError) If numpy is not imported.
+        :raises: (ValueError) If scale or arrow_scale is < 1.
+        :raises: (ValueError) If density or arrow_scale is < 1.
+        :raises: (PlotlyError) If x and y are not evenly spaced lists or
+            1D arrays.
+        :raises: (PlotlyError) If u and v are not the same shape.
+        """
+        if _numpy_imported is False:
+            raise ImportError("To use TraceFactory.create_streamline()"
+                              " please import numpy as np")
```
chriddyp: this isn't strictly correct, the user just needs to have numpy *available*, not necessarily imported. Instead, maybe: `"TraceFactor.create_streamline requires numpy."`

plotly/tools.py
theengineear: yup :smiley_cat: https://github.com/plotly/dev-docs/blob/master/basics/code-review-checklist.md#code-review-animals

plotly/tools.py
cldougl: @chriddyp yes please! - I'm using sublime and thought I had a pep8 validation set up- it's catching some stuff but not everything

plotly/tools.py
cldougl: @theengineear it's expected : )  I got as far as removing numpy from the rk4 method, but it didn't seem very efficient to remove numpy from this type of plot.

theengineear: cool! thanks!

PR #233, opened 2015-06-17
Opt out of special functionality

theengineear:
If validation is off, we now don’t change things to graph objects. If
the user attempts to do certain things, a validation error will likely
pop up. But the normal creation of plots and things shouldn’t cause
issues.

### DIFF
--- a/plotly/graph_objs/graph_objs.py
+++ b/plotly/graph_objs/graph_objs.py
@@ -33,6 +33,7 @@ from plotly.graph_objs.graph_objs_tools import (
 
 from plotly import exceptions
 from plotly import utils
+from plotly.session import get_session_plot_options
 
 import copy
 import sys
@@ -67,6 +68,11 @@ class PlotlyList(list):
 
     def __init__(self, *args):
         super(PlotlyList, self).__init__(*args)
+
+        if not get_session_plot_options().get('validate', True):
+            warnings.warn('Validation off. Defaulting to plain lists/dicts.')
+            return
+
         if args and isinstance(args[0], dict):
             raise exceptions.PlotlyListEntryError(
                 obj=self,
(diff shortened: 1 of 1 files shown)


theengineear:
@chriddyp , got a better idea of how to do this for now? until we get graph reference synced, we want to not raise validation errors.

I didn't want it to be *default*, but it's quite simple to turn them off now:

`py.update_plot_options(validate=False)`. This will now not validate *anything*. You just need to have that once at the top of your script/notebook.

//cc @jackparmer

jackparmer:
Is this going to be a non-blocking a warning instead of a blocking error? It would annoying to have to do this:

`py.update_plot_options(validate=False)`

Every time you make a map.

jackparmer:
Also, can the warning say something to the tune of: "Many maps and 3d features do not yet have syntax validation - These warnings are normal if you're making a map or 3d graph." Its pretty confusing and embarrassing to demo otherwise.

theengineear:
@jackparmer , yup, i can change the warning to be less embarrassing-sounding for sure. Or, would you rather there was *no* warning? I don't really mind since it's set by the user anyhow.

I just really don't want to turn off the errors by default. Then, we can't ever turn them back on because it'd be a massive backwards-incompatible change :(

The way it is now, you have to run `py.update_plot_options(validate=False)` at the beginning of the notebook.

It would take a fair amount of work to *safely* short-circuit all the errors. We can't just take the errors out because then we wouldn't know if we were ever in a stable state or not to run some function.

The *other* thing that we could easily do is allow folks to just set this in their config file (we're planning on doing it anyways).

That way you wouldn't here a peep when you're demoing, but you also won't know if you typed something in wrong.

jackparmer:
Ok, if its a big hassle to make

PR #236, opened 2015-06-25
Plotly Offline

chriddyp:


plotly/offline.py
```diff
@@ -0,0 +1,149 @@
...
+    plotlyjs_path = os.path.expanduser(
+        os.path.join(*'~/.plotly/plotlyjs'.split('/')))
+
+    if not os.path.exists(plotlyjs_path):
+        os.makedirs(plotlyjs_path)
+
+    res = requests.get(download_url + '/sourcefiles.json')
+    res.raise_for_status()
+
+    with open(os.path.join(plotlyjs_path, 'sourcefiles.json'), 'w') as f:
```
etpinard: what are these `sourcefiles.json` ?

plotly/offline.py
```diff
@@ -0,0 +1,149 @@
...
+                        'Questions? support@plot.ly.'
+                        .format(source_path=sourcefiles_path))
+
+    sourcefiles = json.load(open(sourcefiles_path))['files']
+
+    # Include all of the files in the dependencies folder
+    scriptpaths = [os.path.join(plotlyjs_dir, *sourcefile.split('/'))
+                   for sourcefile in sourcefiles]
+
+    unrequirejs = ('<script type="text/javascript">'
```
etpinard: package of the week

plotly/offline.py
```diff
@@ -0,0 +1,149 @@
...
+        pass
+
+    plotdivid = uuid.uuid4()
+    jdata = json.dumps(data, cls=utils.PlotlyJSONEncoder)
+    jlayout = json.dumps(layout, cls=utils.PlotlyJSONEncoder)
+
+    script = '\n'.join([
+        'Plotly.plot("{id}", {data}, {layout}).then(function() {{',
+        '    $(".{id}.loading").remove();',
+        '    $(".link--embedview").text("{link_text}");'
```
etpinard: :+1:

etpinard:
This is the best thing that happen to IPython since IPython notebook.

plotly/offline.py
chriddyp: they contain the list of plotly.js dependencies:
```
{
    "files": ["dependencies/d3.v3.min.js", "dependencies/jquery-latest.js", "dependencies/typedarray.js", "plotly.min.js"]
}
```

that way, if we change the dependencies in the future we don't have to change anything on the python lib side. they just explain which files are necessary to include and to download

chriddyp:
please test and review @aneda @cldougl @theengineear ! thx!

plotly/offline.py
etpinard: LOVE IT

chriddyp:
@BRONSOLO @theengineear - did you guys change the folders endpoint on prod? this branch and master are failing in the `test_duplicate_folders` function.

theengineear:
@chriddyp , it's a minor break, looks like ~~409~~ 400 for the particular issue. I just checked out api tests, the 400 is what we're currently expecting. Mind if we change this to a TestCase while we're in here : D ?:

```
"""
test_meta:
==========

A module intended for use with Nose.

"""
import random
import string
from unittest import TestCase

import plotly.plotly as py
from plotly.exceptions import PlotlyRequestError


class FolderAPITestCase(TestCase):

    def setUp(self):
        py.sign_in('PythonTest', '9v9f20pext')

    def _random_filename(self):
        random_chars = [random.choice(string.ascii_uppercase)
                        for _ in range(5)]
        unique_filename = 'Valid Folder'+''.join(random_chars)
        return unique_filename

    def test_create_folder(self):
        try:
            py.file_ops.mkdirs(self._random_filename())
        except PlotlyRequestError as e:
            self.fail('Expected this *not* to fail! Status: {}'
                      .format(e.status_code))

    def test_create_nested_folders(self):
        first_folder = self._random_filename()
        nested_folder = '{0}/{1}'.format(first_folder, self._random_filename())
        try:
            py.file_ops.mkdirs(nested_folder)
        except PlotlyRequestError as e:
            self.fail('Expected this *not* to fail! Status: {}'
                      .format(e.status_code))

    def test_duplicate_folders(self):
        first_folder = self._random_filename()
        try:
        except PlotlyRequestError as e:
            self.assertTrue(400 <= e.status_code < 500)
        else:
            self.fail('Expected this to fail!')

```

theengineear:
This is the only reference to a 409 i can find:

https://github.com/plotly/python-api/blob/master/plotly/plotly/plotly.py#L774

I don't think the library is breaking because of this test :)

Also, I tested the above test changes in python 2.7, i don't *think* i used TestCase stuff that wasn't around in 2.6... but maybe.

plotly/offline.py
```diff
@@ -0,0 +1,149 @@
...
+from __future__ import absolute_import
+
+from plotly import utils
+import uuid
+import json
+import os
+import requests
+
+
+def download_plotlyjs(download_url, plotlyjs_dir='~/.plotly/plotlyjs'):
```
theengineear: Can we get rid of the `download_url` param and just create a streambed url that we *know* and can authenticate at?

Then, we can just write a function to decide whether the requesting user has access?

Additionally, why do we need the `plotlyjs_dir` arg? is there a good reason for folks to need to change this? Also, i think it's not used in the func?

(obviously, the download_url is perfect for the *current* dev though :))

theengineear: Finally, if this didn't require any arguments, we could just call it the first time a user tries to use it, which would mean they don't need to know about it.

plotly/offline.py
theengineear: could this be a bundle we build in streambed and then a single download file for folks?

plotly/offline.py
```diff
@@ -0,0 +1,149 @@
...
+
+def init_notebook_mode(plotlyjs_dir='~/.plotly/plotlyjs'):
+    # TODO: check if ipython is available...?
+    from IPython.display import HTML, display
+
+    # Split and join the install path for cross-OS directories
+    plotlyjs_dir = os.path.expanduser(os.path.join(*'~/.plotly/plotlyjs'.split('/')))
+    sourcefiles_path = os.path.join(plotlyjs_dir, 'sourcefiles.json')
+    if not os.path.exists(sourcefiles_path):
+        raise Exception('Plotly Offline configuration file at {source_path} '
```
theengineear: PlotlyError?

plotly/offline.py
```diff
@@ -0,0 +1,149 @@
...
+                        'is not found.\n'
+                        'If you have a Plotly Offline license, then try '
+                        'running plotly.offline.configure_offline(url) '
+                        'with a licensed download url.\n'
+                        "Don't have a Plotly Offline license?"
```
theengineear: missing whitespace or newline after the `?`

plotly/offline.py
```diff
@@ -0,0 +1,149 @@
...
+    unrequirejs = ('<script type="text/javascript">'
+                   'require=requirejs=define=undefined;</script>')
+
+    display(HTML(
+        unrequirejs +
+        '\n'.join([
+            '<script type="text/javascript">' + open(script).read() +
+            '</script>' for script in scriptpaths
+        ])
+    ))
```
theengineear: *Could* we just have a `bundle.js` thing here so that we can just do:

```
plotly_bundle = open(bundle_path).read()
display(HTML(plotly_bundle))
```

plotly/offline.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly'):
```
theengineear: :tiger2: , can we define a variable in `session.py` the defines which *mode* we're in so that folks can call iplot as necessary? perhaps a plot_option, which would even make it overridable?

theengineear:
@chriddyp , playing around with this now, nice work!

theengineear:
I noticed that when you open notebooks back up, nothing seems to run until the plot is drawn:

image

theengineear:
that said, super cool that opening them backup just *works*

image

theengineear:
nice! plot loads up in nbviewer, that's super solid!

http://nbviewer.ipython.org/gist/theengineear/3b986f62836453d470ea

plotly/offline.py
chriddyp: :+1:

plotly/offline.py
chriddyp: github.com/chriddyp/unrequire.js

plotly/offline.py
chriddyp: @etpinard - funny thing: this actually gets re-drawn by plotly.js when you zoom, pan.

plotly/offline.py
chriddyp: great idea, going to punt :athletic_shoe:  :football: this one to a future release

plotly/offline.py
chriddyp: :+1:

plotly/offline.py
chriddyp: as discussed today, gonna keep it explicitly in this namespace and keep `plotly.plotly` reserved for network comm

chriddyp:
- [ ] bundle it up
- [ ] test in old ipython notebook
- [ ] investigate moving plotting calls to be async to make the notebook load quick

chriddyp:
@etpinard - due to some conflicts in ipython 2.4.1, I'm going to bundle up plotly.js w/out jquery 1.8.3 and instead rely on ipython's 2.0.3 version of jquery. 

here's the test suite on the different versions:

juypter, jquery 1.8.3 (plotlyjs supplied jquery): http://nbviewer.ipython.org/urls/gist.githubusercontent.com/chriddyp/dfa984ec08e9d627ec2b/raw/9dedcb09d658040439a5088b5fd4b547e39836a3/offline-test-suite-juypter-w-jquery.ipynb
jupyter, jquery 2.0.3 (ipython supplied jquery): http://nbviewer.ipython.org/urls/gist.githubusercontent.com/chriddyp/dfa984ec08e9d627ec2b/raw/adc8c38f9f977ee72ea3a2321700ec42f54e12df/offline-test-suite-jupyter-wout-jquery.ipynb
ipython 2.4.1, jquery 2.0.3 (ipython supplied jquery): http://nbviewer.ipython.org/urls/gist.githubusercontent.com/chriddyp/dfa984ec08e9d627ec2b/raw/e90c15f1e17a050cd94040124a200b9c8e3ed3d4/offline-test-suite-ipython24-wout-query.ipynb

at first glance, versions look and interact the same

chriddyp:
@etpinard try zooming on the `2dhistogram_contour_subplots` example.. do you see the black background? (happens on all versions)

plotly/offline.py
chriddyp: :koala:

chriddyp:
@theengineear nice tinkerbell! throwback!

chriddyp:
alright, @theengineear et all, I'm pretty happy with this. can i get a :dancer:?

chriddyp:
@theengineear - oppositions to adding ipython to the test-core circle suite?

theengineear:
Not opposed to adding ipython to core-tests, *but*, then we need to add it as a requirement in `setup.py`.

I'd suggest we go with `ipython[all]` or at least `ipython[notebook]`.

I just did a quick search for ipython tests and it seems that there's some support out there for testing in a notebook: https://pypi.python.org/pypi/pytest-ipynb

I haven't really tried we might be able to test everything we need to *outside* of a notebook environment. Again, just haven't really tried.

chriddyp:
yeah, nvm, I thought that the test_optional wasn't being run by circle... but it is! @theengineear - can I get a :dancer: ?

plotly/offline.py
```diff
@@ -0,0 +1,178 @@
...
+    os.path.join(*'~/.plotly/plotlyjs'.split('/')))
+PLOTLY_OFFLINE_BUNDLE = os.path.join(PLOTLY_OFFLINE_DIRECTORY,
+                                     'plotly-ipython-offline-bundle.js')
+
+__PLOTLY_OFFLINE_INITIALIZED = False
+
+
+def download_plotlyjs(download_url):
+    if not os.path.exists(plotlyjs_path):
+        os.makedirs(plotlyjs_path)
```
theengineear: can you check for write permissions and gracefully exit here? I think the tools.py module defines it:
https://github.com/plotly/python-api/blob/master/plotly/tools.py#L61-76

`_file_permissions`

plotly/offline.py
```diff
@@ -0,0 +1,178 @@
...
+
+    res = requests.get(download_url)
+    res.raise_for_status()
+
+    with open(PLOTLY_OFFLINE_BUNDLE, 'wb') as f:
```
theengineear: binary? is this necessary?

plotly/offline.py
```diff
@@ -0,0 +1,178 @@
...
+        'code to make your first offline graph:',
+        '',
+        'import plotly',
+        'plotly.offline.init_notebook_mode() '
+        '# run at the start of every ipython notebook',
+        'plotly.offline.iplot([{"x": [1, 2, 3], "y": [3, 1, 6]}])'
+    ]))
+
+
+def init_notebook_mode():
```
theengineear: is this function necessary (non-blocking for now, but just out of curiosity)? since `iplot` is really `offline.iplot`, could we just run this as needed?

plotly/offline.py
```diff
@@ -0,0 +1,178 @@
...
+    if not __PLOTLY_OFFLINE_INITIALIZED:
+        raise PlotlyError('\n'.join([
+            'Plotly Offline mode has not been initialized in this notebook. '
+            'Run: ',
+            '',
+            'import plotly',
+            'plotly.offline.init_notebook_mode() '
+            '# run at the start of every ipython notebook',
+        ]))
+    from IPython.display import HTML, display
```
theengineear: You could protect this with our `_ipython_imported`: https://github.com/plotly/python-api/blob/master/plotly/tools.py#L35-40

plotly/offline.py
```diff
@@ -0,0 +1,178 @@
...
+    else:
+        data = figure_or_data
+        layout = {}
+
+    width = layout.get('width', '100%')
+    height = layout.get('height', 525)
+    try:
+        float(width)
+        width = str(width) + 'px'
+    except:
```
theengineear: :cow2: i think you don't need such a broad exception clause, can we just look before we leap on this one? or, catch `(ValueError, TypeError)`? OR, use this format perhaps?

```
try:
   float(width)
except:
    pass
else:
    width = str(width) + 'px'
```

plotly/offline.py
chriddyp: yup, for python 3.3: 
image

plotly/tests/test_core/test_offline/test_offline.py
```diff
@@ -0,0 +1,18 @@
...
+
+"""
+from __future__ import absolute_import
+
+import os
+
+import plotly
+
+
+def test_downloading_file_saves_it_to_the_disk():
```
theengineear: :cow2: , might be a good habit to start using `TestCase` instead of just bare tests.

plotly/tests/test_optional/test_offline/test_offline.py
```diff
@@ -0,0 +1,31 @@
...
+from __future__ import absolute_import
+from nose.tools import raises
+import os
+
+
+from plotly.exceptions import PlotlyError
+import plotly
+
+
+@raises(PlotlyError)
```
theengineear: ditto in here, re: https://github.com/plotly/python-api/pull/236/files#r33808672

plotly/tests/test_optional/test_offline/test_offline.py
```diff
@@ -0,0 +1,31 @@
...
+def test_calling_iplot_before_initializing_raises_an_error():
+    plotly.offline.iplot([{'x': [1, 2, 3]}])
+
+
+def test_no_errors_are_raised_when_properly_initializing_offline_mode():
+    plotly.offline.init_notebook_mode()
+    plotly.offline.iplot([{'x': [1, 2, 3]}])
```
theengineear: can we test this response at all?

plotly/version.py
```diff
@@ -1 +1 @@
-__version__ = '1.6.17'
+__version__ = '1.7.0'
```
theengineear: :smile_cat:

theengineear:
non-blocking, are we pushing this up to pypi though? i feel like we might want to just allow folks to download from github master for a bit?

:dancers:

plotly/offline.py
chriddyp: yeah, here's a scenario where embedding `init_notebook_mode` would cause unexpected things:

1 - User calls iplot. we check module level variable, see that plotly.js hasn't been inserted, so we drop it in
2 - User is messing around, so deletes the cell. plotly.js is already in memory, so no big deal, iplot will still work. User calls iplot again, since we've already imported plotly.js, we don't insert again.

3 - User shares or re-opens the notebook. The original output cell that contained plotly.js was deleted from (2), so now Plotly is undefined

similarly if someone calls iplot, embedding plotly.js, and then makes a plot in the cell *above* the first cell - when re-opening the notebook the first cell containing the `Plotly.plot` call will be before plotly.js was embedded. (This could still occur with `init_notebook_mode` - the user could call `init_notebook_mode` on cell 10, then call iplot later on cell 3, but this way is a little more explicit)

I also thought about just calling this function when we `import` plotly.offline (since that'll necessarily be the first thing called). however, we don't want to attempt to import the JS before someone calls `download_plotlyjs`...

it's definitely annoying though, I'm not sure if there is a better way to do it.

plotly/offline.py
chriddyp: I thought about that. What do you think is more helpful, an exception: `Exception() Offline only works in an IPython notebook` or just the natural `ImportError` that will occur when this line is executed?

plotly/offline.py
chriddyp: :+1:

plotly/offline.py
chriddyp: :+1:

plotly/tests/test_optional/test_offline/test_offline.py
chriddyp: returns `None`. We could refactor `iplot` to call some private functions that we could test..

plotly/tests/test_core/test_offline/test_offline.py
chriddyp: mmm

chriddyp:
@theengineear - yeah, I'm gonna push it up to pypi as those openssl errors that bubbled up through the git+ install scares me.

plotly/tests/test_core/test_offline/test_offline.py
chriddyp: done

plotly/offline.py
theengineear: ah dang! yeah, i guess i didn't think it through enough.

theengineear:
thx @chriddyp !

azarezade:
Sorry if I ask trivial question
but how can I setup plotly with offline mode available?

PR #238, opened 2015-06-28
Fix failing test and refact to TestCase.

theengineear:
Attempting to fix the failing test due to a 409 --> 400 in plotly's /folders endpoint.

### DIFF
--- a/plotly/tests/test_core/test_file/test_file.py
+++ b/plotly/tests/test_core/test_file/test_file.py
@@ -5,47 +5,47 @@ test_meta:
 A module intended for use with Nose.
 
 """
-
-from nose.tools import raises
-from nose import with_setup
-
 import random
 import string
-import requests
+from unittest import TestCase
 
 import plotly.plotly as py
-import plotly.tools as tls
 from plotly.exceptions import PlotlyRequestError
 
 
-def _random_filename():
-    random_chars = [random.choice(string.ascii_uppercase) for _ in range(5)]
-    unique_filename = 'Valid Folder'+''.join(random_chars)
-    return unique_filename
+class FolderAPITestCase(TestCase):
 
+    def setUp(self):
+        py.sign_in('PythonTest', '9v9f20pext')
 
-def init():
-    py.sign_in('PythonTest', '9v9f20pext')
+    def _random_filename(self):
+        random_chars = [random.choice(string.ascii_uppercase)
+                        for _ in range(5)]
+        unique_filename = 'Valid Folder'+''.join(random_chars)
+        return unique_filename
 
+    def test_create_folder(self):
+        try:
+            py.file_ops.mkdirs(self._random_filename())
+        except PlotlyRequestError as e:
+            self.fail('Expected this *not* to fail! Status: {}'
+                      .format(e.status_code))
 
-@with_setup(init)
-def test_create_folder():
-    py.file_ops.mkdirs(_random_filename())
+    def test_create_nested_folders(self):
+        first_folder = self._random_filename()
+        nested_folder = '{0}/{1}'.format(first_folder, self._random_filename())
+        try:
+            py.file_ops.mkdirs(nested_folder)
+        except PlotlyRequestError as e:
+            self.fail('Expected this *not* to fail! Status: {}'
+                      .format(e.status_code))
 
-
-@with_setup(init)
-def test_create_nested_folders():
-    first_folder = _random_filename()
-    nested_folder = '{0}/{1}'.format(first_folder, _random_filename())
-    py.file_ops.mkdirs(nested_folder)
-
-
-@with_setup(init)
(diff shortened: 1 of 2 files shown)


theengineear:
@chriddyp once tests are passing again, i'm going to go ahead and merge this into master so you can pull the fix into your offline mode branch.

theengineear:
@chriddyp seems to be a random failure in python 2.6? also, looks like that version isn't supported at all anymore:

`DeprecationWarning: Python 2.6 is no longer supported by the Python core team, please upgrade your Python.`

Is is time to ditch 2.6?! w00t?

chriddyp:
huh? did I write that?

theengineear:
Ah, I think it might be related to the crypto stuff?

https://github.com/pyca/cryptography/blob/master/src/cryptography/__init__.py

chriddyp:
oh hm.. uh.. I guess we'll have to follow suit? https://github.com/pyca/cryptography/commit/0857894306d26a655fe06326cd244b4e182c94bf#commitcomment-11225855

plotly/tests/test_core/test_file/test_file.py
```diff
@@ -5,47 +5,47 @@
...
+    def test_duplicate_folders(self):
+        first_folder = self._random_filename()
         py.file_ops.mkdirs(first_folder)
-    except PlotlyRequestError as e:
-        if e.status_code != 409:
-            raise e
+        try:
+            py.file_ops.mkdirs(first_folder)
+        except PlotlyRequestError as e:
+            self.assertTrue(400 <= e.status_code < 500)
```
chriddyp: isn't it sort of helpful to know which error code was thrown? I can imagine writing code against a specific error code (e.g. "if conflict, then the folder exists, so don't make a new folder")

chriddyp:
@theengineear did you see this guy?

```
FAIL: plotly.tests.test_optional.test_utils.test_utils.test_pandas_json_encoding
  File "/home/ubuntu/.pyenv/versions/2.6.8/lib/python2.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/ubuntu/python-api/plotly/tests/test_optional/test_utils/test_utils.py", line 228, in test_pandas_json_encoding
    pd.NaT, np.NaN, np.Inf]))
  File "/home/ubuntu/.pyenv/versions/2.6.8/lib/python2.6/site-packages/pandas/util/testing.py", line 701, in assert_series_equal
    assert_attr_equal('name', left, right)
  File "/home/ubuntu/.pyenv/versions/2.6.8/lib/python2.6/site-packages/pandas/util/testing.py", line 552, in assert_attr_equal
    assert_equal(left_attr,right_attr,"attr is not equal [{0}]" .format(attr))
  File "/home/ubuntu/.pyenv/versions/2.6.8/lib/python2.6/site-packages/pandas/util/testing.py", line 533, in assert_equal
    assert a == b, "%s: %r != %r" % (msg.format(a,b), a, b)
AssertionError: attr is not equal [name]: 'col 1' != None
```

theengineear:
re: https://github.com/plotly/python-api/pull/238#issuecomment-116934789

yeah, i think that's a python 2.6 thing? i ran all of our tests locally without any failures.

theengineear:
i'd happily deprecate the use of 2.6. supporting 2.7, 3.3, and 3.4 is much simpler. 2.6 is like 7 years old now. perhaps this can be in our change to 1.7?

plotly/tests/test_core/test_file/test_file.py
theengineear: yeah, i could definitely see that, it's just that i don't think the python library should couple so closely to streambed. this holds for streaming too, we should try and be as transparent about interacting with the plotly apis as possible. i think :)

Also, the `4xx` is like "hey, you did something wrong, you'll need to change your code to fix it", which is much different from `5xx`, "oops, our bad, you might want to try that again later".

theengineear:
@chriddyp if you're alright with the `400 <= X < 500`, i think this is good to merge, otherwise, we'll need to (a) make a change in streambed or (b) require this to be 400, which might break in the near future since we're working on a lot of api stuff (hint hint, can we keep it as is :smile_cat: ?)

chriddyp:
alright, fair enough @theengineear ! but in the future, once the api is out of beta...!

theengineear:
haha, noted --> https://github.com/plotly/python-api/issues/240

chriddyp:
@theengineear - dropped 2.6 support in this commit, as not being able to import ipython was biting me in this test: https://github.com/plotly/python-api/commit/08a86608e3cbaedfe054cafd9d396c43ee074c77

theengineear:
@chriddyp , wonderful, i may make a pr to clean up some other things related to this drop then :)

PR #239, opened 2015-06-30
Plotly proxy auth

BRONSOLO:
@chriddyp @theengineear. This allows enterprise users working behind a proxy to specify `Plotly-Authorization` header as a `basic auth` header for our `API v2`. This frees up the `Authorization` header for the initial proxy layer. For now, it is up to the server admin to handle converting the `Plotly-Authorization` header to the expected `Authorization` header for basic auth once authorized over the proxy.  In the future, we could simply accept both `Plotly-Authorization` and `Authorization` headers in our back end so the server admin won't need to do anything. I added a flag in the config file for `plotly_proxy_authorization`. 

todo: 
- [x] extend the tests

### DIFF
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -1212,7 +1212,7 @@ class _api_v2:
             headers['authorization'] = 'Basic ' + encoded_proxy_auth
             headers['plotly-authorization'] = 'Basic ' + encoded_api_auth
         else:
-            headers['authorization'] = 'Basic ' + encoded_api_auth,
+            headers['authorization'] = 'Basic ' + encoded_api_auth
 
         return headers
 
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -1201,18 +1201,14 @@ class _api_v2:
             username, api_key))).decode('utf8')
 
         headers = {
-            'authorization': 'Basic ' + encoded_api_auth,
             'plotly-client-platform': 'python {0}'.format(version.__version__)
         }
 
         if get_config()['plotly_proxy_authorization']:
-
-            # custom enterprise + proxy authentication
             proxy_username = credentials['proxy_username']
-            proxy_passwd = credentials['proxy_passwd']
+            proxy_password = credentials['proxy_password']
             encoded_proxy_auth = base64.b64encode(six.b('{0}:{1}'.format(
-                proxy_username, proxy_passwd))).decode('utf8')
-
+                proxy_username, proxy_password))).decode('utf8')
             headers['authorization'] = 'Basic ' + encoded_proxy_auth
             headers['plotly-authorization'] = 'Basic ' + encoded_api_auth
 
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -583,23 +583,9 @@ class image:
                 "https://plot.ly/python/static-image-export/"
             )
 
-        credentials = get_credentials()
-        validate_credentials(credentials)
-        username, api_key = credentials['username'], credentials['api_key']
-        encoded_api_auth = base64.b64encode(six.b('{0}:{1}'.format(
-            username, api_key))).decode('utf8')
-
-        # custom enterprise + proxy authentication
-        proxy_username = credentials['proxy_username']
-        proxy_passwd = credentials['proxy_passwd']
-        encoded_proxy_auth = base64.b64encode(six.b('{0}:{1}'.format(
-            proxy_username, proxy_passwd))).decode('utf8')
-
-        headers = {'Plotly-Version': version.__version__,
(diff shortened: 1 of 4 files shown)


plotly/tools.py
```diff
@@ -48,11 +48,15 @@ def warning_on_one_line(message, category, filename, lineno, file=None, line=Non
 # this sets both the DEFAULTS and the TYPES for these items
 _FILE_CONTENT = {CREDENTIALS_FILE: {'username': '',
                                     'api_key': '',
+                                    'proxy_username': '',
+                                    'proxy_passwd': '',
```
chriddyp: could we change to `proxy_password`?

plotly/tools.py
```diff
@@ -48,11 +48,15 @@ def warning_on_one_line(message, category, filename, lineno, file=None, line=Non
...
+                                    'api_key': '',
                                     'stream_ids': []},
                  CONFIG_FILE: {'plotly_domain': 'https://plot.ly',
                                'plotly_streaming_domain': 'stream.plot.ly',
                                'plotly_api_domain': 'https://api.plot.ly',
-                               'plotly_ssl_verification': True}}
+                               'plotly_ssl_verification': True,
+                               'plotly_proxy_authorization': False}}
```
chriddyp: is there some terminology we can use to specify which kind of proxy this is? I can imagine that some folks might stumble upon this because they have a company proxy which they need to authenticate past just to get outside of their network. like, `plotly_reverse_proxy_authorization`?

plotly/tools.py
```diff
@@ -99,7 +103,8 @@ def ensure_local_plotly_files():
 
 ### credentials tools ###
 
-def set_credentials_file(username=None, api_key=None, stream_ids=None):
+def set_credentials_file(username=None, api_key=None, proxy_username=None,
+                         proxy_passwd=None, stream_ids=None):
     """Set the keyword-value pairs in `~/.plotly_credentials`.
```
chriddyp: add some docstrings describing the keywords here

plotly/tools.py
```diff
@@ -99,7 +103,8 @@ def ensure_local_plotly_files():
 
 ### credentials tools ###
 
-def set_credentials_file(username=None, api_key=None, stream_ids=None):
+def set_credentials_file(username=None, api_key=None, proxy_username=None,
+                         proxy_passwd=None, stream_ids=None):
```
chriddyp: move `proxy_username` and `proxy_password` to the very end for backwards compatibility - some folks might be setting up their creds as positional arguments

plotly/plotly/plotly.py
```diff
@@ -1199,17 +1194,30 @@ def api_url(cls, resource):
...
-            six.b('{0}:{1}'.format(un, api_key))
-        ).decode('utf8')
+        username, api_key = credentials['username'], credentials['api_key']
+        encoded_api_auth = base64.b64encode(six.b('{0}:{1}'.format(
+            username, api_key))).decode('utf8')
 
-        return {
-            'authorization': 'Basic ' + encoded_un_key_pair,
+        headers = {
+            'authorization': 'Basic ' + encoded_api_auth,
```
chriddyp: since this is getting set in the following `if` statement, maybe remove this line and put it in an `else`?

chriddyp:
@BRONSOLO sweet dude! think we should add it to `sign_in` too, for consistency?

BRONSOLO:
@chriddyp boom - thx for the comments/suggestions. Yeah, I'll add it to `sign_in` + include some tests and ping you for another pass.

plotly/tools.py
BRONSOLO: This dev should work for both reverse and forward proxies. I'm tempted to keep it as `plotly_proxy_authorization`, though maybe I'm missing the source of potential confusion. Let me know! I can change it to `plotly_reverse_proxy_authorization` . nbd!

BRONSOLO:
@chriddyp could you take another look through this guy?

plotly/tools.py
```diff
@@ -145,9 +163,16 @@ def reset_credentials_file():
 def set_config_file(plotly_domain=None,
                     plotly_streaming_domain=None,
                     plotly_api_domain=None,
-                    plotly_ssl_verification=None):
+                    plotly_ssl_verification=None,
+                    plotly_proxy_authorization=None):
     """Set the keyword-value pairs in `~/.plotly/.config`.
 
+    :param (str) plotly_domain: ex - https://plot.ly
+    :param (str) plotly_streaming_domain: ex - https://stream.plot.ly
```
chriddyp: I think this is usually set as `stream.plot.ly`

chriddyp:
other than that comment, looks good to me! :dancer:

plotly/tools.py
BRONSOLO: totally - thanks.

PR #243, opened 2015-07-02
first pass at rmving load_plotly

chriddyp:
@theengineear check it!


plotly/offline/offline.py
```diff
@@ -142,8 +122,8 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly'):
     if show_link is False:
         link_text = ''
 
-    plotly_platform_url = py.get_config().get('plotly_domain',
```
chriddyp: struggled with relative imports to `plotly.plotly`...

plotly/utils.py
```diff
@@ -356,3 +356,8 @@ def is_source_key(key):
         return True
     else:
         return False
+
+PLOTLY_OFFLINE_DIRECTORY = plotlyjs_path = os.path.expanduser(
+    os.path.join(*'~/.plotly/plotlyjs'.split('/')))
+PLOTLY_OFFLINE_BUNDLE = os.path.join(PLOTLY_OFFLINE_DIRECTORY,
+                                     'plotly-ipython-offline-bundle.js')
```
chriddyp: .. so that `exceptions.py` *and* `offline.py` can use these variables without circular imports

plotly/offline/__init__.py
```diff
@@ -5,6 +5,5 @@
 """
 from . offline import (
     download_plotlyjs,
-    init_notebook_mode,
```
chriddyp: @theengineear !

plotly/offline/offline.py
```diff
@@ -10,41 +10,37 @@
 import os
 import requests
 
-import plotly.plotly as py
```
theengineear: do you need to import __absolute_import__ from future?

plotly/offline/offline.py
```diff
@@ -142,8 +122,8 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly'):
     if show_link is False:
         link_text = ''
 
-    plotly_platform_url = py.get_config().get('plotly_domain',
-                                              'https://plot.ly')
+    plotly_platform_url = tools.get_config_file().get('plotly_domain',
```
theengineear: this should grab from `session` or else there can be a mismatch after a user sign-in. Just use `py.get_config()` if you can't grab session.

plotly/offline/offline.py
```diff
@@ -185,3 +165,12 @@ def plot():
     """ Configured to work with localhost Plotly graph viewer
     """
     raise NotImplementedError
+
+
+try:
```
theengineear: \o/ nice. i dig.

plotly/offline/offline.py
chriddyp: ahh sweet.

plotly/offline/offline.py
```diff
@@ -185,3 +165,12 @@ def plot():
...
+    _init_notebook_mode()
```
chriddyp: @theengineear - note, this is a little bit scary because this gets run *every* time someone does:

`import plotly`

so we really need to make sure that no unexpected errors will be thrown here. Note that I'm catching the Ipython `ImportError` below, and the `PlotlyOfflineNotFound` error

theengineear: you need to wrap those in `()` i believe: `except (exc1, ecx1, ..):`

plotly/offline/offline.py
```diff
@@ -55,21 +51,13 @@ def init_notebook_mode():
...
-                          'Questions? support@plot.ly.'
-                          .format(source_path=_PLOTLY_OFFLINE_BUNDLE))
+    if not os.path.exists(PLOTLY_OFFLINE_BUNDLE):
+        raise PlotlyOfflineNotFound
 
     global __PLOTLY_OFFLINE_INITIALIZED
     __PLOTLY_OFFLINE_INITIALIZED = True
     display(HTML('<script type="text/javascript">' +
-                 open(_PLOTLY_OFFLINE_BUNDLE).read() + '</script>'))
+                 open(PLOTLY_OFFLINE_BUNDLE).read() + '</script>'))
```
chriddyp: @theengineear w.r.t. https://github.com/plotly/python-api/pull/243/files#r33820323, I'm assuming that that `try` `except` is going to catch everything that could possible go wrong in here. I'm what errors `os.path.exists` might bubble up

plotly/offline/offline.py
theengineear: also, that's what test suites are for! let's make sure that those things aren't cached in circle and we'll be fine!

chriddyp: and a funny caveat is that now this bundle will be shared by anyone who has offline mode, regardless of whether they made an offline graph or not!

plotly/offline/offline.py
chriddyp: https://github.com/plotly/python-api/pull/243/files#diff-3465b0f2dbd6f036db2c0713d3f40872R6

theengineear:
Assuming that this is pretty outdated at this point. Feel free to reopen.

PR #244, opened 2015-07-03
Add subplot titles

cldougl:
PR titles for each subplot

### DIFF
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -586,6 +586,31 @@ def make_subplots(rows=1, cols=1,
     fig['data'] += [Scatter(x=[1,2,3], y=[2,1,2])]
     fig['data'] += [Scatter(x=[1,2,3], y=[2,1,2], xaxis='x2', yaxis='y2')]
 
+    Example 5:
+    # include subplot titles
+    fig = tools.make_subplots(rows=2, subplot_titles=('Plot 1','Plot 2'))
+
+    This is the format of your plot grid:
+    [ (1,1) x1,y1 ]
+    [ (2,1) x2,y2 ]
+
+    fig['data'] += [Scatter(x=[1,2,3], y=[2,1,2])]
+    fig['data'] += [Scatter(x=[1,2,3], y=[2,1,2], xaxis='x2', yaxis='y2')]
+
+    Example 6:
+    # Include subplot title on one plot (but not all)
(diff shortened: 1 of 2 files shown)


plotly/tools.py
```diff
@@ -916,11 +954,13 @@ def _add_domain(layout, x_or_y, label, domain, anchor, position):
         if isinstance(position, float):
             axis['position'] = position
         layout[name] = axis
+        list_of_domains.append(domain)  # added for subplot titles
 
     # Function pasting x/y domains in layout object (3d case)
     def _add_domain_is_3d(layout, s_label, x_domain, y_domain):
         scene = graph_objs.Scene(domain={'x': x_domain, 'y': y_domain})
         layout[s_label] = scene
+        list_of_domains.append(domain)  # added for subplot titles
```
etpinard: @cldougl You'll have to change this in order to get the tests to pass.

As annotations don't work with 3d yet, I would simply skip 3d domains.

cldougl:
@etpinard thanks! it's passing now
cc @chriddyp @theengineear @aneda

etpinard:
@cldougl great! 

Better we :dancer:, could you add two or three tests in [here](https://github.com/plotly/python-api/blob/master/plotly/tests/test_core/test_tools/test_make_subplots.py) ?

etpinard:
Great job!

cldougl:
@theengineear it gets a little too big- do you think I should scale it down when there are more plots or give the user the option to define subplot title size?

screen shot 2015-07-06 at 11 41 54 am

plotly/tools.py
```diff
@@ -1129,6 +1168,53 @@ def _pad(s, cell_len=cell_len):
...
+        if not subplot_titles[index]:
+            pass
+        else:
+            plot_titles.append({'y': subtitle_pos_y[index],
+                                'xref': 'paper',
+                                'x': subtitle_pos_x[index],
+                                'yref': 'paper',
+                                'text': subplot_titles[index],
+                                'showarrow': False,
+                                'font': graph_objs.Font(size=18),
```
chriddyp: @cldougl - I think maybe the subplot titles should be a font-size in between the axes titles (14px) and the the main title (17px): 15px.

chriddyp:
@cldougl - we should dynamically scale down the subplot titles unless we do the same with the axes titles. let's :athletic_shoe:  :football: (punt) that one to the future

PR #248, opened 2015-07-03
Remove optional requests [security] package.

chriddyp:
requests[security] installs pyOpenSSL which installs the latest cryptography.
The latest cryptography fails to install on many mac environments with a traceback like:
`dlopen(/Users/Chelsea/my_project_folder/venv/lib/python2.7/site-packages/cryptography/_Cryptography_cffi_a269d620xd5c405b7.so, 2): Library not loaded: libssl.1.0.0.dylib`

When [security] is included, anyone upgrading plotly with the common:
`$ pip install plotly --upgrade`
or installing plotly fresh from a virtual environment will install the latest cryptography library
which has a high liklihood of breaking for mac users. Once cryptography is broken, everything
that depends on it, like pip, is also broken, which is difficult to get out of!

Hopefully the installation for cryptography on mac is remediable, and we can safely let users
upgrade their security when installing plotly.

cc @cldougl @theengineear @etpinard @aneda


cldougl:
:dancer:

chriddyp:
* cryptography isn't actually the issue, looks like it is pyOpenSSL. Here is the issue: https://github.com/pyca/pyopenssl/issues/290

PR #252, opened 2015-07-07
Remove python 2.6 support

theengineear:
Since we're no longer officially supporting 2.6, instead of allowing half-broken experiences, just be clear about the change by removing our 2.6 code.

Also, a lot of our imports were pretty haggard, so I fixed them up a bit :)

Also, also, some of the test files were suffering badly with a case of the pep:8ball:s. I couldn't help myself.


theengineear:
@chriddyp throwing this into the 1.7 mix, cool? it's just cleanup.

theengineear:
@chriddyp , i'd love to at least get this cleanup merged in as it might cause conflicts down the road if we don't :)

chriddyp:
@theengineear cool :dancer:

PR #253, opened 2015-07-07
Revamp streaming

theengineear:
Todo:
- [ ] setup a mock streaming server to test?
- [x] make a fully async producer/consumer cycle for streaming
- [x] buffer streamed data so that user knows what data wasn't sent
- [x] implement our suggested linear/exponential back off rates (with eventual exceptions)
- [x] allow reconnect-on for both http error codes _and_ socket exceptions

This borrows from chunked requests, but moves the functionality so we don't need that extra submodule. It also attempts to clean up some of the code and make the reconnecting functionality more flexible.

The `asyncio` module comes standard in 3.4 and can be back ported to 3.3. `trollius` is a stand-in for `asyncio` that can be back ported into 2.7.

Though there's still a fair amount of low-level interaction, this attempts to make the interface a bit higher-level.


theengineear:
@chriddyp thoughts on this? I've been playing around with coroutines in some of my hacking-time recently and thought I'd see if I could make our streaming use 'em!

This streaming setup allows users to write to a queue regardless of connection state. You simply use the `.write` and the background thread writes from the queue to the socket as soon as data is available.

Additionally, it handles heart-beats on it's own instead of allowing our servers to shoot back a 408 code.

theengineear:
We get a fair amount of feedback emails related to streaming, so I thought it might be worth looking into for our 1.7 release.

chriddyp:
woahh! here's a test server and file that I made from chunked requests: https://github.com/chriddyp/chunked_requests/tree/master/test, https://github.com/chriddyp/chunked_requests/blob/master/circle.yml

plotly/exceptions.py
```diff
@@ -244,3 +234,20 @@ class PlotlyAccountError(PlotlyServerError):
...
+
+class ClosedConnection(PlotlyConnectionError):
+    """Raised when the Plotly streaming server has disconnected."""
+    def __init__(self, message, status_code=None):
+        self.message = "[Code: {}] {}".format(status_code, message)
+        self.status_code = status_code or None
+        super(ClosedConnection, self).__init__(self.message)
+
+
+class TooManyConnectFailures(PlotlyConnectionError):
```
chriddyp: nice

plotly/exceptions.py
```diff
@@ -244,3 +234,20 @@ class PlotlyAccountError(PlotlyServerError):
 
 class PlotlyRateLimitError(PlotlyServerError):
     pass
+
+
+class ClosedConnection(PlotlyConnectionError):
```
chriddyp: yeah, way better.

chriddyp:
i probably won't have time for a while to do a deeper dive review of this, looks cool though! careful around the socket error numbers, that functionality in chunked_requests took me a while to figure out and get right with all the exception cases on both windows and mac. hopefully that test server with circle will be able to grab most of these.

theengineear:
ah, nice test server!

yeah, i saw a *lot* of commenting around those socket error numbers and figured that some serious time was poured into that :) i'll definitely take a closer peek before trying to get this merged in.

theengineear:
Man, bummed this didn't get in, o well. Maybe i'll have time to revisit down the road.

PR #259, opened 2015-07-14
add ohlc chart to TraceFactory

cldougl:
- add ohlc chart as 2 TraceFactory methods so
`TraceFactory.create_ohlc_increasing()` returns a trace of the
increasing data and `TraceFactory.create_ohlc_decreasing()` returns a
trace of the decreasing data. In these charts increasing and decreasing
periods are typically represented by different colours (default green &
red).
- pull flatten method out of `_Quiver` so other charts can call it


cldougl:
@chriddyp @theengineear @aneda @etpinard Hi guys, ohlc chart is ready for review/feedback

theengineear:
@cldougl your tests are failing (likely) due to the same things affecting @aneda . Once she's merged into master, you should be able to just merge master into this branch and be good to go. Else you can (a) cherry pick the changes or (b) merge her branch *as-is* and then merge master later before you merge into master.

so many options :)

reviewing now!

plotly/tests/test_core/test_tools/test_trace_factory.py
```diff
@@ -100,3 +100,180 @@ def test_more_kwargs(self):
             'type': 'scatter',
             'mode': 'lines', }
         self.assertEqual(quiver, expected_quiver)
+
+
+class TestOHLC(TestCase):
+
+    def test_unequal_o_h_l_c_length(self):
```
theengineear: :cow2: As is well put, [a foolish consistency is the hobgoblin of little minds](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds). It'd be fine to call this:

`test_unequal_ohlc_length`

:smile_cat:

plotly/tests/test_core/test_tools/test_trace_factory.py
```diff
@@ -100,3 +100,180 @@ def test_more_kwargs(self):
...
+
+        # check: PlotlyError if op, hi, lo, cl are not the same length for
+        # TraceFactory.ohlc_increase and TraceFactory.ohlc_decrease
```
theengineear: :cake: You're tests are so easy to read/parse though, thank you!

plotly/tests/test_core/test_tools/test_trace_factory.py
```diff
@@ -100,3 +100,180 @@ def test_more_kwargs(self):
...
+
+        # create_ohlc_increase
+        kwargs = {'op': [2, 3], 'hi': [4, 2], 'lo': [1, 1], 'cl': [1, 2]}
+        self.assertRaises(PlotlyError, tls.TraceFactory.create_ohlc_increase,
+                          **kwargs)
+
+        # create_ohlc_decrease
+        kwargs = {'op': [2, 3], 'hi': [4, 3], 'lo': [1, 1], 'cl': [1, 6]}
+        self.assertRaises(PlotlyError, tls.TraceFactory.create_ohlc_decrease,
+                          **kwargs)
```
theengineear: should we check any of these messages? if they're supposed to be specific about the error, we might want to lock it down.

theengineear: [`self.assertRaisesRegexp`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaisesRegexp)

I was also going to suggest `self.assertRaisesMessage`, but that appears to just be a Django thing!

plotly/tests/test_core/test_tools/test_trace_factory.py
```diff
@@ -100,3 +100,180 @@ def test_more_kwargs(self):
...
+
+        # create_ohlc_increase
+        kwargs = {'op': [2, 3], 'hi': [4, 6], 'lo': [3, 1], 'cl': [1, 2]}
+        self.assertRaises(PlotlyError, tls.TraceFactory.create_ohlc_increase,
+                          **kwargs)
+
+        # create_ohlc_decrease
+        kwargs = {'op': [2, 3], 'hi': [4, 6], 'lo': [1, 5], 'cl': [1, 6]}
+        self.assertRaises(PlotlyError, tls.TraceFactory.create_ohlc_decrease,
+                          **kwargs)
```
theengineear: ditto here with the messages: https://github.com/plotly/python-api/pull/259/files#r34945593

plotly/tests/test_core/test_tools/test_trace_factory.py
```diff
@@ -100,3 +100,180 @@ def test_more_kwargs(self):
...
+
+    def test_one_ohlc_increase(self):
+
+        # This should create one "increase" (i.e. close > open) ohlc stick
+
+        ohlc_incr = tls.TraceFactory.create_ohlc_increase(op=[33.0], hi=[33.2],
```
theengineear: oh man, yeah, love the `TraceFactory.create_*` syntax.

theengineear:
@cldougl imo, having *two* separate calls for ohlc charts seems like a lot. I'm happy to keep them separate in tools, but maybe just create a single way to make an `ohlc` chart using `TraceFactory`? I think, as a user, I'd be more likely to remember `TraceFactory.create_ohlc` than the two alternatives.

Just a thought, happy to be dissuaded.

theengineear:
Also, unless it's *really* common to use `op`, `hi`, `lo`, `cl` in other languages. We usually like to error on the side of explicitness:

```
The Zen of Python, by Tim Peters

Beautiful is better than ugly.
**Explicit is better than implicit.**
Simple is better than complex.
Complex is better than complicated.
```

... etc etc (try doing `import this`) to check them all out.

[`open`, `high`, `low`, `close`](https://en.wikipedia.org/wiki/Open-high-low-close_chart)

plotly/tools.py
```diff
@@ -1395,27 +1395,33 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
 
 class TraceFactory(dict):
 
-    # use as validate method for all traces (not just streamline)
     @staticmethod
-    def validate(x, y, u, v,
-                 scale=.1, density=1,
-                 arrow_scale=.3, **kwargs):
+    def validate_shared(x, y, u, v,
```
theengineear: :cow2: Just a suggestion. Would this be clearer as two functions?

```
def validate_positive_scalars(**kwargs):
    """
    Validates that all values given in key/val pairs are positive.

    Accepts kwargs to improve Exception messages.
    
    :raises: (PlotlyError) If any value is < 0 or raises.
    """
    for key, val in kwargs.items():
        try:
            if val <= 0:
                raise PlotlyError('{} must be > 0, got {}'.format(key, val)
        except TypeError:
            raise PlotlyError('{} must be a number, got {}'.format(key, val)

def validate_array_pair(**kwargs):
    """"
    Validates a pair of named arrays.

    Accepts kwargs to improve Exception messages.

    :raises: (PlotlyError) On failure.
    """"
    for key, val in kwargs.items():
        # do your checks here...
```

I think that after the refactorings, this `validate` method has gotten much simpler than the docstring suggests?

Again, just a :cow2:

plotly/tools.py
```diff
@@ -1460,6 +1468,50 @@ def validate_streamline(x, y, **kwargs):
...
+
+        for lst in [hi, op, cl]:
+            for index in range(len(lo)):
+                if lo[index] > lst[index]:
+                    raise exceptions.PlotlyError("Oops! Looks like some of "
+                                                 "your low values are greater "
+                                                 "than the corresponding high"
+                                                 ", open, or close values. "
+                                                 "Double check that your data "
+                                                 "is entered in O-H-L-C order")
```
theengineear: :smile_cat: great messages!

plotly/tools.py
```diff
@@ -1460,6 +1468,50 @@ def validate_streamline(x, y, **kwargs):
...
+        :param (array): An iterable to flatten
+        :raises (PlotlyError): If iterable is not nested.
+        :rtype (list): The flattened list.
+        """
+        try:
+            return [item for sublist in array for item in sublist]
+        except TypeError:
+            raise exceptions.PlotlyError("Your data array could not be "
+                                         "flattened! Make sure x, y, u, and v "
+                                         "are lists or numpy ndarrays!")
```
theengineear: :cow2: this message could probably be improved. We need a `list or lists` or an `ndarray`? Is that right? Also, does it make sense to talk about `x, y, u, v`?

plotly/tools.py
```diff
@@ -1647,6 +1699,193 @@ def create_streamline(x, y, u, v,
...
+        close_data = [33.0, 32.9, 33.3, 33.1, 33.1]
+
+        # Create ohlc increasing periods
+        ohlc_increase = tls.TraceFactory.create_ohlc_increase(open_data,
+                                                              high_data,
+                                                              low_data,
+                                                              close_data)
+
+        # Plot
+        fig=Figure()
```
theengineear: :dolls: pep8 for docstrings?

plotly/tools.py
```diff
@@ -1647,6 +1699,193 @@ def create_streamline(x, y, u, v,
...
+                                                              close_data)
+
+        # Create ohlc decreasing periods
+        ohlc_decrease = tls.TraceFactory.create_ohlc_decrease(open_data,
+                                                              high_data,
+                                                              low_data,
+                                                              close_data)
+
+        # Plot
+        fig=Figure()
```
theengineear: ditto: https://github.com/plotly/python-api/pull/259/files#r34945854

plotly/tools.py
```diff
@@ -1647,6 +1699,193 @@ def create_streamline(x, y, u, v,
...
+
+        # Create ohlc decreasing periods
+        ohlc_decrease = tls.TraceFactory.create_ohlc_decrease(open_data,
+                                                              high_data,
+                                                              low_data,
+                                                              close_data)
+
+        # Create layout with dates as x-axis labels
+        fig = dict(data=[ohlc_incr, ohlc_decr],
+              layout=dict(xaxis = dict(ticktext = dates,
```
theengineear: should we be supporting xaxis as an array of actual datetime objects?

plotly/tools.py
```diff
@@ -1647,6 +1699,193 @@ def create_streamline(x, y, u, v,
...
+                                       tickvals = [1, 2, 3, 4, 5 ])))
+
+        # Plot
+        url = py.plot(fig, filename='ohlcs_dates',
+                      overwrite=True, validate=False)
```
theengineear: :cow2: `overwrite=True`? Can this just be:

`url = py.plot(fig, filename='ohlcs_dates', validate=False)`

theengineear:
@cldougl yowza! this is super clean! nice job. a couple of comments above and I'm going to hold off on dancing *not literally of course* until tests pass.

cldougl:
@theengineear @chriddyp tests are passing now, can I get another review of this

plotly/tools.py
```diff
@@ -1670,6 +1722,130 @@ def create_streamline(x, y, u, v,
                              mode='lines', **kwargs)
         return streamline
 
+    @staticmethod
+    def create_ohlc(open, high, low, close, direction, **kwargs):
+
```
theengineear: :cow2: I don't believe we typically leave a space between a function and the docstring.

plotly/tools.py
```diff
@@ -1670,6 +1722,130 @@ def create_streamline(x, y, u, v,
...
+        # Create layout with dates as x-axis labels
+        fig = dict(data=[ohlc_increase, ohlc_decrease],
+              layout=dict(xaxis = dict(ticktext = dates,
+                                       tickvals = [1, 2, 3, 4, 5 ])))
+
+        # Plot
+        url = py.plot(fig, filename='ohlcs_dates', validate=False)
+        ```
+        """
+        TraceFactory.validate_equal_length(open, high, low, close)
```
theengineear: :cow2: I think this is OK because we're using args and not kwargs, but `open` is a Python builtin that we'd usually try and not *shadow* in this scope. it shouldn't really make a difference unless we're actually trying to open a file in a child scope though. just a heads up :)

cldougl: @theengineear that's why I labelled them op hi lo cl at first but so far it seems ok

chriddyp:
@cldougl i'll review tonight or tomorrow

theengineear:
only non-blocking comments from me! the interface for creating these looks good!

:dancer: from me, but you should also wait for a green light from @chriddyp

cldougl:
@theengineear @chriddyp sounds good! Thanks!!

cldougl:
@chriddyp @theengineear @aneda @etpinard @jackparmer candlestick charts are ready for review too (they're pretty similar to the ohlc charts but with a shaded middle)

theengineear:
@cldougl I'll review this for now, but it would have been better to merge in the ohlc charts when that review process was over: https://github.com/plotly/python-api/pull/259#issuecomment-123415468

This single commit adds ~600 lines of code!

just a :cow2: :smile_cat: 

Another :cow2: , you can always just branch off a non-master branch and make a new pr out of it. Once the ohlc pr was merged into master *and* you merged master into your other branch, the GH diff would look totally normal.

plotly/tools.py
```diff
@@ -1483,6 +1474,64 @@ def validate_streamline(x, y, **kwargs):
...
 
     @staticmethod
+    def validate_ohlc(open, high, low, close, direction, **kwargs):
+        """
+        ohlc specific validations
+
+        Specifically, this checks that the high value is greatest value and the
+        low value is the lowest value in each unit.
+
+        See TraceFactory.create_streamline() for params
```
chriddyp: typo? wrong function?

theengineear:
@cldougl My only real concern is that we're inheriting from TraceFactory when it doesn't make sense to.

I'm not too worried about it in practice as all this stuff is hidden from the user anyhow.

:dancer:

plotly/tools.py
```diff
@@ -1670,6 +1722,321 @@ def create_streamline(x, y, u, v,
...
+
+        # Create ohlc increasing units
+        ohlc_increase = tls.TraceFactory.create_ohlc(open_data, high_data,
+                                                     low_data, close_data,
+                                                     direction='increasing')
+
+        # Create ohlc decreasing units
+        ohlc_decrease = tls.TraceFactory.create_ohlc(open_data, high_data,
+                                                     low_data, close_data,
+                                                     direction='decreasing')
```
chriddyp: it seems kind of annoying to have to call this twice. i would imagine that most folks would want to see both the increasing and decreasing at once. can we just return both traces from this function, and then tell the user that the first trace will contain the increasing traces, and the second will contain the decreasing?

plotly/tools.py
```diff
@@ -1670,6 +1722,321 @@ def create_streamline(x, y, u, v,
                              mode='lines', **kwargs)
         return streamline
 
+    @staticmethod
+    def create_ohlc(open, high, low, close, direction, **kwargs):
+
+        """
+        Returns data for an ohlc chart
```
chriddyp: right now this actually returns a `trace` that "describes" an ohlc chart (is "describes" the right term to use here? thoughts @theengineear ?). multiple `traces` in a list is what we'd call `data` (but that term out of the figure context is pretty ambiguous)

plotly/tools.py
```diff
@@ -1670,6 +1722,321 @@ def create_streamline(x, y, u, v,
...
+        :param (list) low: low values
+        :param (list) close: closing values
+        :param (string) direction: direction can be 'increasing' or
+            'decreasing'. When the direction is 'increasing', the returned data
+            consists of all units where the close value is greater than the
+            corresponding open value, and when the direction is 'decreasing',
+            the returned data consists of all units where the close value is
+            less than or equal to the corresponding open value.
+        :param (class) kwargs: kwargs passed through plotly.graph_objs.Scatter
+            for more information on valid kwargs call
```
chriddyp: could we add a line before this that says: "These kwargs describe other attributes about the ohlc Scatter trace like the color or the legend name."

plotly/tools.py
```diff
@@ -1670,6 +1722,321 @@ def create_streamline(x, y, u, v,
...
+            help(plotly.graph_objs.Scatter)
+
+        :rtype (trace): returns data for ohlc increasing units or decreasing
```
chriddyp: same comment as above

plotly/tools.py
```diff
@@ -1670,6 +1722,321 @@ def create_streamline(x, y, u, v,
...
+        # Create ohlc decreasing units
+        ohlc_decrease = tls.TraceFactory.create_ohlc(open_data, high_data,
+                                                     low_data, close_data,
+                                                     direction='decreasing')
+
+        # Plot
+        fig = Figure()
+        fig['data'].append(ohlc_increase)
+        fig['data'].append(ohlc_decrease)
+        url = py.plot(fig, filename='ohlc')
```
chriddyp: i wonder if we should include an example of plotting directly as data and plotting as part of a figure. e.g.

```
# Plot
data = [ohlc_increase, ohlc_decrease]
py.plot(data)

# Alternative, if part of a figure:
# fig = Figure(data=[ohlc_increase, ohlc_decrease])
# py.plot(fig, ...)
```

it's so much shorter to just plot `data` which is really nice. but, sometime's people don't quite understand how to add `layout`... so maybe we should just include examples with both call signatures..? or is that too much?

thoughts @aneda @cldougl @theengineear ?

plotly/tools.py
theengineear: ah, good call. yeah we should be careful about using "data", it's so easy to throw around, but probably super confusing for users.

I like `describes` or `represents`.

plotly/tools.py
```diff
@@ -1670,6 +1722,321 @@ def create_streamline(x, y, u, v,
...
+
+        # Create ohlc trace of decreasing units
+        ohlc_decrease = tls.TraceFactory.create_ohlc(open_data, high_data,
+                                                     low_data, close_data,
+                                                     direction='decreasing')
+
+        # Create layout with dates as x-axis labels
+        fig = dict(data=[ohlc_increase, ohlc_decrease],
+              layout=dict(xaxis = dict(ticktext = dates,
+                                       tickvals = [1, 2, 3, 4, 5 ])))
```
chriddyp: yikes, i don't think that this is generic enough. it looks OK for this example because all of the dates that you supplied are evenly spaced but in general they won't be, in which case the labels won't reflect the actual spacing of the dates. This is common enough for OHLC time series data that we should probably figure out a nicer way to encode these traces to handle dates natively, e.g. as part of the `x` inside the traces.

plotly/tools.py
```diff
@@ -1670,6 +1722,321 @@ def create_streamline(x, y, u, v,
...
+        data = candle_incr
+        data.extend(candle_decr)
+        fig = dict(data=data,
+              layout=dict(xaxis = dict(ticktext = dates,
+                                       tickvals = [1, 2, 3, 4, 5 ])))
+
+        # Plot
+        url = py.plot(fig, filename='candle_dates', validate=False)
+        ```
+        """
```
chriddyp: same comments from above apply here

plotly/tools.py
```diff
@@ -1670,6 +1722,321 @@ def create_streamline(x, y, u, v,
...
+
+        # Create layout with dates as x-axis labels
+        fig = dict(data=[ohlc_increase, ohlc_decrease],
+              layout=dict(xaxis = dict(ticktext = dates,
+                                       tickvals = [1, 2, 3, 4, 5 ])))
+
+        # Plot
+        url = py.plot(fig, filename='ohlcs_dates', validate=False)
+        ```
+        """
```
chriddyp: it'd be cool to have a pandas example in here too:

```
import pandas.io.data as web

start = datetime.datetime(2010, 1, 1)
end = datetime.datetime(2013, 1, 27)
df = web.DataReader("aapl", 'yahoo', start, end)
py.iplot([tls.TraceFactory.create_ohlc(df.Open, df.High, df.Low, df.Close)])
```
ohlc-aapl

plotly/tools.py
```diff
@@ -1670,6 +1722,321 @@ def create_streamline(x, y, u, v,
...
+            # Make segments to start fill (individual traces)
+            for index in range(1, len(candlestick), 2):
+                candlestick[index] = dict(type='scatter',
+                                          x=increasing_fill_x[index],
+                                          y=increasing_fill_y[index],
+                                          mode='lines',
+                                          legendgroup='Increasing',
+                                          showlegend=False,
+                                          fill='tonextx',
+                                          **kwargs)
```
chriddyp: I'd also remove the opacity on the fill, seems like most candlestick charts are either totally solid color or totally white, our opacity with the visible line through it is irregular
candlestick

vs e.g.

plotly/tools.py
theengineear: Hmmm, but then we may or may not be returning a list from `TraceFactory`.

Would it make sense to just *always* return a list? Then folks just use `extend` to add it to data?

plotly/tools.py
theengineear: personally, I'd like to move as many examples as possible to using fig. it's more consistent with the db model we use.

*sneak preview* i'd also like to have FileObjects (i think we've chatted about this before @chriddyp ) which just wrap our apis. A whole lot like how you do `resources` in angular.

If we did this, *always* using figure makes a lot more sense.

plotly/tools.py
chriddyp: I don't feel strongly either way.

plotly/tools.py
cldougl: @chriddyp I'll keep brainstorming on that point. So far I couldn't really find a way to make the ohlc & candlestick shapes without using the x grid. I'll look at some other examples and see how unevenly spaced dates are handled.

chriddyp: i think this'll require `x` to be an input argument to the charts (which is OK IMO). then, maybe just look for the smallest interval between all of the x-array, and make the x-coordinates of the left-side of the bar 20% of that spacing

plotly/tools.py
cldougl: @chriddyp if we combine both increasing and decreasing into one call is there an easier way to pass the kwargs for each trace than doing something like: 
```
kwargs.setdefault('name', ('Increasing', 'Decreasing'))
kwargs.setdefault('line', {'color': ('rgb(44, 160, 44)', 'rgb(214, 39, 40)')})
kwargs.setdefault('text', (text_increase, text_decrease))

ohlc_icrease = Scatter(x=flat_increase_x,
                                     y=flat_increase_y,
                                     mode='lines',
                                     name=kwargs['name'][0],
                                     text=kwargs['text'][0],
                                     line=Line(color=kwargs['line']['color'][0])
                        )

ohlc_decrease = Scatter(x=flat_decrease_x,
                                        y=flat_decrease_y,
                                        mode='lines',
                                        name=kwargs['name'][1],
                                        text=kwargs['text'][1],
                                        line=Line(color=kwargs['line']['color'][1])
                         )
```

chriddyp: @cldougl hm.. what do you think about `create_ohlc(..., direction='both') # | 'increasing' | 'decreasing'`?

cldougl: @chriddyp yeah, I think that sounds good. Then the user can easily create both with the default kwargs or call the function twice if they want to change the color or name etc.

chriddyp: yeah, whoops prematuraly deleted my comment. Options:

```
traces = create_ohlc(..., direction='both') # | 'increasing' | 'decreasing'

traces[0].update({'line': {'color': 'blue'}})
traces[1].update({'line': {'color': 'red'}})
```

Other options:
1 - Let user pass through `Scatter` `**kwargs` that apply to both of the traces? e.g. `create_ohlc(..., line=Line(width=4), name='AAPL')`? Not sure. Seems sort of out of line of our declarative philosophy.
2 - Let user apply `kwargs` values as an array, e.g. `create_ohlc(...name=['AAPL Increasing', 'AAPL Decreasing'])`. I vote No: too confusing, too much to teach.

The most common attribute that people will want to edit will be the `name`. I think by default we should `group` the traces and hide one of the legend entries as I imagine most finance folks reading these charts won't need to see the `increasing` and `decreasing` legend labels, but might want to see the tickers (especially if multiple tickers are graphed) e.g. `AAPL`, `COKE`. With grouped and half-hidden legend entries, they'll just have to:
```
traces = create_ohlc(...)
traces[0].update({'name': 'AAPL'})
```

theengineear: I think passing `**kwargs` that apply to *both* traces makes sense. We allow the `**kwargs` for the other `create_*` methods.

I don't think it's terribly confusing. If users *really* want more control, we could allow this:

```
def create_ohlc(..., increasing_kwargs=None, decreasing_kwargs=None, **kwargs)
```

And apply `**kwargs` to *both*, then nested update with the more specific `*_kwargs` dict. But again, just supporting this:

```
def create_ohlc(...,  **kwargs)
```

Is OK for now since you can pretty simply update the individual trace dicts.

theengineear:
Sorry if I missed this discussion, but I think it's pretty confusing to have both `FigureFactory` and `TraceFactory`.

I get that we need to change the layout so `FigureFactory` became necessary, can we then just have *everything* go through figure factory?

The first thing I did in a notebook was try to tab-complete on `TraceFactory` and was surprised not to see  `TraceFactory.create_ohlc`.

I bet most of the use-case for these is creating a single figure, not adding traces to figures.

theengineear:
Should we change hovermode to 'closest' on ohlc charts? The text at the open and close points are hard to hover over when comparing points.

theengineear:
I think I've asked this before, but why are our Factories subclassing `dict`? Here are the methods I see in a notebook:

image

But it's unclear to my why I'd need to `update`, `fromkeys`, `clear`, etc etc. The *object* you get back is a subclass of `dict`, but not the factory.

plotly/tools.py
```diff
@@ -2093,14 +2088,717 @@ def get_streamline_arrows(self):
...
+                         text=text_decrease,
+                         **kwargs)
+
+        layout = dict(xaxis=dict(zeroline=False))
+
+        if direction is 'increasing':
+            return dict(data=[ohlc_incr], layout=layout)
+        elif direction is 'decreasing':
+            return dict(data=[ohlc_decr], layout=layout)
+        elif direction is 'both':
```
theengineear: :cow2: consider something like this? Otherwise, we'd just be returning `None`, right?

```
if direction is 'increasing':
    # blah
elif direction is 'decreasing':
    # blah
elif direction is 'both':
    # blah
else:
    raise SomeException('blah')
```

plotly/tools.py
```diff
@@ -2093,14 +2088,717 @@ def get_streamline_arrows(self):
...
+            return dict(data=[ohlc_incr, ohlc_decr], layout=layout)
+
+    @staticmethod
+    def create_candlestick(open, high, low, close,
```
theengineear: This function is ~140 lines (not including docstrings)! looks like you're using function closures to make the methods defined in here work. However, could you just as simply create another static method and pass the required *state* into it? These static methods should be "private" (`_*` type methods).

plotly/tools.py
theengineear: nvm, i see it in `validate_ohlc` now. However, that means you should just do:

```
if direction is 'increasing':
    # blah
elif direction is 'decreasing':
    # blah
else:
    # blah
```

plotly/tools.py
```diff
@@ -2093,14 +2088,717 @@ def get_streamline_arrows(self):
...
+             candle_bar_decr,
+             candle_line_decr) = make_decreasing_candle(decrease_x,
+                                                        decrease_close,
+                                                        decrease_dif,
+                                                        stick_decrease_y,
+                                                        stick_decrease_x,
+                                                        **kwargs)
+            data = [hidden_bar_decr, candle_bar_decr, candle_line_decr]
+
+        elif direction is 'both':
```
theengineear: Similar to https://github.com/plotly/python-api/pull/259/files#r35625061

you can just use an `else` here, then it's clear that `data` will be defined below.

plotly/tools.py
```diff
@@ -2093,14 +2088,717 @@ def get_streamline_arrows(self):
...
+        self.all_x = []
+        self.all_y = []
+        self.increase_x = []
+        self.increase_y = []
+        self.decrease_x = []
+        self.decrease_y = []
+        self.get_all_xy()
+        self.separate_increase_decrease()
+        self.get_increase()
+        self.get_decrease()
```
theengineear: What's the point of calling `get_increase()` and `get_decrease()` here? They don't seem to change the state and the return values aren't being caught.

plotly/tools.py
```diff
@@ -2093,14 +2088,717 @@ def get_streamline_arrows(self):
...
+                                                        stick_decrease_y,
+                                                        stick_decrease_x,
+                                                        **kwargs)
+            data = [hidden_bar_incr, candle_bar_incr, candle_line_incr,
+                    hidden_bar_decr, candle_bar_decr, candle_line_decr]
+
+        return dict(data=data, layout=layout)
+
+
+class _OHLC(dict):
```
theengineear: :cow2: Why does this need to inherit from `dict`? Wouldn't `object` suffice?

This is "private" though, so we can refactor whenever. nbd :)

theengineear: :cow2: if you *do* want to inherit from something other than `object`, you should typically complete the `super` chain on overridden methods. So here you'd want to make a call to:

`super(_OHLC, self).__init__()` With some arguments.

plotly/tools.py
```diff
@@ -2093,14 +2088,717 @@ def get_streamline_arrows(self):
...
+        :rtype (list, list): streamline_x: all x values for each streamline
+            combined into single list and streamline_y: all y values for each
+            streamline combined into single list
         """
         streamline_x = sum(self.st_x, [])
         streamline_y = sum(self.st_y, [])
         return streamline_x, streamline_y
+
+
+class FigureFactory(dict):
```
theengineear: :dolls: Add help for `FigureFactory`?

theengineear:
@cldougl my main issue is that we now have `TraceFactory` and `FigureFactory`. I think this is more work for our users.

@chriddyp this has been discussed before, but do we have a way of marking new functionality like this as *beta*? I feel like we could potentially get flak for changing this, but it's really not been out there for very long. When we do this in the api, we *at least* have [documentation](https://api.plot.ly/v2/#beta) about it. Perhaps just keeping something like BETA in the docstring would be a good idea for new things like this that are in development? IPython still says that the widgets framework is subject to change... and that's heavily used.

I don't *think* we need to throw up a warning... but it would certainly get the point across...

jackparmer:
+1 BETA in the docstring. Definitely not excited about a big read BETA warning when a user makes a chart, but BETA in the docstring makes sense.

Let's just convert everything to a FigureFactory and get rid of TraceFactory? FigureFactory seems a much more flexible way going forward. There aren't enough (if any) folks using streamline and quiver plots right now that a breaking change is a big deal. Cufflinks is going to use OHLC and candlestick though, so it would be sweet if we had a flexible FigureFactory 

@cldougl Can we do that in 2 PRs to get candlestick and OHLC out the door?
First get candlestick and OHLC out (under the current figure factory), then upgrade streamline and quiver from FT to FF and get rid of FT.

cldougl:
Yep, that sounds good

cldougl:
@theengineear re the Factories subclassing dict: the idea was to make it similar to `Scatter`
We can change it to subclassing from `object` if that makes more sense because we don't need `update`, `fromkeys`, `clear`, etc

theengineear:
Yup. Yeah, we'd only want to subclass `dict` if we're using it like that. It'd be clearer to just use `object`. If it's not too big a change, that'd be great! :monkey:

cldougl:
done :)

theengineear:
\o/

chriddyp:
:dancer: when tests pass

theengineear:
:boom: nice!

PR #260, opened 2015-07-15
Config plot option

aneda:
@theengineear @chriddyp @cldougl 
This is a draft for config_plot_option (world_readable).  I haven't tested for python 3 yet but I wanted check if I am on the right track


plotly/tests/test_core/test_tools/test_file_tools.py
```diff
@@ -1,31 +1,80 @@
...
-    creds = tls.get_credentials_file()
-    assert all(x in creds for x in expected)
-    assert original_creds == creds
+from unittest import TestCase
+
+from plotly import tools
+from plotly import session
+
+
+class FileToolsTest(TestCase):
```
theengineear: \o\ |o| /o/

plotly/tests/test_core/test_tools/test_file_tools.py
```diff
@@ -1,31 +1,80 @@
...
-    reset_creds = tls.get_credentials_file()
-    tls.set_credentials_file(**original_creds)
-    assert all(x in reset_creds for x in expected)
-    creds = tls.get_credentials_file()
-    assert all(x in creds for x in expected)
-    assert original_creds == creds
+from unittest import TestCase
+
+from plotly import tools
+from plotly import session
```
theengineear: :cow2: When you import things from the same place, we usually choose the more compact notation:

`from plotly import session, tools`

plotly/tests/test_core/test_tools/test_file_tools.py
```diff
@@ -1,31 +1,80 @@
...
+
+
+class FileToolsTest(TestCase):
+
+    def test_set_config_file_all_entries(self):
+
+        # Check set_config and get_config return the same values
+
+        pd, ps, pa = 'this', 'thing', 'that'
```
theengineear: :cow2: tests (especially) should be very specific. can you rename these variables to be more descriptive? It's OK if it takes 3 lines :)

plotly/tests/test_core/test_tools/test_file_tools.py
```diff
@@ -1,31 +1,80 @@
...
+        self.assertEqual(config['plotly_ssl_verification'], psv)
+        self.assertEqual(config['plot_options'], po)
+
+    def test_set_config_file_two_entries(self):
+
+        # Check set_config and get_config given only two entries return the
+        # same values
+
+        pd, ps = 'this', 'thing'
+        tools.set_config_file(plotly_domain=pd, plotly_streaming_domain=ps)
```
theengineear: careful, this will set the config file locally for whoever is running the tests!

I'll make a parent testcase with some setup that can cache and take care of this for us. i'll ping you on the pr :)

plotly/tests/test_core/test_tools/test_file_tools.py
```diff
@@ -1,31 +1,80 @@
...
+        self.assertEqual(session_po['world_readable'], po['world_readable'])
+
+    def test_set_config_file_plot_option(self):
+
+        # Return TypeError when plot_option type is not a dict
+
+        kwargs = {'plot_option': False}
+        self.assertRaises(TypeError, tools.set_config_file, **kwargs)
+
+    def reset_config_file_test(self):
```
theengineear: --> `test_reset_config_file`

theengineear:
@aneda you're tests are failing because you have an old version of master.

You should handle it in *this* order:

* I'd accept the merge for my branch into yours in GH *first* (go ahead and click the merge button there is you're OK with it, or drop me a note if you're not, it's just a merge into your branch, not master.)
* Then refresh your local branch `git pull origin config_plot_option` with your branch checked out (which is likely the state)
* Then merge the newest master into your branch (likely a small merge conflict) `git pull origin master` with your branch checked out (again, that's probably how it is)
* Fix the merge conflict, commit and push back up.

Apologies if those instructions are overly detailed. I just want to make sure you're good to go!

theengineear:
@aneda what do you think about https://github.com/plotly/python-api/pull/260#commitcomment-12171141

I apologize for possibly steering you in a slightly wrong direction, but I'd like this to be as simple to implement as possible. Seeing the code so far, I'm realizing that the nested plot options might lead to more complication than it's worth.

I'm really okay either way, as long as the file remains fairly stupid-proof.

plotly/plotly/plotly.py
```diff
@@ -50,7 +50,7 @@
 DEFAULT_PLOT_OPTIONS = {
     'filename': "plot from API",
     'fileopt': "new",
-    'world_readable': True,
+    'world_readable': False,
```
chriddyp: @theengineear - My thought is to include saving the `world_readable` as a "required" step in the new getting-started instructions and not changing the default behaviour. So, everyone who is new will save the default to False, and be aware of privacy, and we don't introduce the backwards incompatibility for everyone else who is upgrading.

thoughts?

plotly/tools.py
```diff
@@ -145,7 +146,8 @@ def reset_credentials_file():
 def set_config_file(plotly_domain=None,
                     plotly_streaming_domain=None,
                     plotly_api_domain=None,
-                    plotly_ssl_verification=None):
+                    plotly_ssl_verification=None,
+                    plot_options=None):
```
chriddyp: @theengineear - agreed: hiding `plot_options` here is confusing. agreed, let's de-nest it and add `world_readable` as an explicit argument (and in the future, `secret_link` :) ). the only benefit to having `plot_options` is that it implicitly matches the kwargs of `plot`, but I don't think anyone besides us will get that.

Just be clear:

- `sign_in` - accepts de-nested arguments of `config`, `credentials`
- `plot`, `iplot`- accepts de-nested `world_readable`, `secret_link`
- `set_credentials` - accepts `credentials`
- `set_config` - accepts `config` and `world_readable`, `secret_link`
- `session.set_plot_options` - (or w/e its called) - accepts `plot_options` including  `world_readable`, `secret_link`

chriddyp:
@theengineear - yeah, changeling through github releases. still seems like the best way, since it includes the exact commit

plotly/plotly/plotly.py
theengineear: yeah, that sounds better to me! To be clear then, @aneda , let's change this back to `'world_readable': True` to maintain backwards compat.

plotly/tools.py

aneda:
@theengineear @chriddyp @cldougl 
I changed the removed the plot_option with world_readable = True.  Please let me know what needs to be modified/ add.  
Btw, thanks for all the reviews and help.  Andrew I don't think it is ever overly detailed for me haha

theengineear:
@aneda this is looking good. if you want me to pull this down and get tests running again let me know. they're failing because of a change @BRONSOLO recently made that we're being a bit too strict about in the api library, imo.

https://github.com/plotly/python-api/blob/config_plot_option/plotly/tests/test_core/test_grid/test_grid.py#L194

That file needs to be updated to use a `TestCase` and it should not depend on the `409`, but rather a `4xx`. Else, you can skip that tests, I'm open to either.

aneda:
@theengineear did I do it right?

plotly/tools.py
```diff
@@ -191,30 +198,46 @@ def set_config_file(plotly_domain=None,
...
-def get_config_file(*args):
-    """Return specified args from `~/.plotly_credentials`. as dict.
+def get_config_file(**kwargs):
+    """Return specified args from `~/.plotly/.config`. as dict.
 
     Returns all if no arguments are specified.
 
     Example:
-        get_credentials_file('username')
+        get_credentials_file('plotly_domain')
```
theengineear: `get_config_file`?

plotly/tools.py
```diff
@@ -191,30 +198,46 @@ def set_config_file(plotly_domain=None,
     settings = get_config_file()
     if isinstance(plotly_domain, six.string_types):
         settings['plotly_domain'] = plotly_domain
+    elif plotly_domain is not None:
+        raise TypeError('Input should be a string')
```
theengineear: this seems like a backwards incompatible change. i think it's better this way.. i like it :)

plotly/tools.py
```diff
@@ -191,30 +198,46 @@ def set_config_file(plotly_domain=None,
...
+        session.update_session_plot_options(**kwargs)
+    elif world_readable is not None:
+        raise TypeError('Input should be a boolean')
     utils.save_json_dict(CONFIG_FILE, settings)
     ensure_local_plotly_files()  # make sure what we just put there is OK
 
 
-def get_config_file(*args):
-    """Return specified args from `~/.plotly_credentials`. as dict.
+def get_config_file(**kwargs):
```
theengineear: why should this be `kwargs` and not `args`? i'd expect to call this like:

`get_config_file(['plotly_domain', 'plotly_api_domain'])`

Or

`get_config_file()['plotly_domain']`

plotly/tools.py
```diff
@@ -2081,4 +2104,3 @@ def sum_streamlines(self):
         streamline_x = sum(self.st_x, [])
         streamline_y = sum(self.st_y, [])
         return streamline_x, streamline_y
-
```
theengineear: :cow2: you should be able to set your editor so that there's always a newline added at the end of a file on save (if it's not already there)

theengineear:
Looking good! One blocking comment about the *args to **kwargs change.

aneda:
@theengineear is this good?  Also would let me know if the issue does make sense or not, thanks!

plotly/tools.py
theengineear: @aneda this is my only blocking comment now :)

aneda: @theengineear yeah, thanks!  just updated the file!  I think I changed this for plot_option but forgot to correct it

aneda:
@theengineear just updated the kwargs to args

theengineear:
hmmm, not sure why this new test is failing? i'll take a closer look tomorrow.

@chriddyp do you think breaking a small amount of backwards compat is OK here:

https://github.com/plotly/python-api/pull/260#discussion_r34914403

aneda:
@theengineear  I added another `@skip` as Chris suggested!  Is it good now!

cc @chriddyp

plotly/tools.py
chriddyp: yes, i think this is fine. in the future, i'd like even more validation here, error-ing if someone includes `https://` in streaming domain

theengineear: yeah, the way we say `domain` but mean `schema + domain` is a little confusing anyhow. we could definitely be smarter about it.

theengineear:
:dancer: !

theengineear:
If we're going to release on pypi, we also need a version bump!

aneda:
@theengineear @chriddyp is it good to go?

theengineear:
si! :dancers:

chriddyp:
merge away!

On Mon, Jul 20, 2015 at 2:37 PM, Andrew <notifications@github.com> wrote:

> si! [image: :dancers:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/plotly/python-api/pull/260#issuecomment-122981897>.
>



-- 
Chris Parmer
Plotly, Chief Product Officer
514 571 5897
https://plot.ly/team

PR #261, opened 2015-07-15
Config file safe tests

theengineear:
Add some protective isolation of session and file config/credentials between unit tests. This only works when you inherit from the new `TestCase` (`PlotlyTestCase`)

### DIFF
--- a/plotly/tests/test_core/test_tools/test_file_tools.py
+++ b/plotly/tests/test_core/test_tools/test_file_tools.py
@@ -1,10 +1,9 @@
-from unittest import TestCase
-
 from plotly import tools
 from plotly import session
+from plotly.tests.utils import PlotlyTestCase
 
 
-class FileToolsTest(TestCase):
+class FileToolsTest(PlotlyTestCase):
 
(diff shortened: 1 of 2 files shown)


theengineear:
@aneda @chriddyp @cldougl , I'd like to merge this into Neda's branch so that we can be safer about credentials/config files during tests. Right now, they're not isolated. Therefore, a change to these files in one test can cause another one to fail :disappointed:. That's the opposite of unit tests! I think this should fix it.

To use this, just inherit from `PlotlyTestCase`. It lives in `plotly.tests.utils`

chriddyp:
once tests pass :dancer:

theengineear:
@chriddyp , this is failing because master needs to be merged into it, i wrote up a procedure for @aneda [here](https://github.com/plotly/python-api/pull/260#issuecomment-121522249). This is merging into her branch, *not* master, so I think it will cause fewer conflicts for her to merge this first and then merge master into the `config_plot_option` branch.

Neda, feel free to merge this into your branch.

PR #264, opened 2015-07-21
Secret key

aneda:
@theengineear @chriddyp @cldougl 
Hey, this is a rough draft for secret_key.  This only works for plot (py.plot(...)).  I haven't done any tests and I am working on iplot (being stuck there).  So I wanted to check if I am on the right track and this is what you had in mind.  Thnaks!


plotly/plotly/plotly.py
```diff
@@ -1246,9 +1274,13 @@ def _send_to_plotly(figure, **plot_options):
     url = get_config()['plotly_domain'] + "/clientresp"
 
     r = requests.post(url, data=payload,
-                      verify=get_config()['plotly_ssl_verification'])
+                      verify=get_config()['plotly_ssl_verification'],)
```
theengineear: did you mean to add this?

plotly/plotly/plotly.py
```diff
@@ -1246,9 +1274,13 @@ def _send_to_plotly(figure, **plot_options):
...
     r.raise_for_status()
     r = json.loads(r.text)
+    # Check if the url needs a secret key
```
theengineear: :cow2: newlines above comments :) https://github.com/plotly/dev-docs/blob/master/reference/styleguides/python.md#spacing

plotly/plotly/plotly.py
```diff
@@ -1216,6 +1218,31 @@ def validate_credentials(credentials):
         raise exceptions.PlotlyLocalCredentialsError()
 
 
+def secret_key(r, **payload):
+    """
+    update plot's url to include the scret key
```
theengineear: :dolls: ~~`update`~~ `Update` and document arguments please :smile_cat: 

Can we also pick a better variable name than `r`?

plotly/plotly/plotly.py
```diff
@@ -1216,6 +1218,31 @@ def validate_credentials(credentials):
...
+                              headers={'Plotly-Client-Platform': 'Python'},
+                              auth=HTTPBasicAuth(payload['un'],
+                                                 payload['key']),
+                              data={"share_key_enabled":
+                                    plot_kwargs['share_key_enabled'],
+                                    "world_readable":
+                                    plot_kwargs['world_readable'],
+                                    "filename": plot_kwargs['filename'],
+                                    "fileopt": plot_kwargs['fileopt']})
+    response = json.loads(response.content)
```
theengineear: :cow2: It's helpful to not change the value type of a variable. So earlier `response` was a `Response` object, and now it's a `dict`. --> `response_data = json.loads(response.content)`?

plotly/plotly/plotly.py
```diff
@@ -1216,6 +1218,31 @@ def validate_credentials(credentials):
...
+def secret_key(r, **payload):
+    """
+    update plot's url to include the scret key
+
+    """
+
+    plot_id = r['url'].replace("https://plot.ly/~" + payload['un'] + "/", "")
+    url = "https://api.plot.ly/v2/files/" + payload['un'] + ":" + plot_id
+    plot_kwargs = json.loads(payload['kwargs'])
+    response = requests.patch(url,
```
theengineear: This is OK for now, but when we move to wrapping the new `/plots` endpoint, this will just be a single api call. We *could* update `clientresp` to handle this now, but we're transitioning away from clientresp so I'd rather not add anything new to it.

plotly/plotly/plotly.py
```diff
@@ -1246,9 +1274,13 @@ def _send_to_plotly(figure, **plot_options):
...
 
     r = requests.post(url, data=payload,
-                      verify=get_config()['plotly_ssl_verification'])
+                      verify=get_config()['plotly_ssl_verification'],)
     r.raise_for_status()
     r = json.loads(r.text)
+    # Check if the url needs a secret key
+    if plot_options['share_key_enabled'] is True:
+        # secret_key updates the url to include the secret key
+        r = secret_key(r, **payload)
```
theengineear: ~~`secrete_key`~~ --> `add_share_key_to_url` ?

plotly/session.py
```diff
@@ -39,7 +39,8 @@
     'fileopt': six.string_types,
     'world_readable': bool,
     'auto_open': bool,
-    'validate': bool
+    'validate': bool,
+    'share_key_enabled': bool,
```
theengineear: :cow2: no need for a coma in the last item of a dict.

theengineear:
The general setup for this looks good to me. Sort of a bummer to have an extra request but it's the safest for all users (enterprise included) Every time we make a change to in streambed, we need to consider when certain users will actually receive that update.

When we start wrapping the other new endpoints I'd really like to move to a more RESTful design which will make this a bit smoother.

Ping me when you want a final review of it : D

aneda:
@theengineear @chriddyp I've added the share_key for iplot but I wasn't able to fix the error.  The error happens on `test_grid.py` (`test_plot_from_grid`).  I think the issue is due of this part : https://github.com/plotly/python-api/blob/master/plotly/plotly/plotly.py#L829-L831
I am not too happy about  `_plot_option_logic`  but wasn't able to come up with another way.

plotly/plotly/plotly.py
```diff
@@ -105,6 +108,21 @@ def _plot_option_logic(plot_options):
...
+
+    # If plot set as public but Share_key_enabled is activated 
+    if ('world_readable' in plot_options and plot_options['world_readable']
+            and plot_options['share_key_enabled']):
+        current_plot_options['world_readable'] = True
+        current_plot_options['share_key_enabled'] = False
+        warnings.warn(
+            "Looks like you've activated the share_key for a public plot. "
+            "share_key can only be activaited for private plots."
+            "\nQuestions? support@plot.ly")
```
theengineear: :cow2: This isn't your fault, but should really be changed. `support@plot.ly` is intended for enterprise-related emails, not for the general public consumers. We should really be pointing users to `feedback@plot.ly`. We do this in a number of places.

@chriddyp actually, we could be a bit sneaky and return different *supporty* emails based on whether the plotly_domain matches the default plotly_domain in the session config? too sneaky? worthwhile?

plotly/plotly/plotly.py
```diff
@@ -119,24 +137,51 @@ def iplot(figure_or_data, **plot_options):
...
+    _logic = True,
+    res = plot(figure_or_data, _logic, **plot_options)
     urlsplit = res.split('/')
-    username, plot_id = urlsplit[-2][1:], urlsplit[-1]  # TODO: HACKY!
+    username = urlsplit[-2][1:]
+
+    # Check if the url contains share_key
+    if ('share_key_enabled' in plot_options and
+            plot_options['share_key_enabled']):
+        plot_id, share_key = urlsplit[-1].split('?share_key=', 1)
```
theengineear: :cow2: yeah, this will be much cleaner if we're using `urlparse`

aneda:
@theengineear Hey thanks for your help and reviews!  I wanted to keep you in the loop, last night @chriddyp had the idea to create a new key , `sharing`, that can be either  `{'public', 'private', 'secret'},  instead of `share_key`.  I am working to add `sharing` now.  Let us know if you think that not a good route

theengineear:
would the new key be a python library thing or something that's going to be added to the backend api?

aneda:
python library

theengineear:
do you mean `share_key_enabled` when you say `share_key`?

theengineear:
yeah, that seems like a pretty user-friendly way to do it. one concern is keeping backwards compat with the `world_readable = False` default. Does this mean we're deprecating this setting?

aneda:
yes, my bad I meant `share_key_enabled`.  We keep `world_readable` and just check if there is any conflict between the two (`world_readable' and `sharing`)

aneda:
@theengineear @chriddyp @cldougl 
I have removed `share_key_enabled` and added `sharing` to `plot_options`.  `sharing` is a dict `{'public', 'private', 'secret'}`.  I am writing the tests now but it failed `test_grid`

theengineear:
sweet, sounds good then!

theengineear:
re: https://github.com/plotly/python-api/pull/264#issuecomment-125682284

try merging in master (you'll need to fix some conflicts) and then pushing back up. I *think* that will be resolved.

aneda:
@theengineear @chriddyp @cldougl  `sharing` key is ready for review

plotly/session.py
```diff
@@ -39,7 +39,8 @@
     'fileopt': six.string_types,
     'world_readable': bool,
     'auto_open': bool,
-    'validate': bool
+    'validate': bool,
+    'share': dict
```
theengineear: Should this match `sharing`?

plotly/tools.py
```diff
@@ -251,7 +251,7 @@ def reset_config_file():
 
 ### embed tools ###
 
-def get_embed(file_owner_or_url, file_id=None, width="100%", height=525):
+def get_embed(file_owner_or_url, file_id=None, share_key=None, width="100%", height=525):
```
theengineear: Does this mean users need to *know* the share key ahead of time? I'm not sure we provide a way to get at that programmatically, so that forces them to need to click around in plot.ly.

@chriddyp thoughts on this one? It's gonna require a quick api hit to get the share key, but I think it'll be easier to work with.

plotly/tools.py
```diff
@@ -269,12 +269,15 @@ def get_embed(file_owner_or_url, file_id=None, width="100%", height=525):
     Keyword arguments:
     file_id (default=None) -- an int or string that can be converted to int
                               if you're using a url, don't fill this in!
+    share_key (default=None) -- a string that enables private plots to be
+                                shared. Note, you must have admin privileges
+                                on the file for this to work
```
theengineear: if we keep this as share_key, we might want to explain how a user would *view* the share key for a plot.

plotly/tools.py
```diff
@@ -307,33 +309,31 @@ def get_embed(file_owner_or_url, file_id=None, width="100%", height=525):
...
             plotly_rest_url=plotly_rest_url,
             file_owner=file_owner, file_id=file_id,
-            plot_width=width - padding, plot_height=height - padding,
             iframe_height=height, iframe_width=width)
     else:
         s = ("<iframe id=\"igraph\" scrolling=\"no\" style=\"border:none;\""
              "seamless=\"seamless\" "
              "src=\"{plotly_rest_url}/"
-             "~{file_owner}/{file_id}.embed\" "
+             "~{file_owner}/{file_id}.embed?{share_key}\" "
```
theengineear: does this work? i thought we needed `share_key={share_key}`?

plotly/tools.py
```diff
@@ -351,12 +351,17 @@ def embed(file_owner_or_url, file_id=None, width="100%", height=525):
...
+                                on the file for this to work
     width (default="100%") -- an int or string corresp. to width of the figure
-    height (default="525") -- same as width but corresp. to the height of the figure
+    height (default="525") -- same as width but corresp. to the height of the
+                              figure
 
     """
     try:
-        s = get_embed(file_owner_or_url, file_id, width, height)
+        s = get_embed(file_owner_or_url, file_id, share_key, width, height)
```
theengineear: :cow2: it's more robust to mimic the call signature of the function, in this case, `share_key=share_key`, etc. If you *don't* do this, someone could just insert a kwarg in the call signature later and break this. In fact, this pr *does* insert the `share_key` kwarg just like that.

In general if the arg is specified as positional (use it as a positional, because the *name* might change) and if it's key-wordy (use it as key-wordy, because the *order* might change)

plotly/tools.py
```diff
@@ -364,17 +369,10 @@ def embed(file_owner_or_url, file_id=None, width="100%", height=525):
...
-            url = "{plotly_domain}/~{un}/{fid}".format(
-                plotly_domain=plotly_domain,
-                un=file_owner_or_url,
-                fid=file_id)
+            return PlotlyDisplay(file_owner_or_url, file_id, share_key, width,
+                                 height)
         else:
-            url = file_owner_or_url
-        return PlotlyDisplay(url, width, height)
+            return PlotlyDisplay(file_owner_or_url, width, height)
```
theengineear: The `file_id` argument is required as defined in `PlotlyDisplay`. Does this work?

plotly/tools.py
aneda: yeah it does cause `six.moves.urllib.parse.urlparse(res).query` returns `share_key=<key>`, so I just need to add `?` to get `?share_key=<key>`

plotly/tools.py
aneda: oops, that's the mistake and I've forgot to correct `PlotlyDisplay` as well

aneda: I've added  `file_id` but the original file didn't have it

https://github.com/plotly/python-api/blob/master/plotly/tools.py#L376

 is there a reason or just a mistake?

plotly/tools.py
theengineear: oh man, that's super confusing (if i'm understanding correctly). so as a user, do i need to know to pass in `share_key='share_key=ASDFIH24HS3'`?

Ideally (imo), our api here is `append_share_key=True` or something and we just handle it.

At least could it be `share_key='ASDFIH24HS3'`?

Also, whatever we need to go with, let's give an example in the docstring for it?

plotly/tools.py
aneda: so just to make sure I understood correctly, we should call it like `get_embed(file_owner_or_url, file_id, share_key=share_key, width=width, height=height)` to avoid any keyword arguments?

plotly/tools.py
theengineear: the get_embed can take a first parameter of `file_owner_or_url`, this is for more flexibility / backwards compat at one point. however, PlotlyDisplay doesn't *do* anything with file_id, this is because it *expects* that the first positional argument is a url, not a username of some kind.

I think we shouldn't add `file_id` to the `PlotlyDisplay`, is there a reason for it?

plotly/tools.py
aneda: well, I can change it to be like `.....embed.?share_key={share_key}`.  I mean we can extract the `<key>` out of  `six.moves.urllib.parse.urlparse(res).query`.  This way users only need to pass the key `ASDFIH24HS3`

plotly/tools.py
aneda: oh nvm, I misunderstood you.  I thought you wanted me to add `file_id` to `PlotlyDisplay`

plotly/tools.py
theengineear: yup, the way it was is totally OK, but this a bit safer. just a :cow2:

you could also do `file_id=file_id` in there. really, it's OK as is, but if you feel like updating, it would be a little stronger.

plotly/tools.py
theengineear: yeah! exactly! i think that's much clearer.

later on, if we want to make this simpler on users by *getting* these keys for them, we can work that in.

plotly/tools.py
aneda: sounds good! I am always up for updates!

plotly/tools.py
aneda: sounds good!

plotly/tools.py
chriddyp: yeah, and "admin privileges" is misleading, if you have the URL of a graph you don't need anything. You need "admin" privledges to "enable private sharing" but not necessarily to embed.

also, `file_owner_or_url` should be able to handle the URLs copied and pasted directly from the firewall:
- https://plot.ly/~christopherp/749/col2/?share_key=wRNj7Du1fvbVtTPQcmd057
- https://plot.ly/~christopherp/749/col2?share_key=wRNj7Du1fvbVtTPQcmd057
- https://plot.ly/~christopherp/749?share_key=wRNj7Du1fvbVtTPQcmd057
- https://plot.ly/~christopherp/749/?share_key=wRNj7Du1fvbVtTPQcmd057

- https://enterprise.plot.ly/~christopherp/749/col2/?share_key=wRNj7Du1fvbVtTPQcmd057
- https://enterprise.plot.ly/~christopherp/749/col2?share_key=wRNj7Du1fvbVtTPQcmd057
- https://enterprise.plot.ly/~christopherp/749?share_key=wRNj7Du1fvbVtTPQcmd057
- https://enterprise.plot.ly/~christopherp/749/?share_key=wRNj7Du1fvbVtTPQcmd057

plotly/plotly/plotly.py
```diff
@@ -1271,6 +1334,50 @@ def validate_credentials(credentials):
...
+                                  headers={'Plotly-Client-Platform':
+                                           payload['platform']},
+                                  auth=HTTPBasicAuth(payload['un'],
+                                                     payload['key']),
+                                  data={"share_key_enabled":
+                                        "True",
+                                        "world_readable":
+                                        plot_kwargs['world_readable'],
+                                        "filename": plot_kwargs['filename'],
+                                        "fileopt": plot_kwargs['fileopt']})
```
chriddyp: need error checking here in case the response fails. `new_response.raise_for_status()`. we should also have friendly error messages around unsupported enterprise releases: if the `url` isn't public plotly *and* the response is a 404 then it's likely the user was using this on an old plotly enterprise without the `share_key` updates. instead of responding with a confusing requests error, we should explicitly tell them that we suspect that they are using an old enterprise version

plotly/plotly/plotly.py
```diff
@@ -1271,6 +1334,50 @@ def validate_credentials(credentials):
...
+        'sharing': ('public' | 'private' | 'secret')
+                    privacy option of the url
+        'layout': (dict) the layout of the figure
+
+    """
+    domain = (get_session_config().get('plotly_domain') or
+              get_config()['plotly_domain'])
+    plot_id = response['url'].replace(domain + "/~"
+                                      + payload['un'] + "/", "")
+    url = "https://api.plot.ly/v2/files/" + payload['un'] + ":" + plot_id
```
chriddyp: needs to be the `config` api url to handle enterprise

plotly/plotly/plotly.py
```diff
@@ -1271,6 +1334,50 @@ def validate_credentials(credentials):
...
+    """
+    Update plot's url to include the secret key
+
+    payload keyword agruments:
+    platform ('python') -- the platform that is associated with this url
+    version (string) -- the Plotly version that was used to create the plot
+    args (dict) -- contains the data from figure
+    un (string) -- username from credentials file
+    key (string) -- api_key from credentials file
+    origin('plot') --
```
chriddyp: is this used?

plotly/plotly/plotly.py
chriddyp: are these kwargs necessary? filename, filept?

plotly/plotly/plotly.py
```diff
@@ -1271,6 +1334,50 @@ def validate_credentials(credentials):
...
+    kwargs (dict) --
+        'filename': (string) the name that is associated with this url
+        'fileopt': ('new' | 'overwrite' | 'extend' | 'append')
+                   file option of the url
+        'world_readable': (bool) make the plot public/ private
+        'sharing': ('public' | 'private' | 'secret')
+                    privacy option of the url
+        'layout': (dict) the layout of the figure
```
chriddyp: i don't think these are necessary

plotly/plotly/plotly.py
```diff
@@ -150,6 +210,8 @@ def plot(figure_or_data, validate=True, **plot_options):
     auto_open (default=True) -- Toggle browser options
         True: open this plot in a new browser tab
         False: do not open plot in the browser, but do return the unique url
+    share_key_enabled (default=False) -- activate the share key for private
+                                         plots
```
chriddyp: it's `sharing` now right?

plotly/plotly/plotly.py
```diff
@@ -119,21 +154,46 @@ def iplot(figure_or_data, **plot_options):
         'extend': add additional numbers (data) to existing traces
         'append': add additional traces to existing data lists
     world_readable (default=True) -- make this figure private/public
+    share_key_enabled (default=False) -- activate the share key for private
+                                         plots
```
chriddyp: it's `sharing` now right?

plotly/plotly/plotly.py
```diff
@@ -105,6 +108,38 @@ def _plot_option_logic(plot_options):
             'fileopt' not in session_plot_options and
             'fileopt' not in plot_options):
         current_plot_options['fileopt'] = 'overwrite'
+
+    # Check for any conflicts between 'sharing' and 'world_readble'
```
chriddyp: * world_readable

plotly/tools.py
chriddyp: Where are they going to get the fid or url in the first place though? either the web browser (which should include `share_key` in the url anyway) or the return of `py.plot`, which should also include the `share_key` in it. i vote `file_owner_or_url` detecting the presence `share_key` and not worrying about folks using the `share_key` keyword argument

plotly/plotly/plotly.py
aneda: no but it's part of `payload` arguments,   should I just put the ones we use in `add_share_key_to_url`?

plotly/plotly/plotly.py
theengineear: to be more precise, get this from `get_config()` which talks to `session.py` to handle the settings.

aneda:
@theengineear @chriddyp @cldougl 

Thank you for your reviews and help!

I've updated `embed` &  `get_embed` so they wont need `share_key`  (oh man,  I don't know why I was spreading it around everywhere, haha)

Also, updated `response_handeler` and added it to `add_share_key_to_url`  but I haven't written a test for it as I wasn't sure where to add my test.  Should I write a test in `test_plot.py` or add a new file?  Does it need to be in `test_core` or `test_optional`? 

Oh, I also check the warning message in `embed` (based on `plotly_domian`)  not sure if we want to keep it or not

plotly/tools.py
```diff
@@ -352,35 +352,36 @@ def embed(file_owner_or_url, file_id=None, width="100%", height=525):
...
-                session.get_session_config().get('plotly_domain') or
-                get_config_file()['plotly_domain']
-            )
-            url = "{plotly_domain}/~{un}/{fid}".format(
-                plotly_domain=plotly_domain,
-                un=file_owner_or_url,
-                fid=file_id)
-        else:
-            url = file_owner_or_url
+        url = file_owner_or_url
```
theengineear: :cactus: we can't just set `file_owner_or_url` to url! We check `file_id` because:

* `if file_id is not None` we assume `file_owner = file_owner_or_url`
* `if file_id is None` we assume `url = file_owner_or_url`

`get_embed` is public and available to users, any code that might be out there using a username instead of a url would break because of this, right?

plotly/tools.py
```diff
@@ -352,35 +352,36 @@ def embed(file_owner_or_url, file_id=None, width="100%", height=525):
...
+                "Looks like you're not using IPython or Sage to embed this "
+                "plot. If you just want the *embed code*, try using "
+                "`get_embed()` instead."
+                "\nQuestions? support@plot.ly")
+        else:
+            warnings.warn(
+                "Looks like you're not using IPython or Sage to embed this "
+                "plot. If you just want the *embed code*, try using "
+                "`get_embed()` instead."
+                "\nQuestions? feedback@plot.ly")
```
theengineear: :cow2: keep it DRY \o/

Duplications here:

* 90% of that string
* `https://plot.ly` is defined as a default elsewhere

Here's a PR to help out with the `https://plot.ly` checking: #272 feel free to merge that in whenever, I believe we'll need it if we want to return these optional feedback emails.

If that's merged, I'd write this as folows:

```
if get_config_defaults()['plotly_domain'] != get_session_config()['plotly_domain']:
    feedback_email = 'feedback@plot.ly'
else:
    feedback_email = 'support@plot.ly'  # different domain likely means enterprise
warnings.warn('...\nQuestions? {}'.format(feedback_email))
```

And if you *really* want to be DRY, you could also centralize the feedback emails! (but i'd suggest another PR for that) I'll create an issue for it \o/

plotly/plotly/plotly.py
```diff
@@ -1209,14 +1275,20 @@ def parse_grid_id_args(cls, grid, grid_url):
...
     def response_handler(cls, response):
-
         try:
             response.raise_for_status()
         except requests.exceptions.HTTPError as requests_exception:
-            plotly_exception = exceptions.PlotlyRequestError(
-                requests_exception
-            )
-            raise(plotly_exception)
+            if (requests_exception == 404 and
```
theengineear: this works? can you compare a requests_exception to an `int`?

theengineear:
@aneda see comments above ^^

re: https://github.com/plotly/python-api/pull/264#issuecomment-126423837

>I've updated embed & get_embed so they wont need share_key (oh man, I don't know why I was spreading it around everywhere, haha)

:+1: 

>Oh, I also check the warning message in embed (based on plotly_domian) not sure if we want to keep it or not

:+1: see comments there

>Also, updated response_handeler ...

Not sure what that update achieves? What's the point of conditionally changing the error bubbled up based on the `plotly_api_domain` setting? (sorry if i'm just not understanding)

plotly/plotly/plotly.py
```diff
@@ -1271,6 +1343,40 @@ def validate_credentials(credentials):
...
+    kwargs (dict) --
+        'world_readable': (bool) make the plot public/ private
+        'sharing': ('public' | 'private' | 'secret')
+                    set privacy for the url
+
+    """
+    domain = (get_session_config().get('plotly_domain') or
+              get_config()['plotly_domain'])
+    plot_id = response['url'].replace(domain + "/~"
+                                      + payload['un'] + "/", "")
```
theengineear: :cow2: mind working some `ulrparse` magic here as well? the replace *works* but is a little fragile, i think.

plotly/plotly/plotly.py
```diff
@@ -1271,6 +1343,40 @@ def validate_credentials(credentials):
...
+                                  headers=_api_v2.headers(),
+                                  data={"share_key_enabled":
+                                        "True",
+                                        "world_readable":
+                                        plot_kwargs['world_readable']})
+
+    _api_v2.response_handler(new_response)
+
+    new_response_data = json.loads(new_response.content)
+    response['url'] += '?share_key=' + new_response_data['share_key']
```
theengineear: :cow2: this could break if the url returned ever contains the `share_key` already. I *think* it's okay as written, but perhaps something to consider.

plotly/session.py
```diff
@@ -39,7 +39,8 @@
     'fileopt': six.string_types,
     'world_readable': bool,
     'auto_open': bool,
-    'validate': bool
+    'validate': bool,
+    'sharing': dict
```
theengineear: :cow2: ha, back to where we were before, we don't have enough information to validate this. I think it's OK as is though. In fact, I'm convinced the *right* way to do this is validation via JSON schemas http://json-schema.org/

I'll open up a ticket for that too. :)

If you *wanted*, you could add a check to ensure that `sharing in ('public', 'private', 'secret')`. It's really good to lock down the JSON, no surprises!

aneda:
@theengineear @chriddyp @cldougl  Hey, secret_key is ready for reviews!

@theengineear I haven't added a check to `session` to check `sharing in {'public', 'private', 'secret'}`  (https://github.com/plotly/python-api/pull/264#discussion-diff-35946052R43).  I was wondering shouldn't `sharing` be a `string` instead of `dict`?  Similar to `fileopt`

Also, I had added the `response_handerler`  incase response fails (@chriddyp comment:  https://github.com/plotly/python-api/pull/264#discussion-diff-35802432R1374 )
Should I keep it or remove it?  Should I write a test for it?

theengineear:
>shouldn't sharing be a string instead of dict? Similar to fileopt

absolutely, yup :)

theengineear:
>Should I keep it or remove it? Should I write a test for it?

Keep it, otherwise I think a weird `requests` error bubbles up.

plotly/plotly/plotly.py
```diff
@@ -150,6 +204,11 @@ def plot(figure_or_data, validate=True, **plot_options):
     auto_open (default=True) -- Toggle browser options
         True: open this plot in a new browser tab
         False: do not open plot in the browser, but do return the unique url
+    sharing ('public' | 'private' | 'secret')
+        'public': set the privacy of the plot to public
+        'private': set the privacy of the plot as private
```
chriddyp:
```
sharing ('public' | 'private' | 'sharing') -- Toggle who can view this graph
- `'public': Anyone can view this graph. It will appear in your profile and can appear in search engines. You do not need to be logged in to Plotly to view this chart.
- `'private': Only you can view and edit this plot. It will not appear in the Plotly feed, your profile, or search engines. You must be logged in to Plotly to view this graph. You can privately share this graph with other Plotly users in your online Plotly account and they will need to be logged in to view this plot.
- 'secret': Anyone with this secret link can view this chart. It will not appear in the Plotly feed, your profile, or search engines. If it is embedded inside a webpage or an IPython notebook, anybody who is viewing that page will be able to view the graph. You do not need to be logged in to view this plot.

File: plotly/plotly/plotly.py [hunk 30]
```diff
@@ -1209,14 +1275,20 @@ def parse_grid_id_args(cls, grid, grid_url):
...
-
         try:
             response.raise_for_status()
         except requests.exceptions.HTTPError as requests_exception:
-            plotly_exception = exceptions.PlotlyRequestError(
-                requests_exception
-            )
-            raise(plotly_exception)
+            if (response.status_code == 404 and
+                    get_config()['plotly_api_domain']
```
chriddyp: is `get_config()[...]` for sure what was used to make this request? Could it have been something different if they `py.sign_in`'d, e.g. something from `session`? (I'm not actually sure, I just thought `get_config` always grabs the config from file)

plotly/plotly/plotly.py
```diff
@@ -1209,14 +1275,20 @@ def parse_grid_id_args(cls, grid, grid_url):
...
+                    != 'https://api.plot.ly'):
+                raise exceptions.PlotlyError(
+                    " "
```
chriddyp: "This endpoint is unavailable at {url}. If you are using Plotly Enterprise, you may need to upgrade your Plotly Enterprise server to request against this endpoint or this endpoint may not be available yet. Questions? support@plot.ly"

plotly/plotly/plotly.py
```diff
@@ -1209,14 +1275,20 @@ def parse_grid_id_args(cls, grid, grid_url):
...
+                )
+            else:
+                plotly_exception = exceptions.PlotlyRequestError(
+                    requests_exception
+                )
+                raise(plotly_exception)
```
chriddyp: otherwise, I think we should just re-raise `requests_exception`. thoughts @theengineear ?

plotly/plotly/plotly.py
```diff
@@ -1271,6 +1343,40 @@ def validate_credentials(credentials):
...
+    Update plot's url to include the secret key
+
+    payload keyword agruments:
+    un (string) -- username from credentials file
+    kwargs (dict) --
+        'world_readable': (bool) make the plot public/ private
+        'sharing': ('public' | 'private' | 'secret')
+                    set privacy for the url
+
+    """
```
chriddyp: could we unpack these input arguments to make it more user-friendly? It'll be easier to test that way, and a user could actually use this function if they wanted to. In other words, take `(url, un, world_readable, sharing)` as the call signature

plotly/plotly/plotly.py
chriddyp: yeah, good point. `urlparse` can get you the query strings: https://docs.python.org/2/library/urlparse.html. This might end up being a useful function for folks once we have a `plotly.files.get` (i.e. polity's `$ ls`), so probably worth making this foolproof.

plotly/session.py
```diff
@@ -115,6 +117,13 @@ def update_session_plot_options(**kwargs):
             raise exceptions.PlotlyError("{} must be of type '{}'"
                                          .format(key, PLOT_OPTIONS[key]))
 
+        # raise exception if sharing is invalid
+        if (key == 'sharing' and not (kwargs[key] in
+                                      ['public', 'private', 'secret'])):
+            raise exceptions.PlotlyError("'{}' must be of either {}"
+                                         .format(key, "'public', 'private'" +
+                                                 " or'secret'"))
```
chriddyp: missing space between or and secret

plotly/tests/test_core/test_plotly/test_plot.py
```diff
@@ -66,3 +66,99 @@ def test_plot_empty_data(self):
...
+
+        plot_option_logic = py._plot_option_logic(kwargs)
+
+        expected_plot_option_logic = {'filename': 'test',
+                                      'auto_open': True,
+                                      'fileopt': 'overwrite',
+                                      'validate': True,
+                                      'world_readable': False,
+                                      'sharing': 'private'}
+        self.assertEqual(plot_option_logic, expected_plot_option_logic)
```
chriddyp: great tests!

plotly/tools.py
```diff
@@ -298,15 +297,21 @@ def get_embed(file_owner_or_url, file_id=None, width="100%", height=525):
...
+        else:
+            share_key = ''
     else:
         file_owner = file_owner_or_url
-    resource = "/apigetfile/{file_owner}/{file_id}".format(file_owner=file_owner,
-                                                           file_id=file_id)
+        share_key = ''
     try:
-        test_if_int = int(file_id)
+        test_if_int = int(0 if file_id is None else file_id)
```
chriddyp: clever, nice

plotly/plotly/plotly.py
```diff
@@ -119,21 +154,40 @@ def iplot(figure_or_data, **plot_options):
...
+        pass
+    else:
+        width = str(width) + 'px'
+
+    try:
+        float(height)
+    except (ValueError, TypeError):
+        pass
+    else:
+        height = str(height) + 'px'
```
chriddyp: nice, thanks for fixing this one. this bug was driving me crazy.

plotly/tools.py
```diff
@@ -285,7 +285,6 @@ def get_embed(file_owner_or_url, file_id=None, width="100%", height=525):
                               figure
 
     """
-    padding = 25
```
chriddyp: nice :)

plotly/tools.py
```diff
@@ -387,11 +394,19 @@ def embed(file_owner_or_url, file_id=None, width="100%", height=525):
             url = file_owner_or_url
         return PlotlyDisplay(url, width, height)
     else:
+        if (get_config_defaults()['plotly_domain']
+                != session.get_session_config()['plotly_domain']):
+            feedback_email = 'feedback@plot.ly'
```
chriddyp: @msund is it *just* you checking these emails?

chriddyp:
@theengineear 
> > Also, updated response_handeler ...

> Not sure what that update achieves? What's the point of conditionally changing the error bubbled up based on the plotly_api_domain setting? (sorry if i'm just not understanding)

this is to provide better "endpoint not supported" behaviour for folks running old versions of enterprise that don't have the new api endpoints on their server

aneda:
@chriddyp 
> is get_config()[...] for sure what was used to make this request? Could it have been something different > if they py.sign_in'd, e.g. something from session?

when I replace `get_config` with `session_get_config` the `grid_test` fails.   What should I do here?

Also,

> could we unpack these input arguments to make it more user-friendly?

Good point!  We don't really need anything but the `url`,  is it ok if I reduce it to one argument? Do I need to check the `url` (in case users call it by the wrong one)

Also, do I need to write tests for stuff I added in `update_session` and  `response_handler` ?

plotly/tools.py
msund: @chriddyp yep, Jack and myself.

aneda:
ok, I think it's ready!  Can I have another review please

Just to keep everyone in the loop @chriddyp and I discuss my questions above

> > is get_config()[...] for sure what was used to make this request? Could it have been something 
> > different  if they py.sign_in'd, e.g. something from session?
> when I replace get_config with session_get_config the grid_test fails. What should I do here?

`get_config` is the correct one 

> >  could we unpack these input arguments to make it more user-friendly?
> Good point! We don't really need anything but the url, is it ok if I reduce it to one argument? 

I changed the argument for `add_share_key_to_url` so it only get the `response['url']` /  plot url

plotly/plotly/plotly.py
theengineear: @chriddyp [`get_config`](https://github.com/plotly/python-api/blob/master/plotly/plotly/plotly.py#L67-L77) should be right.

plotly/plotly/plotly.py
theengineear: yup, i think the requests exception is appropriate for the latter case.

plotly/tools.py
theengineear: ? this will break just below if `file_id=None`, we use `int(file_id)` which can raise an Error?

Why would we want `file_id` to default to `0` anyways?

Did you change this because `None` was raising a `TypeError` instead of a `ValueError`? If that's the case, we should just:

`except (ValueError, TypeError):`

plotly/plotly/plotly.py
```diff
@@ -119,21 +154,52 @@ def iplot(figure_or_data, **plot_options):
...
-        embed_options['height'] = plot_options['height']
+    try:
+        float(height)
+    except (ValueError, TypeError):
+        pass
+    else:
+        height = str(height) + 'px'
 
-    return tools.embed(username, plot_id, **embed_options)
+    return tools.embed(url, file_id, width, height)
```
theengineear: I believe either (a) this should break or (b) we've made `tools.embed` backwards incompatible. The way it *used* to work was to allow either `tools.embed(url, **embed_options)` or `tools.embed(username, file_id, **embed_options)`

By calling it like `tools.embed(url, file_id, **embed_options)`, it *should* think that the url param is a username and I'd imagine it would break.

plotly/plotly/plotly.py
```diff
@@ -212,7 +297,8 @@ def iplot_mpl(fig, resize=True, strip_style=False, update=None, **plot_options):
             fig.update(update)
             fig.validate()
         except exceptions.PlotlyGraphObjectError as err:
-            err.add_note("Your updated figure could not be properly validated.")
+            err.add_note("Your updated figure could not be properly " +
+                         "validated.")
```
theengineear: :cow2: Python's smart enough to concatenate your strings for you:

```
err.add_note("Your updated figure could not be properly"
             "validated")
```

This works because of the `()`

plotly/plotly/plotly.py
```diff
@@ -250,7 +336,8 @@ def plot_mpl(fig, resize=True, strip_style=False, update=None, **plot_options):
...
```
theengineear: Ditto: https://github.com/plotly/python-api/pull/264/files#r36324849

plotly/plotly/plotly.py
```diff
@@ -1271,6 +1372,31 @@ def validate_credentials(credentials):
...
+                                        "True",
+                                        "world_readable":
+                                        "False"})
+
+    _api_v2.response_handler(new_response)
+
+    new_response_data = json.loads(new_response.content)
+    plot_url += '?share_key=' + new_response_data['share_key']
+
+    return plot_url
```
theengineear: \o/ much better!

plotly/session.py
```diff
@@ -115,6 +117,13 @@ def update_session_plot_options(**kwargs):
             raise exceptions.PlotlyError("{} must be of type '{}'"
                                          .format(key, PLOT_OPTIONS[key]))
 
+        # raise exception if sharing is invalid
+        if (key == 'sharing' and not (kwargs[key] in
+                                      ['public', 'private', 'secret'])):
+            raise exceptions.PlotlyError("'{}' must be of either {}"
+                                         .format(key, "'public', 'private'" +
+                                                 " or 'secret'"))
```
theengineear: :dromedary_camel: (DRY :cow2:) You're repeating `'public', 'private', 'secret'`

There should be a variable somewhere that controls that, like `SHARING_OPTIONS`. then you can use it in these two places and it can remain centralized in the future.

plotly/tests/test_core/test_plotly/test_plot.py
```diff
@@ -66,3 +66,99 @@ def test_plot_empty_data(self):
...
+                {
+                    'x': [1, 2, 3],
+                    'y': [2, 1, 2]
+                }
+            ]
+        }
+
+        kwargs = {'filename': 'invalid-sharing-argument',
+                  'sharing': 'privste'}
+        self.assertRaises(PlotlyError, py.plot, fig, **kwargs)
```
theengineear: :tiger2: This can be stronger/more-readable if you use [`self.assertRaisesRegexp`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaisesRegexp) by opening a *context*

```
with self.assertRaisesRegexp(PlotlyError, 'the-exact-message-would-work'):
    py.plot(fig, filename='invalid-sharing-argument', sharing='privste')
```

If you feel comfy with regex's, you can even make it flexible by just making sure the keyword 'sharing' is in there.

plotly/tests/test_core/test_plotly/test_plot.py
```diff
@@ -66,3 +66,99 @@ def test_plot_empty_data(self):
...
+                    'x': [1, 2, 3],
+                    'y': [2, 1, 2]
+                }
+            ]
+        }
+
+        kwargs = {'filename': 'invalid-privacy-setting',
+                  'world_readable': False,
+                  'sharing': 'public'}
+        self.assertRaises(PlotlyError, py.plot, fig, **kwargs)
```
theengineear: ditto: https://github.com/plotly/python-api/pull/264/files#r36326529

plotly/tests/test_core/test_plotly/test_plot.py
```diff
@@ -66,3 +66,99 @@ def test_plot_empty_data(self):
...
+                    'x': [1, 2, 3],
+                    'y': [2, 1, 2]
+                }
+            ]
+        }
+
+        kwargs = {'filename': 'invalid-privacy-setting',
+                  'world_readable': True,
+                  'sharing': 'secret'}
+        self.assertRaises(PlotlyError, py.plot, fig, **kwargs)
```
theengineear: ditto: https://github.com/plotly/python-api/pull/264/files#r36326529

plotly/tests/test_core/test_plotly/test_plot.py
theengineear: ditto: https://github.com/plotly/python-api/pull/264/files#r36220969 :smile_cat:

plotly/tools.py
aneda: I changed it because as I was debugging I used to get a wired case of `int(None)` and wanted to avoid it.  But all is good now so I'll remove it

plotly/tools.py
```diff
@@ -363,18 +366,22 @@ def embed(file_owner_or_url, file_id=None, width="100%", height=525):
...
+
         # see if we are in the SageMath Cloud
         from sage_salvus import html
         return html(s, hide=False)
     except:
         pass
     if _ipython_imported:
-        if file_id:
+        if (file_id and
+                get_config_file()['plotly_domain'] not in file_owner_or_url):
```
theengineear: Is this change necessary? embed isn't meant to be called with a `url` *and* a `file_id` param, it's meant to be called with a `url` param **or** a `file_id` param *and* a `username` param.

plotly/tools.py
```diff
@@ -387,11 +394,19 @@ def embed(file_owner_or_url, file_id=None, width="100%", height=525):
...
+
         warnings.warn(
-            "Looks like you're not using IPython or Sage to embed this plot. "
-            "If you just want the *embed code*, try using `get_embed()` "
-            "instead."
-            "\nQuestions? support@plot.ly")
+            "Looks like you're not using IPython or Sage to embed this "
+            "plot. If you just want the *embed code*,\ntry using "
+            "`get_embed()` instead."
+            '\nQuestions? {}'.format(feedback_email))
```
theengineear: \o/

theengineear:
@aneda , mostly just concerned with the changes related to `embed` function calls. sorry if i'm missing something obvious there

plotly/tools.py
aneda: I thought when we call `embed` from `iplot` we give it both the `url` and `file_id`.   Can I remove that?  I used to get an error that's why I checked the url

theengineear: The old call looked like [this](https://github.com/plotly/python-api/blob/master/plotly/plotly/plotly.py#L136):

`return tools.embed(username, plot_id, **embed_options)`

Note that we're sending `username` and `plot_id` *not* `url`. It's *better* to use these since the logic for this case is simpler, but *just* sending `url` (no `file_id`) is OK too.

aneda:
@theengineear 

Thanks for your reviews!

>@aneda , mostly just concerned with the changes related to embed function calls.

I change the `embed` call to `return tools.embed(url, **embed_options)` in `iplot`.  I had to keep `url` since we do all the checking for `sharing` in `iplot`.  I can change that so we check everything in `embed` but I thought it's better to keep everything in one file.  Any suggestions?

aneda:
Oh, also, can I everyone get another review please?!

chriddyp:
- [ ] test on enterprise

plotly/plotly/plotly.py
```diff
@@ -105,6 +108,38 @@ def _plot_option_logic(plot_options):
...
+                raise exceptions.PlotlyError(
+                    "Looks like you are setting your plot privacy to both "
+                    "public and private.\n If you set world_readable as "
+                    "False, sharing can only be set to 'private' or 'secret'")
+        else:
+            raise exceptions.PlotlyError(
+                "The 'sharing' argument only accepts one of the following "
+                "strings:\n'public' -- for public plots\n"
+                "'private' -- for private plots\n"
+                "'secret' -- for private plots that can be shared \n"
```
chriddyp: "for private plots that can be shared with a secret url"

plotly/plotly/plotly.py
chriddyp: @aneda ^^

plotly/plotly/plotly.py
```diff
@@ -1209,14 +1298,25 @@ def parse_grid_id_args(cls, grid, grid_url):
...
-            raise(plotly_exception)
+            if (response.status_code == 404 and
+                    get_config()['plotly_api_domain']
+                    != tools.get_config_defaults()['plotly_api_domain']):
+                raise exceptions.PlotlyError(
+                    "This endpoint is unavailable at {url}. If you are using "
+                    "Plotly Enterprise, you may need to upgrade your Plotly "
+                    "Enterprise server to request against this endpoint or "
+                    "this endpoint may not be available yet.\nQuestions? "
+                    "support@plot.ly".format(url=get_config()
```
chriddyp: Questions? support@plot.ly or your plotly administrator.

plotly/tests/test_core/test_plotly/test_plot.py
chriddyp: @aneda - can we add a test that this actually works? i.e. 
- make a request to plotly with the share key
- confirm that the returned url has a share key in it
- make another get request to that url and make sure we get a 200 level response
- make another request to the url *without* the share key and make sure we get a 404 level response
it's technically testing stuff on the plotly server side, but it's good for this package and its tests to be aware that it is interfacing with plotly as expected.

chriddyp: similarly for the `add_share_key_to_url` function: make a private plot, check that you can't access it, then add a share key, and check that you can access it if you have the share key

aneda:
Not sure why both 8d13f50 and dcaacfd are failing.  They pass both python 2.7 and python 3.4 on my end...

chriddyp:
@aneda - did you see the errors on circle? looks like 1 error with 3.3:

```
ERROR: test_plot_url_given_sharing_key (plotly.tests.test_core.test_plotly.test_plot.TestPlot)
  File "/home/ubuntu/python-api/plotly/tests/test_core/test_plotly/test_plot.py", line 194, in test_plot_url_given_sharing_key
    plot_url = py._send_to_plotly(fig, **kwargs)['url']
  File "/home/ubuntu/python-api/plotly/plotly/plotly.py", line 1440, in _send_to_plotly
    r['url'] = add_share_key_to_url(r['url'])
  File "/home/ubuntu/python-api/plotly/plotly/plotly.py", line 1394, in add_share_key_to_url
    new_response_data = json.loads(new_response.content)
  File "/home/ubuntu/.pyenv/versions/3.3.3/lib/python3.3/json/__init__.py", line 319, in loads
    return _default_decoder.decode(s)
  File "/home/ubuntu/.pyenv/versions/3.3.3/lib/python3.3/json/decoder.py", line 352, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
nose.proxy.TypeError: can't use a string pattern on a bytes-like object
-------------------- >> begin captured logging << --------------------
```

aneda:
@chriddyp yeah but I don't know how to fix it.  It passes on my laptop, is it because I have 3.4?

aneda:
I will look more into it tonight

chriddyp:
@aneda yeah, it could be a 3.3 issue

aneda:
Ok, finally it passed all tests!  Big thanks to @chriddyp  for  https://github.com/plotly/python-api/commit/abc71799388ca756241678d9576db9d5506ff5a6

Can I get another review please?!

chriddyp:
looks great to me! :dancer:

chriddyp:
(and after this gets merged, lets update the appropriate documentation: the [getting started](https://github.com/plotly/documentation/blob/source/_includes/python-getting-started.html) and the [privacy section](https://github.com/plotly/documentation/blob/source/_posts/pandas/privacy/2015-04-09-privacy_python_index.html)

theengineear:
Re: https://github.com/plotly/python-api/pull/264#issuecomment-130103729 , @aneda can you either open a PR for this or open an issue in the 1.7 milestone so we don't forget?

theengineear:
@chriddyp 's comment here still needs addressing:

https://github.com/plotly/python-api/pull/264/files#r36436323

plotly/session.py
```diff
@@ -115,6 +119,12 @@ def update_session_plot_options(**kwargs):
             raise exceptions.PlotlyError("{} must be of type '{}'"
                                          .format(key, PLOT_OPTIONS[key]))
 
+        # raise exception if sharing is invalid
+        if (key == 'sharing' and not (kwargs[key] in SHARING_OPTIONS)):
+            raise exceptions.PlotlyError("'{0}' must be of either '{1}', '{2}'"
+                                         " or '{3}'"
+                                         .format(key, *SHARING_OPTIONS))
```
theengineear: so DRY! :smile_cat:

plotly/tests/test_core/test_plotly/test_plot.py
```diff
@@ -66,3 +70,187 @@ def test_plot_empty_data(self):
...
+        # correct arguments {'public, 'private', 'secret'}
+
+        fig = {
+            'data': [
+                {
+                    'x': [1, 2, 3],
+                    'y': [2, 1, 2]
+                }
+            ]
+        }
```
theengineear: :cow2: mostly personal preference, but i we usually favor tighter formatting for this:

```
fig = {'data': [{'x': [1, 2, 3], 'y': [2, 1, 2]}]}
```

Fits on a single line i think and would probably improve readability.

plotly/tests/test_core/test_plotly/test_plot.py
```diff
@@ -66,3 +70,187 @@ def test_plot_empty_data(self):
...
+        # Raise an error if world_readable=False but sharing='public'
+
+        fig = {
+            'data': [
+                {
+                    'x': [1, 2, 3],
+                    'y': [2, 1, 2]
+                }
+            ]
+        }
```
theengineear: ditto https://github.com/plotly/python-api/pull/264/files#r36833746

8 lines could be on 1 with similar readability, imo.

plotly/tests/test_core/test_plotly/test_plot.py
```diff
@@ -66,3 +70,187 @@ def test_plot_empty_data(self):
...
+        # Raise an error if world_readable=True but sharing='secret'
+
+        fig = {
+            'data': [
+                {
+                    'x': [1, 2, 3],
+                    'y': [2, 1, 2]
+                }
+            ]
+        }
```
theengineear: Looks like you use this *same* figure a lot, could you consider putting it in the `TestCase.setUp` method?

```
class TestPlot(TestCase):

    def setUp(self):
        self.simple_figure = {'data': [{'x': [1, 2, 3], 'y': [2, 1, 2]}]}
        super(TestPlot, self).setUp()

    # yadda, yadda
```

`unittest` will run `setUp` *before* each test method.

Making changes like these makes it easier to distinguish setup-logic from testing logic.

plotly/tests/test_core/test_plotly/test_plot.py
```diff
@@ -66,3 +70,187 @@ def test_plot_empty_data(self):
...
+
+        kwargs = {'filename': 'invalid-privacy-setting',
+                  'world_readable': True,
+                  'sharing': 'secret'}
+
+        with self.assertRaisesRegexp(PlotlyError, 'sharing'):
```
theengineear: :smile_cat:

plotly/tests/test_core/test_plotly/test_plot.py
```diff
@@ -66,3 +70,187 @@ def test_plot_empty_data(self):
...
+
+        expected_plot_option_logic = {'filename': 'test',
+                                      'auto_open': True,
+                                      'fileopt': 'overwrite',
+                                      'validate': True,
+                                      'world_readable': False,
+                                      'sharing': 'private'}
+        self.assertEqual(plot_option_logic, expected_plot_option_logic)
+
+    # @skip('added to see if everything else passes on circle')
```
theengineear: can we get rid of this comment then?

plotly/tests/test_core/test_plotly/test_plot.py
```diff
@@ -66,3 +70,187 @@ def test_plot_empty_data(self):
...
+        # the share_key
+
+        data = {
+            'data': [
+                {
+                    'x': [1, 2, 3],
+                    'y': [2, 1, 2]
+                }
+            ]
+        }
```
theengineear: ditto: https://github.com/plotly/python-api/pull/264/files#r36834089

plotly/tests/test_core/test_plotly/test_plot.py
```diff
@@ -66,3 +70,187 @@ def test_plot_empty_data(self):
...
+        fig = tls.return_figure_from_figure_or_data(data, validate)
+        kwargs = {'filename': 'is_share_key_included',
+                  'fileopt': 'overwrite',
+                  'world_readable': False,
+                  'sharing': 'secret'}
+
+        self.assertTrue(200 ==
+                        requests.get(
+                            py._send_to_plotly(fig, **kwargs)['url'])
+                        .status_code)
```
theengineear: :cow2: This makes debugging the test pretty hard. It'd be better to write it like this:

```
# explain why we need to use _send_to_plotly instead of .plot or something?
plot_url = py._send_to_plotly(fig, **kwargs)['url']
response = requests.get(plot_url)
self.assertEqual(response.status_code, 200)
```

Same amount of lines and it will be much clearer to understand and debug.

plotly/tests/test_core/test_session/test_session.py
```diff
@@ -0,0 +1,17 @@
...
+
+class TestSession(TestCase):
+
+    def test_update_session_plot_options_sharing(self):
+
+        # Return PlotlyError when sharing arguement is not
+        # 'public', 'private' or 'secret'
+
+        kwargs = {'sharing': 'priva'}
+        self.assertRaises(PlotlyError, update_session_plot_options, **kwargs)
```
theengineear: This looks like it tests that it fails, can we also test that it works for all SHARING_OPTIONS?

plotly/tools.py
```diff
@@ -363,18 +366,21 @@ def embed(file_owner_or_url, file_id=None, width="100%", height=525):
...
+                      height=height)
+
         # see if we are in the SageMath Cloud
         from sage_salvus import html
         return html(s, hide=False)
     except:
         pass
     if _ipython_imported:
-        if file_id:
+        if (file_id):
```
theengineear: :cat: why the parentheses?

theengineear:
non-blocking from me, i think there's a comment from chris that still needs to be checked and some general code style things that could be improved.

Feel free to :dancer: after you've taken care of https://github.com/plotly/python-api/pull/264#issuecomment-130103089 and https://github.com/plotly/python-api/pull/264/files#r36436323

PR #268, opened 2015-07-28
Decrease chance of collisions… by a lot. (more)

theengineear:
Now we need > 10,000,000 attempts before having a .01% collision chance.
I.e., should be good.

Referencing: https://github.com/plotly/python-api/commit/e59f6ad3736b5d19b92fc85c8713c6092b51693f

### DIFF
--- a/plotly/tests/test_core/test_file/test_file.py
+++ b/plotly/tests/test_core/test_file/test_file.py
@@ -8,7 +8,6 @@ A module intended for use with Nose.
 import random
 import string
 from unittest import TestCase
-from nose.tools import nottest
 
 import plotly.plotly as py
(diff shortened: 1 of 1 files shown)


theengineear:
@chriddyp @BRONSOLO 

This should really not be statistically possible now. In the future, it'd be nice to just batch delete all the test-users files between tests... this is a pain right now though... well, sorta. It's just a bit time consuming--requiring about `2N + 1` requests where N is the number of files the test user has. One request to get all the files, one request *per file* to trash and one request *per file* to permanent delete after it's been trashed.

For example the `PythonTest` user has > 10,000 folders : p

BRONSOLO:
:dancer: good catch!

PR #272, opened 2015-07-31
Add defaults check

theengineear:
Convenience method to help us check differences between config defaults and user config.


theengineear:
@aneda

theengineear:
@chriddyp , i think this will help us sort out the email issue with feedback, which would be nice :)

PR #274, opened 2015-08-04
Dendrogram class

OxanaSachenkova:
Hi all, please have a look at the dendrogram wrapper and let me know if you want to add anything else at this point. 
All feedback is welcome!

### DIFF
--- a/plotly/tests/test_core/test_tools/test_figure_factory.py
+++ b/plotly/tests/test_core/test_tools/test_figure_factory.py
@@ -4,6 +4,8 @@ from unittest import TestCase
 import datetime
 from nose.tools import raises
 
+import numpy as np 
+
 import plotly.tools as tls
 from plotly.exceptions import PlotlyError
 from plotly.graph_objs import graph_objs
@@ -801,4 +803,50 @@ class TestFinanceCharts(TestCase):
                                                      35.831999999999994]}}}
 
         self.assertEqual(candle, exp_candle)
+
+class TestDendrogram(TestCase):
+
+    def test_default_dendrogram(self):
+        dendro = tls.TraceFactory.create_dendrogram(X=[[1, 2, 3, 4],
+                                                       [1, 1, 3, 4],
+                                                       [1, 2, 1, 4],
+                                                       [1, 2, 3, 1]])
+        expected_dendro_data = [{'marker': {'color': 'rgb(255,133,27)'},
+                                 'mode': 'lines', 'xaxis': 'xs',
+                                 'yaxis': 'ys',
+                                 'y': np.array([0.,  1.,  1.,  0.]),
+                                 'x': np.array([25.,  25.,  35.,  35.]),
+                                 'type': u'scatter'},
+                                {'marker': {'color': 'rgb(255,133,27)'},
+                                 'mode': 'lines',
+                                 'xaxis': 'xs',
+                                 'yaxis': 'ys',
+                                 'y': np.array([0., 2.23606798, 2.23606798, 1.]),
+                                 'x': np.array([15., 15., 30., 30.]),
+                                 'type': u'scatter'},
+                                {'marker': {'color': 'blue'},
+                                 'mode': 'lines',
+                                 'xaxis': 'xs',
+                                 'yaxis': 'ys',
+                                 'y': np.array([0., 3.60555128, 3.60555128, 2.23606798]),
+                                 'x': np.array([5., 5., 22.5, 22.5]), 'type': u'scatter'}]
+       
+        self.assertEqual(len(dendro.data), len(expected_dendro_data))
+        self.assertTrue(np.array_equal(dendro.labels, np.array(['3', '2', '0', '1'])))
+
+        for i in range(1,len(dendro.data)):
+          self.assertTrue(np.allclose(dendro.data[i]['x'], expected_dendro_data[i]['x']))
+          self.assertTrue(np.allclose(dendro.data[i]['y'], expected_dendro_data[i]['y']))
+
+    def test_dendrogram_random_matrix(self):
+        # create a random uncorrelated matrix
+        X = np.random.rand(5,5) 
+        # variable 2 is correlated with all the other variables
+        X[2,:] = sum(X,0)
+
+        dendro = tls.TraceFactory.create_dendrogram(X)
+
+        # Check that 2 is in a separate cluster
+        self.assertEqual(dendro.labels[0], '2')
+
 
--- a/plotly/tests/test_core/test_tools/test_figure_factory.py
+++ b/plotly/tests/test_core/test_tools/test_figure_factory.py
@@ -807,10 +807,10 @@ class TestFinanceCharts(TestCase):
 class TestDendrogram(TestCase):
 
     def test_default_dendrogram(self):
-        dendro = tls.TraceFactory.create_dendrogram(X=[[1, 2, 3, 4],
+        dendro = tls.FigureFactory.create_dendrogram(X=np.array([[1, 2, 3, 4],
                                                        [1, 1, 3, 4],
                                                        [1, 2, 1, 4],
-                                                       [1, 2, 3, 1]])
+                                                       [1, 2, 3, 1]]))
         expected_dendro_data = [{'marker': {'color': 'rgb(255,133,27)'},
                                  'mode': 'lines', 'xaxis': 'xs',
                                  'yaxis': 'ys',
@@ -844,7 +844,7 @@ class TestDendrogram(TestCase):
         # variable 2 is correlated with all the other variables
         X[2,:] = sum(X,0)
 
-        dendro = tls.TraceFactory.create_dendrogram(X)
+        dendro = tls.FigureFactory.create_dendrogram(X)
 
         # Check that 2 is in a separate cluster
         self.assertEqual(dendro.labels[0], '2')
--- a/plotly/tests/test_core/test_tools/test_figure_factory.py
+++ b/plotly/tests/test_core/test_tools/test_figure_factory.py
@@ -3,9 +3,6 @@ from unittest import TestCase
 
 import datetime
 from nose.tools import raises
-
-import numpy as np 
-
 import plotly.tools as tls
 from plotly.exceptions import PlotlyError
 from plotly.graph_objs import graph_objs
@@ -803,50 +800,4 @@ class TestFinanceCharts(TestCase):
                                                      35.831999999999994]}}}
 
         self.assertEqual(candle, exp_candle)
-
-class TestDendrogram(TestCase):
-
-    def test_default_dendrogram(self):
-        dendro = tls.FigureFactory.create_dendrogram(X=np.array([[1, 2, 3, 4],
-                                                       [1, 1, 3, 4],
-                                                       [1, 2, 1, 4],
-                                                       [1, 2, 3, 1]]))
-        expected_dendro_data = [{'marker': {'color': 'rgb(255,133,27)'},
-                                 'mode': 'lines', 'xaxis': 'xs',
-                                 'yaxis': 'ys',
-                                 'y': np.array([0.,  1.,  1.,  0.]),
-                                 'x': np.array([25.,  25.,  35.,  35.]),
-                                 'type': u'scatter'},
-                                {'marker': {'color': 'rgb(255,133,27)'},
-                                 'mode': 'lines',
-                                 'xaxis': 'xs',
-                                 'yaxis': 'ys',
-                                 'y': np.array([0., 2.23606798, 2.23606798, 1.]),
-                                 'x': np.array([15., 15., 30., 30.]),
-                                 'type': u'scatter'},
-                                {'marker': {'color': 'blue'},
-                                 'mode': 'lines',
-                                 'xaxis': 'xs',
-                                 'yaxis': 'ys',
-                                 'y': np.array([0., 3.60555128, 3.60555128, 2.23606798]),
-                                 'x': np.array([5., 5., 22.5, 22.5]), 'type': u'scatter'}]
-       
-        self.assertEqual(len(dendro.data), len(expected_dendro_data))
-        self.assertTrue(np.array_equal(dendro.labels, np.array(['3', '2', '0', '1'])))
-
-        for i in range(1,len(dendro.data)):
-          self.assertTrue(np.allclose(dendro.data[i]['x'], expected_dendro_data[i]['x']))
-          self.assertTrue(np.allclose(dendro.data[i]['y'], expected_dendro_data[i]['y']))
-
-    def test_dendrogram_random_matrix(self):
-        # create a random uncorrelated matrix
-        X = np.random.rand(5,5) 
-        # variable 2 is correlated with all the other variables
-        X[2,:] = sum(X,0)
-
-        dendro = tls.FigureFactory.create_dendrogram(X)
-
-        # Check that 2 is in a separate cluster
-        self.assertEqual(dendro.labels[0], '2')
-
 
--- a/plotly/tests/test_optional/test_opt_tracefactory.py
+++ b/plotly/tests/test_optional/test_opt_tracefactory.py
@@ -136,9 +136,9 @@ class TestDendrogram(TestCase):
                           'x': np.array([5., 5., 22.5, 22.5]),
                           'type': u'scatter'}]
         expected_layout = {'width': '100%',
-                            'showlegend': False,
-                            'autoscale': False,
-                            'x': {'showticklabels': True,
+                           'showlegend': False,
+                           'autoscale': False,
+                           'xaxis': {'showticklabels': True,
                                      'tickmode': 'array',
                                      'ticks': 'outside',
                                      'showgrid': False,
@@ -152,15 +152,15 @@ class TestDendrogram(TestCase):
                                      'type': 'linear',
                                      'tickvals': np.array([5.0, 15.0,
                                                           25.0, 35.0])},
-                            'y': {'showticklabels': True,
-                                  'ticks': 'outside',
-                                  'showgrid': False,
-                                  'mirror': 'allticks',
-                                  'zeroline': False,
-                                  'showline': True,
-                                  'rangemode': 'tozero',
-                                  'type': 'linear'},
-                            'hovermode': 'closest'}  
+                           'yaxis': {'showticklabels': True,
+                                     'ticks': 'outside',
+                                     'showgrid': False,
+                                     'mirror': 'allticks',
+                                     'zeroline': False,
+                                     'showline': True,
+                                     'rangemode': 'tozero',
+                                     'type': 'linear'},
+                           'hovermode': 'closest'}  
 
         # Make sure data is as expected
         self.assertEqual(len(dendro['data']), len(expected_data))
@@ -171,12 +171,12 @@ class TestDendrogram(TestCase):
                             expected_data[i]['y']))
 
         # Make sure layout is as expected
-        self.assertTrue(np.array_equal(dendro['layout']['x']['ticktext'],
-                                       expected_layout['x']['ticktext']))
-        self.assertTrue(np.array_equal(dendro['layout']['x']['tickvals'],
-                                       expected_layout['x']['tickvals']))
-        self.assertEqual(dendro['layout']['x']['ticks'], 'outside')
-        self.assertEqual(dendro['layout']['y']['ticks'], 'outside')
+        self.assertTrue(np.array_equal(dendro['layout']['xaxis']['ticktext'],
+                                       expected_layout['xaxis']['ticktext']))
+        self.assertTrue(np.array_equal(dendro['layout']['xaxis']['tickvals'],
+                                       expected_layout['xaxis']['tickvals']))
+        self.assertEqual(dendro['layout']['xaxis']['ticks'], 'outside')
+        self.assertEqual(dendro['layout']['yaxis']['ticks'], 'outside')
         self.assertEqual(dendro['layout']['width'], expected_layout['width'])
 
     def test_dendrogram_random_matrix(self):
@@ -189,28 +189,28 @@ class TestDendrogram(TestCase):
         dendro = tls.FigureFactory.create_dendrogram(X, labels=names)
 
         # Check that 2 is in a separate cluster and it's labelled correctly
-        self.assertEqual(dendro['layout']['x']['ticktext'][0], 'John')
+        self.assertEqual(dendro['layout']['xaxis']['ticktext'][0], 'John')
 
     def test_dendrogram_orientation(self):
         X = np.random.rand(5, 5) 
 
         dendro_left = tls.FigureFactory.create_dendrogram(X, orientation='left')
 
-        self.assertEqual(len(dendro_left['layout']['y']['ticktext']), 5)
-        tickvals_left = np.array(dendro_left['layout']['y']['tickvals'])
+        self.assertEqual(len(dendro_left['layout']['yaxis']['ticktext']), 5)
+        tickvals_left = np.array(dendro_left['layout']['yaxis']['tickvals'])
         self.assertTrue((tickvals_left <= 0).all())
 
         dendro_right = tls.FigureFactory.create_dendrogram(X, orientation='right')
-        tickvals_right = np.array(dendro_right['layout']['y']['tickvals'])
+        tickvals_right = np.array(dendro_right['layout']['yaxis']['tickvals'])
         self.assertTrue((tickvals_right >= 0).all())
 
         dendro_bottom = tls.FigureFactory.create_dendrogram(X, orientation='bottom')
-        self.assertEqual(len(dendro_bottom['layout']['x']['ticktext']), 5)
-        tickvals_bottom = np.array(dendro_bottom['layout']['x']['tickvals'])
+        self.assertEqual(len(dendro_bottom['layout']['xaxis']['ticktext']), 5)
+        tickvals_bottom = np.array(dendro_bottom['layout']['xaxis']['tickvals'])
         self.assertTrue((tickvals_bottom >= 0).all())
 
         dendro_top = tls.FigureFactory.create_dendrogram(X, orientation='top')
-        tickvals_top = np.array(dendro_top['layout']['x']['tickvals'])
+        tickvals_top = np.array(dendro_top['layout']['xaxis']['tickvals'])
         self.assertTrue((tickvals_top <= 0).all())
 
       
--- a/plotly/tests/test_optional/test_opt_tracefactory.py
+++ b/plotly/tests/test_optional/test_opt_tracefactory.py
@@ -105,3 +105,47 @@ class TestStreamline(TestCase):
         self.assertListEqual(strln['data'][0]['x'][0:100],
                              expected_strln_0_100['x'])
 
+class TestDendrogram(TestCase):
+
+    def test_default_dendrogram(self):
+        dendro = tls.FigureFactory.create_dendrogram(X=np.array([[1, 2, 3, 4],
+                                                       [1, 1, 3, 4],
+                                                       [1, 2, 1, 4],
+                                                       [1, 2, 3, 1]]))
+        expected_dendro_data = [{'marker': {'color': 'rgb(255,133,27)'},
+                                 'mode': 'lines', 'xaxis': 'xs',
+                                 'yaxis': 'ys',
+                                 'y': np.array([0.,  1.,  1.,  0.]),
+                                 'x': np.array([25.,  25.,  35.,  35.]),
+                                 'type': u'scatter'},
+                                {'marker': {'color': 'rgb(255,133,27)'},
+                                 'mode': 'lines',
+                                 'xaxis': 'xs',
+                                 'yaxis': 'ys',
+                                 'y': np.array([0., 2.23606798, 2.23606798, 1.]),
+                                 'x': np.array([15., 15., 30., 30.]),
+                                 'type': u'scatter'},
+                                {'marker': {'color': 'blue'},
+                                 'mode': 'lines',
+                                 'xaxis': 'xs',
+                                 'yaxis': 'ys',
+                                 'y': np.array([0., 3.60555128, 3.60555128, 2.23606798]),
+                                 'x': np.array([5., 5., 22.5, 22.5]), 'type': u'scatter'}]
+       
+        self.assertEqual(len(dendro.data), len(expected_dendro_data))
+        self.assertTrue(np.array_equal(dendro.labels, np.array(['3', '2', '0', '1'])))
+
+        for i in range(1,len(dendro.data)):
+          self.assertTrue(np.allclose(dendro.data[i]['x'], expected_dendro_data[i]['x']))
+          self.assertTrue(np.allclose(dendro.data[i]['y'], expected_dendro_data[i]['y']))
+
+    def test_dendrogram_random_matrix(self):
+        # create a random uncorrelated matrix
+        X = np.random.rand(5,5) 
+        # variable 2 is correlated with all the other variables
+        X[2,:] = sum(X,0)
+
+        dendro = tls.FigureFactory.create_dendrogram(X)
+
+        # Check that 2 is in a separate cluster
+        self.assertEqual(dendro.labels[0], '2')
--- a/plotly/tests/test_optional/test_opt_tracefactory.py
+++ b/plotly/tests/test_optional/test_opt_tracefactory.py
@@ -113,38 +113,71 @@ class TestDendrogram(TestCase):
                                                                  [1, 1, 3, 4],
                                                                  [1, 2, 1, 4],
                                                                  [1, 2, 3, 1]]))
-        expected_dendro_data = [{'marker': {'color': 'rgb(255,133,27)'},
-                                 'mode': 'lines', 'xaxis': 'xs',
-                                 'yaxis': 'y',
-                                 'y': np.array([0.,  1.,  1.,  0.]),
-                                 'x': np.array([25.,  25.,  35.,  35.]),
-                                 'type': u'scatter'},
-                                {'marker': {'color': 'rgb(255,133,27)'},
-                                 'mode': 'lines',
-                                 'xaxis': 'x',
-                                 'yaxis': 'y',
-                                 'y': np.array([0., 2.23606798,
-                                               2.23606798, 1.]),
-                                 'x': np.array([15., 15., 30., 30.]),
-                                 'type': u'scatter'},
-                                {'marker': {'color': 'blue'},
-                                 'mode': 'lines',
-                                 'xaxis': 'x',
-                                 'yaxis': 'y',
-                                 'y': np.array([0., 3.60555128,
-                                               3.60555128, 2.23606798]),
-                                 'x': np.array([5., 5., 22.5, 22.5]),
-                                 'type': u'scatter'}]
+        expected_data = [{'marker': {'color': 'rgb(255,133,27)'},
+                          'mode': 'lines', 'xaxis': 'xs',
+                          'yaxis': 'y',
+                          'y': np.array([0.,  1.,  1.,  0.]),
+                          'x': np.array([25.,  25.,  35.,  35.]),
+                          'type': u'scatter'},
+                         {'marker': {'color': 'rgb(255,133,27)'},
+                          'mode': 'lines',
+                          'xaxis': 'x',
+                          'yaxis': 'y',
+                          'y': np.array([0., 2.23606798,
+                                        2.23606798, 1.]),
+                          'x': np.array([15., 15., 30., 30.]),
+                          'type': u'scatter'},
+                         {'marker': {'color': 'blue'},
+                          'mode': 'lines',
+                          'xaxis': 'x',
+                          'yaxis': 'y',
+                          'y': np.array([0., 3.60555128,
+                                        3.60555128, 2.23606798]),
+                          'x': np.array([5., 5., 22.5, 22.5]),
+                          'type': u'scatter'}]
+        expected_layout = {'width': '100%',
+                            'showlegend': False,
+                            'autoscale': False,
+                            'x': {'showticklabels': True,
+                                     'tickmode': 'array',
+                                     'ticks': 'outside',
+                                     'showgrid': False,
+                                     'mirror': 'allticks',
+                                     'zeroline': False,
+                                     'showline': True,
+                                     'ticktext': np.array(['3', '2',
+                                                           '0', '1'],
+                                                          dtype='|S1'),
+                                     'rangemode': 'tozero',
+                                     'type': 'linear',
+                                     'tickvals': np.array([5.0, 15.0,
+                                                          25.0, 35.0])},
+                            'y': {'showticklabels': True,
+                                  'ticks': 'outside',
+                                  'showgrid': False,
+                                  'mirror': 'allticks',
+                                  'zeroline': False,
+                                  'showline': True,
+                                  'rangemode': 'tozero',
+                                  'type': 'linear'},
+                            'hovermode': 'closest'}  
 
-        self.assertEqual(len(dendro['data']), len(expected_dendro_data))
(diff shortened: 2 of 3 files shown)


plotly/tests/test_core/test_tools/test_figure_factory.py
```diff
@@ -802,3 +804,49 @@ def test_datetime_candlestick(self):
...
+        X = np.random.rand(5,5) 
+        # variable 2 is correlated with all the other variables
+        X[2,:] = sum(X,0)
+
+        dendro = tls.FigureFactory.create_dendrogram(X)
+
+        # Check that 2 is in a separate cluster
+        self.assertEqual(dendro.labels[0], '2')
+
+
```
chriddyp: We don't require our users to have `numpy` or `scipy` installed to use this package. For this dendrogram class, it's OK to rely on these packages, you'll just have to make a few changes in this package organization to make this work:
1 - Move these tests into `test_optional`. That's the folder that we use for all the functionality that uses optional dependencies like numpy, matplotlib, scipy. Here is a file that you can place these tests into, instead: https://github.com/plotly/python-api/blob/master/plotly/tests/test_optional/test_opt_tracefactory.py
2 - Protect imports. More on that below!
We split our tests into two folders: `test_core` and `test_optional`. The only difference is that `test_optional` contains tests that

plotly/tools.py
```diff
@@ -15,14 +15,16 @@
 import six
 
 import math
+import scipy
 
+import scipy.cluster.hierarchy as sch
```
chriddyp: We'll have to protect these imports into the actual dendrogram function. Not everyone using Plotly will have these packages installed, and it's a dependency that we would like to avoid for as long as possible! However, it's OK for the dendrogram class to use these functions, you'll just have to protect these imports. Here is how we import numpy, scipy and scipy.cluster.hierarcy can use similar logic:

Protected import: https://github.com/plotly/python-api/blob/master/plotly/tools.py#L48-L52
Raise exception if the import fails and they try to use the function that requires it: https://github.com/plotly/python-api/blob/master/plotly/tools.py#L1534-L1535

plotly/tools.py
```diff
@@ -2275,6 +2277,25 @@ def create_candlestick(open, high, low, close,
         return dict(data=data, layout=layout)
 
 
+    @staticmethod
+    def create_dendrogram(X, orientation="bottom", labels=None,
+                          colorscale=None, **kwargs):
+        """
+        Returns a dendrogram Plotly figure object.
+
+        X: Heatmap matrix as array of arrays
```
chriddyp: Is a more common convention for matrices, "Z"? That's what we use in our heatmaps, but I'm not sure if there is a dendrogram convention that uses "X"

plotly/tools.py
```diff
@@ -2275,6 +2277,25 @@ def create_candlestick(open, high, low, close,
...
+        orientation: 'top', 'right', 'bottom', or 'left'
+        labels: List of axis category labels
+        colorscale: Optional colorscale for dendgrogram tree clusters
```
chriddyp: Spelling: *dendgrogram :)

plotly/tools.py
```diff
@@ -2275,6 +2277,25 @@ def create_candlestick(open, high, low, close,
...
+        """
```
chriddyp: It'd be great to have one or two simple, copy-and-pastable examples in here!

plotly/tools.py
```diff
@@ -2871,3 +2891,149 @@ def get_candle_decrease(self):
...
+                "rgb(0,116,217)",
+                "rgb(255,65,54)",
+                "rgb(133,20,75)",
+                "rgb(255,133,27)",
+                "rgb(255,220,0)",
+                "rgb(61,153,112)"]
+        for i in range(len(default_colors.keys())):
+            k = default_colors.keys()[i] 
+            if i < len( colorscale ):
+                default_colors[k] = colorscale[i]
```
chriddyp: nice :) I might be missing something here, but since `default_colors` is a dict, we don't know the order of `default_colors.keys()`. Instead, an [OrderedDict](https://docs.python.org/2/library/collections.html#collections.OrderedDict) might be a better solution, or at least `sorted(default_colors.keys())`

plotly/tools.py
```diff
@@ -2871,3 +2891,149 @@ def get_candle_decrease(self):
...
+        self.set_axis_layout(self.xaxis)
+        self.set_axis_layout(self.yaxis)
+
+        return self.layout
+
+    def get_dendrogram_traces( self, X, colorscale ):
+        ''' Returns a tuple with:
+            (a) List of Plotly trace objects for the dendrogram tree
+            (b) icoord: All X points of the dendogram tree as array of arrays with lenght 4 
+            (c) dcoord: All Y points of the dendogram tree as array of arrays with lenght 4 '''
```
chriddyp: *lenght -> length

plotly/tools.py
```diff
@@ -2871,3 +2891,149 @@ def get_candle_decrease(self):
...
+                'ticks': 'inside',
+                'mirror': 'allticks',
+                'rangemode': 'tozero',
+                'showticklabels': True,                 
+                'zeroline': False,
+                'showgrid': False,              
+                'showline': True,
+            }
+        
+        if self.labels != None:
```
chriddyp: I get a 

```
plotly/tools.py:2978: FutureWarning:

comparison to `None` will result in an elementwise object comparison in the future.
```

is there a better alternative comparison?

plotly/tools.py
```diff
@@ -2871,3 +2891,149 @@ def get_candle_decrease(self):
...
+        self.zero_vals = []
+
+        for i in range(len(yvals_flat)):
+            if yvals_flat[i] == 0.0 and xvals_flat[i] not in self.zero_vals:
+                self.zero_vals.append( xvals_flat[i] )
+
+        self.zero_vals.sort() 
+                
+        self.layout = self.set_figure_layout( width, height )
+        self.data = Data( dd_traces )
```
chriddyp: Our latest convention for `FigureFactory` is to actually return a figure object (a `dict`) that can be plotted right away. Here is an example from ohlc: https://github.com/plotly/python-api/blob/master/plotly/tools.py#L2001

So here, add: `return {'data': layout, 'data': [dd_traces]}`

plotly/tools.py
```diff
@@ -2871,3 +2891,149 @@ def get_candle_decrease(self):
...
+                "rgb(255,65,54)",
+                "rgb(133,20,75)",
+                "rgb(255,133,27)",
+                "rgb(255,220,0)",
+                "rgb(61,153,112)"]
+        for i in range(len(default_colors.keys())):
+            k = default_colors.keys()[i] 
+            if i < len( colorscale ):
+                default_colors[k] = colorscale[i] 
+
```
chriddyp: Also, looks like `blue` wasn't getting replaced:
```
>> X=np.array([[1, 2, 3, 4],
    [1, 1, 3, 4],
    [1, 2, 1, 4],
    [1, 2, 3, 1]])

>> dendro = FF.create_dendrogram(X)
>> dendro.data[2]
{'marker': {'color': 'blue'},
 'mode': 'lines',
 'type': u'scatter',
 'x': array([  5. ,   5. ,  22.5,  22.5]),
 'xaxis': 'xs',
 'y': array([ 0.        ,  3.60555128,  3.60555128,  2.23606798]),
 'yaxis': 'ys'}
```

plotly/tools.py
```diff
@@ -2871,3 +2891,149 @@ def get_candle_decrease(self):
...
 
+class _Dendrogram(FigureFactory):
+    ''' Returns a Dendrogram figure object
+        Example usage:
+        D = Dendrogram( Z )
+        fig = { 'data':D.data, 'layout':D.layout }
+        py.iplot( fig, filename='Dendro', validate=False )'''
+    
+    def __init__(self, X, orientation='bottom', labels=None, colorscale=None, \
+                 width=700, height=700, xaxis='xaxis', yaxis='yaxis' ):
```
chriddyp: thoughts on keeping `width` and `height` out of the defaults and using the `autosize` default instead? I much prefer the full-width view inside ipython notebooks:

image

plotly/tools.py
```diff
@@ -2871,3 +2891,149 @@ def get_candle_decrease(self):
...
+                
+        return default_colors
+   
+    def set_axis_layout( self, axis_key ):
+        ''' Sets and returns default axis object for dendrogram figure 
+            axis_key: "xaxis", "xaxis1", "yaxis", yaxis1", etc '''
+        
+        axis_defaults = {
+                'type': 'linear',
+                'ticks': 'inside',
```
chriddyp: thoughts on removing the ticks? at least for the x-axis? it seems like the leafs of the dendrogram will always intersect with the axis labels which makes them a bit redundant. Also, when the leaves are small, the ticks cover the leaves up!

image

chriddyp:
Looks great, thanks for submitting! A few other comments besides those in the code:
- We try to follow pep8 pretty closely so that all of our contributions look and read the same. Code editors usually have a pep8 style plug-in (I use sublime and this plug-in: https://github.com/dreadatour/Flake8Lint).
- Thoughts on adding an option, or a separate FigureFactory, to include a heatmap and a second horizontal dendrogram?

plotly/tools.py
cldougl: It might be good to keep the docstrings similar across the FigureFactory methods. I tried to use this guideline https://github.com/plotly/dev-docs/blob/master/reference/styleguides/python.md#function-docstrings for the others.

plotly/tools.py
```diff
@@ -2275,6 +2277,25 @@ def create_candlestick(open, high, low, close,
         return dict(data=data, layout=layout)
 
 
+    @staticmethod
+    def create_dendrogram(X, orientation="bottom", labels=None,
+                          colorscale=None, **kwargs):
+        """
```
cldougl: It might be good to keep the docstrings similar across the FigureFactory methods. I used this guideline https://github.com/plotly/dev-docs/blob/master/reference/styleguides/python.md#function-docstrings for the others.

plotly/tools.py
```diff
@@ -2871,3 +2891,149 @@ def get_candle_decrease(self):
         return (decrease_x, decrease_close, decrease_dif,
                 stick_decrease_y, stick_decrease_x)
 
+class _Dendrogram(FigureFactory):
+    ''' Returns a Dendrogram figure object
+        Example usage:
```
cldougl: It's better to move the example to the `create_dendrogram()` docstring because that's the one that users will see if they call help: help(FigureFactory.create_dendrogram) 
Our idea is that users will only use the create_ methods so the FigureFactory classes are hidden

Then this docstring can be as simple as: Refer to FigureFactory.create_dendrogram() for docstring

OxanaSachenkova:
@chriddyp and @cldougl thanks! will fix/respond on each comment 

as for 

>> Thoughts on adding an option, or a separate FigureFactory, to include a heatmap and a second horizontal dendrogram?

You can already create a horizontal dendrogram using this class and adding them to a heatmap is not that difficult using subplots. Although an option of including dendrograms added to the existing heatmap function would come handy.

plotly/tools.py
cldougl: I like removing them (or at least moving them to the outside). I think it looks better when they're not covering the leaves.

plotly/tools.py
cldougl: @chriddyp  `return {'layout': layout, 'data': [dd_traces]}` right?

plotly/tools.py
theengineear: :+1: also, tiny style change. We tend to use `'''` over `"""` on docstrings.

plotly/tools.py
```diff
@@ -2871,3 +2891,149 @@ def get_candle_decrease(self):
...
+        D = Dendrogram( Z )
+        fig = { 'data':D.data, 'layout':D.layout }
+        py.iplot( fig, filename='Dendro', validate=False )'''
+    
+    def __init__(self, X, orientation='bottom', labels=None, colorscale=None, \
+                 width=700, height=700, xaxis='xaxis', yaxis='yaxis' ):
+        ''' Draw a 2d dendrogram tree
```
theengineear: We've been adding things like this with a `BETA` tag in the docstring. This way we don't have to promise backwards compatibility on these new wrappers. We simply use:

```
"""
Draw a 2d dendrogram tree.

BETA: May change without warning in a backwards incompatible way.

...
"""
```

plotly/tools.py
```diff
@@ -2871,3 +2891,149 @@ def get_candle_decrease(self):
...
+        self.leaves = []
+        self.sign = { self.xaxis:1, self.yaxis:1 }        
+        self.layout = { self.xaxis:{}, self.yaxis:{} }
+        
+        self.sign[self.xaxis] = 1 if self.orientation in ['left','bottom'] else -1
+        self.sign[self.yaxis] = 1 if self.orientation in ['right','bottom'] else -1        
+        
+        dd_traces, xvals, yvals, ordered_labels, leaves = self.get_dendrogram_traces( X, colorscale )
+        
+        self.labels = ordered_labels
```
theengineear: How is the passed in `labels` used? Seems like it's just reset to this `ordered_labels` value.

plotly/tools.py
```diff
@@ -2871,3 +2891,149 @@ def get_candle_decrease(self):
...
+
+        self.zero_vals = []
+
+        for i in range(len(yvals_flat)):
+            if yvals_flat[i] == 0.0 and xvals_flat[i] not in self.zero_vals:
+                self.zero_vals.append( xvals_flat[i] )
+
+        self.zero_vals.sort() 
+                
+        self.layout = self.set_figure_layout( width, height )
```
theengineear: Style note, pep:8ball: on the whitespace around these args: https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements

Rather:

`self.layout = self.set_figure_layout(width, height)`

plotly/tools.py
```diff
@@ -2871,3 +2891,149 @@ def get_candle_decrease(self):
...
+        trace_list = []
+        
+        for i in range(len(icoord)):
+            # xs and ys are arrays of 4 points that make up the '∩' shapes of the dendrogram tree
+            xs = icoord[i] if self.orientation in ['top','bottom'] else dcoord[i]
+            ys = dcoord[i] if self.orientation in ['top','bottom'] else icoord[i]
+            color_key = color_list[i]   
+            trace = Scatter(x=np.multiply(self.sign[self.xaxis],xs), \
+                            y=np.multiply(self.sign[self.yaxis],ys), \
+                            mode='lines', marker=Marker(color=colors[color_key]) )
```
theengineear: nice :smiley_cat: !

plotly/tools.py
```diff
@@ -2275,6 +2277,25 @@ def create_candlestick(open, high, low, close,
         return dict(data=data, layout=layout)
 
 
+    @staticmethod
+    def create_dendrogram(X, orientation="bottom", labels=None,
+                          colorscale=None, **kwargs):
```
theengineear: Any use for the `**kwargs`? It doesn't look like it's been used.

We *may* want to allow updates like these, but maybe we should hold off for now.

plotly/tools.py

plotly/tests/test_core/test_tools/test_figure_factory.py
OxanaSachenkova: Done with 378190691dd9fec92e62a232c1f4d5385c71085e

plotly/tools.py
OxanaSachenkova: Fixed in 31efa33c29e0ce4ad02a03b4fecd7ab8fa5683c7

plotly/tools.py
OxanaSachenkova: In this case it's to distinguish between a linkage matrix (Z) and the initial array of observations(X), like in http://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.pdist.html#scipy.spatial.distance.pdist

plotly/tools.py
OxanaSachenkova: Fixed in 02beb6ce250fd963d5c86723256c38deb1ba7bca

plotly/tools.py
OxanaSachenkova: Fixed in 02beb6ce250fd963d5c86723256c38deb1ba7bca

plotly/tools.py
OxanaSachenkova: Fixed in 02beb6ce250fd963d5c86723256c38deb1ba7bca

plotly/tools.py
OxanaSachenkova: Fixed in 02beb6ce250fd963d5c86723256c38deb1ba7bca

plotly/tools.py
OxanaSachenkova: fixed in ff98cb7c71241809a6f6e3dcaeb1a3354841b357

plotly/tools.py
OxanaSachenkova: should be fixed by ff98cb7c71241809a6f6e3dcaeb1a3354841b357 also

plotly/tools.py
OxanaSachenkova: @chriddyp is it ok to put 100% in default like here 4afc8c99c6a4b1e6585a37ecacc439bfed4027d5

plotly/tools.py
OxanaSachenkova: I agree, it's better on the outside in this case, I fixed here 4afc8c99c6a4b1e6585a37ecacc439bfed4027d5 
But I wouldn't remove them

plotly/tools.py
OxanaSachenkova: @cldougl seems like I can't access this link, but I tried to comply anyways =) have a look at 57e15bdffbd5f4ffd00d0e349e8346116ab4ee56

plotly/tools.py
OxanaSachenkova: fixed in 57e15bdffbd5f4ffd00d0e349e8346116ab4ee56

plotly/tools.py
OxanaSachenkova: fixed in 57e15bdffbd5f4ffd00d0e349e8346116ab4ee56

plotly/tools.py
OxanaSachenkova: agreed! fixed in 57e15bdffbd5f4ffd00d0e349e8346116ab4ee56

OxanaSachenkova:
OK, guys, I think I have addressed all of your comments :sweat_drops: , please have another look! 
:gift: @chriddyp @cldougl @theengineear

theengineear:
@OxanaSachenkova going to go through and review this again now, thanks for the updates!

If you see things marked as :cow2: it just means is a style-related tip (poor pun intended). There's no change required, but an update would be appreciated :)

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
         return (decrease_x, decrease_close, decrease_dif,
                 stick_decrease_y, stick_decrease_x)
 
+class _Dendrogram(FigureFactory):
+    
+    """
+    Refer to FigureFactory.create_dendrogram() for docstring.
+    """
+
+    def __init__(self, X, orientation='bottom', labels=None, colorscale=None, \
```
theengineear: :cow2: (style tip) the `\` for a line break while inside parentheses is redundant.

plotly/tools.py
```diff
@@ -8,21 +8,20 @@
...
 import math
 
-
 from plotly import utils
 from plotly import exceptions
 from plotly import session
 
 from plotly.graph_objs import graph_objs
-from plotly.graph_objs import Scatter, Marker
+from plotly.graph_objs import Scatter, Marker, Line, Data
```
theengineear: :cow2: these can be imported on the same line:

`from plotly.graph_objs import Data, Line, Marker, Scatter, graph_objs`

plotly/tools.py
```diff
@@ -51,6 +50,15 @@ def warning_on_one_line(message, category, filename, lineno,
...
     _numpy_imported = False
 
+try:
+    import scipy as scp
+    import scipy.spatial as scs 
+    import scipy.cluster.hierarchy as sch 
+
+    _scipy_imported = True
+except ImportError:
+    _scipy_imported = False
```
theengineear: thanks!

plotly/tools.py
```diff
@@ -2275,6 +2283,62 @@ def create_candlestick(open, high, low, close,
...
+        names_X = ['Jack', 'Oxana', 'John', 'Chelsea', 'Mark']
+        dendro_X = FF.create_dendrogram(X, orientation='right', labels=names_X)
+
+        py.iplot(dendro_X, validate=False, height=1000, width=300)
+        ```
+        """
+
+        if _scipy_imported is False:
+            raise ImportError("FigureFactory.create_dendrogram requires scipy,
+                              scipy.spatial and scipy.hierarchy")
```
theengineear: :smile_cat: perfect!

plotly/tools.py
```diff
@@ -2275,6 +2283,62 @@ def create_candlestick(open, high, low, close,
...
+
+        s = X.shape
+        if len(s) != 2:
+            exceptions.PlotlyError("X should be 2-dimensional array.")
+
+        dendrogram = _Dendrogram(X, orientation, labels, colorscale)
+
+        return {'layout': dendrogram.layout,
+                'data': dendrogram.data,
+                'labels': dendrogram.labels}
```
theengineear: Apologies if I'm missing something here, but shouldn't these labels end up as trace `name` attributes or something? Where are these `labels` supposed to show up in a Plotly plot?

@chriddyp ?

plotly/tests/test_optional/test_opt_tracefactory.py
```diff
@@ -105,3 +105,54 @@ def test_simple_streamline(self):
         self.assertListEqual(strln['data'][0]['x'][0:100],
                              expected_strln_0_100['x'])
 
+
```
theengineear: Can you make sure that:

1. The label argument *in* works as expected
2. The orientation *in* works as expected
3. colorscale *in* works as expected

It'd be best if these were tested separately to keep these unit tests from overlapping as much as possible.

plotly/tests/test_optional/test_opt_tracefactory.py
```diff
@@ -105,3 +105,54 @@ def test_simple_streamline(self):
...
+class TestDendrogram(TestCase):
+
+    def test_default_dendrogram(self):
+        dendro = tls.FigureFactory.create_dendrogram(X=np.array([[1, 2, 3, 4],
+                                                                 [1, 1, 3, 4],
+                                                                 [1, 2, 1, 4],
+                                                                 [1, 2, 3, 1]]))
+        expected_dendro_data = [{'marker': {'color': 'rgb(255,133,27)'},
+                                 'mode': 'lines', 'xaxis': 'xs',
+                                 'yaxis': 'ys',
```
theengineear: These 'yaxis' and 'xaxis' strings are off. I think it's because your default string is `'xaxis'` and you use `xaxis[-1]` which ends up being the string `'s'`. These should probably default to `'x'` and `'y'`, right @chriddyp?

plotly/tests/test_optional/test_opt_tracefactory.py
```diff
@@ -105,3 +105,54 @@ def test_simple_streamline(self):
...
+        self.assertEqual(len(dendro['data']), len(expected_dendro_data))
+        self.assertTrue(np.array_equal(dendro['labels'],
+                                       np.array(['3', '2', '0', '1'])))
+
+        for i in range(1, len(dendro['data'])):
+            self.assertTrue(np.allclose(dendro['data'][i]['x'],
+                            expected_dendro_data[i]['x']))
+            self.assertTrue(np.allclose(dendro['data'][i]['y'],
+                            expected_dendro_data[i]['y']))
+
```
theengineear: Can you also test that the output `layout` is as expected?

plotly/tools.py
```diff
@@ -2275,6 +2283,62 @@ def create_candlestick(open, high, low, close,
...
+        :param (ndarray) X: Matrix of observations as arrray of arrays
+        :param (str) orientation: 'top', 'right', 'bottom', or 'left'
+        :param (list) labels: List of axis category labels(observation labels)
+        :param (list) colorscale: Optional colorscale for dendrogram tree
+        clusters
+
+        X: Heatmap matrix as array of arrays
+        orientation: 'top', 'right', 'bottom', or 'left'
+        labels: List of axis category labels
+        colorscale: Optional colorscale for dendrogram tree clusters
```
theengineear: :cow2: this is redundant, you've already added it with the `:param:` info beautifully above :)

plotly/tools.py
```diff
@@ -2275,6 +2283,62 @@ def create_candlestick(open, high, low, close,
...
+
+
+        Example 1: Simple bottom oriented dendrogram
+        ```
+        import plotly.plotly as py
+        from plotly.tools import FigureFactory as FF
+
+        import numpy as np
```
theengineear: :cow2: typically, we order the less-specific import above:

```python
import numpy as np

import plotly.plotly as py
from plotly.tools import FigureFactory as FF
```

plotly/tools.py
```diff
@@ -2275,6 +2283,62 @@ def create_candlestick(open, high, low, close,
...
+
+        X = np.random.rand(5,5)
+        dendro_X = FF.create_dendrogram(X)
```
theengineear: IMO, the `dendro_X` variable seems like it would be clearer to read as a user if just `dendro`? Super minor, and maybe it's my misunderstanding.

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
...
                 stick_decrease_y, stick_decrease_x)
 
+class _Dendrogram(FigureFactory):
+    
+    """
+    Refer to FigureFactory.create_dendrogram() for docstring.
+    """
+
+    def __init__(self, X, orientation='bottom', labels=None, colorscale=None, \
+                 width="100%", height="100%", xaxis='xaxis', yaxis='yaxis' ):
```
theengineear: See https://github.com/plotly/python-api/pull/274/files#r37722528

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
...
+        self.orientation = orientation
+        self.labels = labels
+        self.xaxis = xaxis
+        self.yaxis = yaxis
+        self.data = []
+        self.leaves = []
+        self.sign = {self.xaxis: 1, self.yaxis: 1}
+        self.layout = {self.xaxis: {}, self.yaxis: {}}
```
theengineear: Hm, maybe I'm unclear on which xaxis/yaxis you're trying to store here. There's the *axis reference* which is when you say something like `'xaxis': 'x1'` and then there's the *layout axis object* which is when you say something like `'xaxis2': {'domain': [0, 1]}`.

So, I *think* you're making them serve two purposes, using the full string to name the layout object and trying to take the last digit and use it as a *reference*. I'd suggest you write a bit of logic in there that handles the case when the last character *isn't* a digit:

```python
try:
    axis_index = int(axis_string[-1])
except ValueError:
    axis_index = ''
else:
    axis_char = axis_string[0]
    axis_reference = '{}{}'.format(axis_char, axis_index)
```

Again, sorry if I'm misinterpreting. I think just clearing up the test suite will sort this out.

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
...
+
+        if self.orientation in ['left', 'bottom']:
+            self.sign[self.xaxis] = 1
+        else:
+            self.sign[self.xaxis] = -1
+
+        if self.orientation in ['right', 'bottom']:
+            self.sign[self.yaxis] = 1
+        else:
+            self.sign[self.yaxis] = -1
```
theengineear: This orientation logic doesn't seem to match the `left`, `right`, `bottom`, `top` that your documentation suggests?

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
...
+            # of the dendrogram tree
+            if self.orientation in ['top', 'bottom']:
+                xs = icoord[i]
+            else:
+                xs = dcoord[i]
+
+            if self.orientation in ['top', 'bottom']:
+                ys = dcoord[i]
+            else:
+                ys = icoord[i]
```
theengineear: Where does the `left` and `right` come in here?

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
...
+            color_key = color_list[i]
+            trace = Scatter(x=np.multiply(self.sign[self.xaxis], xs),
+                            y=np.multiply(self.sign[self.yaxis], ys),
+                            mode='lines',
+                            marker=Marker(color=colors[color_key]))
+            trace['xaxis'] = 'x'+self.xaxis[-1]
+            trace['yaxis'] = 'y'+self.yaxis[-1]
```
theengineear: pep:8ball: spaces around operators

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
         return (decrease_x, decrease_close, decrease_dif,
                 stick_decrease_y, stick_decrease_x)
 
+class _Dendrogram(FigureFactory):
+
```
theengineear: :cow2: Just some unnecessary whitespace :)

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
...
+                
+        self.layout = self.set_figure_layout(width, height)
+        self.data = Data(dd_traces)
+
+    def get_color_dict(self, colorscale):
+        """ 
+            Returns colorscale used for dendrogram tree clusters
+            :param (list) colorscale: colors to use for the plot,
+            in rgb format
+        """
```
theengineear: :cow2: formatting on this docstring is a little off:

```
"""
Returns colorscale used for dendrogram tree clusters

:param (list) colorscale: colors to use for the plot in rgb format
:return: (?)

"""
```

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
...
+        if colorscale is None:
+            colorscale = [
+                "rgb(0,116,217)",   # blue
+                "rgb(35,205,205)",  # cyan
+                "rgb(61,153,112)",  # green
+                "rgb(40,35,35)",    # black
+                "rgb(133,20,75)",   # magenta
+                "rgb(255,65,54)",   # red
+                "rgb(255,255,255)",   # white
+                "rgb(255,220,0)"]   # yellow
```
theengineear: :100: !

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
...
+            if i < len(colorscale):
+                default_colors[k] = colorscale[i]
+                
+        return default_colors
+   
+    def set_axis_layout(self, axis_key):
+        """ 
+            Sets and returns default axis object for dendrogram figure
+            :param (str) axis_key: "xaxis", "xaxis1", "yaxis", yaxis1", etc.
+        """
```
theengineear: :cow2: docstring again:

```
"""
Sets and returns default axis object for dendrogram figure

:param (str) axis_key: "xaxis", "xaxis1", "yaxis", yaxis1", etc.

"""
```

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
...
+                'ticks': 'outside',
+                'mirror': 'allticks',
+                'rangemode': 'tozero',
+                'showticklabels': True,                 
+                'zeroline': False,
+                'showgrid': False,              
+                'showline': True,
+            }
+        
+        if len(self.labels) != 0:
```
theengineear: :cow2: `if self.labels:`?

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
...
+            axis_key_labels = self.xaxis
+            if self.orientation in ['left', 'right']:
+                axis_key_labels = self.yaxis
+            if axis_key_labels not in self.layout:
+                self.layout[axis_key_labels] = {}
+            self.layout[axis_key_labels]['tickvals'] = [ea*self.sign[axis_key]
+                                                        for ea in self.zero_vals]
```
theengineear: :cow2: I don't understand what the `ea` variable means here? Just a little tough to read.

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
...
+            self.layout[axis_key_labels]['tickmode'] = 'array'
+            
+        self.layout[axis_key].update(axis_defaults)
+            
+        return self.layout[axis_key]
+
+    def set_figure_layout(self, width, height):
+        """ 
+            Sets and returns default layout object for dendrogram figure 
+        """
```
theengineear: :cow2: doctstring just has extra indentation.

plotly/tools.py
```diff
@@ -2871,3 +2934,188 @@ def get_candle_decrease(self):
...
+            (a) trace_list: List of Plotly trace objects for the dendrogram
+            tree
+            (b) icoord: All X points of the dendogram tree as array of arrays
+            with length 4
+            (c) dcoord: All Y points of the dendogram tree as array of arrays
+            with length 4
+            (d) ordered_labels: leaf labels in the order they are going to
+            appear on the plot
+            (e) P['leaves']: left-to-right traversal of the leaves
+        """
```
theengineear: :cow2: again, I think you can save your space the indentation in these docstrings by just unindenting the whole thing.

theengineear:
Thanks for putting all this work in! There are still a couple things that could use some addressing. Mostly, I think adding more tests to lock down the expected functionality would do.

plotly/tools.py
theengineear: Just as a preventative measure for conflicting optional import needs, would you mind incorporating a syntax as suggested in https://github.com/plotly/python-api/pull/275/files#r37724351 ?

@chriddyp does this seem reasonable? The other option I see is *tagging* import patterns as `_dendrogram_imports` or something... but that seems like it'd make the code harder to read?

It feels strange to me to force folks who have scipy and *aren't* using dendrograms to have to install that functionality... but maybe this is yet another viable route...

theengineear: Yet another, other option would be to *not* automatically import this, so that user's will naturally get a Python `ImportError` when they attempt to use it because *they* need to manually import this... I think this isn't as user-friendly, but perhaps more robust.

I still think we should be *nice* to users and continue with our protected imports pattern...

plotly/tools.py
OxanaSachenkova: @theengineear 
from https://docs.python.org/2/library/exceptions.html
"exception ImportError
Raised when an import statement fails to find the module definition or when a from ... import fails to find a name that is to be imported." 

so I think our _scipy_imported will be false when from ... import statements fail in the beginning of the script

plotly/tools.py
OxanaSachenkova: fixed in 44cdece2c3b91c8cb57ad0ddb3fff9283ff0f989

plotly/tools.py
OxanaSachenkova: fixed in 44cdece2c3b91c8cb57ad0ddb3fff9283ff0f989

plotly/tools.py
OxanaSachenkova: I prefer protected imports as a user  for sure. 
refined in 44cdece2c3b91c8cb57ad0ddb3fff9283ff0f989

plotly/tools.py
OxanaSachenkova: refined in 44cdece2c3b91c8cb57ad0ddb3fff9283ff0f989

plotly/tests/test_optional/test_opt_tracefactory.py
OxanaSachenkova: @theengineear right now the format of colorscale is different from plotly JSON, as is the goal of using it. Colorscale in this case only replaces default colors that scipy assigns to cluster elements. There is no control over to which element the color will be assigned as it is usually is with Plotly's colorscale key-value pairs. 
Should I perhaps communicate this in a docstring/example?

plotly/tools.py
OxanaSachenkova: fixed in 16c74c1da67c4301e96c4a5ea0612ca405f1c7fe

plotly/tools.py
OxanaSachenkova: you're right, this isn't actually doing anything. fixed and adjusted the tests  in this commit 16c74c1da67c4301e96c4a5ea0612ca405f1c7fe

plotly/tools.py
OxanaSachenkova: @theengineear I am actually a bit confused about this one, but I think  I fixed it in this ebff53e5c2367b42679f2a4ddb49018ab9b93e2e commit and cleaned up here 16c74c1da67c4301e96c4a5ea0612ca405f1c7fe

plotly/tests/test_optional/test_opt_tracefactory.py
OxanaSachenkova: fixed in ebff53e5c2367b42679f2a4ddb49018ab9b93e2e

plotly/tests/test_optional/test_opt_tracefactory.py
OxanaSachenkova: added layout test in ebff53e5c2367b42679f2a4ddb49018ab9b93e2e

plotly/tools.py
OxanaSachenkova: @theengineear I don't really understand what does not match, could you clarify? I am attaching how the dendrograms look with different orientations now also
orientations

plotly/tools.py
OxanaSachenkova: @theengineear in the "else" statement

plotly/tools.py
OxanaSachenkova: @cldougl @theengineear P.S. any tips how to move the axis labels together with the actual axis when changing orientation?

plotly/tests/test_optional/test_opt_tracefactory.py
OxanaSachenkova: added a test 20fc40c48a42b74915575d8dacec125e81b06c35

plotly/tools.py
OxanaSachenkova: fixed here 88842b1a9c8d6e15b71aaa01d8f1ad67aceb7377

plotly/tools.py
OxanaSachenkova: fixed here 88842b1a9c8d6e15b71aaa01d8f1ad67aceb7377

plotly/tools.py
OxanaSachenkova: fixed here 88842b1a9c8d6e15b71aaa01d8f1ad67aceb7377

plotly/tools.py
OxanaSachenkova: fixed here 88842b1a9c8d6e15b71aaa01d8f1ad67aceb7377

plotly/tools.py
OxanaSachenkova: fixed here 88842b1a9c8d6e15b71aaa01d8f1ad67aceb7377

OxanaSachenkova:
@theengineear @cldougl Let's go for round 3 and perhaps you have some useful tips for improving orientation plots (see image in the comments above)

plotly/tools.py
theengineear: Apologies, I think I may have just read over that too fast, looks good. Also, these are really great plots!!

I wouldn't worry about that for now, let's get this merged in! I'll be more present so that we can get this in soon so we don't keep leaving you hanging.

theengineear:
@OxanaSachenkova sorry for the radio silence! Let me know if you want me to pull down the branch to fix any conflicts! I'm happy to help with those.

OxanaSachenkova:
@theengineear  yes, I think it's better if you take over to fix the last things and resolve the conflicts. probably will be more efficient

theengineear:
Great, I'll do that today and we can get this in!

theengineear:
Closing this in favor of #312. Thanks @OxanaSachenkova !!!

PR #275, opened 2015-08-05
Distplot

cldougl:
- add distplot to FigureFactory
- add `validate_displot()` method to tools.py
- add core tests for distplot parameters


cldougl:
@chriddyp @theengineear @aneda @etpinard 
Hey everyone, here's a draft of the distplot (http://stanford.edu/~mwaskom/software/seaborn/generated/seaborn.distplot.html) FigureFactory addition.
One issue though - I'm not sure if the 'protected import' of scipy I tried is how we want to add it / correct in general- the tests won't pass on circle.

chriddyp:
@cldougl from the Circle error:
```
ERROR: test_distplot_more_args (plotly.tests.test_optional.test_opt_tracefactory.TestDistplot)
  File "/home/ubuntu/python-api/plotly/tests/test_optional/test_opt_tracefactory.py", line 109, in test_distplot_more_args
    show_rug=False, bin_size=.2)
  File "/home/ubuntu/python-api/plotly/tools.py", line 2418, in create_distplot
    FigureFactory.validate_distplot(hist_data, curve_type)
  File "/home/ubuntu/python-api/plotly/tools.py", line 1541, in validate_distplot
    raise ImportError("FigureFactory.create_distplot requires scipy")
ImportError: FigureFactory.create_distplot requires scipy
```

it looks like our Circle build doesn't have scipy. Circle will only build the packages that we explicitly configure. Here is the circle configuration file: https://github.com/plotly/python-api/blob/master/circle.yml#L7, note the "optional-requirements.txt" reference. if you add `scipy` to that file: https://github.com/plotly/python-api/blob/master/optional-requirements.txt, then circle should build it and tests should pass

cldougl:
ahh thank you @chriddyp !!

plotly/tools.py
```diff
@@ -51,6 +51,13 @@ def warning_on_one_line(message, category, filename, lineno,
 except ImportError:
     _numpy_imported = False
 
+try:
+    import scipy
+    import scipy.stats
+    _scipy_imported = True
+except ImportError:
+    _scipy_imported = False
+
```
theengineear: \o/

plotly/tools.py
```diff
@@ -1501,6 +1508,39 @@ def validate_ohlc(open, high, low, close, direction, **kwargs):
...
+                                             "sure your hist_data variable is "
+                                             "still a list of lists, i.e. x = "
+                                             "[1, 2, 3] -> x = [[1, 2, 3]]")
+        elif type(hist_data[0]) is not list:
+            raise exceptions.PlotlyError("Oops, this function was written "
+                                         "to handle multiple datasets, if "
+                                         "you want to plot just one, make "
+                                         "sure your hist_data variable is "
+                                         "still a list of lists, i.e. x = "
+                                         "[1, 2, 3] -> x = [[1, 2, 3]]")
```
theengineear: :camel: (DRY-tip), can you use `isinstance` here and save some text?

```
allowed_types = [list]
if _numpy_imported:
    allowed_types.append(np.ndarray)

if not isinstance(hist_data[0], allowed_types):
    raise :)
```

plotly/tools.py
```diff
@@ -2274,6 +2314,160 @@ def create_candlestick(open, high, low, close,
...
+            hist_data, group_labels, bin_size,
+            curve_type, **kwargs).make_rug()
+
+        data = []
+        if show_hist:
+            data.append(hist)
+        if show_curve:
+            data.append(curve)
+        if show_rug:
+            data.append(rug)
```
theengineear: would it be cool to add all this data by default, but hide it in the figure? then they could always turn it on in the gui later?

@chriddyp thoughts?

plotly/tools.py
```diff
@@ -2274,6 +2314,160 @@ def create_candlestick(open, high, low, close,
...
+                            anchor='y2',
+                            zeroline=False),
+                yaxis1=dict(domain=[0.35, 1],
+                            anchor='free',
+                            position=0.0),
+                yaxis2=dict(domain=[0, 0.25],
+                            anchor='x1',
+                            dtick=1))
+        else:
+                layout = graph_objs.Layout(
```
theengineear: over-indentation?

plotly/tools.py
cldougl: @theengineear  would that work with how I set the layout? My concern would be if we did that and they don't want the rug, they'll still get the subplot for the rug if the rug traces are hidden https://plot.ly/~chelsea_lyn/5569/dist-plot/ vs https://plot.ly/~chelsea_lyn/5567/dist-plot/

theengineear: Cool, I trust you on that!

plotly/tools.py
theengineear: Note that this is going to be a merge conflict with #274 . I guess we'll need to have some specific requirements specified... How about a syntax like:

```
_scipy_imported
_scipy__spatial_imported
_scipy__cluster__hierarchy_imported
```

I'm not sure which is going to land first, but the latter will have to implement this.

plotly/tools.py
```diff
@@ -2871,3 +3061,152 @@ def get_candle_decrease(self):
...
+            self.colors = kwargs['colors']
+        else:
+            self.colors = [
+                "rgb(31, 119, 180)", "rgb(255, 127, 14)",
+                "rgb(44, 160, 44)", "rgb(214, 39, 40)",
+                "rgb(148, 103, 189)", "rgb(140, 86, 75)",
+                "rgb(227, 119, 194)", "rgb(127, 127, 127)",
+                "rgb(188, 189, 34)", "rgb(23, 190, 207)"]
+        self.curve_x = [None]*self.trace_number
+        self.curve_y = [None]*self.trace_number
```
theengineear: pep:8ball: spaces around operators

plotly/tests/test_optional/test_opt_tracefactory.py
```diff
@@ -9,6 +9,148 @@
...
+                                   'x': [0.8, 1.5, 1.5, 0.6, 0.59, 1.0, 0.8,
+                                         1.7, 0.5, 0.8, -0.3, 1.2, 0.56, 0.3,
+                                         2.2],
+                                   'xaxis': 'x1',
+                                   'xbins': {'end': 2.2, 'size': 0.2,
+                                             'start': -0.3},
+                                   'yaxis': 'y1'}
+        self.assertEqual(dp['data'][1], expected_dp_data_hist_2)
+
+
```
theengineear: \o/ tests!

theengineear:
@cldougl awesome! I'm ok with :dancer: this. I'd like to let @chriddyp check over it again too though.

plotly/tools.py
```diff
@@ -1501,6 +1508,35 @@ def validate_ohlc(open, high, low, close, direction, **kwargs):
                                          "'both'")
 
     @staticmethod
+    def validate_distplot(hist_data, curve_type):
```
chriddyp: can we make these guys hidden? (by adding a `_`). there's hardly anything I like more than seeing the minimal number of functions in the ipython tab-tab helper:
image

plotly/tools.py
```diff
@@ -2274,6 +2310,160 @@ def create_candlestick(open, high, low, close,
...
+        :param (string) curve_type: 'kde' or 'normal', default = 'kde'
+        :param (bool) show_hist: True/False defines if histogram is added to
+            distplot, default = True
+        :param (bool) show_curve: True/False defines if curve is added to
+            distplot, default = True
+        :param (bool) show_rug: True/False defines if rug plot is added to
+            distplot, default = True
+        :param (list) colors: list of strings of color values,
+            default = first 10 Plot.ly colors
+        :param (list) rug_text: list of strings of hovertext for rug_plot,
```
chriddyp: `rug_text`, `colors` aren't in the call signature

plotly/tools.py
```diff
@@ -2274,6 +2310,160 @@ def create_candlestick(open, high, low, close,
 
         return dict(data=data, layout=layout)
 
+    @staticmethod
+    def create_distplot(hist_data, group_labels,
+                        bin_size=1., curve_type='kde',
+                        show_hist=True, show_curve=True,
+                        show_rug=True, **kwargs):
```
chriddyp: it doesn't look like `kwargs` is used below - If so, I'd remove it from the call signature and let the user modify the traces as a return argument

plotly/tools.py
```diff
@@ -2274,6 +2310,160 @@ def create_candlestick(open, high, low, close,
...
+        BETA function that creates a distplot similar to seaborn.distplot
+
+        The distplot can be composed of all or any combination of the following
+        3 components: (1) histogram, (2) curve: (a) kernal density estimation
+        or (b)normal curve, and (3)rug plot. Additionally, multiple distplots
+        (from multiple datasets) can be created in the same plot.
+
+        :param (list) hist_data: list of histogram data, should be a list of
+            lists if multiple data sets are plotted on the same plot
+        :param (list) group_labels: list of strings of the name of each data
```
chriddyp: hah @theengineear me and you are never going to agree on docstrings. I find this format so difficult to read! I guess it's a nice standard if we ever make auto HTML versions of these docs though...

plotly/tools.py
```diff
@@ -2274,6 +2310,160 @@ def create_candlestick(open, high, low, close,
...
+                      3.5, 4.1, 4.4, 4.5, 4.5,
+                      5.0, 5.0, 5.2, 5.5, 5.5,
+                      5.5, 5.5, 5.5, 6.1, 7.0]]
+
+        group_labels = ['distplot example']
+
+        fig = FF.create_distplot(hist_data, group_labels)
+
+        url = py.plot(fig, filename='Simple distplot', validate=False)
+        ```
```
chriddyp: the left-most annotation is cut-off in this example:
image

plotly/tools.py
cldougl: @chriddyp yeah maybe this wasn't a good way to write it: `rug_text` and `colors` are optional so I was including them as **kwargs

plotly/tools.py
chriddyp: I think that the annotation is probably redundant, since it's already color coded in the legend

plotly/tools.py
```diff
@@ -2274,6 +2310,160 @@ def create_candlestick(open, high, low, close,
...
+
+        # Group data together
+        hist_data = [hist1_x] + [hist2_x]
+
+        group_labels = ['2012', '2013']
+
+        rug_text = ['a', 'b', 'c', 'd', 'e',
+                    'f', 'g', 'h', 'i', 'j',
+                    'k', 'l', 'm', 'n', 'o']
+
```
chriddyp: at first glance this is a bit confusing, because I don't know how the which set of data the `rug_text` corresponds to

plotly/tools.py
cldougl: @chriddyp agreed, I'll remove that

plotly/tools.py
```diff
@@ -2274,6 +2310,160 @@ def create_candlestick(open, high, low, close,
...
+        # Create distplot
+        fig = FF.create_distplot(
+            hist_data, group_labels, rug_text=rug_text, bin_size=.2)
```
chriddyp: this example didn't work for me: 
image

plotly/tools.py
chriddyp: Oh I see, it's not an annotation, it's just the tick labels. OK, what if we removed the y-axis tick labels by default? I think the colors do a nicer job of coordinating the traces.

cldougl: yep exactly. I think the color coordination is enough information

plotly/tools.py
```diff
@@ -2274,6 +2310,160 @@ def create_candlestick(open, high, low, close,
...
+        else:
+            layout = graph_objs.Layout(
+                barmode='overlay',
+                hovermode='closest',
+                xaxis1=dict(domain=[0.0, 1.0],
+                            anchor='y2',
+                            zeroline=False),
+                yaxis1=dict(domain=[0., 1],
+                            anchor='free',
+                            position=0.0))
```
chriddyp: Could we reverse the legend? That way way the rugs will be displayed in the same order as the legend entries. Right now they're in opposite orders
image

plotly/tools.py
```diff
@@ -2274,6 +2310,160 @@ def create_candlestick(open, high, low, close,
...
+                        show_rug=True, **kwargs):
+        """
+        BETA function that creates a distplot similar to seaborn.distplot
+
+        The distplot can be composed of all or any combination of the following
+        3 components: (1) histogram, (2) curve: (a) kernal density estimation
+        or (b)normal curve, and (3)rug plot. Additionally, multiple distplots
+        (from multiple datasets) can be created in the same plot.
+
+        :param (list) hist_data: list of histogram data, should be a list of
```
chriddyp: @theengineear @cldougl @etpinard - we need to converge on a good term for these arguments. Strictly speaking, it's `Encodable as JSON array` which uh, is too much. Options:
- vague: `list-like`
- hella explicit: `list, 1D numpy array, or pandas series of numbers, strings, or datetimes`

can i get a vote? I vote *explicit*

plotly/tools.py
```diff
@@ -2274,6 +2310,160 @@ def create_candlestick(open, high, low, close,
...
+        fig['layout'].update(title='Dist Plot')
+
+        # Plot!
+        url = py.plot(fig, filename='Distplot with rug text', validate=False)
+        ```
+
+        Example 3: Plot with normal curve and hide rug plot
+        ```
+        import plotly.plotly as py
+        from plotly.tools import FigureFactory as FF
```
chriddyp: add a `import numpy as np`

plotly/tools.py
cldougl: explicit for me too: `list-like` seems like it would be a bit confusing for users

plotly/tools.py
```diff
@@ -2274,6 +2310,160 @@ def create_candlestick(open, high, low, close,
...
+
+        x1 = np.random.randn(20)
+        x2 = np.random.randn(40)+1
+        x3 = np.random.randn(10)-1
+        x4 = np.random.randn(10)+2
```
chriddyp: could we move these up to size `200` and bin size `0.2`? It's so much more pretty :)

image

plotly/tools.py
```diff
@@ -2871,3 +3061,152 @@ def get_candle_decrease(self):
...
+        for index in range(self.trace_number):
+            curve[index] = dict(type='scatter',
+                                x=self.curve_x[index],
+                                y=self.curve_y[index],
+                                xaxis='x1',
+                                yaxis='y1',
+                                mode='lines',
+                                name=self.group_labels[index],
+                                legendgroup=self.group_labels[index],
+                                showlegend=False,
```
chriddyp: funny side affect. if `show_hist` is `False` then no legend appears. So, we'll have to set `showlegend=True` to the histogram if it's shown, otherwise the KDE if it's shown, otherwise the rug if it's shown

image

chriddyp: kde: `False if show_hist else True`
rug: `False if show_hist or show_curve else True`

plotly/tools.py
cldougl: @chriddyp sure - 
I was trying to show an example that the data sets don't all need to be the same size, or is that obvious?

plotly/tools.py
```diff
@@ -1501,6 +1508,35 @@ def validate_ohlc(open, high, low, close, direction, **kwargs):
...
+        if _numpy_imported:
+            hist_data_types += (np.ndarray,)
+
+        if not isinstance(hist_data[0], hist_data_types):
+                raise exceptions.PlotlyError("Oops, this function was written "
+                                             "to handle multiple datasets, if "
+                                             "you want to plot just one, make "
+                                             "sure your hist_data variable is "
+                                             "still a list of lists, i.e. x = "
+                                             "[1, 2, 3] -> x = [[1, 2, 3]]")
```
chriddyp: ahhhh. this'll fail with pandas series. The easiest thing to do is probably add:

```
if _pandas_imported:
    hist_data_types ++ (pd.Series, )
```

plotly/tools.py
```diff
@@ -1501,6 +1508,35 @@ def validate_ohlc(open, high, low, close, direction, **kwargs):
...
+
+        curve_opts = ('kde', 'normal')
+        if curve_type not in curve_opts:
+            raise exceptions.PlotlyError("curve_type must be defined as "
+                                         "'kde' or 'normal'")
+
+        if _scipy_imported is False:
+            raise ImportError("FigureFactory.create_distplot requires scipy")
```
chriddyp: it'd be great to add a test in `test_core` that tries this function, and makes sure that this is the exception that is thrown

plotly/tools.py
```diff
@@ -2274,6 +2310,160 @@ def create_candlestick(open, high, low, close,
...
+
+        hist_data = [x1] + [x2] + [x3] + [x4]
+        group_labels = ['2012', '2013', '2014', '2015']
+
+        fig = FF.create_distplot(
+            hist_data, group_labels, curve_type='normal',
+            show_rug=False, bin_size=.4)
+
+        url = py.plot(fig, filename='hist and normal curve', validate=False)
+        """
```
chriddyp: it'd be great to add one more example that uses pandas:

# Example 4 - Distplot with Pandas
```
import pandas as pd
df = pd.DataFrame({'2012': np.random.randn(200), '2013': np.random.randn(200)+1})
iplot(FF.create_distplot([df[c] for c in df.columns], df.columns))
```

chriddyp:
otherwise, sweet!

plotly/tools.py
chriddyp: ah OK. maybe center them around 200? eg 180, 190, 200, 210

cldougl: :+1:

plotly/tools.py
theengineear: I'll be honest, I'm *super* biased because of the IDE I use. It does full type-checking and will warn me when things are wrong. Also, I don't often have an issue writing/reading in this format though. Docstrings are just hard to get right.

This one, for example, could be more tersely *and* explicitly written:

```
:param (list[str]) group_labels: Names for each data set.
```

Here are some other factors to keep in mind:

* This function has *tons* of input parameters, which is inherently going to be confusing (not a dig on the function, btw)
* These explanations are a bit redundant.

Here's the same information in a terser (though just as informative) form:

```
        """
        BETA function that creates a distplot similar to seaborn.distplot

        The distplot can be composed of all or any combination of the following
        3 components: (1) histogram, (2) curve: (a) kernal density estimation
        or (b) normal curve, and (3) rug plot. Additionally, multiple distplots
        (from multiple datasets) can be created in the same plot.

        :param (list, list[list]) hist_data: Use list of lists to plot multiple data sets on
            the same plot.
        :param (list[str]) group_labels: Names for each data set.
        :param (float) bin_size: Size of histogram bins. Default = 1.
        :param (str) curve_type: 'kde' or 'normal'. Default = 'kde'.
        :param (bool) show_hist: Add histogram to distplot? Default = True.
        :param (bool) show_curve: Add curve to distplot? Default = True.
        :param (bool) show_rug: Add rug to distplot? Default = True.
        :param (list[str]) colors: Colors for traces.
        :param (list[str]) rug_text: Hovertext values for rug_plot.
        :return: (dict) Representation of a distplot figure.

        # other stuff
        """
```

As long as our docstrings are machine-parseable, I think we're doing it right :) If that means we switch to a different docstring format, I'd be totally fine with it. Here's a [SO question on it](http://stackoverflow.com/questions/3898572/what-is-the-standard-python-docstring-format). It seems to suggest that there are tools to *convert* from one machine-parseable docstring convention to another.

plotly/tools.py
theengineear: :+1: they should be in the call signature, not grouped into `kwargs`.

plotly/tools.py
theengineear: I'm indifferent on this one, I feel like a user is just going to *try stuff*. I definitely don't see any harm using the explicit approach though. My indifference comes from the duck-typing nature of Python where I'd assume that if I saw `list`, I could use any list-y thing in its place.

plotly/tools.py
theengineear: One more note (man, you guys probably think I'm so nit-picky :p)

Our style guide for JS documentation and Python in streambed follow similar conventions, *JS Docs* for javascript and *reST* for Python:

* [JS](https://github.com/plotly/dev-docs/blob/master/reference/styleguides/javascript.md#comments)
* [Python](https://github.com/plotly/dev-docs/blob/master/reference/styleguides/python.md#function-docstrings)

Not that that's held as a commandment by the team, but seeing consistency in our code is pretty nice, imo.

cldougl:
@chriddyp can I merge this?

chriddyp:
Looking good! It doesn't look like the `colors` argument is working for me:

```
import plotly.plotly as py
from plotly.tools import FigureFactory as FF
import numpy as np

x1 = np.random.randn(200)
x2 = np.random.randn(200)+1
x3 = np.random.randn(200)-1
x4 = np.random.randn(200)+2

hist_data = [x1] + [x2] + [x3] + [x4]
group_labels = ['2012', '2013', '2014', '2015']

fig = FF.create_distplot(
    hist_data, group_labels, curve_type='normal',
    bin_size=.2, colors=['yellow', 'pink', 'grey', 'aqua'])

iplot(fig)
```

image

chriddyp:
Thx for fixing! All good by me now :dancer: Can you add some python examples as ipython notebooks to /documentation?

cldougl:
Thanks! yep doc's ready

PR #284, opened 2015-09-03
Download and cache plot schema (or always download if no file permissions).

theengineear:
This PR handles getting the *new* plot schema downloaded/cached/backed-up within the old framework. There's no change to code usage here, just JSON handling.

Note that if enterprise doesn't have the new endpoint defined, the default one can be used. This *doesn't* guarantee that the plotlyjs running on the enterprise machine is in sync with the plotlyjs used to create the default though! In particular, the `autotick` to `tickmode` change on axes inside the mpl renderer may be problematic. Hopefully this will get knocked to a *warning* though (`_deprecated`) and will eventually stabilize once enterprise instances get updated versions.

Todo:

- [x] :tiger2: ~~for no file permissions and loaded graph reference~~ this will get checked on circle
- [x] :tiger2: for proper caching
- [x] :tiger2: for working offline if it's been downloaded
- [x] ship a default copy of plot-schema per version, actually this is vital to keeping enterprise customers working! super important! otherwise they won't have this new url when they upgrade!
- [x] manually update our default schema before we close this pr.


theengineear:
@chriddyp @etpinard @cldougl , can I get a review on this?

etpinard:
Looking good.

About

> ship a default copy of plot-schema per version, actually this is vital to keeping enterprise customers working! super important! otherwise they won't have this new url when they upgrade!

How often should we update the *default* schema?

chriddyp:
I'm not opposed to bumping the minor version number as often as plot schema changes - a robot could make PRs.

theengineear:
No real opinion on the upgrade frequency. A bot could load the default json dict, hash it and check if it needs updating daily, pushing to master and then uploading to PyPI on changes. I'll open up an issue to discuss on the side, #296.

plotly/graph_reference.py
```diff
@@ -0,0 +1,64 @@
...
+        plotly_api_domain = config.get('plotly_api_domain',
+                                       default_config['plotly_api_domain'])
+    else:
+        graph_reference = {}
+        plotly_api_domain = default_config['plotly_api_domain']
+
+    sha1 = hashlib.sha1(six.b(str(graph_reference))).hexdigest()
+
+    graph_reference_url = '{}{}?sha1={}'.format(plotly_api_domain,
+                                                GRAPH_REFERENCE_PATH, sha1)
```
chriddyp: Yikes, I didn't think about this. A few things to consider:
- Is this going to break for the plotly enterprises out there that don't have `v2` endpoints?
- If we keep seperate versions of plot-schema for whichever plotly.js is being used, we'll need another version for plotly offline and whatever version of plot-schema *it* uses.

plotly/graph_reference.py
```diff
@@ -0,0 +1,64 @@
...
+
+    try:
+        response = requests.get(graph_reference_url)
+        response.raise_for_status()
+    except requests.exceptions.RequestException:
+        if not graph_reference:
+            path = os.path.join('graph_reference', 'default-schema.json')
+            s = resource_string('plotly', path).decode('utf-8')
+            graph_reference = json.loads(s)
```
chriddyp: OK, I see. Old enterprise will be using the local, shipped version.

plotly/tools.py
```diff
@@ -95,28 +61,30 @@ def get_config_defaults():
...
                 if key not in contents:
                     contents[key] = val
             contents_keys = list(contents.keys())
             for key in contents_keys:
-                if key not in _FILE_CONTENT[fn]:
+                if key not in FILE_CONTENT[fn]:
                     del contents[key]
             utils.save_json_dict(fn, contents)
+
+        # make a request to get graph reference if DNE.
```
chriddyp: but it will always exist right, since we ship with a version? In the future, this will change to `# make a request to get graph reference if outdated` presumably?

plotly/tools.py
```diff
@@ -95,28 +61,30 @@ def get_config_defaults():
...
+        utils.ensure_file_exists(GRAPH_REFERENCE_FILE)
+        utils.save_json_dict(GRAPH_REFERENCE_FILE,
+                             graph_reference.GRAPH_REFERENCE)
```
chriddyp: I find this confusing - doesn't `from plotly import graph_reference`  call `graph_reference.get_graph_reference`? So hasn't the request already been made?

plotly/graph_reference.py
```diff
@@ -0,0 +1,64 @@
...
+            content = str(response.content, encoding='utf-8')
+        else:
+            content = response.content
+        data = json.loads(content)
+        if data['modified']:
+            graph_reference = data['schema']
+
+    return utils.decode_unicode(graph_reference)
+
+GRAPH_REFERENCE = get_graph_reference()
```
chriddyp: I forget how imports work, but if I do something like:

```
import plotly.plotly as py
from plotly.offline import iplot
from plotly.tools import FF
```
is that going to call this file 3 times (and perform 3 downloads?)

plotly/graph_reference.py
```diff
@@ -0,0 +1,64 @@
...
+        graph_reference = {}
+        plotly_api_domain = default_config['plotly_api_domain']
+
+    sha1 = hashlib.sha1(six.b(str(graph_reference))).hexdigest()
+
+    graph_reference_url = '{}{}?sha1={}'.format(plotly_api_domain,
+                                                GRAPH_REFERENCE_PATH, sha1)
+
+    try:
+        response = requests.get(graph_reference_url)
```
chriddyp: I wonder if we should place a client-side timeout on this. I'm imagining the spotty internet connections (conferences), where the request doesn't fail right away, but might take eg 30 seconds. Restarting the ipython kernal and waiting for a  >5 sec import would be a bummer!

chriddyp: related: https://github.com/plotly/python-api/issues/297

plotly/graph_reference.py
theengineear: Yup, exactly. Note that this *doesn't* mean they'll be in sync though :(

plotly/graph_reference.py
theengineear: nop, this runs *once* only.

plotly/graph_reference.py
theengineear: It won't break enterprise solutions (more than they're already broken). They have a default schema to fall back on. This also has the deprecated stuff worked in, so it'll be allowed. A funny situation would be if the plotlyjs they are using hasn't deprecated a key and we throw a warning suggesting they use a key that actually doesn't exist yet...

theengineear: @chriddyp you're gonna need to refresh me on the plotly offline here :cat:. Do we have a plot schema for the plotly-offline that we can get at somewhere?

plotly/graph_reference.py
theengineear: Bueno, timeout sounds great.

Alternatively, we could run it in the background and raise a runtime warning that the user needs to manually update the graph reference with some command? That'd be the *fastest* approach and seems like it wouldn't be too much to ask our users? Dunno how I feel about that...

plotly/tools.py
theengineear: Yeah, I actually had a hesitation where I thought "this documentation might be confusing" and then I totally didn't change it. I'll fix that :smile_cat:

plotly/graph_reference.py
chriddyp: Timeout: `3 seconds`?

theengineear: Definitely agreed about the complexity-factor. Yeah 3 seconds seems reasonable I guess, I'll try it out a bunch and see how often I hit that threshold.

plotly/graph_reference.py
chriddyp: Presumably, when plotly offline downloads (the latest) `plotly.js`, it can also download that version's `plot-schema`. This schema won't need updating, since it is locked in with the local `plotly.js`.

plotly/graph_reference.py
chriddyp: Cool :) This should pass the "I think I'll work out on my back patio, even though the internet sucks, test"

theengineear: Based on this study done ~20 seconds ago : P

https://plot.ly/~theengineear/2413/stats-on-wait-times-for-loading-plot-schema-50-times/

I think we should have a timeout around `5s`.

theengineear:
Anything blocking this beyond the timeout discussion?

I think it will make the update-graph-reference pr a little cleaner if I can get this merged into master and then master merged into that branch.

cldougl:
no objections here.. 
I merged another chart into FigureFactory yesterday so I think we'll have to add 
`from plotly.graph_objs import graph_objs` to the `distplot` function

chriddyp:
boom!

theengineear:
@cldougl , yep, the merge conflict is taken care of, thanks for the reminder!

just let me know if/when it can :dancer: :) @chriddyp also looking for confirmation on the `5s` timeout in https://github.com/plotly/python-api/pull/284#discussion_r39859104

plotly/graph_reference.py
chriddyp: Looks good to me!
How did you do this test btw? was this all in the same location or did you move around? (14 sec is crazy!)

chriddyp:
:dancer:

theengineear:
K, timeout added and default schema updated. mergin'

plotly/graph_reference.py
theengineear: All from the same location. I used a crappy network I have access to. And yeah, 14 is definitely crazy... which is exactly having *some* sane timeout is a great call ;)

PR #286, opened 2015-09-04
Test plot schema

theengineear:
JUST A TEST! DON'T REVIEW!


theengineear:
@chriddyp, @cldougl, @etpinard  I wanted to push *something* up here before I head out next week.

I'll create a proper PR with added tests / documentation when I return. You should be able play around with the auto-completion, which is a new feature that sort of came along for the ride.

When I dish this into a nice/sane commit history, I think it'll all make sense.

theengineear:
See #292

Issue #291, opened 2015-09-15
Be smarter about our `__copy__` and `__deepcopy__` methods.


theengineear:
Because we override some special methods we need to define `__copy__` and `__deepcopy__` methods on `PlotlyDict` and `PlotlyList`. Getting the object references straight is non-trivial, but isn't high-priority. Hence the separate issue / pr.

jonmmease:
No longer applicable to version 3

Issue #290, opened 2015-09-15
Add key ordering back to python api.


theengineear:
There's no order in the new plot schema.

etpinard:
Wouldn't loading up the plot-schema json with as an ordered dict suffice?

theengineear:
sorta. i think this is still hard to test though, if we made a change in the plot-schema and the new hashes changed the order of the stored json, then loading this new json as an ordered dict would break tests, right? we *have* to set an order in `plot-schema.json` before we can start using the `OrderedDict` `ordered_pairs` hook. Right?

etpinard:
@theengineear That sounds like another syncing nightmare.

Why would we need to test the order of the plot-schema?

Doesn't validation only care of the existence of the attributes, not their position in the plot-schema?

Are you concerned about the output of the `help` method?

theengineear:
Testing `to_string` is nearly impossible without having a predictable order.

Just in case we're not on the same page. As far as updating the python api goes, I really don't care or need the plot schema to have a consistent order. I do some *exceedingly* simple ordering in the python api so that the `to_string` test result will pass. But that's all. This is highly :football:-able as far as I'm concerned.

I created an issue because we *would* ideally like better `help` and I think there's some documentation-related needs for ordering in the plot schema.

The python api has no intention of testing full ordering, but just a *small* subset needs to be predictable in order to test the `to_string` method.

@chriddyp :wind_chime:-ed in on this earlier though and might have more opinions on it.

etpinard:
@theengineear all good :+1:

Issue #289, opened 2015-09-15
Add pretty doc strings back after graph reference upgrade


theengineear:
The old docs were based on the old version of graph reference. This has changed, but belongs in a separate PR.

chriddyp:
by docs, you mean doc strings?

theengineear:
yeah, sorry, i'll correct that title

chriddyp:
and fyi python reference pages have been updated, there might be some things that you'll want to copy: https://plot.ly/python/reference/

In particular, I think the explanation of the types is pretty good, like `flaglist`: https://plot.ly/python/reference/#scatter-mode, and `graph_objs`: https://plot.ly/python/reference/#scatter-marker-line.

Code is here: https://github.com/plotly/documentation/blob/source/_includes/reference-block.html

theengineear:
yeah, those docs are wonderful! i'd love to have something like that in our 'help'

jonmmease:
Version 3 docstrings are very pretty 🙂

PR #292, opened 2015-09-15
Update graph reference

theengineear:
Alright, the real deal!

TODO:

- [x] add back deprecation warnings with `_deprecated`
- [x] ~~use `hrName`~~
- [x] wait, don't use hrname :) (it adds unnecessary maintenance overhead)
- [x] use `_isSubplotObj`
- [x] stop hard coding meta keys
- [x] skipped tests need attached GH issues
- [x] :tiger2: for `graph_objs_tools.py`
- [x] discuss #291 (copy stuff), #290 (key ordering), #289 (pretty `help`), #293 (schema as singleton)
- [x] :dolls: need a final revisit
- [x] fix up some test code that uses deprecated keys
- [x] incorporate new deprecated `role` definitions
- [x] ~~use descriptions from deprecated fields in warnings~~ i'm now pointing users to call help appropriately. i think this is good enough.
- [ ] gotta make sure this doesn't break anything in cufflinks
- [x] don't add *new* class definitions (only for traces)


plotly/graph_objs/graph_objs.py
```diff
@@ -297,67 +348,175 @@ def force_clean(self, caller=True):
...
+
+    def __setitem__(self, key, value, _raise=True):
+
+        if not isinstance(key, six.string_types):
+            if _raise:
+                raise TypeError('Key must be string, not {}'.format(type(key)))
+            return
+
+        if key.endswith('src') and key in self._attributes:
+            value = graph_objs_tools.assign_id_to_src(key, value)
```
etpinard: why do we need this?

plotly/graph_objs/graph_objs.py
```diff
@@ -297,67 +348,175 @@ def force_clean(self, caller=True):
...
+        if subplot_key is not None:
+            value = GraphObjectFactory.create(subplot_key)
+            value.__dict__['_parent'] = self
+            value.__dict__['_parent_key'] = key
+            super(PlotlyDict, self).__setitem__(key, value)
+
+    def _get_subplot_key(self, key):
+
+        # TODO: this can use _isSubplotObj instead and won't require subclass!
+        subplot_key_strings = ('xaxis', 'yaxis', 'zaxis', 'lataxis', 'lonaxis',
```
etpinard: I'm repeating myself, but for reference this should be `layout.xaxis`, `layout.yaxis` (e.g. `layout.scene.xaxis is NOT a subplot object), `scene` and `geo`

theengineear:
FWIW tests are failing right now because the newest graph reference isn't on streambed production.

https://github.com/plotly/streambed/pull/3802 needs to land before these things will work.

plotly/graph_objs/graph_objs.py
theengineear: @chriddyp ? i believe this is something he added a while back, just maintaining.

plotly/graph_objs/graph_objs.py
```diff
@@ -429,30 +585,21 @@ def strip_style(self):
...
-                        # TODO: use graph_objs_tools.value_is_data
-                        if isinstance(self[key], six.string_types):
-                            del self[key]
-                        elif not hasattr(self[key], '__iter__'):
-                            del self[key]
-                except KeyError:  # TODO: Update the JSON
-                    # print("'type' not in {0} for {1}".format(obj_key, key))
-                    pass
+                role = graph_objs_tools.get_role(self, key, self[key])
+                if role == 'style':
```
etpinard: :tada:

plotly/graph_objs/graph_objs.py
theengineear: Yep. This needs a revisit. I need to go back in and check the `_isSubplotObj`. These hard-coded strings will go away after that.

plotly/graph_objs/graph_objs.py
etpinard: great, thanks

plotly/graph_objs/graph_objs.py
etpinard: :balloon:

plotly/graph_objs/graph_objs.py
```diff
@@ -429,30 +585,21 @@ def strip_style(self):
...
+                    del self[key]
+
+                # this is for backwards compat when we updated graph reference.
+                if self._name == 'layout' and key == 'autosize':
```
etpinard: Not too bad. I thought there would be more of these. :+1:

plotly/graph_objs/graph_objs.py
```diff
@@ -464,12 +611,13 @@ def get_data(self, flatten=False):
...
-                    if graph_objs_tools.value_is_data(obj_key, key, val):
-                        d[key] = val
-                except KeyError:
-                    pass
+                role = graph_objs_tools.get_role(self, key, val)
+                if role == 'data':
+                    d[key] = val
+
+                # we use the name to help make data frames
+                if self._name in graph_reference.TRACE_NAMES and key == 'name':
```
etpinard: not too bad here too

plotly/graph_objs/graph_objs.py
```diff
@@ -634,110 +641,72 @@ def to_string(self, level=0, indent=4, eol='\n',
...
-                    else:  # they want it all!
-                        string += repr(self[key])
-                if index < len(self):
-                    string += ","
-                if index == len(self):  # TODO: extraneous...
-                    break
-        string += "{eol}{indent})".format(eol=eol, indent=' ' * indent * level)
-        return string
+        if self._name in graph_reference.TRACE_NAMES:
+            keys = [key for key in self.keys() if key != 'type']
```
etpinard: not too bad either

plotly/graph_reference.py
```diff
@@ -0,0 +1,321 @@
...
+    :param (str) string: Presumably an object_name from GRAPH_REFERENCE.
+    :return: (str)
+
+    """
+
+    # capitalize first letter
+    string = re.sub(r'[A-Za-z]', lambda m: m.group().title(), string, count=1)
+
+    # replace `*_<c>` with `*<C>` E.g., `Error_x` --> `ErrorX`
+    string = re.sub(r'_[A-Za-z]+', lambda m: m.group()[1:].title(), string)
```
etpinard: don't forget about 

image

plotly/graph_objs/graph_objs.py
```diff
@@ -79,32 +126,96 @@ def __init__(self, *args):
...
+        for value in iterable:
+            index = len(self)
+            value = self.value_to_graph_object(index, value)
+            super(PlotlyList, self).append(value)
+
+    def insert(self, index, value):
+        """Override to enforce validation."""
+        value = self.value_to_graph_object(index, value)
+        value.__dict__['_parent'] = self
+        super(PlotlyList, self).insert(index, value)
```
chriddyp: do you also need to overwrite __iadd__?

plotly/graph_objs/graph_objs_tools.py
```diff
@@ -1,190 +1,84 @@
...
-    for key in keys:
-        if key in translations:
-            updated_keys += [translations[key]]
-        else:
-            updated_keys += [key]
-    return updated_keys
+def _make_dict_doc(name):
 
-translations = dict(
-    scl="colorscale",
```
etpinard: also @theengineear these deprecated keys will be part of the plot-schema. PR today.

plotly/graph_objs/graph_objs.py
chriddyp: yeah, something weird w grid_objs

plotly/graph_reference.py
```diff
@@ -0,0 +1,321 @@
...
+from __future__ import absolute_import
+
+import json
+import re
+import requests
+
+from plotly import exceptions, files, utils
+
+GRAPH_REFERENCE_PATH = '/plot-schema.json'
+META_KEYS = ['_isLinkedToArray', '_isSubplotObj', 'description', 'role']
```
etpinard: done in https://github.com/plotly/streambed/commit/81aa2710a55b50efb913efe757082d803c9aabd2

chriddyp:
on my phone, can't figure out how to leave inline comments:
- can we pip ship a copy of plotschema? some companies have wicked firewall rules that might require them to install from source, and restrict all downloads

plotly/graph_reference.py
chriddyp: hrname is great, I'll be able to use the same names in /reference

plotly/graph_objs/graph_objs_tools.py
etpinard: https://github.com/plotly/streambed/pull/3862

plotly/graph_objs/graph_objs.py
theengineear: nice! yeah, that's a great call, in the back of my head i was equating that to an `extend`, but it's totally different. thanks!

theengineear: yuck there are actually a whole slew of these `isub`, `imul`, `ifloordiv`, not sure if they're supported by list, i'll check now :)

theengineear: `iadd` seems like the only thing we need to add `imul` is also there, but there's no way to invalidate the list just by making copies.

theengineear:
@chriddyp yeah, we can probably pip-shipit i'll add it to the TODO, thx!

theengineear:
@chriddyp @etpinard small question on naming conventions. Python's 'title' naturally wants to change '2d' to '2D', so we'd get 'Histogram2DContour', not 'Histogram2dContour'. I think this is nice and it's slightly easier. Seem OK? (backwards compat is kept, no worries there)

etpinard:
> small question on naming conventions. Python's 'title' naturally wants to change '2d' to '2D', so we'd get 'Histogram2DContour', not 'Histogram2dContour'. I think this is nice and it's slightly easier. Seem OK? (backwards compat is kept, no worries there)

I have no preference. @theengineear your call.

chriddyp:
+1

theengineear:
Great, re: https://github.com/plotly/python-api/pull/292#issuecomment-140734678, we'll move to the capital letter after a number then :).

theengineear:
@chriddyp and @etpinard, this experience seem *OK* for deprecated keys?

image

theengineear:
Sorry (not sorry?) to add https://github.com/plotly/python-api/commit/56100a7199afef809e898bd69ec9c62d3de2aaac. It's really a bummer to want to run tests quickly locally and have to wait ~4 minutes for them to complete. That commit gives us the ability to do:

`nosetests -a '!slow' plotly/tests`

Which completes in about 8-10s. I was really getting bogged down making small changes and wanting to make sure it didn't break some unexpected part of code *quickly*, which was really impossible before.

Hope it's alright, it's speeding up my dev process now!

theengineear:
@chriddyp and @etpinard There are some issues listed in https://github.com/plotly/python-api/pull/292#issue-106556486 that I wanted to document here (at least). I think that it's not worth fixing them in this PR. Is it okay to skip them for now?

Right now the Python API isn't really doing anything with the `descriptions`. We should probably add that back in to make `help` better. I'll do a *little* bit more work on the `help`, but I'm not going to throw any more time than is necessary into that for now.

theengineear:
I just played around with this, no issues so far, but I'd love it if it got a bit more use before merging this PR in. If you do end up pulling this down and checking it out, I'd love to know (it'd instil some extra confidence!).

You should really consider setting `%config IPCompleter.greedy = True` in your notebook as it greatly improves the tab-completion, making tabbing into nested objects a really nice experience.

Here's a quick plot I made, https://plot.ly/~theengineear/5432

Here's the NB, http://nbviewer.ipython.org/gist/theengineear/87bbe323a506ca43f9f1

etpinard:
@theengineear 

> @chriddyp and @etpinard, this experience seem OK for deprecated keys?

Would it be too hard to incorporate the deprecated key `description` from the plot-schema:


image

etpinard:
@theengineear something is up with how nested attributes are listed:

image

compared to 

image

theengineear:
@etpinard yeah, great point, i didn't think it would come up too much, but the objects allow a combination  of possible attributes from their locations. I can make this smarter though since i recently made objects aware of their parents.

So, we'll end up with this:

* `Marker()` defined alone allows *any* of the possible attributes
* `Marker()` defined inside `Pie()` only allows a subset of that

A bit trickier, but I think I can work it in.

One thing I'm *not* going to work in is validating the figure's layout based on the different traces included in the data array. I think the logic would get pretty hairy and there are really not many wins for the user experience.

etpinard:
> So, we'll end up with this:
Marker() defined alone allows any of the possible attributes
Marker() defined inside Pie() only allows a subset of that

This is how plotlyjs works; this is how the python api should work.

> One thing I'm not going to work in is validating the figure's layout based on the different traces included in the data array. I think the logic would get pretty hairy and there are really not many wins for the user experience.

:+1: This is not necessary. In fact plotlyjs, doesn't try to be [smart about it](https://github.com/plotly/streambed/blob/master/shelly/plotlyjs/static/plotlyjs/src/graph_obj.js#L1859-L1875) either.

plotly/graph_objs/graph_objs.py
chriddyp: The alternative way to do this would be to make this a `MutableSequence` and only write out `__getitem__`, `__setitem__`, `__delitem__`, `__len__`, and `insert`: https://docs.python.org/2/library/collections.html#collections-abstract-base-classes
 and get `append`, `reverse`, `extend`, `pop`, `remove`, and `__iadd__` for free.

theengineear: Not 100% sure I follow on that. I agree that we could do this with abstract base classes though. What do you mean "for free"? Wouldn't we still need to override those methods to ensure that they convert dicts to graph objects appropriately?

chriddyp:
@theengineear - lemme know when you incorporate the parent-aware keys, and then I'll take this for a spin

plotly/graph_objs/graph_objs.py
chriddyp: by "for free" I mean that if you define those methods (`__getitem__`, `__setitem__`, ...) and subclass `MutableSequence`, then implementations of `append`, `reverse`, ... are provided as part of the super and they just use the `__getitem__`, `__setitem__` methods that were provided.

theengineear: Ah, sweet! I didn't really know if Python did something different under the hood with those. I'll keep this in mind. It won't change the user-facing interactions so it should be an easy swap if it works out!

theengineear:
yowza, doing attributes dynamically via paths is tougher than i expected!

getting there!

image

Not going to have this in the PR until tomorrow. Overall, I'm glad we're catching this now. It's a little harder to do, but definitely the more correct approach.

theengineear:
I think that I'll just *always* lookup attributes/items dynamically instead of keeping the `_attributes` `_items` `attr`s on each class definition. (mostly just a note for myself)

theengineear:
@chriddyp @etpinard @cldougl definitions should be more intelligently gathered now. overall, i think the change ended up making the code a bit clearer too!

wanna give it a :car:?

chriddyp:
sweet! I'll give it a :blue_car: . vroom vroom

chriddyp:
Another fun auto-completion feature we could add (sometime) would be to generate named arguments in the call signatures of these objects, that way if you start typing:

```
Layout(xa
```

and then tab-tab, you see `xaxis`.

I'm doing this type of named argument class factory generation in the next version of dash, here is the code: https://github.com/chriddyp/messin/blob/master/backend/components.py#L111-L151, which copies a lot of python's standard lib implementation of NamedTuple.

chriddyp:
So this is sweet:

```
> Scatter(marker=Marker(symbol='+'))
{'marker': {'symbol': '+'}, 'type': 'scatter'}

> Bar(marker=Marker(symbol='+'))
PlotlyDictKeyError: Invalid key, 'symbol', for class, 'Marker'.

Run 'help(plotly.graph_objs.Marker)' for more information.

Path To Error:
['marker']['symbol']
```

Although the error message is misleading:
- Can we add `for class, 'Marker', inside 'Bar'`?
- Can we add these notes in the docstrings so that `help(plotly.graph_objs.Marker)` doesn't mislead?

```
> help(Marker)
Valid Attributes:

* color
* cauto
* cmax
* cmin
* symbol (only valid inside `Scatter`)
* sizeref (only valid inside `Scatter`)
...
```

chriddyp:
Now that we know the enumeration types, thoughts on valtype validation?

```
> Scatter(mode='asdf')
```

chriddyp:
I can't believe that this works!!

```
t1 = Scatter()
t1.marker.color = 'blue'
```

That autocomplete dot notation is sweet. I wish that we could hide the non-attributes from the autocomplete: 
image

chriddyp:
I wonder if we could subclass `None` and provide help on keys:

```
> l = Layout()
> help(l.dragmode)
Help on NoneType object:

class NoneType(object)
 |  Methods defined here:
 |  
 |  __hash__(...)
 |      x.__hash__() <==> hash(x)
 |  
 |  __repr__(...)
 |      x.__repr__() <==> repr(x)
```

chriddyp:
- [ ] gotta make sure this doesn't break anything in cufflinks

chriddyp:
looking good so far!

plotly/graph_reference.py
```diff
@@ -0,0 +1,355 @@
...
+    """
+    string = _get_hr_name(object_name)
+
+    # capitalize first letter
+    string = re.sub(r'[A-Za-z]', lambda m: m.group().title(), string, count=1)
+
+    # replace `*_<c>` with `*<C>` E.g., `Error_x` --> `ErrorX`
+    string = re.sub(r'_[A-Za-z0-9]+', lambda m: m.group()[1:].title(), string)
+
+    return string
```
chriddyp: this is going to be so annoying to rewrite in liquid templating

etpinard:
Awesome stuff

plotly/graph_reference.py
```diff
@@ -0,0 +1,355 @@
...
+    """Get human readable object name from reference ('hrName')."""
+    if object_name in ARRAYS:
+        return object_name  # TODO: how to handle hr name on array?
+
+    object_dict = OBJECTS[object_name]
+    meta_paths = object_dict['meta_paths']
+    for meta_path in meta_paths:
+        meta_dict = utils.get_by_path(GRAPH_REFERENCE, meta_path)
+        if meta_dict.get('_isLinkedToArray'):
+            return object_name  # TODO: how to handle hr name on array?
```
chriddyp: https://github.com/plotly/documentation/commit/76f273577d16df346b13480b0958cdae9a251cc1#commitcomment-13317046

chriddyp: Let me know when you resolve this, and I'll make sure that documentation logic mirrors it.

plotly/graph_reference.py
```diff
@@ -0,0 +1,355 @@
...
+    objects = {}
+    for node, path in utils.node_generator(GRAPH_REFERENCE):
+        if any([key in path for key in GRAPH_REFERENCE['defs']['metaKeys']]):
+            continue  # objects don't exist under nested meta keys
+
+        # note that arrays are *not* stored in objects! they're arrays!
+        if node.get('role') == 'object':
+            object_name = path[-1]
+            if node.get('_isLinkedToArray'):
+                object_name = object_name[:-1]
```
chriddyp: Ah, yup, that's how you're handling it. Me too in documentation: https://github.com/plotly/documentation/commit/76f273577d16df346b13480b0958cdae9a251cc1#commitcomment-13317046. cool.

plotly/graph_reference.py
chriddyp: Not so bad: https://github.com/plotly/documentation/commit/76f273577d16df346b13480b0958cdae9a251cc1

chriddyp: So this depreciates `XBins` in favor of `Xbins`, and `XAxis` in favor of `Xaxis`, eh?

plotly/graph_reference.py
theengineear: Ha, yeah. This is pretty fragile though. @etpinard can we add a test that ensures that `_isLinkedToArray` objects in the schema end in an `s` or something? Else, can we enforce that such arrays define some meta information about the items they have? `_itemName` or something? I don't have a good sense of how this works out will arrays that have many items. For example, would `data` (not that it's explicitely defined) use `"_itemName": "scatter"` as a default?

theengineear: no need for immediate action, just as a discussion piece.

plotly/graph_reference.py
theengineear: yeah, exactly. The funky thing about this is *if* we add an hr_name of `x_axis` at some point `XAxis` would go away, making a backwards incompatible change... one way to prevent it is to use the object name *and* the hrName to create classes... which does a *lot* of doubling...

I'm not sure what the right path is here.

theengineear:
Making `help()` return dynamic information seems pretty rough :(

How does this experience look?

image

image

It's pretty messy for things that have *many* definitions:

image

But, I think with some work I could consolidate *repeated* definitions and just show a list of the parent paths they can exist under.

theengineear:
Aight, help formatting is good to go, imo:

image

image

image

theengineear:
Cruising through a cufflinks notebook: http://nbviewer.ipython.org/gist/santosjorge/b278ce0ae2448f47c31d

Issues right now:

## Histogram doesn't have `line` in plot schema, but it's used in `cufflinks`:

image

theengineear:
@chriddyp @etpinard @cldougl , final review? `cufflinks` is broken, see https://github.com/plotly/python-api/pull/292#issuecomment-141948216, but it's unclear to me what needs to be changed/updated?

I've added the `.help()` function into the graph objects so that users can more easily ask for help on objects in general, or attributes of objects.

All suggestions/comments welcome. I tried to make the commit history cleaner, I felt bad about dropping this change a little while back (https://github.com/plotly/python-api/commit/eccb46940af0937231569cdbd79d5051ecbad186). Hopefully it's understandable.

chriddyp:
beautiful!! I'll fix up the stuff for cufflinks

etpinard:
@theengineear 

> @etpinard are these breaks correct? were any of these things deprecated?

These *might* be deprecated, but they are definitely not handled in `cleanData` at the moment.

I'd vote for updating cufflinks.

theengineear:
:+1: sg!

cldougl:
It's working well for me- I think the help formatting looks good/clear!

chriddyp:
fixed the histogram line issue for cufflinks here: https://github.com/chriddyp/cufflinks/commit/e66bafe55f9c2cf86eea08978fb68d04dac47501, just waiting on a graph reference update.

theengineear:
@etpinard OK, trying to put your suggestions into motion. Here's how things work now:

image


Notes:

* Adding/overwriting as a dict works now (just as it did before)
* Adding/editing via the `.` works now (just as it did before)
* The only new classes created are *traces*, no other sub-objects
* Old sub-objects are still accessible for backwards compat
* We use `list` or `dict` to explain sub-object instantiation to users. This is to hide the `PlotlyDict` and `PlotlyList` objects from users, who shouldn't know about these. Note that upon being added as a sub-object, a `dict` gets converted to a `PlotlyDict

theengineear:
Thoughts on trying to get this merged and done with?

etpinard:
> Internally, should we now change to never using things like Marker, Data, etc? We can just instantiate from Figure, use dict/list or use GraphObjectFactory.create with a hidden _name argument, which seems alright to do internally. E.g. GraphObjectFactory.create(_name='marker'). I wouldn't mind if all our object creation went through this... it also makes our reference errors go away for dev stuff.

Yes. In version 2.0.0 I'd vote for having `Marker`, `Data` and all the graph objects other than the trace objects, `Layout` and `Figure` out of the picture. Using `GraphObjectFactory.create` sounds like the easiest and most scalable way for keeping the python api in-sync.

> Should we not expose objects like Marker in the to_string call? I think we should! It has the unexpected, but nice side effect of making the instantiation path errors auto-magically work out :) **see below

I'd vote for the full-path solution. Plotly attributes are setup to depend on the root object; python api users should get to know that fact.

In that case, how hard would it be to change the error message

image

so that things like `PlotlyDict` aren't exposed to the users?

plotly/graph_reference.py
```diff
@@ -7,17 +7,62 @@
...
 
 GRAPH_REFERENCE_PATH = '/v2/plot-schema'
 GRAPH_REFERENCE_DOWNLOAD_TIMEOUT = 5  # seconds
 
 
+# For backwards compat, we keep this list of previously known objects.
+# Moving forward, we only add new trace names.
+# {<ClassName>: {'object_name': <object_name>, 'base_type': <base-type>}
+_BACKWARDS_COMPAT_CLASS_NAMES = {
+    'AngularAxis': {'object_name': 'angularaxis', 'base_type': dict},
```
etpinard: Nice. Maybe @chriddyp could use this for the https://plot.ly/python/reference/

etpinard:
:dancer: The dream is now reality.

chriddyp:
:dancer: looks good to me!

theengineear:
:boom: thanks guys. i'll clean up some things related to https://github.com/plotly/python-api/pull/292#issuecomment-143221317 tomorrow and merge this guy in! woot!

theengineear:
second thought, this pr is wayy to massive to keep working on. i'll open up new changes in new prs.

PR #299, opened 2015-09-21
Raise for bad plot options in `_send_to_plotly`

### LINKED ISSUE #298: py.iplot should throw error if unexpected number of args
I was trying to reproduce someone else's work from a python file. 

original code:
`fig = Figure (data = data, layout = layout)`

In IPython notebook, when I did:

`py.iplot(data=data, layout=layout, filename = 'testing')`

some of the plot showed up (the data), and there were no errors, and some of the layout features did not show up, but most did not, which was confusing. I think incorrect syntax should throw an error. 

Ultimately I decided it probably wasn't using the layout argument at all (?), since this works as expected:

`py.iplot(fig, filename='testing')`

theengineear:
Fixes #298

This is backwards incompatible as obsolete `plot_options` have previously been silently ignored. It _is_ a little misleading and has caused confusion for users (#298).

Another option is to allow `layout` again, but _only_ if (1) data is given as the first argument or (2) layout is not defined in the figure. We could just throw a deprecation warning and then stop supporting it in `2.x`?

### DIFF
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -1390,6 +1390,11 @@ def _send_to_plotly(figure, **plot_options):
     """
 
     """
+    invalid_plot_options = [plot_option for plot_option in plot_options
+                            if plot_option not in DEFAULT_PLOT_OPTIONS]
+    if invalid_plot_options:
+        raise exceptions.PlotlyError('Invalid plot options, {}.'
+                                     .format(invalid_plot_options))
     fig = tools._replace_newline(figure)  # does not mutate figure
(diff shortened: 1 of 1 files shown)


theengineear:
@chriddyp @cldougl 

I don't want to make an incompatible change in the middle of `1.x`, but maybe the updated graph reference pr could make a good change to `2.x`? Then I think we're alright *all-of-a-sudden* erring on this .

Thoughts?

theengineear:
Old PR. It's pretty simple and we should just open up a new PR if we still want to fix.

PR #305, opened 2015-09-29
Dataframes have a `to_string` method!

theengineear:
We try to use duck-typing here, but it doesn’t quite make sense.


theengineear:
@etpinard | @chriddyp | @cldougl quickie :)

Anything that happens to define a `to_string` method will be used here, but since we pass arguments to `to_string`, that's almost always going to fail. This just checks that the val is the right instance instead of doing it by the [EAFP Principle](http://stackoverflow.com/questions/11360858/what-is-the-eafp-principle-in-python).

chriddyp:
:dancer:

PR #307, opened 2015-09-29
Add a test to force us to update default schema!

theengineear:
Gah, we'll never ever do this if it doesn't error on circle. The failure prints out instructions to fix it... so maybe it won't even be confusing when our future selves encounter it? :)

### DIFF
--- a/plotly/graph_reference/default-schema.json
+++ b/plotly/graph_reference/default-schema.json
@@ -1,8 +1,8 @@
 {
     "defs": {
         "metaKeys": [
-            "_isLinkedToArray",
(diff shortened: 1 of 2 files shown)


theengineear:
@chriddyp @etpinard @cldougl I realized I was about to pip-push a broken version to PyPI because we recently changed the structure of `annotations` and `shapes` in plotlyjs!

I was *just* going to fix it and move on, but I think that's not a great solution, this keeps us a little more honest.

Seem like a good idea?

etpinard:
:dancer: good idea

PR #316, opened 2015-10-02
Move figure factory

theengineear:
PR 1: Move FigureFactory out of tools.py!

We can be nice and raise a deprecation warning. Here's what that looks like:

image


theengineear:
:/ this never made it in, so I'm going to close it down since it's so old. However, I'd still really like to untangle all our imports and shift some things around.

PR #317, opened 2015-10-02
Cleaner import checking

theengineear:
PR 2: Use a single, central place to check whether optional deps are imported and _get_ them from a central source.

(WIP still based off another cleanup branch #316)

This attempts to de-uglify how we do protected imports. One module to rule them all :ring: 

Todo:
- [ ] definitely need some :tiger2:s on the new module


theengineear:
@chriddyp @cldougl @etpinard does this seem reasonable? I'm just using `importlib` to do some dynamic importing.

This is the important commit if you feel like checking this out https://github.com/plotly/python-api/commit/7cbd3550de8095d0a71a93ebc1fc08f3c8cfa6aa

Chris, I'm trying to follow your momentum with untangling plot_options logic. I think there's a lot more work to be done :sweat_smile:

theengineear:
I'm going to try and revisit this one if I end up with some quiet weekends. The way we currently do optional imports is sorta frustrating. 7cbd355 introduces a new way to tap into the `import` pathways in Python that allows a cleaner way to handle optional deps in our code.

theengineear:
Woo. Done in a different PR.

Issue #324, opened 2015-10-09
update `chunked_requests` submodule


chriddyp:
hey @theengineear - I just merged a PR in to `chunked_requests` (https://github.com/chriddyp/chunked_requests/pull/5) - how do I update this repo? Are we still doing

```
$ make sync_chunked
```

a la https://github.com/plotly/python-api/blob/master/makefile#L46-L49?

theengineear:
Yep, that should do it, you'll have to also make a commit afterwards since it hangs out as a separate code blob, fyi.

PR #325, opened 2015-10-10
sync chunked_requests after chriddyp/chunked_requests#5 merge 62581cf3a1 and address #324

kiwiroy:
looks correct, though I modified URLs in `.gitmodules` as I can not connect GitHub otherwise :blush:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "submodules/mplexporter"]
 	path = submodules/mplexporter
-	url = git://github.com/mpld3/mplexporter.git
+	url = git@github.com:mpld3/mplexporter.git
 [submodule "submodules/chunked_requests"]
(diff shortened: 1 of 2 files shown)


chriddyp:
@theengineear - look right to you?

kiwiroy:
I'm happy to revert the `.gitmodules` change for convenience.

theengineear:
The call to `make` should have made actual code changes in `plotly/plotly/chunked_requests`. In order to make it simpler for contributors (or this was the intent), we actually keep the code inside the python-api repo as a synced copy.

So, yeah, the changes in there look right, but you should have a bunch more code changes.

kiwiroy:
@theengineear, that makes sense, and should be achieved by 4f8186e.

theengineear:
:+1: lookin' good now! Thanks @kiwiroy!

PR #341, opened 2015-10-28
Add back ordering

### LINKED ISSUE #290: Add key ordering back to python api.

theengineear:
Closes #290

I won't say fixes, but it does add back some predictability, which is needed for testing in https://github.com/plotly/streambed/pull/4063.


theengineear:
I'm going to merge this in when it passes and push to pip so that the changes can be used in other repos.

theengineear:
//cc @chriddyp @etpinard

PR #348, opened 2015-11-15
Updated offline mode

chriddyp:


chriddyp:
hey @etpinard  @theengineear @cpsievert @cldougl @jackparmer - new offline functionality

plotly/offline/offline.py
```diff
@@ -8,40 +8,29 @@
...
-        'code to make your first offline graph:',
-        '',
-        'import plotly',
-        'plotly.offline.init_notebook_mode() '
-        '# run at the start of every ipython notebook',
-        'plotly.offline.iplot([{"x": [1, 2, 3], "y": [3, 1, 6]}])'
-    ]))
+def get_plotlyjs():
+    path = os.path.join('offline', 'plotly.min.js')
+    plotlyjs = resource_string('plotly', path).decode('utf-8')
```
chriddyp: in the future, I'd also like to ship an un-synced version of the `plot-schema.json` that corresponds to this particular `plotly.min.js` and validate offline mode against that `plot-schema.json`.

theengineear:
@chriddyp is this reviewable? Mind adding a little context-providing note in the top comment?

chriddyp:
@theengineear - yup, reviewable. here's some context: https://github.com/plotly/plotly.py/commit/6c5f172323893c155f7dc227280ab4239a17db5a

theengineear:
@chriddyp thanks! i'll check this out today.

plotly/offline/offline.py
```diff
@@ -8,40 +8,29 @@
 import json
 import os
 import uuid
-
-import requests
+from pkg_resources import resource_string
+import warnings
```
theengineear: :cow2: mind flipping these import statements? We typically do `from ...` after `import ...`.

plotly/offline/offline.py
```diff
@@ -8,40 +8,29 @@
...
-        os.makedirs(PLOTLY_OFFLINE_DIRECTORY)
-
-    res = requests.get(download_url)
-    res.raise_for_status()
+    warnings.warn('''
+        `download_plotlyjs` is deprecated and will be removed in the
+        next release. plotly.js is shipped with this module, it is no
+        longer necessary to download this bundle separately.
+    ''', DeprecationWarning)
+    pass
```
theengineear: Wait, if it's just deprecated, why are we getting rid of this function? Shouldn't we do that in 2.x?

plotly/offline/offline.py
theengineear: I'm a little confused how this is working? I thought you needed a line in `setup.py` telling the setup script to include the `.js` file in the package information. Like [here](https://github.com/plotly/plotly.py/blob/master/setup.py#L43).

I think you need a `'offline/*.js'` item in that list?

plotly/offline/offline.py
```diff
@@ -55,24 +44,20 @@ def init_notebook_mode():
...
     __PLOTLY_OFFLINE_INITIALIZED = True
     display(HTML('<script type="text/javascript">' +
-                 open(PLOTLY_OFFLINE_BUNDLE).read() + '</script>'))
+                 # ipython's includes `require` as a global, which
+                 # conflicts with plotly.js. so, unrequire it.
+                 'require=requirejs=define=undefined;' +
+                 '</script>' +
+                 '<script type="text/javascript">' +
+                 get_plotlyjs() +
+                 '</script>'))
```
theengineear: :( I'm not super familiar with the repercussions of this. Is this a safe thing to do?

plotly/offline/offline.py
```diff
@@ -90,6 +75,11 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly'):
                                 of the chart that will export the chart to
                                 Plotly Cloud or Plotly Enterprise
     link_text (default='Export to plot.ly') -- the text of export link
+    validate (default=True) -- validate that all of the keys in the figure
+                               are valid? omit if your version of plotly.js
+                               has become outdated with your version of
+                               graph_reference.json or if you need to include
+                               extra, unnecessary keys in your figure.
```
theengineear: This sounds like it could potentially be a bad user experience. Might we be better off just forcing users to keep their `plotly.min.js` file updated?

Rationale:

* say there *is* a conflict between your local version and plotly, when you finally want to push up a file... it would break, which is pretty confusing!
* plotly.js tries very hard to provide backwards compatibility, I think we might be better off just allowing these rare instances to raise errors.
* if we do have an out of sync `plotly.min.js`, then we also need to provide an out of sync `plot-schema`, i feel like we're potentially creating tough environments to recreate for debugging.

Would it be feasible to get away from this pattern?

I like how the graph_reference is setup, where we have the following:

* a method to download the *most recent*
* a copy that lives in a users local `~` directory
* a versioned default (for offline users and such)

The thing I *don't* like is that the initial download is frustratingly slow. Is there any way we can provide this without the long import lag?

plotly/offline/offline.py
```diff
@@ -186,4 +172,3 @@ def plot():
     """ Configured to work with localhost Plotly graph viewer
     """
     raise NotImplementedError
-
```
theengineear: Think you might have :hocho:'d the newline at the end of the file?

plotly/tests/test_optional/test_offline/test_offline.py
```diff
@@ -4,41 +4,21 @@
...
 from nose.tools import raises
 from unittest import TestCase
 
 import plotly
-from plotly.exceptions import PlotlyError
-
-dummy_js_url = ('https://gist.githubusercontent.com/chriddyp/'
-                'f40bd33d1eab6f0715dc/raw/'
-                '24cd2e4e62ceea79e6e790b3a2c94cda63510ede/'
-                'test.js')
```
theengineear: Just wondering if we want to keep these since we're only deprecating this functionality, not removing it yet?

plotly/version.py
```diff
@@ -1 +1 @@
-__version__ = '1.8.12'
```
theengineear: :+1:

setup.py
```diff
@@ -20,7 +20,7 @@ def readme():
                   "interactive, publication-quality graphs.",
       long_description=readme(),
       classifiers=[
-          'Development Status :: 4 - Beta',
```
theengineear: I don't know that it makes sense to upgrade this from beta to stable on a minor version release. Maybe we should wait for 2.x. I think it would be a good idea to have a meeting on goals for 2.x with this switch from Beta to Production in mind?

I also don't know how much weight this little line holds, so I'm not that worried about it.

plotly/offline/offline.py
chriddyp: Oh, interesting. OK, I'll add it there.

plotly/offline/offline.py
chriddyp: ah yhea, you're right.

plotly/offline/offline.py
chriddyp: yup, I've been doing it since the release of offline.

plotly/offline/offline.py
chriddyp: Here's my rational: 

- It's easier for users to understand what's going on. Upgrading the python client is a natural action taken by users who want to new functionality. Upgrading `plotly.js` is separately is another step to teach and document. 
- We can formally introduce new functionality as new releases. A new chart type in `plotly.js` will trigger new chart type releases in Python, and update the changelog accordingly.
- We can test the integration of R and Python with `plotly.js` independently, and be confident about our releases (and documentation)
- It's easier for us to document. If a user views our docs, they can check their version of the python client against the documentation's versions. If something doesn't work with their chart, it's easier for them to rule-out it being an issue with an incompatible version.
- It's easier for IT depts to manage - version x of the package will work exactly like every other user's version x.

theengineear: sold. okie dokie. so then `/offline` will now have:

* `offline/plotly.min.js`
* `offline/plot-schema.json`

And somehow we'll need to route validation through the appropriate `plot-schema.json`?

plotly/offline/offline.py
theengineear: :+1: perfect. thanks.

theengineear:
K, final major comment. Will this play well with on-prem versions? So say an on-prem user using an old version of plotly.js installs the newest version of `plotly`, they will be able to do things offline that their on-premise machine won't recognize. Is this just handled in documentation?

chriddyp:
@theengineear - yeah, handled in documentation. Eventually, maybe also handled in shareplot with validation.

theengineear:
triple :+1: sg.

plotly/offline/offline.py
chriddyp: I can't really keep this functionality in here *and* include a shipped version of `plotly.js`. I can either remove this function, or keep the warning in case someone finds an old reference of this function. In either case, the purpose of this function is redundant - a `pass` or performing the download have the same affect. 
Keep the function, or keep the warning?

plotly/offline/offline.py
chriddyp: Agreed. Won't get to this until the next version though. For now, `offline.iplot` will have slightly more validation than it used to (none), but not quite as much as it needs to.

theengineear: yep, sounds good!

chriddyp:
@theengineear - just added some more comments, and a couple fixes.

chriddyp:
image

theengineear:
^^ :heart_eyes_cat:

plotly/offline/offline.py
theengineear: OK, nbd. I think it's alright to leave it as-is then.

plotly/offline/offline.py
theengineear: :cat2: ?

plotly/tests/test_optional/test_offline/test_offline.py
theengineear: resolved, see https://github.com/plotly/plotly.py/pull/348/files#r45068339

theengineear:
Okay! As long as you confirm that you didn't take out that newline at the end of the file noted [here](https://github.com/plotly/plotly.py/pull/348/files#r44941797), :dancer: away :)

PR #358, opened 2015-11-24
Add offline plotting that doesn't depend on Jupyter

### LINKED ISSUE #353: Add offline plotting that doesn't depend on Ipython/Jupyter
It would be good to have two extra ```plotly.offline``` API entry points that return the rendered html for a given plot, and a second that rendered a given plot in a standalone local server.

```python
def plot_html(figure_or_data)
    pass

class PlotServer(object):
    #starts a simple python http server to serve the plotly js / html
    def plot(figure_or_data, return_url=False, open_url=True):
        pass
```

nzjrs:
For discussion. Fixes #353

### DIFF
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -131,13 +131,6 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
 
     plotly_platform_url = session.get_session_config().get('plotly_domain',
                                                            'https://plot.ly')
-    if (plotly_platform_url != 'https://plot.ly' and
-            link_text == 'Export to plot.ly'):
-
-        link_domain = plotly_platform_url\
-            .replace('https://', '')\
-            .replace('http://', '')
-        link_text = link_text.replace('plot.ly', link_domain)
 
     display(HTML(
         '<script type="text/javascript">'
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -56,6 +56,74 @@ def init_notebook_mode():
                  '</script>'))
 
 
+def plot(figure_or_data, html_style='standalone', show_link=True, link_text='Export to plot.ly', validate=True):
+    """ Configured to work with localhost Plotly graph viewer
+    """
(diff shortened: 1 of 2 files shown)


chriddyp:
nice, thanks @nzjrs !

I'd like `plotly.offline.plot` to also handle a few more things:
- `filename="temp-plotly.html"` argument that saves the graph to the directory
- `auto_open=True` argument which will open the plot in your web browser by default. We should piggyback on the config settings that a user can save in their config file: https://github.com/plotly/plotly.py/blob/master/plotly/plotly/plotly.py#L240
- For `standalone` sizing, I think we should make the plot full window, make the div `100% x 100%` of the body.
- I love the idea of `html_style` idea! I'm not sure about what the use case is for `body`, it seems like `div` serves that purpose

plotly/offline/offline.py
```diff
@@ -56,6 +56,74 @@ def init_notebook_mode():
...
+def plot(figure_or_data, html_style='standalone', show_link=True, link_text='Export to plot.ly', validate=True):
+    """ Configured to work with localhost Plotly graph viewer
+    """
+    EMBED_TYPES = ('standalone', 'body', 'div')
+    if html_style not in EMBED_TYPES:
+        raise ValueError('html_style must be one of %s' % ', '.join(EMBED_TYPES))
+
+    html = ""
+    if html_style == 'standalone':
+        html += '<html><head><script type="text/javascript">' + get_plotlyjs() + '</script></head><body>'
```
chriddyp: we should include
`<!DOCTYPE html>` and  `<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">`

jackgolding:
If we wanted to create an internal dashboard (i.e. not plots hosted on plot.ly) using plot.ly would using iframes to these generated HTML files be a good idea or is this what we need the enterprise mode for?

baobobs:
Any progress on this? I'm excited to test this out once it's been successfully merged.

chriddyp:
Closing this in favor of https://github.com/plotly/plotly.py/pull/378

PR #361, opened 2015-11-25
go back to using git protocol

BRONSOLO:
looking to revert the protocol change. context: https://github.com/plotly/plotly.py/pull/325#issuecomment-147120176

ping @theengineear

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "submodules/mplexporter"]
 	path = submodules/mplexporter
-	url = git@github.com:mpld3/mplexporter.git
+	url = git://github.com/mpld3/mplexporter.git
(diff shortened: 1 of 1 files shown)


theengineear:
The url here [according to these docs](https://git-scm.com/docs/gitmodules) is passed directly to `clone`. Therefore, if the `git` is an issue, perhaps we should really go to the simplest form:

`https://github.com/chriddyp/chunked-requests`

??

theengineear:
This can :dancer: or we can attempt to go to `https` as well.

BRONSOLO:
Great. I think https should be fine. I say we revert to what we were using and if we ever have a problem with this, we should consider using https.

theengineear:
^^ :+1: :dancer:

PR #377, opened 2015-12-20
Fix height to px conversion in iplot function

teklinp:


theengineear:
Fixed in https://github.com/plotly/plotly.py/commit/6cd1e9e7ef8d44ca2db36be80f962c3fe0d01413

PR #379, opened 2016-01-02
Change to offline.init_notebook_mode()

joshburkart:
Make `init_notebook_mode()` only import the JS library once. Otherwise if someone puts it at the top of a Jupyter cell and it gets executed every time (typical pattern of usage?), it adds ~1 s to the cell's execution time.


jackparmer:
+1 Thanks!

PR #380, opened 2016-01-07
Conda build

rgbkrk:
Somewhat of a follow on to #136.

All of our (Jupyter/IPython) Docker environments rely on `conda`, which means we need this setup in order for it to be packaged with our conda environments.

### DIFF
--- a/bld.bat
+++ b/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
+:: http://docs.continuum.io/conda/build.html
+:: for a list of environment variables that are set during the build process.
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,9 @@
 #!/bin/bash
 
-${PYTHON} setup.py install || exit 1;
+$PYTHON setup.py install
 
+# Add more build steps here, if they are necessary.
+
+# See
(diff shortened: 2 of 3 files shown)


nehalecky:
What's the status of this? Would be great to officially support conda. :)

theengineear:
Great. Yeah, I remember putting a little effort into this before but ended up never merging it. @rgbkrk any thoughts on maintenance? I'm wondering if we can do the following:

1. create a makefile target (or something) that basically automates the creation of the `meta.yaml` file
2. add a test that enforces the version in the `meta.yaml` file matches the version in `setup.py`. this test should be helpful in suggesting we run the command from (1).

Does that make sense? Is there a better way to keep these things in sync?

rgbkrk:
You'd think for a package like this that's pure Python that Conda would have something to auto generate this at release time. Have you submitted this to a plotly channel on Anaconda.org?

theengineear:
>Have you submitted this to a plotly channel on Anaconda.org?

Nope, haven't looked into doing that yet. Please do! Or, if you don't have time, I can look into it.

chohner:
Just found this, sorry for duplication. I did the same over at https://github.com/plotly/plotly.py/pull/523 with the current 1.12.4 version and a small readme. I can confirm that my build runs (https://anaconda.org/chohner/plotly), at least on the current OSX with python 3.5.

rgbkrk:
Closing in favor of #523.

PR #390, opened 2016-01-19
Mpl offline

cldougl:
@chriddyp @theengineear I forgot to update the changelog :disappointed:


theengineear:
:dancer: ! thanks for not letting it fall through the cracks!

PR #392, opened 2016-01-20
Added a space between attributes so that picky HTML parsers don't freak out

chrstphe:
The lack of space between the style attribute value and the seamless attribute in the following embed_code causes some parsers to consider the html code malformed:

```<iframe id="igraph" scrolling="no" style="border:none;"seamless="seamless" src="https://plotly.skd.midasplayer.com/~christophe.carvenius/28.embed" height="525px" width="100%"></iframe>```


chriddyp:
thanks @chrstphe !

Issue #399, opened 2016-02-11
Config options for offline plots


icaromedeiros:
There is no way to set the 4th parameter in plotly.js when using offline plots as it is hardcoded in `_plot_html` (https://github.com/plotly/plotly.py/blob/master/plotly/offline/offline.py#L100)

``` python
config = {}
config['showLink'] = show_link
config['linkText'] = link_text
jconfig = json.dumps(config)
```

ispmarin:
:+1: very relevant for offline plotting.

cudmore:
I would like to hide the top plotly toolbar in my plotly offline plots (python). I see this is done with a 'config' option in other versions of the plotly library. How can I do this in plotly offline for python?

I am using this function in [offline.py](https://github.com/plotly/plotly.py/blob/master/plotly/offline/offline.py) to make my plots:

```
def plot(figure_or_data,
         show_link=True, link_text='Export to plot.ly',
         validate=True, output_type='file',
         include_plotlyjs=True,
         filename='temp-plot.html',
         auto_open=True):
```

mrslezak:
Same issue here - the config options for the modebar available in plotly.js aren't implemented in plotly.py https://plot.ly/javascript/configuration-options/

Rikorose:
@yankev will this be merged into master?

gte620v:
This seems to have broken offline mode, which is still expecting the old argument signature for `_plot_html`.  See https://github.com/plotly/plotly.py/blob/4420d794ed58f31cd2c3a1f13558a9e9da798335/plotly/offline/offline.py#L311-L313

Compared to https://github.com/plotly/plotly.py/blob/4420d794ed58f31cd2c3a1f13558a9e9da798335/plotly/offline/offline.py#L163-L164

In this PR, I think line 311 should have been changed to:
```
    plot_html, plotdivid, width, height = _plot_html(
        figure_or_data, {'showLink':show_link, 'linkText':link_text}, validate,
        '100%', 525, global_requirejs=True)
```

Without this, even the most basic usage fails.

vctrd:
Hi, I'm trying to use plotly with django and found this help: https://stackoverflow.com/questions/35092571/how-to-create-charts-with-plotly-on-django
But the signature of _plot_html seems to have changed in plotly 2.0.11

Do you know if this has already been solved in a different version?

thanks

jonmmease:
[...]

PR #406, opened 2016-02-16
Issue399

yankev:
Changed the parameters for all the functions from taking on `linkText` and `showLink` to take on `config` (which is a dictionary of possible configurations).

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -98,7 +98,7 @@ To install the optional dependencies:
 pip install -r optional-requirements.txt
 ```
 
-##Testing
+##Testing 
 
 Our API uses Nose to run tests. (https://nose.readthedocs.org/en/latest/)
 
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -126,13 +126,15 @@ def _plot_html(figure_or_data, config,
     # really be somewhere other than plotly.plotly
     plotly_platform_url = plotly.plotly.get_config().get('plotly_domain',
                                                          'https://plot.ly')
+
     if (plotly_platform_url != 'https://plot.ly' and
-            link_text == 'Export to plot.ly'):
+            'linkText' in config):
 
         link_domain = plotly_platform_url\
             .replace('https://', '')\
             .replace('http://', '')
-        link_text = link_text.replace('plot.ly', link_domain)
+
+        config['linkText'] = config['linkText'].replace('plot.ly', link_domain)
 
     script = 'Plotly.newPlot("{id}", {data}, {layout}, {config})'.format(
         id=plotdivid,
@@ -157,7 +159,7 @@ def _plot_html(figure_or_data, config,
     return plotly_html_div, plotdivid, width, height
 
 
-def iplot(figure_or_data, config = {'show_link': True, 'link_text': 'Export to plot.ly'},
+def iplot(figure_or_data, config = {'showLink': True, 'linkText': 'Export to plot.ly'},
           validate=True):
     """
     Draw plotly graphs inside an IPython notebook without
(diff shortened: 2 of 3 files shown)


yankev:
@cldougl @chriddyp

plotly/offline/offline.py
```diff
@@ -97,22 +99,44 @@ def _plot_html(figure_or_data, show_link, link_text,
...
+        'linkText',
+        'showSources',
+        'displayModeBar',
+        'modeBarButtonsToRemove',
+        'modeBarButtonsToAdd',
+        'modeBarButtons',
+        'displaylogo',
+        'plotGlPixelRatio',
+        'setBackground',
+        'topojsonURL')
```
etpinard: What if we add another config option in plotly.js? This list can't stay in-sync.

etpinard: @cpsievert got this right, by not validating the config options: https://github.com/ropensci/plotly/pull/446/files

At some point, I want to make the config option part of the plot schema, where they could be use to validate user input from the python api. Before that happens though, I would much prefer not validating the config options.

jackparmer:
Related: https://github.com/plotly/plotly.py/issues/399

plotly/offline/offline.py
```diff
@@ -137,7 +161,8 @@ def _plot_html(figure_or_data, show_link, link_text,
     return plotly_html_div, plotdivid, width, height
 
 
-def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
+def iplot(figure_or_data,
+          config={'showLink': True, 'linkText': 'Export to plot.ly'},
```
chriddyp: we try really hard not to break backwards incompatibility. if we remove `show_link` and `link_text` then our user's code could break when they upgrade plotly. in this case, we might want to do something like:

```
iplot(figure_or_data, show_link=True, link_text='Export to plot.ly', **config_options)
```

plotly/offline/offline.py
```diff
@@ -152,10 +177,12 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
-    show_link (default=True) -- display a link in the bottom-right corner of
-                                of the chart that will export the chart to
-                                Plotly Cloud or Plotly Enterprise
-    link_text (default='Export to plot.ly') -- the text of export link
+    config -- a dictionary with configuration attributes for the plot,
+        default contains the following two elements:
+        showLink (default=True) -- display a link in the bottom-right corner of
+                                    of the chart that will export the chart to
+                                    Plotly Cloud or Plotly Enterprise
+        linkText (default='Export to plot.ly') -- the text of export link
```
chriddyp: we'll want to describe that these are `plotly.js` configuration options and provide a link to our documentation: https://plot.ly/javascript/configuration-options/

contributing.md
```diff
@@ -98,7 +98,7 @@ To install the optional dependencies:
 pip install -r optional-requirements.txt
 ```
 
-##Testing
+##Testing
```
chriddyp: I'd set up your editor to trim trailing white spaces on save

plotly/offline/offline.py
```diff
@@ -9,9 +9,11 @@
 import os
 import uuid
 import warnings
-from pkg_resources import resource_string
 import webbrowser
 
+from pkg_resources import resource_string
+
+
```
chriddyp: it looks like this got accidentally committed - GitHub Desktop makes it easy to not commit and discard these types of changes

plotly/offline/offline.py
yankev: Ok.  Something to note though is that the keywords for show_link ad link_text aren't in sync with the ones on the `plotly.js config options`, which could lead to some confusion.

yankev:
Made the changes in a new branch, and create a new pull request here:
[https://github.com/plotly/plotly.py/pull/410](https://github.com/plotly/plotly.py/pull/410)

PR #410, opened 2016-02-18
Issue399 (updated)

yankev:
Fix for Issue 399, with updates recommended in pull request #406.


icaromedeiros:
:+1:

davidism:
I ended up writing my own version of `_plot_html` that allows passing `config`.  I only needed a simplified version of the function so I was ok with copying it, but it would be nice to allow passing `config` to `plot` in the actual library.

theengineear:
Looks like this got handled elsewhere. Closing.

PR #417, opened 2016-03-04
Adding Scatterplot Matrix to FigureFactory

Kully:
@cldougl 

This is the first PR for the Scatterplot Matrix in the FigureFactory class in tools.py. The function currently takes in only pandas dataframes, does sanity checks on the dataframe (more than one column, either strings or numbers within a column, etc.) and then builds a scatterplot matrix based on this input. There is an option for indexing by one of the columns which must be inputted by the user (parameter: index) and for either a scatterplot or histogram along the main diagonal (parameter: diagonal). 

Things I am working on implementing:
- I want to integrate a jitter parameter that adds random noise to the scatterplot so that densities can be easily seen in a plot. I already have a jitter parameter, but it only works in certain places.
- I also am working to add to more options for diagonal plots, such as a 2d density and box plot. 
- Adding more parameters and options to the scatterplot (eg. showgrid) is also another thing I am working on expanding.
dkjb stats
heart rate against time


theengineear:
Hi @Kully, I've committed a bunch of stuff to this repo, I'll take a little weight off @cldougl 's shoulders and give this a quick review!

FWIW, for things like this, I'd be really great to get a little pic of the graphs this is making in the top comment.

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
     """
 
     @staticmethod
+    def create_scatterplotmatrix(df, useindex = False, index = " ",
+                                 diagonal = "Scatter", symbol = 0, size = 6,
+                                 height = " ", width = " ", jitter = 0,
+                                 title = "Scatterplot Matrix"):
```
theengineear: We typically keep pretty lengthy docs for [things like this](https://github.com/plotly/plotly.py/blob/master/plotly/tools.py#L1663-LL1742)

This way, when users run `help`, the information is right there.

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+
+        from plotly import tools
+        import plotly.plotly as py
+        import plotly.graph_objs as go
+        import pandas as pd
+        import random as r
```
theengineear: In general, it's a good idea to import things at the top of a file, when possible, not dynamically like this.

AFAIK, the only thing that *needs* to be imported down here is https://github.com/plotly/plotly.py/blob/master/plotly/tools.py#L1743-L1744

Additionally, we protect `pandas` imports (among other things) so that we don't need to bring the big dependency into our `setup.py` file.

Can you first add a conditional check for [`_pandas_imported`](https://github.com/plotly/plotly.py/blob/master/plotly/tools.py#L54) and then raise a helpful warning instead of importing pandas here?

Finally, I think we typically just `import random`, without changing the namespace to `r`.

Some of this stuff is a little confusing, let me know if you have questions.

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+                  "rgb(227, 119, 194)", "rgb(127, 127, 127)",
+                  "rgb(188, 189, 34)", "rgb(23, 190, 207)"]
+
+        matrix = []
+        col_names = []
+        index_vals = []
+        diagonal_choices = ["Scatter", "Histogram"]
+
+        # Check if pandas dataframe
+        if type(df) != pd.core.frame.DataFrame:
```
theengineear: You should try to prefer `isInstance(df, pd.core.frame.DataFrame)` here.

https://stereochro.me/ideas/type-vs-isinstance

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+            raise exceptions.PlotlyError("Dataframe not inputed. Please " 
+                                         "use a pandas dataframe to pro" 
+                                         "duce a scatterplot matrix.")
```
theengineear: Great messages!

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+        # Check if dataframe is 1 row or less.
+        if len(df.columns) <= 1:
+            raise exceptions.PlotlyError("Dataframe has only one column. To "
+                                         "use the scatterplot matrix, use at "
+                                         "least 2 columns.")
+
+        # Check that diagonal parameter is selected properly
+        if diagonal not in diagonal_choices:
+            raise exceptions.PlotlyError("Make sure diagonal is set to "
+                                         "either Scatter or Histogram.")
```
theengineear: Can you condense all the input checks into a single private `_` function to reduce the size of this method? It's hard to consider in it's entirety with so many lines (and hard to test!).

Here's an [example](https://github.com/plotly/plotly.py/blob/master/plotly/tools.py#L1501)

Note that you might be able to reuse other validation methods in `FigureFactory`.

theengineear: Also, to ensure that these messages don't get out-of-sync with the actual variables, you should do something like:

`'Make sure diagonal is set to one of {}'.format(diagonal_choices)`

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+                if name != index:
+                    col_names.append(name)
+
+            # Populate matrix = []
+            for name in col_names:
+                matrix.append(df[name].values.tolist())
+
+            # Check if Matrix Values are either all strings or numbers
+            for vector in matrix:   
+                if ((type(vector[0]) == float) or (type(vector[0]) == int)):
```
theengineear: `isInstance`

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+                    for entry in vector:
+                        if not ((type(entry) == float) or (type(entry) == int)):
+                            raise exceptions.PlotlyError("Error in data"
+                                                         "frame. Make sure "
+                                                         "that all entries "
+                                                         "of each column are "
+                                                         "either numbers or "
+                                                         "strings.")
```
theengineear: Just a style thing, but this would be easier to read if written:

```
raise exceptions.PlotlyError(
    'Error in dataframe. Make sure that all entries '
    'of each column are either numbers or strings.'
)
```

That goes for a bunch of these messaged.

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+                                
+                if (type(vector[0]) == str):
```
theengineear: I don't understand this, how can `vector[0]` be a `str` when you've checked `if not ((type(entry) == float) or (type(entry) == int))`?

theengineear: Oops, sorry, misread the indentation.

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+                            
+            if (type(index_vals[0]) == str):
+                for entry in index_vals:
+                    if (type(entry) != str):
+                        raise exceptions.PlotlyError("Error in data"
+                                                     "frame. Make sure "
+                                                     "that all entries "
+                                                     "of each column are "
+                                                     "either numbers or "
+                                                     "strings.")
```
theengineear: Lots of duplication here, you could write two functions `_is_string_vector` and `_is_number_vector` or something. Try and keep it [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). :smile_cat:

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+
+        if useindex == False:
```
theengineear: Always use `is` not `==` when comparing with `False`, `True`, and `None`. If you don't actually require a boolean here, you could also `if not useindex`, which is much more readable.

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+                                                         "strings.")
+                                
+                if (type(vector[0]) == str):
+                    for entry in vector:
+                        if (type(entry) != str):
+                            raise exceptions.PlotlyError("Error in dataframe. "
+                                                         "Make sure that all "
+                                                         "entries of each col"
+                                                         "umn are either "
+                                                         "numbers or strings.")
```
theengineear: Again, [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+
+        # Main Code
```
theengineear: Instead of using a `#Main Code` comment, you could consider splitting this into separate methods with clear method names. This is the way most of the other `FigureFactory` methods have been created.

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+        dim = len(matrix)
+        trace_list = []
+
+        if index_vals == []:
+            fig = tools.make_subplots(rows=dim, cols=dim)
```
theengineear: nice reuse!

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+            
+            # Insert traces into trace_list
+            for listy in matrix:
+                for listx in matrix:
+                    if (listx == listy) and (diagonal == "Histogram"):
+                        trace = go.Histogram(
+                            x = listx,
+                            showlegend = False
+                            )
```
theengineear: :smile: lookin' like JS code with all those spaces. I think Python code typically suppresses such [extraneous whitespace](https://www.python.org/dev/peps/pep-0008/#other-recommendations)

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
     """
 
     @staticmethod
+    def create_scatterplotmatrix(df, useindex = False, index = " ",
```
theengineear: Can you use `None`, not `" "` as default values for these things? `" "` is a truthy value, which is confusing, IMO.

plotly/tools.py
```diff
@@ -1443,6 +1443,376 @@ class FigureFactory(object):
...
+                    showlegend = True,
+                    barmode = "stack")
+                return fig
+
+            if diagonal == "Scatter":
+                fig['layout'].update(
+                    height = height, width = width,
+                    title = title,
+                    showlegend = True)
+                return fig
```
theengineear: If for some reason `diagonal` isn't `Histogram` or `Scatter`, `None` gets returned. This could be more robust using an `if`-`else` here.

theengineear:
@Kully thanks for submitting this! The logic flow is super clear. I do have some general concerns:

* Tests! This 100% needs to be tested before we can pull it into the repo
* Refact to reuse more of the code (DRY comments)
* Some Python convention changes (general and specific to this repo)

Let me know when you want another round of review, happy to cruise through again. Thanks for the awesome work so far!

Kully:
@theengineear Thank you so much for the feedback. I would definitely want some more feedback in another round of review.

theengineear:
@Kully please feel free to reach out with any questions Git, Python, otherwise!

jackparmer:
Looking good!

I'd love to see an optional `theme` kwarg that styles the subplots. A lot of folks like the seaborn and ggplot2 style, but it takes a lot of lines of `layout` code to achieve in plotly, especially with subplots.

Here are some nice examples.

image
via http://stackoverflow.com/questions/28197901/how-to-use-color-coding-for-intervals-for-scatter-plot-matrix-in-a-python-pandas

image
via https://plot.ly/ipython-notebooks/cufflinks/

image
via https://www.pindrop.com/visualizing-data-in-the-space-it-resides-in/

theengineear:
@Kully Awesome! I'm a little swamped until Tuesday, I probably won't be able to give this a good look until then. Feel free to give me a poke if I don't respond by Tuesday night though! I don't want this PR to loose steam!

cldougl:
@Kully awesome -- I'll take a look at this tonight as well!!

theengineear:
@Kully Taking a look now.

theengineear:
@Kully The tests are failing because you need to update the default schema. The short story is that we force new merges to keep `master1

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
     """
 
     @staticmethod
-    def create_scatterplotmatrix(df, useindex = False, index = " ",
-                                 diagonal = "Scatter", symbol = 0, size = 6,
-                                 height = " ", width = " ", jitter = 0,
-                                 title = "Scatterplot Matrix"):
+    def _scatterplot_no_index(matrix, headers,
```
theengineear: :cow2: , any reason you didn't use the full 79 characters of line length here?

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
...
-    def create_scatterplotmatrix(df, useindex = False, index = " ",
-                                 diagonal = "Scatter", symbol = 0, size = 6,
-                                 height = " ", width = " ", jitter = 0,
-                                 title = "Scatterplot Matrix"):
+    def _scatterplot_no_index(matrix, headers, 
+                              diag, size, symbol,
+                              height, width,
+                              jitter, title):
+        from plotly.graph_objs import graph_objs
+        from random import gauss
```
theengineear: The `random` import doesn't need to be protected, can you put it at the top?

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
...
-                  "rgb(227, 119, 194)", "rgb(127, 127, 127)",
-                  "rgb(188, 189, 34)", "rgb(23, 190, 207)"]
+        # Insert traces into trace_list
+        for listy in matrix:
+            for listx in matrix:
+                if (listx == listy) and (diag == "Histogram"):
+                    trace = graph_objs.Histogram(
+                        x = listx,
+                        showlegend = False
+                        )
```
theengineear: Just another style point, but this parens is over-indented. There are a few spots where you do this.

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
...
-       
-        colors = ["rgb(31, 119, 180)", "rgb(255, 127, 14)",
-                  "rgb(44, 160, 44)", "rgb(214, 39, 40)",
-                  "rgb(148, 103, 189)", "rgb(140, 86, 75)",
-                  "rgb(227, 119, 194)", "rgb(127, 127, 127)",
-                  "rgb(188, 189, 34)", "rgb(23, 190, 207)"]
+        # Insert traces into trace_list
+        for listy in matrix:
+            for listx in matrix:
+                if (listx == listy) and (diag == "Histogram"):
```
theengineear: Use single quotes `''` when possible.

theengineear: We reserve double quotes for triple-quoted strings or for strings that include the `'` character to avoid escaping.

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
...
+                        y = listx,
+                        name = None,
+                        showlegend = False
+                        )
+                else:
+                    # Add Jitter
+                    if not isinstance(listx[0], basestring):
+                        spanx = abs(max(listx) - min(listx))
+                        for j in range(len(listx)):
+                            listx[j] = listx[j] + (spanx/2)*jitter*gauss(0,1)
```
theengineear: :8ball: more spacing around operators.

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
...
+                    if not isinstance(listy[0], basestring):
+                        spany = abs(max(listy) - min(listy))
```
theengineear: Do you need to check both `listy` and `listx` here? It's possible that the division below will fail, not sure if that's what you want.

I'd either (1) check that they're *both* what you want or (2) wrap this is a try-except block. It's unclear why we know to do this operation is something is *not* a string.

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
...
+                            listy[j] = listy[j] + (spany/2)*jitter*gauss(0,1)
+
+                    trace = graph_objs.Scatter(
+                            x = listx,
+                            y = listy,
+                            mode = "markers",
+                            marker = dict(
+                                symbol = symbol,
+                                size = size),
+                            showlegend = False
```
theengineear: No extra space required in function calls.

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
...
+        if (height is None):
+            if dim <= 4:
+                height = 985
+            else:
+                height = 1231
+        if (width is None):
+            if dim <= 4:
+                width = 985
+            else:
+                width = 1231
```
theengineear: Can we create module-scoped constants for the width and height? Or, something like that to remove the need for magic :tophat: numbers here?

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
...
 
-        matrix = []
-        col_names = []
-        index_vals = []
-        diagonal_choices = ["Scatter", "Histogram"]
+        fig['layout'].update(
+        height = height, width = width,
+        title = title,
+        showlegend = True
```
theengineear: :8ball: these three lines need an indentation.

plotly/tools.py
```diff
@@ -1477,339 +1767,155 @@ def create_scatterplotmatrix(df, useindex = False, index = " ",
...
+        : param (pandas dataframe) df: array of the data with column headers
+        : param (string) index: name of the index column in data array
+        : param (string) diag: sets graph type for the main diagonal plots
+        : param (number >= 0) size: sets marker size (in px)
+        : param (string|int) symbol: sets the marker symbol type
+        : param (int|float) height: sets the height of the graph 
+        : param (int|float) width: sets the width of the graph
+        : param (number in [0,1]) jitter: adds noise to points in scatterplots
+        : param (string) title: the title label of the scatterplot matrix
+        : param (dict) **scatter_kws: a dictionary of scatter attributes
```
theengineear: No space between `:` and `p` in our docs.

theengineear: Also, try to use real types in the `()`, so `string` --> `str`.

plotly/tools.py
```diff
@@ -1477,339 +1767,155 @@ def create_scatterplotmatrix(df, useindex = False, index = " ",
...
-                                                     "of each column are "
-                                                     "either numbers or "
-                                                     "strings.")
-
-        if useindex == False:
-            # Make list of column names
-            for name in df:
-                col_names.append(name)
+        # Plot
+        py.iplot(fig, filename='ScatterPlot Matrix')
```
theengineear: Thx for the example!!

plotly/tools.py
```diff
@@ -1477,339 +1767,155 @@ def create_scatterplotmatrix(df, useindex = False, index = " ",
...
+        # Plot
+        fig = py.iplot(fig, filename='ScatterPlot Matrix')
+        ```
+        """
+        # TODO: protected until #282
+        matrix = []
+        headers = []
+        index_vals = []
+        global diag_choices
+        diag_choices = ["Scatter", "Histogram", "box"]
```
theengineear: Can we lowercase all these options? I find them hard to remember.

theengineear: Also, try not to use `global`, just define this at the top of the module:

`DIAG_CHOICES = ('scatter', 'histogram', 'box')`

plotly/tools.py
```diff
@@ -1477,339 +1767,155 @@ def create_scatterplotmatrix(df, useindex = False, index = " ",
...
+        colors = ["rgb(31, 119, 180)", "rgb(255, 127, 14)",
+                  "rgb(44, 160, 44)", "rgb(214, 39, 40)",
+                  "rgb(148, 103, 189)", "rgb(140, 86, 75)",
+                  "rgb(227, 119, 194)", "rgb(127, 127, 127)",
+                  "rgb(188, 189, 34)", "rgb(23, 190, 207)"]
```
theengineear: You can do the same with this `colors` declaration.

`DEFAULT_SCATTER_MATRIX_COLORS = ..`

plotly/tools.py
```diff
@@ -1477,339 +1767,155 @@ def create_scatterplotmatrix(df, useindex = False, index = " ",
...
+                                                         size, symbol,
+                                                         height, width,
+                                                         jitter, title)
+            return figure
+
+        if index:
+            if index not in df:
+                raise exceptions.PlotlyError("Make sure you set the index " 
+                                             "input variable to one of the "
+                                             "column names of your matrix.")
```
theengineear: matrix --> dataframe?

plotly/tools.py
```diff
@@ -1477,339 +1767,155 @@ def create_scatterplotmatrix(df, useindex = False, index = " ",
...
+            index_vals = df[index].values.tolist()
+            for name in df:
+                if name != index:
+                    headers.append(name)
+            for name in headers:
+                matrix.append(df[name].values.tolist())
+            # Check for same data-type in df columns
+            FigureFactory._validate_matrix(matrix)
+            FigureFactory._validate_index(index_vals)
```

plotly/tools.py
```diff
@@ -1477,339 +1767,155 @@ def create_scatterplotmatrix(df, useindex = False, index = " ",
...
+
+            figure = FigureFactory._scatterplot_no_index(matrix,
+                                                         headers, 
+                                                         diag,
+                                                         size, symbol,
+                                                         height, width,
+                                                         jitter, title)
+            return figure
+
+        if index:
```
theengineear: Use `else`, not `if index`?

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
...
+                            name = name,
+                            marker = dict(
+                                symbol = symbol,
+                                size = size,
+                                color = colors[color_index]),
+                            showlegend = True,
+                            )
+                    
+                    # Generate trace with INVISIBLE icon
+                    if legend_param != 1:
```
theengineear: `else`?

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
...
+                height = height, width = width,
+                title = title,
+                showlegend = True)
+            return fig
+
+
+    @staticmethod
+    def _validate_index(index_vals):
+        import types
+        NumberTypes = (types.IntType, types.LongType, types.FloatType)
```
theengineear: Can you use:

```
import number
if isinstance(x, number.Numbers):
```

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
...
+                                                 "numbers or strings.")
+            elif isinstance(vector[0], basestring):
+                if not all(isinstance(item, basestring) for item in vector):
+                    raise exceptions.PlotlyError("Error in dataframe. "
+                                                 "Make sure all entries of "
+                                                 "each column are either "
+                                                 "numbers or strings.")
+
+    @staticmethod
+    def _validate_scatterplotmatrix(df, jitter, index, diag, **scatter_kws):
```
theengineear: Typically it makes code more readable not to change the names of `*args` and `**kwargs` in call signatures.

plotly/tools.py
```diff
@@ -1443,30 +1443,320 @@ class FigureFactory(object):
...
+        if _pandas_imported is False:
```

plotly/tools.py
```diff
@@ -1477,339 +1767,155 @@ def create_scatterplotmatrix(df, useindex = False, index = " ",
...
                                          "least 2 columns.")
 
-        # Check that diagonal parameter is selected properly
-        if diagonal not in diagonal_choices:
-            raise exceptions.PlotlyError("Make sure diagonal is set to "
-                                         "either Scatter or Histogram.")
+        # Check that diag parameter is selected properly
+        if diag not in diag_choices:
+            raise exceptions.PlotlyError("Make sure diag is set to "
+                                         "one of {}".format(diag_choices))
```

theengineear:
@Kully , a bunch more style/implementation/etc comments for ya.

In general:

1. I'm still not seeing any tests in here, can you add some?
2. Do you have a [Pep8](https://www.python.org/dev/peps/pep-0008) linter set up?

Let me know if you need help with either or both of those! I'm happy to do a quick video chat to get your dev environment set up better.

Kully:
@theengineear Thank you again for the comments. I will work through them and let you know if I need any help as well.

And thanks for pointing things out in the code I actually did okay in. The positive reinforcement is helpful for morale.

theengineear:
@Kully sounds good! Yeah, keep on :truck:-in' this is great work! The resulting graphs are looking awesome.

theengineear:
@Kully any updates on this one?

Kully:
@cldougl @theengineear @jackparmer 

The 3rd PR is finally finished. I've implemented a sweet palette system where the user can input either a string  (a plotly_scale name like 'Blues', 'Portland', etc.) or a list of 2 triplets, either in RGB normalized space (0,1) or regular RGB space (0,255). Also **kwargs is added to the function.

The only things that I have to do are tidy up the code a little (a few places I can compact things into one method, some lines past 79 characters) and add a 2D Density plot capability.

Lastly, I wrote some tests in both test_optional and test_core and they all passed. I am looking at fixing my stuff wrt CircleCI on Monday.

Here are some pictures of the beautiful outputs (inspect their titles for a description):
scatterplot matrix

scatterplot matrix 1

scatterplot matrix 2

jackparmer:
Boom! 💥 This is looking real awesome!

theengineear:
💥:boom: I'll give this a review tonight!

theengineear:
@Kully Looks like you have merge conflicts and there's a failing test. The test that's failing should get resolved if you merge `master` into your branch and then push the result back up here. Are you comfortable with doing conflict resolution for this merge in git?

theengineear:
Actually, I bet the conflict is *in* the default schema file. I'd suggest just selecting all of `master`s changes and then updating that file again if things are still failing.

cldougl:
If you want in-person help with that let me know next time you're in the office @Kully !

plotly/tests/test_core/test_tools/test_figure_factory.py
```diff
@@ -1126,6 +1125,19 @@ def test_table_with_index(self):
...
+    def test_dataframe_input(self):
+
+        # check: dataframe is imported
+        df = 'foo'
+
+        self.assertRaisesRegexp(PlotlyError, "Dataframe not inputed. Please "
+                                             "use a pandas dataframe to pro"
+                                             "duce a scatterplot matrix.",
+                                tls.FigureFactory.create_scatterplotmatrix,
+                                df)
```
theengineear: 🐄 I'd suggest writing it like this:

```
pattern = (
    "Dataframe not inputed. Please use a pandas dataframe to "
    "produce a scatterplot matrix."
)
with self.assertRaisesRegexp(PlotlyError, pattern):
    tls.FigureFactory.create_scatterplotmatrix(df)
```

plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -529,3 +530,149 @@ def test_dendrogram_colorscale(self):
...
+    def test_one_column_dataframe(self):
+
+        # check: dataframe has 1 column or less
+        df = pd.DataFrame([1, 2, 3])
+
+        self.assertRaisesRegexp(PlotlyError, "Dataframe has only one column. "
+                                             "To use the scatterplot matrix, "
+                                             "use at least 2 columns.",
+                                tls.FigureFactory.create_scatterplotmatrix,
+                                df)
```
theengineear: See https://github.com/plotly/plotly.py/pull/417/files#r60284049

plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -529,3 +530,149 @@ def test_dendrogram_colorscale(self):
...
+        self.assertRaisesRegexp(PlotlyError, "Your kwargs dictionary cannot "
+                                             "include the 'size', 'color' or "
+                                             "'colorscale' key words inside "
+                                             "the marker dict since 'size' is "
+                                             "already an argument of the "
+                                             "scatterplot matrix function and "
+                                             "both 'color' and 'colorscale "
+                                             "are set internally.",
+                                tls.FigureFactory.create_scatterplotmatrix,
+                                df, **kwargs)
```
theengineear: See https://github.com/plotly/plotly.py/pull/417/files#r60284049

plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -529,3 +530,149 @@ def test_dendrogram_colorscale(self):
...
+                                palette='fake_scale')
+
+        self.assertRaisesRegexp(PlotlyError,
+                                "The items of 'palette' must be "
+                                "tripets of the form a,b,c or "
+                                "'rgbx,y,z' where a,b,c belong "
+                                "to the interval 0,1 and x,y,z "
+                                "belong to 0,255.",
+                                tls.FigureFactory.create_scatterplotmatrix,
+                                df, use_theme=True, palette=1, index='c')
```
theengineear: https://github.com/plotly/plotly.py/pull/417/files#r60284049 (all of these can be made a lot more readable)

theengineear:
@Kully Looking great. Before we merge, mind adding a test that checks the *result* of a complete function call? You tests only check paths that end in exceptions being raised. That's good, but we need to check a success path too. Even if it's just one solid test, it'd go a long way to preventing regressions in your code.

theengineear:
Super close, once that test is in and master gets merged into this successfully, we should be good!

Kully:
@theengineear Thanks again for the feedback. I'll be in the office tomorrow so hopefully I can wrap everything up then.

jackparmer:
@Kully Congrats on nearing the finish line with this one! 🐎 
If you haven't done these before, here are two useful things to learn how to do once this is merged:
- Upgrade Plotly's Python package on pip (@cldougl  can show you this one)
- Create an IPython notebook about this feature and publish it on nbviewer. The notebook should show the basic usage as well as several "flashy", publication-quality examples. The docs team will then take this notebook and publish it to Plotly's official docs. Here are some great examples:
https://plot.ly/python/table/
https://plot.ly/python/distplot/

Kully:
@theengineear  @jackparmer  @cldougl  I implemented the comments above (pattern = ..., function call test) and updated the version number.

plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -672,3 +678,42 @@ def test_valid_endpts(self):
                                 tls.FigureFactory.create_scatterplotmatrix,
                                 df, use_theme=True, index='a',
                                 palette='Blues', endpts=[2, 1])
+
+    def test_scatter_plot_matrix_kwargs(self):
```
theengineear: 💯 this is great! thanks!

plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -672,3 +678,42 @@ def test_valid_endpts(self):
...
+        df = pd.DataFrame([[2, 'Apple'], [6, 'Pear'],
+                          [-15, 'Apple'], [5, 'Pear'],
+                          [-2, 'Apple'], [0, 'Apple']],
+                          columns=['Numbers', 'Fruit'])
+
+        test_scatter_plot_matrix = tls.FigureFactory.create_scatterplotmatrix(df, index='Fruit', endpts=[-10, -1],
+                                                                              diag='histogram', height=1000, width=1000,
+                                                                              size=13, title='Scatterplot Matrix',
+                                                                              use_theme=True, palette='YlOrRd',
+                                                                              marker=dict(symbol=136))
```
theengineear: PEP8 for line length here, you can just wrap this all. It might seem fussy, but it's super helpful to follow these style rules!

plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -672,3 +678,42 @@ def test_valid_endpts(self):
...
+                                   'x': [2, -15, -2, 0],
+                                   'xaxis': 'x1',
+                                   'yaxis': 'y1'}],
+                                   'layout': {'barmode': 'stack',
+                                              'height': 1000,
+                                              'showlegend': True,
+                                              'title': 'Scatterplot Matrix',
+                                              'width': 1000,
+                                              'xaxis1': {'anchor': 'y1', 'domain': [0.0, 1.0], 'title': 'Numbers'},
+                                              'yaxis1': {'anchor': 'x1', 'domain': [0.0, 1.0], 'title': 'Numbers'}}}
```
theengineear: pep8 again here, you should allow the original `'data'` to wrap onto a new line.

plotly/tests/test_core/test_tools/test_figure_factory.py
```diff
@@ -1,4 +1,3 @@
-import math
```
theengineear: This is why your test seems to be failing, you need to put this import back.

theengineear:
@Kully excellent! Once you get the test suite passing again you can merge! 💃 !

This isn't blocking from me, but it would be great if you could get *all* the new code you've added to be pep8 compliant. There are various programs that you can run to check your files. Additionally, the editor you're using (not sure what it is) likely has a plugin that will do these checks live for you.

Let me know if you'd like help setting any of that up!

Nice job on this!

Kully:
@theengineear  Thanks to you too.
I installed a Sublime plugin that is supposed to check PEP8 things for me. Line length is not one its checks I guess but I can manual go through and fix that.

Just to be clear, would a proper wrap for your first comment look like this?

        test_scatter_plot_matrix = tls.FigureFactory.create_scatterplotmatrix(
            df, index='Fruit', endpts=[-10, -1], diag='histogram', height=1000,
            width=1000, size=13, title='Scatterplot Matrix', use_theme=True,
            palette='YlOrRd', marker=dict(symbol=136)
        )

theengineear:
@Kully https://github.com/plotly/plotly.py/pull/417#issuecomment-212692920 looks good to me!

theengineear:
One thing to note thought in your code is that you have a default mutable argument. Careful ;)

http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

Kully:
@theengineear Thanks. I'll look at it tomorrow. I have exams to study for!

Fun stuff.

Kully:
@theengineear @jackparmer @cldougl I finished all the PEP-8 stuff and added the math import to the top of test_core file. Chelsea helped me look at the mutable objects in my code and everything seemed to be working just fine; all my tests are passing.

@theengineear Do you want me to do anything with these dataframe = [], headers = [] and index_vals = [] lists?

theengineear:
@Kully , 

>Do you want me to do anything with these dataframe = [], headers = [] and index_vals = [] lists?

The safest thing to do is `dataframe=None`, etc. and then assign the values in the function. This guarantees that you can't have an issue where the same default gets used twice down the line. I'd prefer you did that. But as I said, I'm not going to block this PR anymore at this point.

theengineear:
@Kully, your test is failing because it depends on `pandas` being imported to work. However, you've got the test under `test_core`. Our *core* functionality doesn't rely on pandas (it's not in our setup.py), so it won't be imported for those tests. You'll have to move those tests to https://github.com/plotly/plotly.py/blob/master/plotly/tests/test_optional/test_figure_factory.py

Kully:
@theengineear When you say 'The safest thing to do is dataframe=None, etc. and then assign the values in the function.' how exactly is that implemented?

Do I do something like:
dataframe = None
dataframe = list(...)

or do I somehow check if isinstance(dataframe, list)?

theengineear:
>Do I do something like:

Yep! Exactly as you suggest. To be clear:

```
def foo(dataframe=None):
    if dataframe is None:
        dataframe = []

    # blah, blah...
```

cldougl:
@Kully based on what we discussed today:
- [ ] edits so this works (& tests pass) w/ python3 (you can check in your virtual env that we set up)
- [ ] update version.py 
- [ ] update CHANGELOG.md (adding your new functionality!)

cldougl:
💃💃💃 when tests pass

theengineear:
Ditto: 👯 after tests!

plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -709,33 +709,32 @@ def test_scatter_plot_matrix_kwargs(self):
...
+                       'xaxis1': {'anchor': 'y1',
+                                  'domain': [0.0, 1.0],
+                                  'title': 'Numbers'},
+                       'yaxis1': {'anchor': 'x1',
+                                  'domain': [0.0, 1.0],
+                                  'title': 'Numbers'}}
+        }
+
+        self.maxDiff = None
+        self.assertEqual(test_scatter_plot_matrix.data,
```
cldougl: @Kully looks like you left `.data` appended to `test_scatter_plot_matrix`

Kully:
@cldougl  Just removed the .data

theengineear:
@Kully and @cldougl let me know if you'd like any help debugging test failures. One quick tip to checkout would be the 2-->3 gotcha with `.keys` and `.values` for dicts. Things like that behave in *slightly* different ways from python 2 to python 3. http://blog.labix.org/2008/06/27/watch-out-for-listdictkeys-in-python-3

Have you been able to run the Python 3 test suite locally successfully?

Kully:
@theengineear We tried 3.4 and it was working locally. But we haven't tried 3.3 yet.

Kully:
@cldougl @theengineear 
Very weird.

I changed the diag param from 'histogram' to both 'scatter' and 'box' (and the respective expected scatter plot matrices) and it worked! Will check out 'histogram' option again and see if it fails.

Kully:
@cldougl @theengineear @jackparmer 

Good and ready for a merge!

theengineear:
Not a *huge* deal, but in the future, it's helpful for reviewing when the commit history is clean:

image

I'm guessing those commits were related to failures only occurring on Circle?

Even if that's the case, you can rewrite git history (as long as no one is branching off your branch) with `git rebase -i` https://www.atlassian.com/git/tutorials/rewriting-history/

That's a little advanced `git` stuff, but it can come in really handy for things like this.

Again, just *in the future* 😸. Don't worry about it now.

theengineear:
💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃 💃

PR #412, opened 2016-02-23
Don't clobber the Notebook's require/requirejs/define instances

jdfreder:
It doesn't make sense to to clobber the require.js globals that live in the notebook for Plotly's offline mode.  Nor does it make sense to load the Plotly JS from the CDN when running in offline mode.  This PR uses requirejs named definitions and requirejs commonjs support to embed the plotly JS on the page.

This PR does not solve the workflow of getting the plotly.js build into the plotly.py repo.  In a separate PR that should be automated.

related to #408
closes #373

Issue #394, opened 2016-01-28
init_notebook_mode() breaks conversion of notebook to slides


llambourne:
Hi,

If I try and convert a test notebook to slides using 
`jupyter nbconvert --to slides test.ipynb --post serve`

if the notebook contains a call to plotly.init_notebook_mode() then the resulting slides are just a blank page and I get the following javascript error:
`TypeError: require is not a function`
from where the require function is called at the bottom of the test.slides.html file that nbconvert produces. 

the issue seems to be as a result of the javascript:
`require=requirejs=define=undefined;`
in init_notebook_mode()

jonmmease:
`require=requirejs=define=undefined` is no longer performed in `init_notebook_mode()`

PR #420, opened 2016-03-09
Jupyter Notebook: Fix window.Plotly not defined error when running all cells

dalogsdon:
In Dynamic Dashboards (https://github.com/jupyter-incubator/dashboards_server/issues/108) we need to be able to execute all code cells of a Jupyter Notebook file. Currently Plotly does not work, with the error that `window.Plotly` is not defined. This problem can also be observed in a notebook when performing a "Run All" of all cells.

@jdfreder's fix in #412 preserves requirejs when plotting in offline mode in the notebook, but causes an error if doing a "Run All" cells. The reason for this is the initialization of `window.Plotly` [occurs in a requirejs function](https://github.com/plotly/plotly.py/blob/127ebc5f0be979c52847a6427a2ada3b11fc4352/plotly/offline/offline.py#L69) and is immediately followed by an attempt to access it [outside of a requirejs function](https://github.com/plotly/plotly.py/blob/127ebc5f0be979c52847a6427a2ada3b11fc4352/plotly/offline/offline.py#L132). At this point `window.Plotly` won't yet be defined since the definition will execute asynchronously after the attempt to access it.

To fix, the call to `Plotly.newPlot()` could happen inside its own requrejs function, making use of the 'plotly' requirejs module. I have made the minimal change in this PR to demonstrate. (I kept the requirejs function that defines `window.Plotly`, but it could potentially be removed in favor of using the requirejs module only.) It works for me when doing a "Run All" in the notebook and also works when deploying a dashboard read from the notebook file.

This may also be related to #394 if the conversion of notebook to slides is doing a "Run All".

### DIFF
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -127,9 +127,11 @@ def _plot_html(figure_or_data, show_link, link_text,
         'class="plotly-graph-div">'
         '</div>'
         '<script type="text/javascript">'
+        'require(["plotly"], function(Plotly) {{'
         'window.PLOTLYENV=window.PLOTLYENV || {{}};'
         'window.PLOTLYENV.BASE_URL="' + plotly_platform_url + '";'
         '{script}'
+        '}});'
         '</script>'
         '').format(
         id=plotdivid, script=script,


jdfreder:
Thanks @dalogsdon , :+1:

theengineear:
@jdfreder , does this need to wait on any dev you're doing, or can we merge this in now?

jdfreder:
@theengineear thanks for asking, but no, it doesn't, go ahead and merge away :)

theengineear:
great, thx!

PR #424, opened 2016-03-17
Require plotly before attempting to show a plot

Konubinix:
When refreshing a notebook, some plot may not be shown because the plotly
initialization script did not finish before the plot was shown. It results in a
console error like 'Plotly is not defined'. This patch fixes this error by
making every plot require plotly.


Konubinix:
Duplicate of #420

Issue #340, opened 2015-10-26
Warn if using old Plotly/Python version


jackparmer:
We answer several support emails where users simply have ancient versions of the Plotly python client. Maybe should consider more aggressive update suggestions.

jonmmease:
I don't *think* this is necessary now that plotly.py has an offline mode and the documentation indicates the required plotly.py version for each documentation page.  Feel free to reopen if you don't agree 🙂

PR #436, opened 2016-04-12
Check Version on `import plotly`

yankev:
Related to Issue https://github.com/plotly/plotly.py/issues/340.  
I'm reading the version from version.py from the plotly.py repo and comparing to the local version.py. 
@chriddyp , @cldougl


yankev:
I also tried retrieving the version number using pip (I have another branch with this one here), but there was some weird issue when testing under Python 3. I've ran the tests on both 2.7 and 3.4 locally and it seems fine. So in the end I returned to this method, however the test that's failing is a call to py.image.get()

yankev:
Updated PR with conflicts resolved: https://github.com/plotly/plotly.py/pull/451

Issue #439, opened 2016-04-14
Offline mode: Uncaught ReferenceError: require is not defined


ricleal:
Hello all,

I just updated to plotly from 1.9.5 to 1.9.8 and the offline mode is not working anymore...

Now I have a javascript error:
`Uncaught ReferenceError: require is not defined`

Which arises from the rendered code:
```javascript
require(["plotly"], function(Plotly) {
    window.PLOTLYENV = window.PLOTLYENV || {};
```
Note that in version 1.9.5, the `require(["plotly"], function(Plotly) {` did not exist!

Any workaround this issue?

Thanks in advance!

yankev:
Can you provide some more detail of what you're trying to do (ie: the code that generated this error).

ricleal:
I'm embedding plotly in Django.

I generate the javascript as:
```python
    fig = go.Figure(data=data, layout=layout)
    plot_div = plot(fig, output_type='div', include_plotlyjs=False)
```

In my html I have:
```html
<script src="https://cdn.plot.ly/plotly-latest.min.js"></script>

<!-- Renders the plot_div variable -->
<div >{{plot_div|safe}}</div>
```

It used to work in version 1.9.5...

Any javascript I could include for the `require` function?

Thanks

cldougl:
This should work for you now with 1.9.9 -- but please let us know if you're still having an issue!

cldougl:
Closing due to update. @ricleal feel free to let us know if you're still having issues with the newest version (currently 1.12.7)

gerard0315:
@cldougl hi I'm using plotly 2.0.9 but still having this issue. I was using plotly to render a static matplotlib chart, in [nteract/display-area](https://github.com/nteract/nteract/tree/master/packages/display-area).

PR #440, opened 2016-04-15
Fix for 1.9.8 offline plotting issue

yankev:
https://github.com/plotly/plotly.py/issues/439
@cldougl, @chriddyp


plotly/offline/offline.py
```diff
@@ -121,17 +121,27 @@ def _plot_html(figure_or_data, show_link, link_text,
         layout=jlayout,
         config=jconfig)
 
+    if kwargs and kwargs['browser']:
```
chriddyp: instead of "browser" maybe something more explicit, like "global_requirejs"

cldougl:
+1 for using `global_requirejs` over `**kwargs` thanks for changing that @yankev 
💃 (when the tests ✔️ )

Issue #447, opened 2016-04-27
Different error fields in error response


AltxSF:
When generating images from a batch script, I get the following error:

```
Traceback (most recent call last):
  File "utils/generate_graphs.py", line 414, in <module>
    manager.run()
  File "/usr/local/lib/python2.7/dist-packages/flask_script/__init__.py", line 412, in run
    result = self.handle(sys.argv[0], sys.argv[1:])
  File "/usr/local/lib/python2.7/dist-packages/flask_script/__init__.py", line 383, in handle
    res = handle(*args, **config)
  File "/usr/local/lib/python2.7/dist-packages/flask_script/commands.py", line 216, in __call__
    return self.run(*args, **kwargs)
  File "utils/generate_graphs.py", line 123, in generate_graph
    _generate_dot_chart(correlations, obj['ObjID'] + '_postfix')
  File "utils/generate_graphs.py", line 337, in _generate_dot_chart
    _save_image(fig, filename)
  File "utils/generate_graphs.py", line 400, in _save_image
    py.image.save_as(fig, filename=absolute_path)
  File "/usr/local/lib/python2.7/dist-packages/plotly/plotly/plotly.py", line 760, in save_as
    img = cls.get(figure_or_data, format, width, height, scale)
  File "/usr/local/lib/python2.7/dist-packages/plotly/plotly/plotly.py", line 683, in get
    raise exceptions.PlotlyError(return_data['error'])
KeyError: 'error'
```

`return_data` does not have an `error` key for that specific exception. `return_data` is

```
{u'errors': [{u'path': None, u'message': u'Unknown Image Server Error', u'code': u'UNKNOWN', u'field': None}], u'detail': u'Unknown Image Server Error'}
```

I will add a pull request with defensive code to handle this, though a better solution would be to standardize the response to use `error` instead of `errors`.

yankev:
@AltxSF Do you happen to have some shareable code that'll reproduce the issue?

AltxSF:
Yes. It happened originally when getting my requests throttled (professional version), but you can also reproduce it if you try using the wrong credentials at `~/.plotly/credentials`. Then run:

```
import random
import plotly.plotly as py
import plotly.graph_objs as go

data = go.Bar(
    x=random.sample(range(100), 25),
    y=random.sample(range(100), 25),
    name='My Graph'
)

py.image.save_as([data], filename='out.png')
```

```
Traceback (most recent call last):
  File "test.py", line 13, in <module>
[...]

PR #448, opened 2016-04-27
Check for different error fields in error response (#447).

AltxSF:
Fix for #447.

### DIFF
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -680,7 +680,14 @@ class image:
                 raise exceptions.PlotlyError("The response "
                                              "from plotly could "
                                              "not be translated.")
-            raise exceptions.PlotlyError(return_data['error'])
+
+            error_msg = return_data.get('error')
+            if error_msg is None:
+                error_msg = return_data.get('errors')
+            if error_msg is None:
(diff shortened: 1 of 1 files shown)


plotly/plotly/plotly.py
```diff
@@ -680,7 +680,14 @@ def get(figure_or_data, format='png', width=None, height=None, scale=None):
                 raise exceptions.PlotlyError("The response "
                                              "from plotly could "
                                              "not be translated.")
-            raise exceptions.PlotlyError(return_data['error'])
+
+            error_msg = return_data.get('error')
+            if error_msg is None:
+                error_msg = return_data.get('errors')
+            if error_msg is None:
```
yankev: You can avoid the extra if statement by setting 
Line 688 as the default value in the get method.

plotly/plotly/plotly.py
```diff
@@ -680,7 +680,14 @@ def get(figure_or_data, format='png', width=None, height=None, scale=None):
...
+                error_msg = 'Unknow error: ' + return_data
```
yankev: Unknown is misspelled here.

theengineear:
Closing, I think this should be handled by #640.

PR #453, opened 2016-05-03
First Push for Trisurf Plots

Kully:
@jackparmer @theengineear @cldougl 

This is the PR for create_trisurf() in FigureFactory.

The functionality so far is this: it takes arrays x, y and z (the vertices of the points) which are to be created outside the function. This requires a _import numpy _and _from scipy.spatial import Delaunay_ in the notebook.

It looks pretty cool right now. Like the scatterplotmatrix, you can enter a Plotly colorscale name in colorscale= to get your heatmap as a function of the z-coordinate. Or you can opt to enter a list of 2 rgb tuples (or normalized rgb tuples, where each coordinate is between 0 and 1).

I also included a doc string which outputs some cool looking plots.

Things I need to do:
-add tests (which I can do now!)
-potential **kwargs

Bring forth your comments!


Kully:
trisurf plot - sphere

trisurf plot - mobius band

trisurf plot - light cone

trisurf plot - calamari

Kully:
Some more color variety for y'all.

trisurf plot - viridis

jackparmer:
HECK YEAH! These are looking great. Can't wait to replace the mammoth 🐘 we have here:
https://plot.ly/python/trisurf/
With some slick 1 liners :)

plotly/tools.py
```diff
@@ -1425,6 +1425,12 @@ def return_figure_from_figure_or_data(figure_or_data, validate_figure):
 _DEFAULT_INCREASING_COLOR = '#3D9970'  # http://clrs.cc
 _DEFAULT_DECREASING_COLOR = '#FF4136'
 
+DEFAULT_PLOTLY_COLORS = ['rgb(31, 119, 180)', 'rgb(255, 127, 14)',
```
theengineear: 🐄 This should go to the top of the file. The above constants should have been at the top as well.

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
     """
 
     @staticmethod
+    def _unlabel_rgb(colors):
```
theengineear: :dolls:

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+            unlabelled_tuple = (numbers[0], numbers[1], numbers[2])
+            unlabelled_colors.append(unlabelled_tuple)
+
+        return unlabelled_colors
+
+    @staticmethod
+    def _find_intermediate_color(tuple1, tuple2, t):
+        # Given two color tuples (in normalized RGB space)
+        # and a value 0 < t < 1, spit out a color that is
+        # t percent from tuple1 to tuple2.
```
theengineear: :dolls: Comment --> docstring here.

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+        diff_1 = float(tuple2[1] - tuple1[1])
+        diff_2 = float(tuple2[2] - tuple1[2])
+
+        new_tuple = (tuple1[0] + t*diff_0,
+                     tuple1[1] + t*diff_1,
+                     tuple1[2] + t*diff_2)
+        return new_tuple
+
+    @staticmethod
+    def _unconvert_from_RGB_255(colors):
```
theengineear: `normalize_rgb_colors`?

theengineear: 🐄 You might also consider having it handle *one* color iterable at a time, then you'd just use it in a list comprehension:

`blah = [normalize_rgb_color(color) for color in colors]`

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+        un_rgb_colors = []
+        for color in colors:
+            un_rgb_color = (color[0]/(255.0),
+                            color[1]/(255.0),
+                            color[2]/(255.0))
```
theengineear: any fear that `color[i] > 255`?

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+
+            un_rgb_colors.append(un_rgb_color)
+        return un_rgb_colors
+
+    @staticmethod
+    def _map_z2color(zval, colormap, vmin, vmax):
+        # Map the normalized value zval to a
+        #corresponding color in the colormap
```
theengineear: :dolls: comment --> docstring

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+        if vmin > vmax:
+            raise ValueError("Incorrect relation between vmin and vmax."
+                             " The vmin cannot be bigger than vmax.")
```
theengineear: Is it OK if `vmin == vmax`?

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+        t = (zval - vmin)/float((vmax - vmin))  # normalize val
+        t_color = FigureFactory._find_intermediate_color(colormap[0],
+                                                         colormap[1],
+                                                         t)
+        t_color = (t_color[0]*255.0, t_color[1]*255.0, t_color[2]*255.0)
+        labelled_color = 'rgb{}'.format(t_color)
```
theengineear: Is `a` accepted anywhere, is it something we need to handle? It seems like we'd just ignore `rgba` --> `rgb`. Is that right?

Related, should we accept `hsl`/`hsla` values?

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+
+        return labelled_color
+
+    @staticmethod
+    def _tri_indices(simplices):
+        return ([triplet[c] for triplet in simplices] for c in range(3))
```
theengineear: Is this just a transpose function? If so, would a dependency like `numpy` be able to handle it? (looks like you're requiring it anyhow.

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+
+    @staticmethod
+    def _plotly_trisurf(x, y, z, simplices, colormap=None,
+                        plot_edges=None):
+        import numpy as np
```
theengineear: Don't protect this import. you can use the `_numpy_imported` or whatever it's called.

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+        tri = Delaunay(points2D)
+        simplices = tri.simplices
+
+        # Create a figure
+        fig1 = FF.create_trisurf(x=x, y=y, z=z,
+                                 colormap=[(0.2, 0.4, 0.6),(1, 1, 1)],
+                                 simplices=simplices)
+        # Plot the data
+        py.iplot(fig1, filename='Trisurf Plot')
+        ```
```
theengineear: 🎉 amazin' docs!

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+                         'Blues': ['rgb(5,10,172)', 'rgb(220,220,220)'],
+                         'Picnic': ['rgb(0,0,255)', 'rgb(255,0,0)'],
+                         'Rainbow': ['rgb(150,0,90)', 'rgb(255,0,0)'],
+                         'Portland': ['rgb(12,51,131)', 'rgb(217,30,30)'],
+                         'Jet': ['rgb(0,0,131)', 'rgb(128,0,0)'],
+                         'Hot': ['rgb(0,0,0)', 'rgb(255,255,255)'],
+                         'Blackbody': ['rgb(0,0,0)', 'rgb(160,200,255)'],
+                         'Earth': ['rgb(0,0,130)', 'rgb(255,255,255)'],
+                         'Electric': ['rgb(0,0,0)', 'rgb(255,250,220)'],
+                         'Viridis': ['rgb(68,1,84)', 'rgb(253,231,37)']}
```
theengineear: Should we move this to a constant somewhere?

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+
+        if simplices is None:
+            raise exceptions.PlotlyError("Make sure you enter 'simplices' "
+                                         "in the trisurf function.")
```
theengineear: If we require this, should it just be a positional argument in the call signature?

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+        if colormap is None:
+            colormap = [DEFAULT_PLOTLY_COLORS[0],
+                        DEFAULT_PLOTLY_COLORS[1]]
+            colormap = FigureFactory._unlabel_rgb(colormap)
+            colormap = FigureFactory._unconvert_from_RGB_255(colormap)
+
+        if isinstance(colormap, str):
+            if colormap not in plotly_scales:
+                raise exceptions.PlotlyError("You must pick a valid "
+                                             "plotly colorscale name.")
```
theengineear: Can you print out a list of `valid` names here to be more helpful for users?

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+        labelled_color = 'rgb{}'.format(t_color)
+
+        return labelled_color
+
+    @staticmethod
+    def _tri_indices(simplices):
+        return ([triplet[c] for triplet in simplices] for c in range(3))
+
+    @staticmethod
+    def _plotly_trisurf(x, y, z, simplices, colormap=None,
```
theengineear: 🐄 `_get_trisurf_data`?

I don't think we need to keep `_plotly_*`-ing the names of these. It's clear that the user is using plotly to generate this stuff, IMO.

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+                      max_zmean) for zz in zmean])
+        I, J, K = FigureFactory._tri_indices(simplices)
+
+        triangles = graph_objs.Mesh3d(x=x, y=y, z=z, facecolor=facecolor,
+                                      i=I, j=J, k=K, name='')
+
+        if plot_edges is None:  # the triangle sides are not plotted
+            return graph_objs.Data([triangles])
+
+        else:
```
theengineear: 🐄 you don't need an `else` after the `return` in the `if` above.

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+    def create_trisurf(x, y, z, colormap=None, simplices=None,
+                       title='Trisurf Plot',
+                       showbackground=True,
+                       backgroundcolor='rgb(230, 230, 230)',
+                       gridcolor='rgb(255, 255, 255)',
+                       zerolinecolor='rgb(255, 255, 255)',
+                       height=800, width=800,
+                       aspectratio=dict(x=1, y=1, z=1)):
+        """
+        Returns data for a triangulated surface plot.
```
theengineear: Returns a whole figure, not just the `data`, right?

plotly/tools.py
```diff
@@ -1443,6 +1449,333 @@ class FigureFactory(object):
...
+                xaxis=graph_objs.XAxis(axis),
+                yaxis=graph_objs.YAxis(axis),
+                zaxis=graph_objs.ZAxis(axis),
+                aspectratio=dict(
+                    x=aspectratio['x'],
+                    y=aspectratio['y'],
+                    z=aspectratio['z']),
+                )
+        )
+        fig1 = graph_objs.Figure(data=data1, layout=layout)
```
theengineear: 🐄 just return this directly?

theengineear:
@Kully Looking amazing! Mostly small comments about style and organization. Ping me when tests are up for another review!

empet:
I'm really depressed seeing the code, some comments and reactions.

plotly/tools.py
jackparmer: @theengineear If these are going somewhere special, what do you think of also making the other ColorBrewer scales available by default (and going to a 9 color scale instead of 2)?

https://github.com/jackparmer/colorlover/blob/master/colorlover/__init__.py#L1353

Maybe also these 4 cubic helix scales?
http://www.ifweassume.com/2014/04/cubehelix-colormap-for-python.html

theengineear: >going to a 9 color scale instead of 2

That's fine by me.

>what do you think of also making the other ColorBrewer scales available by default

We shouldn't pull in `colorlover` as a dep, but I'm definitely not opposed to adding an example that shows how you can leverage that library. Same with `cubehelix`. I don't want to pull in that dep, but it'd be great to show an example using that scale.

Kully: > going to a 9 color scale instead of 2

Do we need to have a 9 color scale? My stuff so far (scatterplotmatrix and trisurf) only requires and uses 2 color lists and just computes the distance between the two colors for heatmapping numeric values, etc. Is it computationally more efficient to go with a larger number, or do other functions use a larger number of colors?

jackparmer: > Do we need to have a 9 color scale?

We don't need to have a 9 color scale, totally optional. 

> Is it computationally more efficient

Instead of interpolating between 2 values in RGB colorspace, you'd be interpolating between 9, which could give some more interesting scales.

This isn't a sticking point for me at all though, sorry for the distraction. As @theengineear mentioned, I'm just as happy with some examples in the docs that show how to use colorbrewer and cubic helix scales.

Kully:
trisurf plot - boys surface

jackparmer:
Haha, awesome, boy's surface is the logo for Mayavi: 
http://docs.enthought.com/mayavi/mayavi/auto/examples.html#example-gallery
I've tried to make some of their other example plots... You got much further than me!

Kully:
@theengineear @jackparmer @cldougl @chriddyp

Final something with no errors. Thanks a lot to Andrew for helping me out with this last night.
Checklist:
-Pep-8 compliant
-tests are thorough and work

theengineear:
Great! I'll take a peek today!

plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -529,3 +530,451 @@ def test_dendrogram_colorscale(self):
...
+                             'yaxis': {'backgroundcolor': 'rgb(230, 230, 230)',
+                              'gridcolor': 'rgb(255, 255, 255)',
+                              'showbackground': True,
+                              'zerolinecolor': 'rgb(255, 255, 255)'},
+                             'zaxis': {'backgroundcolor': 'rgb(230, 230, 230)',
+                              'gridcolor': 'rgb(255, 255, 255)',
+                              'showbackground': True,
+                              'zerolinecolor': 'rgb(255, 255, 255)'}},
+                            'title': 'Trisurf Plot',
+                            'width': 800}}
```
theengineear: 😿 Can you reformat this? You could even save on indentation by defining it at the top of the file or something. For example, this would be much better:

```
{
    'data': [
        {
            'facecolor': ['rgb(143.0, 123.0, 97.000000000000014)',
                          'rgb(255.0, 127.0, 14.000000000000007)',
                          'rgb(143.0, 123.0, 97.000000000000014)',
                          'rgb(31.0, 119.0, 180.0)',
                          'rgb(143.0, 123.0, 97.000000000000014)',
                          'rgb(31.0, 119.0, 180.0)',
                          'rgb(143.0, 123.0, 97.000000000000014)',
                          'rgb(255.0, 127.0, 14.000000000000007)'],
            'i': [3, 1, 1, 5, 7, 3, 5, 7],
            'j': [1, 3, 5, 1, 3, 7, 7, 5],
            'k': [4, 0, 4, 2, 4, 6, 4, 8],
            'name': '',
            'type': 'mesh3d',
            'x': np.array([-1.,  0.,  1., -1.,  0.,  1., -1.,  0.,  1.]),
            'y': np.array([-1., -1., -1.,  0.,  0.,  0.,  1.,  1.,  1.]),
            'z': np.array([ 1., -0., -1., -0.,  0.,  0., -1.,  0.,  1.])
        },
        {
        'line': {'color': 'rgb(50, 50, 50)', 'width': 1.5},
        'mode': 'lines',
        'type': 'scatter3d',
        'x': [-1.0, 0.0, 0.0, -1.0, None, 0.0, -1.0, -1.0, 0.0, None, 0.0,
              1.0, 0.0, 0.0, None, 1.0, 0.0, 1.0, 1.0, None, 0.0, -1.0, 0.0,
              0.0, None, -1.0, 0.0, -1.0, -1.0, None, 1.0, 0.0, 0.0, 1.0,
              None, 0.0, 1.0, 1.0, 0.0, None],
        'y': [0.0, -1.0, 0.0, 0.0, None, -1.0, 0.0, -1.0, -1.0, None, -1.0,
              0.0, 0.0, -1.0, None, 0.0, -1.0, -1.0, 0.0, None, 1.0, 0.0,
              0.0, 1.0, None, 0.0, 1.0, 1.0, 0.0, None, 0.0, 1.0, 0.0, 0.0,
              None, 1.0, 0.0, 1.0, 1.0, None],
        'z': [-0.0, -0.0, 0.0, -0.0, None, -0.0, -0.0, 1.0, -0.0, None,
              -0.0, 0.0, 0.0, -0.0, None, 0.0, -0.0, -1.0, 0.0, None, 0.0,
              -0.0, 0.0, 0.0, None, -0.0, 0.0, -1.0, -0.0, None, 0.0, 0.0,
              0.0, 0.0, None, 0.0, 0.0, 1.0, 0.0, None]
        }
    ],
    'layout': {
        'height': 800,
        'scene': {'aspectratio': {'x': 1, 'y': 1, 'z': 1},
        'xaxis': {'backgroundcolor': 'rgb(230, 230, 230)',
        'gridcolor': 'rgb(255, 255, 255)',
        'showbackground': True,
        'zerolinecolor': 'rgb(255, 255, 255)'},
        'yaxis': {'backgroundcolor': 'rgb(230, 230, 230)',
        'gridcolor': 'rgb(255, 255, 255)',
        'showbackground': True,
        'zerolinecolor': 'rgb(255, 255, 255)'},
        'zaxis': {'backgroundcolor': 'rgb(230, 230, 230)',
        'gridcolor': 'rgb(255, 255, 255)',
        'showbackground': True,
        'zerolinecolor': 'rgb(255, 255, 255)'}},
        'title': 'Trisurf Plot',
        'width': 800
    }
}
```

plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -529,3 +530,451 @@ def test_dendrogram_colorscale(self):
...
+                                'gridcolor': 'rgb(2, 2, 2)',
+                                'showbackground': False,
+                                'zerolinecolor': 'rgb(3, 3, 3)'}},
+                              'title': 'Fun',
+                              'width': 500}}
+        """
+
+        print(test_trisurf_plot)
+
+        print(exp_trisurf_plot)
```
theengineear: You can remove these now :)

plotly/tools.py
```diff
@@ -1443,6 +1449,398 @@ class FigureFactory(object):
     """
 
     @staticmethod
+    def _unlabel_rgb(colors):
+        """
+        This function takes a list of two 'rgb(a, b, c)' color strings and
```
theengineear: 🐄 Python doc strings typically follow one of the following patterns:

## Single line docstring

```
"""I fit on one line!"""
```

## Multi-line docstring

```
"""
Short summary

A more detailed explanation following a blank line.

"""
```

plotly/tools.py
```diff
@@ -1443,6 +1449,398 @@ class FigureFactory(object):
...
+        returns a list of the color tuples in tuple form without the 'rgb'
+        label. In particular, the output is a list of two tuples of the form
+        (a, b, c)
+        """
+        unlabelled_colors = []
+        for color in colors:
+            str_vals = ''
+            for index in range(len(color)):
```
theengineear: 🐄 i think this would read better as:

```
for character in color:
    # ... etc, etc
```

plotly/tools.py
```diff
@@ -1443,6 +1449,398 @@ class FigureFactory(object):
...
+                else:
+                    numbers.append(float(str_num))
+                    str_num = ''
+            unlabelled_tuple = (numbers[0], numbers[1], numbers[2])
+            unlabelled_colors.append(unlabelled_tuple)
+
+        return unlabelled_colors
+
+    @staticmethod
+    def _find_intermediate_color(tuple1, tuple2, t):
```
theengineear: 🐄 can we give `tuple1` and `tuple2` and `t` better variable names? Naming things is always more time-consuming than you want it to be, but it *is* helpful for code readability.

plotly/tools.py
```diff
@@ -1443,6 +1449,398 @@ class FigureFactory(object):
...
+    @staticmethod
+    def _trisurf(x, y, z, simplices, colormap=None,
+                 plot_edges=None, Xe=None, Ye=None, Ze=None):
+        """
+        Refer to FigureFactory.create_trisurf() for docstring
+        """
+        # numpy import check
+        if _numpy_imported is False:
+            raise ImportError("FigureFactory._trisurf() requires "
+                              "numpy imported.")
```
theengineear: 👍

plotly/tools.py
```diff
@@ -1443,6 +1449,398 @@ class FigureFactory(object):
...
+
+        if plot_edges is None:  # the triangle sides are not plotted
+            return graph_objs.Data([triangles])
+
+        # define the lists Xe, Ye, Ze, of x, y, resp z coordinates of
+        # edge end points for each triangle
+        # None separates data corresponding to two consecutive triangles
+        lists_coord = ([[[T[k % 3][c] for k in range(4)]+[None]
+                        for T in tri_vertices] for c in range(3)])
+        if Xe is None:
```
theengineear: 🐄 try not to use `TitleCase` for anything but `class`es in Python. Can we rename this something like `x_edge`?

plotly/tools.py
```diff
@@ -1443,6 +1449,398 @@ class FigureFactory(object):
...
+
+        if isinstance(colormap, str):
+            if colormap not in plotly_scales:
+                raise exceptions.PlotlyError("You must pick a valid "
+                                             "plotly colorscale "
+                                             "name from {}"
+                                             .format(
+                                                 list(plotly_scales.keys())
+                                             )
+                                             )
```
theengineear: 🐄 formatting on this is wonky.

theengineear:
Still looking good to me! Ping me when you're all finished and I'll give it a final review.

Kully:
@theengineear

jackparmer:
It would be great to do a sanity check making sure this `figure_factory`
still works smoothly with PLY files.

http://moderndata.plot.ly/visualizing-56-ply-files-in-python/

Is custom surface coloring handled?

https://plot.ly/python/3d-surface-coloring/

Kully:
> It would be great to do a sanity check making sure this `figure_factory`
> still works smoothly with PLY files.
> 
> http://moderndata.plot.ly/visualizing-56-ply-files-in-python/
> 
> Is custom surface coloring handled?
> 
> https://plot.ly/python/3d-surface-coloring/

No, custom coloring isn't handled yet, but I can add.

I can check the PLT files as well.

Kully:
@jackparmer It's more than working... =D

trisurf plot - chopper

trisurf plot - beethoven

jackparmer:
Awesome - Beethoven in the house! Last thing we need is support for custom
coloring. Lots of customers need coloring to change with a variable that is
not the Z axis.

Kully:
> Awesome - Beethoven in the house! Last thing we need is support for custom
> coloring. Lots of customers need coloring to change with a variable that is
> not the Z axis.

Makes sense and I'm on it.

Kully:
@jackparmer @theengineear @cldougl 

I have added the option to write a function dist_func(x, y, z) and input that into the Trisurf function.

A couple notes:
-For my examples in the create_trisurf doc string, they only work in my notebook if I do the usual `cd adam/plotly/plotly.py` (i.e. move to the appropriate directory). I don't think I had to add something like that for a doc like this, as I imagine the user will have access to the plotly module from anywhere.
-I had previously added a `from scipy.spatial import Delaunay` line to test_figure_factory.py in test_optional (which was there in the previous PR update) so that my trisurf tests would work. Just a note.

theengineear:
>For my examples in the create_trisurf doc string, they only work in my notebook if I do the usual cd adam/plotly/plotly.py (i.e. move to the appropriate directory). I don't think I had to add something like that for a doc like this, as I imagine the user will have access to the plotly module from anywhere.

AH! That's why you don't need the special `activate` file for your `venv`. Makes sense! Also, nope, don't include that in the docstring. The user should be A-OK.

>I had previously added a from scipy.spatial import Delaunay line to test_figure_factory.py in test_optional (which was there in the previous PR update) so that my trisurf tests would work. Just a note.

Okie dokie.

theengineear:
No new notes from me, I think there are a few code comments in https://github.com/plotly/plotly.py/pull/453/files that were not addressed but nothing major.

Not gonna block this one --> 💃

PR #456, opened 2016-05-10
In python 3 `map` is a generator, `list()` it!

theengineear:
In python 2, you typically get `list` objects. For efficiency (AFAIK) in
python 3 you’ll get a generator. Note that you can’t iterate through
a generator more than once though!

### DIFF
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -1562,7 +1562,7 @@ class FigureFactory(object):
         points3D = np.vstack((x, y, z)).T
 
         # vertices of the surface triangles
-        tri_vertices = map(lambda index: points3D[index], simplices)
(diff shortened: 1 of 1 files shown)


theengineear:
@Kully you had a spent `generator`. This is a tricky issue between Python 2 and Python 3--good we have tests ;)

The problem is that when a `generator` is iterated over to completion, you typically can't iterate over it again. *Many* of the functions in Python 2 return `lists

theengineear:
@Kully, as a heads up, please delete branches you've merged after your finished with them (i.e., after you've merged them). Otherwise they'll stick around in https://github.com/plotly/plotly.py/branches forever 😸 

image

PR #464, opened 2016-05-18
Reverted the check to see if offline mode was ever initialized. 

yankev:
@chriddyp: Reverts #379 
Just commented out the check, could just delete it though. Also when I created a branch of master, it came with a change in `chunked requests`.


chriddyp:
You can just delete it, the history is in git.

Can you confirm that it works as expected? If so, bump the version, update the changelog, and let's release.

PR #467, opened 2016-05-19
updated changelog to reflect version update for offline fix

yankev:
@BRONSOLO 
Going to update Pip with the new version that includes the offline bug fix


BRONSOLO:
💃 after ⚪

Issue #458, opened 2016-05-13
No plotly.js found if init_notebook_mode run twice.


raphtown:
Hello,

I have been running into issues when using the offline notebook mode of plotly with jupyter.  Specifically, if I run init_notebook_mode twice, and then save and close the notebook with a plotted figure, upon re-opening the figure no longer shows and I can no longer get any plots to show.

Below is a notebook that reproduces the error.  To break it, run all cells, then run cell 2 again, save, and close.  Upon reload, it is not possible to get a plotly plot to show.  Interestingly, if you run cell 1 and 2 again, save, and close, the issue is fixed upon next re-open.

The issue does not show up at all if you do not include a plot in your saved notebook.

Cell 1:
```
from plotly.offline import init_notebook_mode
```
Cell 2:
```
init_notebook_mode()
```
Cell 3:
```
import plotly as py
import plotly.graph_objs as go

trace = go.Scatter3d(
    x=[1,0],
    y=[1,0],
    z=[1,0]
)
fig = go.Figure(data=[trace])
py.offline.iplot(fig)
```

My jupyter logs show the following:

> [W 17:32:55.779 NotebookApp] 404 GET /static/plotly.js?v=20160512163326 (127.0.0.1) 4.02ms referer=http://localhost:8895/notebooks/experiments/Test.ipynb

I am running version 1.9.10 as installed from pip on python 2.7.

empet:
This bug is present in version 1.9.11, too.
`404 GET /static/plotly.js?v=20160517213924 (::1) 8.00ms referer=http://localhost:8888/notebooks/Plotly-tst.ipynb`

PR #468, opened 2016-05-19
make init_notebook_mode idempotent

chriddyp:
Fixes https://github.com/plotly/plotly.py/issues/458

Running `init_notebook_mode` injects plotly.js into the output cell.
Before, if you ran it twice, it would clear the output cell on the
second call and remove plotly.js from DOM.
This was deceiving, because any existing plots on the page wouldn’t
disappear even though their underlying plotly.js was no longer present.
But, if you refresh the page, the plots would not render (as plotly.js
was no longer in the output cell)


chriddyp:
Here is the notebook that I am using for QA and testing: https://gist.github.com/chriddyp/170f46b36a9a3be7caee96eb3d4118bc

@yankev - can you please confirm that this fixes the issue?

PR #469, opened 2016-05-19
V1.10.0 - Inject plotly.js into the output cell on every `init_notebook_mode` call

### LINKED ISSUE #458: No plotly.js found if init_notebook_mode run twice.

chriddyp:


chriddyp:
Full description of changes (from changelog):
> Version 1.9.13 fixed an issue in offline mode where if you ran `init_notebook_mode`
more than once the function would skip importing (because it saw that it had
already imported the library) but then accidentally clear plotly.js from the DOM.
This meant that if you ran `init_notebook_mode` more than once, your graphs would
not appear when you refreshed the page.
Version 1.9.13 solved this issue by injecting plotly.js with every iplot call.
While this works, it also injects the library excessively, causing notebooks
to have multiple versions of plotly.js inline in the DOM, potentially making
notebooks with many `iplot` calls very large.
Version 1.10.0 brings back the requirement to call `init_notebook_mode` before
making an `iplot` call. It makes `init_notebook_mode` idempotent: you can call
it multiple times without worrying about losing your plots on refresh.

chriddyp:
cc @yankev @theengineear @mdtusz @BRONSOLO @aggFTW

This fixes #458 in a slightly different way than version 1.9.13 proposed. This change fixes an issue in 1.9.13 introduced where `plotly.js` is injected into the DOM excessively.

This is the notebook that I am using for QA. Unfortunately, this stuff is still hard to catch in our automated tests. https://gist.github.com/chriddyp/170f46b36a9a3be7caee96eb3d4118bc - @yankev - could you make sure that this works as intended? @mdtusz - can we add a test that ensures that plotly.js isn't injected multiple times?

Thank you @aggFTW again for bringing #458 to our attention!

plotly/offline/offline.py
```diff
@@ -50,16 +52,11 @@ def init_notebook_mode():
...
-        next release. Notebook mode is now automatically initialized when
-        notebook methods are invoked, so it is no
-        longer necessary to manually initialize.
-    ''', DeprecationWarning)
-
     if not _ipython_imported:
         raise ImportError('`iplot` can only run inside an IPython Notebook.')
 
+    global __PLOTLY_OFFLINE_INITIALIZED
+    # Inject plotly.js into the output cell
```
aggFTW: You are missing

`if not __PLOTLY_OFFLINE_INITIALIZED:`

no?

chriddyp: I'm not actually, that's the heart of the fix. Before, we had this check:
```
if not __PLOTLY_OFFLINE_INITIALIZED:
    return
```

which meant that if you ran this function twice, we return `None` the second time which cleared the output cell (which previously had `plotly.js` in it). Now, we just inject `plotly.js` into the output cell every time.

theengineear: *__* oh, sweet! sorry, def confused by that on the first read.

theengineear:
💃 for the change after we hear back from @yankev on https://github.com/plotly/plotly.py/pull/469#issuecomment-220408990

aggFTW:
What we do is we use plotly inside of an `Output` ipywidget. Can you check in your notebook that this still works after this change? Widgets are not persisted to the output fields.

Basically, add this to your test notebook code as a use case:
```
from ipywidgets import Output
from IPython.core.display import display
from plotly.offline import init_notebook_mode
from plotly.graph_objs import Pie, Figure, Data
from plotly.offline import iplot

o = Output()
display(o)

with o:
    print("I'm not persisted to output area!")
    init_notebook_mode()
    data = [Pie(values=[1,2,3])]
    fig = Figure(data=Data(data))
    iplot(fig, show_link=False)
```

yankev:
@aggFTW so it seems that the `Output widget` persists, but the text and the pie chart don't after a refresh. Were the text and charts persisting in a previous version of `plotly.py`?

aggFTW:
No, it has never persisted. When plotly was broken, a refresh would show an exception though. We should avoid that behavior.

chriddyp:
@yankev sounds like we're good. can you fix or update the tests and deploy?

yankev:
@chriddyp yep got it

Issue #459, opened 2016-05-14
Wrong setting of histnorm in the method  make_hist, FigureFactory.create_distplot


empet:
`FigureFactory.create_distplot` is intended to compare the  histogram of a data set with  the kde estimation of the probability density function, and more. But Plotly Histogram can create  5 types of histograms, each one set via the key `histnorm`.
 The method `create_hist` has a drawback. It does not choose the right value for the histnorm key.  Its default  value is `histnorm='probability'`:
[https://github.com/plotly/plotly.py/blob/master/plotly/tools.py#L5084](https://github.com/plotly/plotly.py/blob/master/plotly/tools.py#L5084),  and this  is contrary to the theoretical definition of this kind of histogram, and that of the  probability density function (pdf).

When `histnorm='probability'`, the height  of a bar in histogram equals the probability that data fall within the corresponding bin.  Comparing the kde estimation of the pdf with such a histogram means that we admit that the pdf takes only values in [0,1], and this  is not right.

 In a histogram plotted with histnorm='probability density', the height `h`   of a bar is such that h*bin_size=probability that data fall in that bin=bar area. The bar area approximates the area under the pdf graph above that bin.
Hence the probability density function (pdf) or its kde estimation should be compared to the Plotly Histogram corresponding to `histnorm='probability density'`.
I illustrate below the  plot of the pdf of  Beta distribution over the two types of histograms, to point out the drawback of the `make_hist` method.
![beta_probability](https://cloud.githubusercontent.com/assets/3627253/15268520/7694b5a8-19ea-11e6-88cc-8c7f022f13f4.png)
![beta_probability_density](https://cloud.githubusercontent.com/assets/3627253/15268521/7b43e4c0-19ea-11e6-9911-4f6f4eac030f.png)

 See also in [https://plot.ly/~chelsea_lyn/11601/group-1-group-2-group-3-group-4-group-1-group-2-group-3-group-4-group-1-group-2-/](https://plot.ly/~chelsea_lyn/11601/group-1-group-2-group-3-group-4-group-1-group-2-group-3-group-4-group-1-group-2-/)  how distant from the corresponding histograms are the  last two pdfs.
Here [http://nbviewer.ipython.org/0f42b607de8f0d0c50b0ffb0ccfdff08](http://nbviewer.ipython.org/0f42b607de8f0d0c50b0ffb0ccfdff08)  is the updated plot with `histnorm='probability density'`.

yankev:
[...]

PR #473, opened 2016-05-20
Distplot fix

yankev:
Fixes issue #459 
screen shot 2016-05-20 at 2 45 18 pm

### DIFF
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -4019,7 +4019,7 @@ class FigureFactory(object):
     @staticmethod
     def create_distplot(hist_data, group_labels,
                         bin_size=1., curve_type='kde',
-                        colors=[], rug_text=[], histnorm='probability density',
+                        colors=[], rug_text=[], histnorm='probability',
                         show_hist=True, show_curve=True,
                         show_rug=True):
         """
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -5621,3 +5621,4 @@ class _Table(FigureFactory):
                         font=dict(color=font_color),
                         showarrow=False))
         return annotations
+
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -5081,7 +5081,7 @@ class _Distplot(FigureFactory):
                                x=self.hist_data[index],
                                xaxis='x1',
                                yaxis='y1',
-                               histnorm='probability',
+                               histnorm='probability density',
                                name=self.group_labels[index],
                                legendgroup=self.group_labels[index],
                                marker=dict(color=self.colors[index]),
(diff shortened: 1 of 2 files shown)


yankev:
Going to include a keyword argument to swap between the new way (where histnorm = 'probability density'), and the original (histnorm = 'probability')

yankev:
@kully @cldougl @theengineear @empet
Added an extra parameter called `histnorm` to deal with the two different types of distplots (the seaborn-like method/the original and probability density method)
I've kept the default for `histnorm` as `probability` for backwards compatibility.

plotly/tools.py
```diff
@@ -4019,7 +4019,7 @@ def create_candlestick(open, high, low, close,
     @staticmethod
     def create_distplot(hist_data, group_labels,
                         bin_size=1., curve_type='kde',
-                        colors=[], rug_text=[],
+                        colors=[], rug_text=[], histnorm='probability',
                         show_hist=True, show_curve=True,
                         show_rug=True):
         """
```
cldougl: @yankev can you add that param to the docstring as well (so around line 4039)

plotly/tools.py
```diff
@@ -4019,7 +4019,7 @@ def create_candlestick(open, high, low, close,
     @staticmethod
     def create_distplot(hist_data, group_labels,
                         bin_size=1., curve_type='kde',
-                        colors=[], rug_text=[],
+                        colors=[], rug_text=[], histnorm='probability',
```
cldougl: @yankev can you add that param to the docstring as well (so around line 4039)

empet:
@yankev The seaborn  distplot is equivalent to plotly distplot corresponding to `histnorm=probability density`.  Leaving the default as `probability` it's not a good idea because a lot of plots on Plotly cloud used wrong this key and people that take as examples these older plots can think that this version leads to a histogram comparable with an estimate of the probbaility density associated to data.

empet:
Here is the proof that seaborn distplot works like Plotly distplot with `histnorm=probability density` (see the last plot):[http://nbviewer.jupyter.org/gist/empet/1bf38aea056f1d60481e600fc68c21be?flush_cache=true](http://nbviewer.jupyter.org/gist/empet/1bf38aea056f1d60481e600fc68c21be?flush_cache=true)

yankev:
@empet oh I see, thanks. Initially I was trying to keep some backwards compatibility, but it appears as though setting `histnorm = 'probability density'` is the correct approach.

empet:
@yankev Yes, this is the recommended default value for histnorm in distplot.

plotly/tools.py
```diff
@@ -5108,7 +5111,9 @@ def make_kde(self):
             self.curve_y[index] = (scipy.stats.gaussian_kde
                                    (self.hist_data[index])
                                    (self.curve_x[index]))
-            self.curve_y[index] *= self.bin_size[index]
+
+            if self.histnorm == 'probability':
```
theengineear: Can you pull these into constants at the top of the file? Hard-coded things like this always raise 🚨's for me.

theengineear:
@yankev tests look great, thanks for adding those! 💃 me after pulling those `histnorm` values out into some constants somewhere!

PR #475, opened 2016-05-22
Allow for the option of a CDN in offline mode

yankev:
@chriddyp First attempt at it. Couldn't get it to work out. 
Notes: 
- When I view the html for the ipynb, I can see the `<script src ..>` for loading plotlyjs. 
- the connected parameter is to see if a user wants to use the cdn (could name it differently, but for now connected=True --> use the cdn)

What happens: 

screen shot 2016-05-22 at 12 13 50 am

### DIFF
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -67,10 +67,15 @@ def init_notebook_mode(connected=False):
         script_inject = (
             ''
             '<script>'
-            'if(!window.Plotly) {{'
-            'require([\'https://cdn.plot.ly/plotly-latest.min.js\'],'
+            # 'if(!window.Plotly) {{'
+            'requirejs.config({'
+            'paths: { '
+            '\'plotly\': [\'https://cdn.plot.ly/plotly-latest.min\']},'
+            '});'
+            'require([\'plotly\'],'
             'function(plotly) {window.Plotly=plotly;});'
-            '}} </script>'
+            # '}}'
+            '</script>'
             )
     else:
         # Inject plotly.js into the output cell
@@ -142,9 +147,8 @@ def _plot_html(figure_or_data, show_link, link_text,
         config=jconfig)
 
     optional_line1 = ('require(["plotly"], function(Plotly) {{ '
-                      if (global_requirejs and (not __PLOTLY_USE_CDN)) else '')
-    optional_line2 = ('}});' if (global_requirejs and (not __PLOTLY_USE_CDN))
-                      else '')
+                      if global_requirejs else '')
+    optional_line2 = ('}});' if global_requirejs else '')
 
     plotly_html_div = (
         ''
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -67,14 +67,14 @@ def init_notebook_mode(connected=False):
         script_inject = (
             ''
             '<script>'
-            # 'if(!window.Plotly) {{'
             'requirejs.config({'
             'paths: { '
             '\'plotly\': [\'https://cdn.plot.ly/plotly-latest.min\']},'
             '});'
+            'if(!window.Plotly) {{'
             'require([\'plotly\'],'
             'function(plotly) {window.Plotly=plotly;});'
-            # '}}'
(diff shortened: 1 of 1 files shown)


yankev:
After latest commit:
screen shot 2016-05-22 at 12 35 58 am

fperez:
Pinging @briehl, in case you have any extra feedback on the API for the KBase uses...

plotly/offline/offline.py
```diff
@@ -193,7 +208,7 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
 
     plot_html, plotdivid, width, height = _plot_html(
         figure_or_data, show_link, link_text, validate,
-        '100%', 525, global_requirejs=True)
+        '100%', 525, global_requirejs=__PLOTLY_USE_CDN)
```
tarzzz: @yankev : Should not this be opposite? I havent looked at the complete code, but if `__PLOTLY_USE_CDN` is `True`, should not `global_requirejs=False`?

mdtusz:
I'm pretty sure @tarzzz's comment is right - might want to switch, or change the check for `optional_line`'s to only check `_PLOTLY_USE_CDN`. Other than this, looks to me like it should be fine - just make sure that all the offline methods work with both CDN and script string injection, and that you can successfully load/reload/restart/share and all that jazz too!

yankev:
@chriddyp @tarzzz @mdtusz If you guys have time, can you look over this. I ran through the QA-offline notebook and it seems to work (just have to make sure that you start from a clean dom where `window.Plotly = undefined`).

plotly/offline/offline.py
```diff
@@ -56,20 +58,40 @@ def init_notebook_mode():
...
+    global __PLOTLY_USE_CDN
+
+    __PLOTLY_USE_CDN = connected
+
+    if connected:
+        # Inject plotly.js into the output cell
+        script_inject = (
+            ''
+            '<script>'
+            # 'if(!window.Plotly) {{'
```
mdtusz: Are these meant to be uncommented?

yankev: There were some persistence and notebook sharing issues from time to time when I left them in. So I just took them out for now.

tarzzz:
@yankev : Looks good..

cldougl:
@yankev looks good to me too 💃

jackparmer:
Hey @fperez and @briehl - This should be on `pip` now. Let us know if we can help out on this anymore!

plotly/offline/offline.py
```diff
@@ -56,20 +58,40 @@ def init_notebook_mode():
...
+    __PLOTLY_USE_CDN = connected
+
+    if connected:
+        # Inject plotly.js into the output cell
+        script_inject = (
+            ''
+            '<script>'
+            'requirejs.config({'
+            'paths: { '
+            '\'plotly\': [\'https://cdn.plot.ly/plotly-latest.min\']},'
```
chriddyp: isn't this supposed to be `https://cdn.plot.ly/plotly-latest.min.js`? Are you missing the "`.js`"?

plotly/offline/offline.py
```diff
@@ -46,7 +48,7 @@ def get_plotlyjs():
     return plotlyjs
 
 
-def init_notebook_mode():
+def init_notebook_mode(connected=False):
```
chriddyp: Can you add a docstring comment explaining what `connected` is?

plotly/offline/offline.py
```diff
@@ -56,20 +58,40 @@ def init_notebook_mode():
...
-        'define(\'plotly\', function(require, exports, module) {{'
-        '{script}'
-        '}});'
-        'require([\'plotly\'], function(Plotly) {{'
-        'window.Plotly = Plotly;'
-        '}});'
-        '}}'
-        '</script>'
-        '').format(script=get_plotlyjs())
+    global __PLOTLY_USE_CDN
```
chriddyp: why is this a `global` variable? it looks like you're only using it in this function context

PR #477, opened 2016-05-24
Plus 2 colorscale

Kully:
@theengineear  @jackparmer 

This pull request is for updating all my figurefactories (trisurf, scatterplotmatrix, violin, gantt eventually) so that they support more than 2 color lists as inputs.

I've only dealt with trisurf here, changing the framework for how colormap is validated.

Should I make separate branches for each of the figurefactories, or just work in this PR?

### DIFF
--- a/plotly/tests/test_optional/test_figure_factory.py
+++ b/plotly/tests/test_optional/test_figure_factory.py
@@ -627,18 +627,6 @@ class TestTrisurf(NumpyTestUtilsMixin, TestCase):
         self.assertRaises(PlotlyError, tls.FigureFactory.create_trisurf,
                           x, y, z, simplices, colormap='foo')
 
-        # check that colormap is a list, if not a string
-
-        pattern1 = (
-            "If 'colormap' is a list, then its items must be tripets of the "
-            "form a,b,c or 'rgbx,y,z' where a,b,c are between 0 and 1 "
-            "inclusive and x,y,z are between 0 and 255 inclusive."
-        )
-
-        self.assertRaisesRegexp(PlotlyError, pattern1,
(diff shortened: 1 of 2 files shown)


theengineear:
>Should I make separate branches for each of the figurefactories, or just work in this PR?

Just doing it in *this* PR works for me.

Kully:
@theengineear @jackparmer @choldgraf 

So Scatterplotmatrix has been completely reworked to allow for divergent and categorical colormaps. In particular, you can give it a list of an arbitrary number of colors (with mixing and matching color types) and it cat make a divergent colormap between these colors. 

If you want to assign specific index values to a specific color, just input colormap (new param) as a dictionary and assign a color to each value.

The tests should pass and some pictures are posted below.

I want to get this thing up and merged as soon as we can. Some of the trisurf code is old and has already been updated in master iirc.

scatterplot matrix - colormap dictionary

scatterplot matrix - intervals

Kully:
divergent scale

choldgraf:
yep, the trisurf code is already updated and there's a PR just needs some tiny fixes and will probably be ready to go today.

I am generally +1 to colormap improvements, as I've found it to be a little clunky. I basically just use helpers that take `matplotlib` colormaps and return `rgb` strings a ton (maybe that already exists in plotly though). There's also a big chunk of `create_trisurf` that is just choosing what colormap to use, and my guess is that could be wrapped up in a hidden function and applied to lots of other things in FF too.

choldgraf:
Cool, lmk when you've had time to peek through mine and then it'll be ready
too

Kully:
> Cool, lmk when you've had time to peek through mine and then it'll be ready
> too

Was about to check it out.

Issue #478, opened 2016-05-25
Not able to import plotly module on windows


ghtmtt:
Hi,

I'm trying to install the `plotly` library via `pip` on some windows machines.

In all the machines there is python **2** installed:

    python -V
    Python 2.7.4

running

    pip install plotly

works fine, **BUT** I'm not able to import plotly in a python shell:

    Traceback (most recent call last):
    File "<stdin>", line 1, in <module>
    File "C:\OSGeo4W\apps\Python27\lib\site-packages\plotly\__init__.py", line 31, 
    in <module>
    from plotly import (plotly, graph_objs, grid_objs, tools, utils, session,
    File "C:\OSGeo4W\apps\Python27\lib\site-packages\plotly\plotly\__init__.py", 
    line 10, in <module>
    from . plotly import (
    File "C:\OSGeo4W\apps\Python27\lib\site-packages\plotly\plotly\plotly.py",  
    line 32, in <module>
    from plotly.plotly import chunked_requests
    File "C:\OSGeo4W\apps\Python27\lib\site-packages\plotly\plotly\chunked_requests\__init__.py",    line 1, in <module>
    from . chunked_request import Stream
    File "C:\OSGeo4W\apps\Python27\lib\site-packages\plotly\plotly\chunked_requestschunked_request.py", line 5, in <module>
    from six.moves.urllib.parse import urlparse
    ImportError: No module named urllib.parse

After some researches I discovered that this error is due to the `urllib.parse` named different in python 2 or python 3.

If in the file `chuncked_request` I change the line

    from six.moves.urllib.parse import urlparse

with

    import urlparse

plotly will be correctly imported in the python envirnonment.

In all the Windows machine I tried to import plotly I always get the same error (Windows XP, 7 and 10). 

While no error encountered in Linux machines.

Any ideas on how to handle this problem? 

Maybe a check in the `chuncked_request` file?

Thanks

ghtmtt:
I kind of workaround (the code is really ugly and I don't know if it can be used anywhere) could be change the `chuncked_request` file in this way:

    from platform import python_version
    ver =  python_version()
    if ver[0] == '2':
        import urlparse
    else:
        from six.moves.urllib.parse import urlparse

I tried that and it works.. 

Maybe there is a better way to do that (sorry I'm not a developer :) )

jackparmer:
[...]

PR #479, opened 2016-05-25
python version check in order to import urlparse

ghtmtt:
This small patch should fix issue #478 
Just check the python version and run the related code so that urlparse is installed on both python2 and python3
Tested on Windows XP, Windows 10, Linux Debian Stable. All with python 2 installed.
Maybe some test on machines with python3


theengineear:
@ghtmtt sorry for the delay, is this still a problem for you in the new `2.0.0` version?

ghtmtt:
It seems OK now. Thanks!

PR #481, opened 2016-05-25
Choldgraf optimize trisurf

Kully:
@theengineear @jackparmer @choldgraf @yankev

I made a new branch off of my duplicate of choldgraf:optimize_trisurf

Modified the main test for trisurf and hopefully will pass on circle. They passed locally for my on both Python 2 and 3.


plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -753,6 +753,17 @@ def test_trisurf_all_args(self):
         self.assert_dict_equal(test_trisurf_plot['data'][1],
                                exp_trisurf_plot['data'][1])
 
+    def test_optimize_trisurf(self):
```
choldgraf: I take it you'd like to do this in a later PR?

plotly/tools.py
```diff
@@ -1508,21 +1507,14 @@ def _map_z2color(zval, colormap, vmin, vmax):
...
-        """
-        Returns a triplet of lists containing simplex coordinates
-        """
-        return ([triplet[c] for triplet in simplices] for c in range(3))
+        t = (zvals - vmin) / float((vmax - vmin))
+        t_colors = FigureFactory._find_intermediate_color(colormap[0],
+                                                          colormap[1],
+                                                          t)
+        t_colors = t_colors * 255.
+        labelled_colors = ['rgb(%s, %s, %s)' % (i, j, k)
```
choldgraf: random speed thing I noticed - it turns out that python is a fair bit faster in generating strings with the % operator as opposed to `.format`, check out [this](http://tomatohater.com/2013/08/30/python-performance-string-formatting/).

plotly/tools.py
```diff
@@ -1559,38 +1551,43 @@ def _trisurf(x, y, z, simplices, colormap=None, dist_func=None,
 
         min_mean_dists = np.min(mean_dists)
         max_mean_dists = np.max(mean_dists)
-        facecolor = ([FigureFactory._map_z2color(zz, colormap, min_mean_dists,
-                      max_mean_dists) for zz in mean_dists])
-        ii, jj, kk = FigureFactory._tri_indices(simplices)
+        facecolor = FigureFactory._map_z2color(mean_dists, colormap,
```
choldgraf: This seems to be the way that trisurf defines color of the faces, no? It looks here like it will expect `dist_func` to be callable on an `x, y, z` tuple, so don't we want to support someone giving an array of shape (n_triangles,) that just has rgb strings in it? Alternatively, if it's relatively easy to change the colors after the data has been created, maybe it's better to have a different high level function that changes them post-hoc, so that we don't re-do the triangle computations each time the colors change. In either case, I'd vote for a different variable name than `dist_func`...maybe `color_func` or just `colors`, and then explain in the docs that it could be a function. WDYT?

plotly/tools.py
```diff
@@ -1559,38 +1551,43 @@ def _trisurf(x, y, z, simplices, colormap=None, dist_func=None,
...
-        for array in lists_coord[1]:
-            for item in array:
-                y_edge.append(item)
-
-        if z_edge is None:
-            z_edge = []
-        for array in lists_coord[2]:
-            for item in array:
-                z_edge.append(item)
+        is_none = [ii is None for ii in [x_edge, y_edge, z_edge]]
```
choldgraf: If we wanted to be extra careful we could add more checks here (e.g., shape is the same for x_edge, y_edge, z_edge)

plotly/tools.py
```diff
@@ -5621,4 +5618,3 @@ def make_table_annotations(self):
                         font=dict(color=font_color),
                         showarrow=False))
         return annotations
-
```
choldgraf: Did I accidentally delete this line?

Kully:
@theengineear

plotly/tools.py
theengineear: Yikes. Interesting. Thanks for linking that!

Kully:
@choldgraf

Sounds all good. I can fix this all up now.
Did you push anything just now? I'm wondering why the tests are running after they looked finished.

choldgraf:
I haven't (and won't) push anything now that this PR is up and running!

theengineear:
@Kully , GH seems to think you made different pushes. 

image

Not sure if that's the case or not.

At any rate. I'll leave it to @Kully to fix-up/discuss as necessary. As long as tests are passing, This can 💃 into `master` whenever.

Thanks for all the help @choldgraf!

plotly/tools.py
theengineear: Looks like it 😺. @Kully mind adding the newline at the end of `plotly/tools.py` back?

Kully:
> Not sure if that's the case or not.

I actually changed version.py once and tried to push but got an error, so I changed it back and tried again with success.

choldgraf:
thanks for the help as well! I imagine that we'll need to make tweaks here and there, and can open (hopefully tiny) PRs as necessary.

@jackparmer any idea on whether the time to first plot can be improved on the JS side?

Also, once we get this working I can write some high-level functions to calculate facecolors based on neural activation and write a blog post of some sort if it's useful.

choldgraf:
ps @theengineear where are those stacks in your profile picture? Reminds me of when I was a radio DJ back in college :D

choldgraf:
ack crap - I missed one thing. I just noticed that the `create_trisurf` function still doesn't have a `plot_edges` argument. We should add that in and then pass it in the call to `_trisurf`. I can make a PR but it's a pretty simple fix if someone wants to just edit master.

Kully:
> ack crap - I missed one thing. I just noticed that the create_trisurf function still doesn't have a plot_edges argument. We should add that in and then pass it in the call to _trisurf. I can make a PR but it's a pretty simple fix if someone wants to just edit master.

I can do now. Besides, I forgot to switch dist_func to color_func ;)

choldgraf:
ok cool :)  I swear PRs are like emails...every time I send one off I have an "oh shoot I missed XXX" moment

choldgraf:
A couple other random things:

1. The `plot_edges` argument should default to `False` IMO, not `None`, since it's a bool
2. We should let the user specify an array/list of rgb colors instead of `color_func`, in case they've got pre-computed colors (e.g., activation patterns relative to certain electrodes). Maybe just a quick `isinstance(color_func, (np.ndarray, list))` then a check to make sure it's full of rgb strings. Or is that for another PR?

Kully:
> The plot_edges argument should default to False IMO, not None, since it's a bool

Okay, easy.

> We should let the user specify an array/list of rgb colors instead of color_func, in case they've got pre-computed colors (e.g., activation patterns relative to certain electrodes). Maybe just a quick isinstance(color_func, (np.ndarray, list)) then a check to make sure it's full of rgb strings. Or is that for another PR?

You want both functionalities right? I think I can do this in the same PR.

choldgraf:
Yeah I think it should be "check if the input is a list / array. if yes, then assert that it's the same length as `simplices`. Something like:

```python
        if isinstance(color_func, (list, np.ndarray)):
            if len(color_func) != len(simplices):
                raise ValueError('If colors is array-like, '
                                 'it must be the same length as simplices')
            mean_dists = color_func
        elif not dist_func:
```
around line `1536`

choldgraf:
Though note that we'll probably lose all the benefits of the optimization if someone gives a function, since it's a looping and appending to lists. It'd be faster to expect a function that takes an (n x 3) array and returns an array of shape (n,). That way the user could decide to optimize the function however they like. So you could just collapse that code to:

`mean_dists = colors(tri_vertices)`

choldgraf:
(but I think that might be worth a different PR if it seems like people are often using that functionality...my guess is that people would often pre-compute their colors)

Kully:
> (but I think that might be worth a different PR if it seems like people are often using that functionality...my guess is that people would often pre-compute their colors)

Fair enough. Yes, let's make it a separate PR then.

I was a little confused by your penultimate message:  `mean_dists = colors(tri_vertices)`

Is colors supposed to be the colors_func?

choldgraf:
ah yeah, sorry...locally I've been just calling it `colors` since it is allowed to be a list or array.

choldgraf:
Been spending the morning writing some activity functions too, so when the lighting is fixed it should look relatively purdy:

image

PR #482, opened 2016-05-25
Fixed show_edges and renamed dist_func to color_func

Kully:
@choldgraf

PR #484, opened 2016-05-26
Custom colors trisurf

choldgraf:
This is a quick add-on to #481 and #482. It basically does 3 things:

1. Add support for passing a list or array to `color_func`. In this case, it assumes the list/array has pre-computed color values and it won't try to calculate it from the z-coordinates etc.
2. Ensure that everything piped into the `Mesh3d` class is an array. This makes plotting *much* faster.
3. Rename the `_map_z2color` function because we're using this with more than just the z-values (shouldn't affect the functionality at all).

I also added a few tests, which are passing for me. @Kully @theengineear @jackparmer

### DIFF
--- a/plotly/tests/test_optional/test_figure_factory.py
+++ b/plotly/tests/test_optional/test_figure_factory.py
@@ -753,6 +753,26 @@ class TestTrisurf(NumpyTestUtilsMixin, TestCase):
         self.assert_dict_equal(test_trisurf_plot['data'][1],
                                exp_trisurf_plot['data'][1])
 
+        # Test passing custom colors
+        colors_raw = np.random.randn(simplices.shape[0])
+        colors_str = ['rgb(%s, %s, %s)' % (i, j, k)
+                      for i, j, k in np.random.randn(simplices.shape[0], 3)]
+
+        # Color == strings should be kept the same
+        test_colors_plot = tls.FigureFactory.create_trisurf(
+            x, y, z, simplices, color_func=colors_str)
+        self.assertListEqual(list(test_colors_plot['data'][0]['facecolor']),
+                             list(colors_str))
+        # Colors must match length of simplices
+        colors_bad = colors_str[:-1]
+        self.assertRaises(ValueError, tls.FigureFactory.create_trisurf, x, y,
+                          z, simplices, color_func=colors_bad)
+        # Check converting custom colors to strings
+        test_colors_plot = tls.FigureFactory.create_trisurf(
+                    x, y, z, simplices, color_func=colors_raw)
+        self.assertTrue(isinstance(test_colors_plot['data'][0]['facecolor'][0],
+                                   str))
+
 
 class TestScatterPlotMatrix(NumpyTestUtilsMixin, TestCase):
 
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -1533,9 +1533,14 @@ class FigureFactory(object):
         # vertices of the surface triangles
         tri_vertices = points3D[simplices]
 
-        if not color_func:
+        if color_func is None:
             # mean values of z-coordinates of triangle vertices
             mean_dists = tri_vertices[:, :, 2].mean(-1)
+        elif isinstance(color_func, (list, np.ndarray)):
+            if len(color_func) != len(simplices):
+                raise ValueError('If color_func is a list/array, must'
+                                 ' be the same length as simplices')
+            mean_dists = color_func
         else:
             # apply user inputted function to calculate
             # custom coloring for triangle vertices
@@ -1548,13 +1553,15 @@ class FigureFactory(object):
                     dists.append(dist)
 
                 mean_dists.append(np.mean(dists))
-
-        min_mean_dists = np.min(mean_dists)
-        max_mean_dists = np.max(mean_dists)
-        facecolor = FigureFactory._map_z2color(mean_dists,
-                                               colormap,
-                                               min_mean_dists,
-                                               max_mean_dists)
+        if isinstance(mean_dists[0], str):
+            facecolor = mean_dists
+        else:
+            min_mean_dists = np.min(mean_dists)
+            max_mean_dists = np.max(mean_dists)
+            facecolor = FigureFactory._map_z2color(mean_dists,
+                                                   colormap,
+                                                   min_mean_dists,
+                                                   max_mean_dists)
 
         ii, jj, kk = zip(*simplices)
         triangles = graph_objs.Mesh3d(x=x, y=y, z=z, facecolor=facecolor,
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -1490,11 +1490,11 @@ class FigureFactory(object):
         return un_rgb_colors
 
     @staticmethod
-    def _map_z2color(zvals, colormap, vmin, vmax):
+    def _map_array2color(array, colormap, vmin, vmax):
         """
-        Returns the color corresponding zval's place between vmin and vmax
+        Normalize values in array by vmin/vmax and return plotly color strings.
 
-        This function takes a z value (zval) along with a colormap and a
+        This function takes an array of values along with a colormap and a
         minimum (vmin) and maximum (vmax) range of possible z values for the
         given parametrized surface. It returns an rgb color based on the
         relative position of zval between vmin and vmax
@@ -1507,7 +1507,7 @@ class FigureFactory(object):
                                          "of vmax.")
         # find distance t of zval from vmin to vmax where the distance
         # is normalized to be between 0 and 1
-        t = (zvals - vmin) / float((vmax - vmin))
+        t = (array - vmin) / float((vmax - vmin))
         t_colors = FigureFactory._find_intermediate_color(colormap[0],
                                                           colormap[1],
                                                           t)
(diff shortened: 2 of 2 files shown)


plotly/tools.py
```diff
@@ -1529,34 +1529,46 @@ def _trisurf(x, y, z, simplices, colormap=None, color_func=None,
...
-        if not color_func:
+        # Define colors for the triangle faces
+        if color_func is None:
             # mean values of z-coordinates of triangle vertices
             mean_dists = tri_vertices[:, :, 2].mean(-1)
+        elif isinstance(color_func, (list, np.ndarray)):
+            # Pre-computed list / array of values to map onto color
+            if len(color_func) != len(simplices):
+                raise ValueError('If color_func is a list/array, must'
+                                 ' be the same length as simplices')
```
theengineear: 🐐 (grammar goat), sorry we rely heavily on emojis for code reviews at Plotly.

```
'If color_func is a list/array, it must be the same length as simplices.'
```

plotly/tools.py
```diff
@@ -1529,34 +1529,46 @@ def _trisurf(x, y, z, simplices, colormap=None, color_func=None,
...
-        min_mean_dists = np.min(mean_dists)
-        max_mean_dists = np.max(mean_dists)
-        facecolor = FigureFactory._map_z2color(mean_dists,
-                                               colormap,
-                                               min_mean_dists,
-                                               max_mean_dists)
-
-        ii, jj, kk = zip(*simplices)
+        # Check if facecolors are already strings and can be skipped
+        if isinstance(mean_dists[0], str):
```
theengineear: Will this fail if folks are using python 2.7 with `unicode`? can we check against `basestring` here or use `six`?

plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -753,6 +753,26 @@ def test_trisurf_all_args(self):
...
+                             list(colors_str))
+        # Colors must match length of simplices
+        colors_bad = colors_str[:-1]
+        self.assertRaises(ValueError, tls.FigureFactory.create_trisurf, x, y,
+                          z, simplices, color_func=colors_bad)
+        # Check converting custom colors to strings
+        test_colors_plot = tls.FigureFactory.create_trisurf(
+                    x, y, z, simplices, color_func=colors_raw)
+        self.assertTrue(isinstance(test_colors_plot['data'][0]['facecolor'][0],
+                                   str))
```
theengineear: thanks for adding these 😺

theengineear:
Just a couple of comments @choldgraf, thanks for adding this! 👍 from me after you address those little nit-picks.

@Kully can you also give it a review and pass off if you're happy with it?

plotly/tools.py
choldgraf: haha, will do. I think I left off the `it` because I was trying to keep the line under PEP8 max, but ended up splitting it anyway. I had a PR last week where someone got me for including a dangling participle :D

plotly/tools.py
choldgraf: that's a good question....I'm open to another way of doing it, you might be right

choldgraf:
ok, I'll wait until another pair of eyes sees it and will then make changes. Feel free to suggest how to deal with the string checking situation.

For me, the array-forcing got the JSON packaging down to ~4 seconds per hemisphere (down from about 16 seconds).

btw, it looks like the plotly.js repo has been updated with lighting fixes today. Any idea when that'll make its way into the python plotly library?

jackparmer:
> btw, it looks like the plotly.js repo has been updated with lighting fixes today. Any idea when that'll make its way into the python plotly library?

Offline mode should download the latest version of plotly.js from the plotly.js CDN (am I remembering this right @etpinard ?)

etpinard:
Hmm @jackparmer I don't remember.

choldgraf:
I don't think so anymore. If you look at L34 of offline.py, it says that
downloading is deprecated in favor of shipping a mini plotly.js with
plotly.py

plotly/tools.py
Kully: > can we check against basestring here or use six

I don't know if this is relevant/has been answered, but I know for sure that basestring won't work between 2 and 3. str works fine though.

I'll have a look tomorrow though at the rest.

choldgraf: cool so then it's OK as is?

cldougl:
We're manually updating plotly.js for offline mode.
Here's some reasoning: https://github.com/plotly/plotly.py/pull/348#discussion_r44943281

@choldgraf we'll add a new version today with the updates

choldgraf:
thanks! :)

Kully:
#486

choldgraf:
OK, just tried the latest master and the plotting seems to work as planned. Now just waiting for the JS update!

Kully:
> OK, just tried the latest master and the plotting seems to work as planned. Now just waiting for the JS update!

There's one thing I noticed that is lacking, and that is support for hex and tuple colors in color_func (if a list).

Right now I am:
1. adding another example in the trisurf doc for color_func as a list
2. adding hex and tuple compat. in color_func

I finished 1. but I can't seem to solve `tuple` compatibility at the moment.

Kully:
@choldgraf I am going to make a new PR for what I am doing. But I will hold off on allowing for tuple compatibility. Need to prioritize updating the docs.

choldgraf:
I think it's worth having a `handle_colors` function that basically allows you to intelligently deal with all of these usecases (rgb strings, rgba strings, hex strings, tuples, an array of rgb 0-1 floats, an array of 0-255 floats, or a 1-D array w/ a colormap specified elsewhere). That'd save lines within the trisurf function and might be useful elsewhere.

Kully:
> I think it's worth having a handle_colors function that basically allows you to intelligently deal with all of these usecases (rgb strings, rgba strings, hex strings, tuples, an array of rgb 0-1 floats, an array of 0-255 floats, or a 1-D array w/ a colormap specified elsewhere). That'd save lines within the trisurf function and might be useful elsewhere.

That sounds like a good idea. The only thing is that not all functions in FigureFactory use 'rgb' as the canonical color type, to my knowledge. One of them uses array of 0-255 iirc

choldgraf:
Good point - do you want to handle a PR for a function like this? I am
happy to give thoughts, review, etc from my use cases.

choldgraf:
guys this looks awesome

https://www.dropbox.com/s/9ozy6rv92vjr8cg/moarbrainz.mov?dl=0

seriously

choldgraf:
@Kully @theengineear @jackparmer @cldougl @etpinard

Kully:
> guys this looks awesome
> 
> https://www.dropbox.com/s/9ozy6rv92vjr8cg/moarbrainz.mov?dl=0
> 
> seriously

Dayummmm!

monfera:
@cldougl here's an example for darkened deep grooves: intensity gradient depends on the euclidean distance from an ellipsoid that represents the central region of the brain. Of course an ellipsoid shape is simplistic, I'm sure there are established ways for fitting an anatomy-identifying model to a specific brain.
image
image

choldgraf:
@monfera that's beautiful. clever work!

I showed these images to one of the devs at pysurfer, and they're interested in exploring using plotly as the backend for their 3D visualizations. pysurfer is a package for visualizing brain surfaces and plotting activity on top of them. Check out their repo [here](https://github.com/nipy/PySurfer). It integrates pretty heavily with `freesurfer` which is sort of the gold standard for creating these surfaces in the first place.

They're using Mayavi right now which is pretty clunky, it'd be great if there were a faster / more stable / prettier plotting solution (e.g., these plots get pretty close). I'll probably open an issue there to discuss switching the backend to plotly and see what people think. Will keep you guys updated.

ntfrgl:
_This is a `Plotly` beginner's question. Please let me know if it fits better elsewhere._

Custom colours on a highly interactive `trisurf` rendering within the browser are exactly what I need, but for many applications including mine, a colour bar is essential for visual interpretation. Based on the documentation on [trisurf](https://plot.ly/python/trisurf/), [surface coloring](https://plot.ly/python/3d-surface-coloring/) and [meshes](https://plot.ly/python/3d-mesh/), I guessed that something like
``` python
fig.data[0].update(colorbar=go.ColorBar(...))
```
would do the job, but this was ineffectual. What would be the right way? I would have expected it to be a parameter for `create_trisurf()`.
<hr>
Also, I find a regular oscillation with high frequency and high amplitude in my RAM consumption when rendering a big `trisurf` inside an `ipython notebook` within Chrome. Not so for Firefox. Is this a general issue with the underlying `js` libraries?

monfera:
@ntfrgl this PR and associated PRs indeed deal with mesh coloring. On the plotly.js side of things it involved an update that solved lighting and reflection issues that in the past turned the mesh show up in a subpar color or no color. Since color representation, annotation and data flow haven't changed, i.e. the possible causes seem unrelated, it would be better to document new issues (of course they canlink to this or other PRs). 

When creating a new issue, it's helpful if a minimal example is developed that demonstrates the problem. If the problems were shown via the Python API, then it's useful to reproduce it with the JS API directly as well, because this can suggest the location of the issue. 

The utility of minimal problem reproductions also mean that it's useful to separate issues, e.g. one for the legend (with your example of what didn't work) and one for the memory fluctuation (maybe with a screenshot of Chrome vs Firefox). Easiest thing might be if there's a link to a demonstrating example.

A comment about memory fluctuations is that we started looking into some aspects of memory management, which have a speed impact as well, so any report like yours is very useful.

ntfrgl:
Opened #517, commented in #487.

PR #486, opened 2016-05-27
Choldgraf custom colors trisurf

Kully:
@choldgraf @theengineear 

Added the 'it' to that error message. Everything worked on my end and it looks good.

Plus, I made some modern art with `import random`:

modern art

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [1.10.0] - 2016-05-19
 ### Fixed
-Version 1.9.13 fixed an issue in offline mode where if you ran `init_notebook_mode`
+- Version 1.9.13 fixed an issue in offline mode where if you ran `init_notebook_mode`
 more than once the function would skip importing (because it saw that it had
 already imported the library) but then accidentally clear plotly.js from the DOM.
 This meant that if you ran `init_notebook_mode` more than once, your graphs would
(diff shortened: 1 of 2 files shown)


theengineear:
Great! 💃

choldgraf:
hah that's awesome. (so I shouldn't worry about my PR anymore, right?)

theengineear:
Yah @choldgraf to speed things up, we're just knabbing your commits and throwing them into a `plotly/plotly.py` repo branch to iterate a bit more collaboratively here as necessary.

Note that you *can* still add updates to your branch as necessary and we can either rebase or merge them into the new branch we've created to include them.

We're just avoiding having weird code state in our `master` branch as much as possible.

choldgraf:
no problem, if you guys are happy w/ this codebase then I say go for it. I'll push a little repo to do brain plotting, and we can see how the performance holds up. I got it down to ~4-6 seconds on the python side on my browser, and probably another 4-5 seconds for the browser to actually render the scene and plot. The biggest challenge I came across was that it looks like something isn't getting garbage collected on the JS side, because after a few plots the browser becomes really slow and uses up a ton of memory. But that's probably for a different PR.

cldougl:
@Kully (cc @choldgraf ) I'm updating our pip version now, could you update the [changelog](https://github.com/plotly/plotly.py/blob/master/CHANGELOG.md#change-log) when you have a minute (your changes will be applied to [v1.11.0](https://github.com/plotly/plotly.py/blob/master/CHANGELOG.md#1110---2016-05-27)) :)

PR #492, opened 2016-05-30
Updating cdn-options PR

yankev:
- removed the use of global variable __PLOTLY_USE_CDN.
- added keyword description to docstring

Note: For `requirejs.config`, the js will added to the end of the path. 
There were issues of plot persistence when I did include the .js. 

@chriddyp

### DIFF
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -30,7 +30,6 @@ except ImportError:
     _matplotlib_imported = False
 
 __PLOTLY_OFFLINE_INITIALIZED = False
-__PLOTLY_USE_CDN = False
 
 
 def download_plotlyjs(download_url):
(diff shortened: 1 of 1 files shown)


chriddyp:
> Note: For requirejs.config, the js will added to the end of the path. 
There were issues of plot persistence when I did include the .js.

Got it. Can you add a comment to the code to explain this?

plotly/offline/offline.py
```diff
@@ -53,14 +52,17 @@ def init_notebook_mode(connected=False):
     Initialize plotly.js in the browser if it hasn't been loaded into the DOM
     yet. This is an idempotent method and can and should be called from any
     offline methods that require plotly.js to be loaded into the notebook dom.
+
+    Keyword arguments:
+
+    connected (default=False) -- if connected is True, this means that the
+    plotly.js library will be loaded from a CDN(online) rather than the
+    local file from pip.
```
chriddyp: Although this explains what it is doing, it doesn't really explain why. What if instead:

```
Arguments:

connected (default=False) -- If True, plotly.js library will be loaded from an online CDN. If False, plotly.js is loaded locally from this python package. 

Use connected = True if you want your notebooks to have a smaller file size. plotly.js is a multi-megabyte file and when connected = False, the entire library is included inline inside the notebook. When connected = True, the library is downloaded from the web when you view the notebook. Since the library will be downloaded from the web, you and your viewers must be connected to the internet to be able to view this notebook's charts.

Use connected = False if you want you and your collaborators to be able to view and create charts regardless of your internet connection. This is the default option since it is the most predictable. Since the library will be included inline inside your notebook, this makes your notebooks considerably larger than if you were using connected = True. 
```

chriddyp:
💃

yankev:
Cool, merging.

PR #493, opened 2016-05-30
Allow more than 2 colormaps again

Kully:
@choldgraf @theengineear 

Guys,

So Chris, I think when you made the efficiency improvements to trisurf, you got rid of the capability of mapping divergent or colormaps of more than 2 colors. I have put that feature back now while trying to maintain the efficiency as best I could.

I put the 'rgb(%s, %s, %s)' stuff in the FigureFactory function _label_rgb(), which I was using anyways to convert a tuple to an rgb tuple.

Also, facecolor is not done in one line anymore, but is an empty list that keeps appending.

Let me know if you have any improvements to add.

(Also rewrote some doc strings for clarity)

### DIFF
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -1518,33 +1518,32 @@ class FigureFactory(object):
                                          "and vmax. The vmin value cannot be "
                                          "bigger than or equal to the value "
                                          "of vmax.")
-        # find the normalized distance t of a triangle face between vmin and
-        #vmax where the distance is normalized between 0 and 1
-        t = (face - vmin) / float((vmax - vmin))
 
-        if len(colormap) <= 1:
-            t_color = colormap[0]
-            color = FigureFactory._convert_to_RGB_255(t_color)
-            color = FigureFactory._label_rgb(color)
+        if len(colormap) == 1:
+            # color each triangle face with the same color in colormap
+            face_color = colormap[0]
+            face_color = FigureFactory._convert_to_RGB_255(face_color)
+            face_color = FigureFactory._label_rgb(face_color)
         else:
-            # account for colormaps with more than one color
-            incr = 1./(len(colormap) - 1)
-            low_color_index = int(t/incr)
+            # find the normalized distance t of a triangle face between vmin
+            # and vmax where the distance is normalized between 0 and 1
+            t = (face - vmin) / float((vmax - vmin))
+            low_color_index = int(t / (1./(len(colormap) - 1)))
 
             if t == 1:
-                t_color = colormap[low_color_index]
-                color = FigureFactory._convert_to_RGB_255(t_color)
-                color = FigureFactory._label_rgb(color)
+                face_color = colormap[low_color_index]
+                face_color = FigureFactory._convert_to_RGB_255(face_color)
+                face_color = FigureFactory._label_rgb(face_color)
 
             else:
-                t_color = FigureFactory._find_intermediate_color(
+                face_color = FigureFactory._find_intermediate_color(
                     colormap[low_color_index],
                     colormap[low_color_index + 1],
-                    (t - low_color_index * incr) / incr)
-                color = FigureFactory._convert_to_RGB_255(t_color)
-                color = FigureFactory._label_rgb(color)
+                    t * (len(colormap) - 1) - low_color_index)
+                face_color = FigureFactory._convert_to_RGB_255(face_color)
+                face_color = FigureFactory._label_rgb(face_color)
 
-        return color
+        return face_color
 
     @staticmethod
     def _trisurf(x, y, z, simplices, colormap=None, color_func=None,
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -1468,9 +1468,9 @@ class FigureFactory(object):
         diff_1 = float(highcolor[1] - lowcolor[1])
         diff_2 = float(highcolor[2] - lowcolor[2])
 
-        inter_colors = np.array([lowcolor[0] + intermed * diff_0,
-                                 lowcolor[1] + intermed * diff_1,
-                                 lowcolor[2] + intermed * diff_2])
+        inter_colors = (lowcolor[0] + intermed * diff_0,
+                        lowcolor[1] + intermed * diff_1,
+                        lowcolor[2] + intermed * diff_2)
         return inter_colors
 
     @staticmethod
@@ -1503,14 +1503,14 @@ class FigureFactory(object):
             return un_rgb_colors
 
     @staticmethod
-    def _map_array2color(array, colormap, vmin, vmax):
+    def _map_face2color(face, colormap, vmin, vmax):
         """
-        Normalize values in array by vmin/vmax and return plotly color strings.
+        Normalize facecolor values by vmin/vmax and return rgb-color strings
 
-        This function takes an array of values along with a colormap and a
-        minimum (vmin) and maximum (vmax) range of possible z values for the
-        given parametrized surface. It returns an rgb color based on the
-        relative position of zval between vmin and vmax
+        This function takes a tuple color with elements between 0 and 1, along
+        with a colormap and a minimum (vmin) and maximum (vmax) range of
+        possible mean distances for the given parametrized surface. It returns
+        an rgb color based on the mean distance between vmin and vmax
 
         """
         if vmin >= vmax:
@@ -1518,20 +1518,38 @@ class FigureFactory(object):
                                          "and vmax. The vmin value cannot be "
                                          "bigger than or equal to the value "
                                          "of vmax.")
-        # find distance t of zval from vmin to vmax where the distance
-        # is normalized to be between 0 and 1
-        t = (array - vmin) / float((vmax - vmin))
-        t_colors = FigureFactory._find_intermediate_color(colormap[0],
-                                                          colormap[1],
-                                                          t)
-        t_colors = t_colors * 255.
-        labelled_colors = ['rgb(%s, %s, %s)' % (i, j, k)
-                           for i, j, k in t_colors.T]
-        return labelled_colors
+        # find the normalized distance t of a triangle face between vmin and
+        #vmax where the distance is normalized between 0 and 1
+        t = (face - vmin) / float((vmax - vmin))
+
+        if len(colormap) <= 1:
+            t_color = colormap[0]
+            color = FigureFactory._convert_to_RGB_255(t_color)
+            color = FigureFactory._label_rgb(color)
+        else:
+            # account for colormaps with more than one color
+            incr = 1./(len(colormap) - 1)
+            low_color_index = int(t/incr)
+
+            if t == 1:
+                t_color = colormap[low_color_index]
+                color = FigureFactory._convert_to_RGB_255(t_color)
+                color = FigureFactory._label_rgb(color)
+
+            else:
+                t_color = FigureFactory._find_intermediate_color(
+                    colormap[low_color_index],
+                    colormap[low_color_index + 1],
+                    (t - low_color_index * incr) / incr)
+                color = FigureFactory._convert_to_RGB_255(t_color)
+                color = FigureFactory._label_rgb(color)
+
+        return color
 
     @staticmethod
     def _trisurf(x, y, z, simplices, colormap=None, color_func=None,
-                 plot_edges=False, x_edge=None, y_edge=None, z_edge=None):
+                 plot_edges=False, x_edge=None, y_edge=None, z_edge=None,
+                 facecolor=None):
         """
         Refer to FigureFactory.create_trisurf() for docstring
         """
@@ -1556,11 +1574,11 @@ class FigureFactory(object):
             if len(color_func) != len(simplices):
                 raise ValueError("If color_func is a list/array, it must "
                                  "be the same length as simplices.")
-                # convert all colors to rgb
-                for index in range(len(color_func)):
-                    if '#' in color_func[index]:
-                        foo = FigureFactory._hex_to_rgb(color_func[index])
-                        color_func[index] = FigureFactory._label_rgb(foo)
+            # convert all colors to rgb
+            for index in range(len(color_func)):
+                if '#' in color_func[index]:
+                    foo = FigureFactory._hex_to_rgb(color_func[index])
+                    color_func[index] = FigureFactory._label_rgb(foo)
 
             mean_dists = np.asarray(color_func)
         else:
@@ -1581,10 +1599,16 @@ class FigureFactory(object):
         else:
             min_mean_dists = np.min(mean_dists)
             max_mean_dists = np.max(mean_dists)
-            facecolor = FigureFactory._map_array2color(mean_dists,
-                                                       colormap,
-                                                       min_mean_dists,
-                                                       max_mean_dists)
+
+            if facecolor is None:
+                facecolor = []
+            for index in range(len(mean_dists)):
+                color = FigureFactory._map_face2color(mean_dists[index],
+                                                      colormap,
+                                                      min_mean_dists,
+                                                      max_mean_dists)
+                facecolor.append(color)
+
         # Make sure we have arrays to speed up plotting
         facecolor = np.asarray(facecolor)
         ii, jj, kk = simplices.T
@@ -2931,11 +2955,13 @@ class FigureFactory(object):
 
         """
         if isinstance(colors, tuple):
-            return 'rgb{}'.format(colors)
+            return ('rgb(%s, %s, %s)' % (colors[0], colors[1], colors[2]))
         else:
             colors_label = []
             for color in colors:
-                color_label = 'rgb{}'.format(color)
+                color_label = ('rgb(%s, %s, %s)' % (color[0],
+                                                    color[1],
+                                                    color[2]))
                 colors_label.append(color_label)
 
             return colors_label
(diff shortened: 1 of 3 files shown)


choldgraf:
hey - you might be right, I'll take a look through the changes now and add
thoughts

plotly/tools.py
```diff
@@ -1468,9 +1468,9 @@ def _find_intermediate_color(lowcolor, highcolor, intermed):
         diff_1 = float(highcolor[1] - lowcolor[1])
         diff_2 = float(highcolor[2] - lowcolor[2])
 
-        inter_colors = np.array([lowcolor[0] + intermed * diff_0,
```
choldgraf: What is the benefit gained by making it a list and looping through? IME using arrays instead of lists almost always makes things much faster.

plotly/tools.py
```diff
@@ -1503,35 +1503,53 @@ def _unconvert_from_RGB_255(colors):
             return un_rgb_colors
 
     @staticmethod
-    def _map_array2color(array, colormap, vmin, vmax):
```
choldgraf: Is support for multiple colormaps mentioned in the docstring for trisurf anywhere? (or is it supported in a different function that calls this function?) I didn't see any mention of multiple colormaps. What's the usecase for including multiple colormaps in the call to trisurf, e.g., how would people specify which colormap goes with which datapoint?

plotly/tools.py
Kully: > What is the benefit gained by making it a list and looping through? IME using arrays instead of lists almost always makes things much faster.

I don't necessarily think it's faster or anything like that. I actually just switched it back to this because I wanted it to work with my FF converting code

plotly/tools.py
```diff
@@ -1503,35 +1503,53 @@ def _unconvert_from_RGB_255(colors):
...
+            incr = 1./(len(colormap) - 1)
+            low_color_index = int(t/incr)
+
+            if t == 1:
+                t_color = colormap[low_color_index]
+                color = FigureFactory._convert_to_RGB_255(t_color)
+                color = FigureFactory._label_rgb(color)
+
+            else:
+                t_color = FigureFactory._find_intermediate_color(
```
choldgraf: ah, so by "multiple colormaps" do you just mean "will allow 2 colormaps and assume these are the min/max ends of a spectrum, interpolating colors in between"?

Kully: > ah, so by "multiple colormaps" do you just mean "will allow 2 colormaps and assume these are the min/max ends of a spectrum, interpolating colors in between"?

I mean more than 2 colors and yes, will interpolate between all the colors linearly, first deciding which two colors it is 'in between'. I should make all this more clear in the doc, you're right. I'll do that now after the test hopefully passes

plotly/tools.py
choldgraf: I'm wondering if there's a benefit other than speed. E.g., if you thought the user might pass in tuples of variable shapes then it's probably better for the input to be a list or tuple or something (since arrays require consistency across dimensions). If you like, maybe I can take a pass through and make it work w/ arrays similar to the last PR. I don't think the syntax would be any different, and it'll likely be an order of magnitude faster to use arrays.

plotly/tools.py
```diff
@@ -1503,35 +1503,53 @@ def _unconvert_from_RGB_255(colors):
...
+    def _map_face2color(face, colormap, vmin, vmax):
         """
-        Normalize values in array by vmin/vmax and return plotly color strings.
+        Normalize facecolor values by vmin/vmax and return rgb-color strings
 
-        This function takes an array of values along with a colormap and a
-        minimum (vmin) and maximum (vmax) range of possible z values for the
-        given parametrized surface. It returns an rgb color based on the
-        relative position of zval between vmin and vmax
+        This function takes a tuple color with elements between 0 and 1, along
```
choldgraf: distances don't need to be 0 < color < 1, right? Since you're standardizing them by vmin and vmax

plotly/tools.py
choldgraf: Ah I see - I guess I'm having a hard time following the function. I don't see any `for` loops here, so how will it iterate through all of the colors if you pass a colormap with, e.g., 4 values?

Kully: > ah, so by "multiple colormaps" do you just mean "will allow 2 colormaps and assume these are the min/max ends of a spectrum, interpolating colors in between"?

I thought I responded to this message.

Oh well, I said that I want a user to be able to enter a sequence of n colors and this can create a divergent type of colormap, where each facecolor is placed between its closest colors in colormap based on its relative position between vmin and vmax. So you'll get a flow from one color to the next in the plot. I already updated the string docs locally and will update once the tests pass.

choldgraf: Also, just my intuition here, I think it's better to have this handled by the colormap itself, rather than internally in a function like this. E.g., a matplotlib colormap will accept any value from 0 to 1, and will linearly interpolate according to the colormap's parameters. Is there some kind of functionality like this in plotly? (this PR is implementing something like this, but IMO it should be a more general function that is used elsewhere in the codebase and not just for trisurf)

choldgraf: I think I understand the functionality you want here, I just wasn't understanding how you were choosing which colormap within the list to use. Looking through the code then I think I get it what you're doing - it'll basically choose the index to use in the colormap list based on where from 0 to 1 `t` is. That's a clever way of doing things. I think it could be done pretty easily w/ an array without changing the functionality or syntax...should I make a PR making those changes after this one?

Kully: > Ah I see - I guess I'm having a hard time following the function. I don't see any for loops here, so how will it iterate through all of the colors if you pass a colormap with, e.g., 4 values?

The looping is happening here:

```Python
for index in range(len(mean_dists)):
    color = FigureFactory._map_face2color(mean_dists[index],
                                          colormap,
                                          min_mean_dists,
                                          max_mean_dists)
    facecolor.append(color)
```

I am looping the calling of _mapface2color and appending to the function

plotly/tools.py
Kully: > distances don't need to be 0 < color < 1, right? Since you're standardizing them by vmin and vmax

Right again but they will all be converted to colors with elements between 0 and 1 anyways. I updated the string doc again. =)

plotly/tools.py
choldgraf: I meant the looping to tell the function which two indices of `colormap` to use, but I now see that it's not designed to be looped in the way I thought it was :)

Kully: > should I make a PR making those changes after this one?

That would be fantastic, yes please!

choldgraf: Mk, I'll wait until this is pushed and then I can iterate off of that. Just a thought - this kind of functionality might be useful in the `colorlover` package. It'd be useful if there was a `ColorMapper` class that would do the kinds of interpolation that you're talking about here. That might be a more appropriate place to handle interpolating within or between an arbitrary list of colors. So you could initialize it with a list of colors (like we have here) and an arbitrary number of bins (e.g., `mymap = cl.ColorMapper(color_list, n_bins=1000)`) and then you could pass a number between 0 and 1 to return the color for that bin (e.g.,  `mymap(.2)` and it'd return the index according to int(n_bins * .2)).

Kully: > Mk, I'll wait until this is pushed and then I can iterate off of that. Just a thought - this kind of functionality might be useful in the colorlover package. It'd be useful if there was a ColorMapper class that would do the kinds of interpolation that you're talking about here. That might be a more appropriate place to handle interpolating within or between an arbitrary list of colors. So you could initialize it with a list of colors (like we have here) and an arbitrary number of bins (e.g., mymap = cl.ColorMapper(color_list, n_bins=1000)) and then you could pass a number between 0 and 1 to return the color for that bin (e.g., mymap(.2) and it'd return the color indexed according to int(n_bins * .2)).

This sounds like a good idea. Much more structured.

@theengineear  Can you have a quick look at this PR? It looks finished to me and I'd like to push it to master.

theengineear: Sure, looking now.

plotly/tools.py
theengineear: @Kully and @choldgraf did you guys resolve this one?

plotly/tools.py
```diff
@@ -1503,35 +1503,53 @@ def _unconvert_from_RGB_255(colors):
...
-        t = (array - vmin) / float((vmax - vmin))
-        t_colors = FigureFactory._find_intermediate_color(colormap[0],
-                                                          colormap[1],
-                                                          t)
-        t_colors = t_colors * 255.
-        labelled_colors = ['rgb(%s, %s, %s)' % (i, j, k)
-                           for i, j, k in t_colors.T]
-        return labelled_colors
+        # find the normalized distance t of a triangle face between vmin and
+        #vmax where the distance is normalized between 0 and 1
```
theengineear: :dolls: missing a ` ` in this comment.

plotly/tools.py
```diff
@@ -1503,35 +1503,53 @@ def _unconvert_from_RGB_255(colors):
...
+        t = (face - vmin) / float((vmax - vmin))
+
+        if len(colormap) <= 1:
+            t_color = colormap[0]
```
theengineear: If `len(colormap) == 0` here, you'll get an `IndexError`. Does this condition need to be updated to `len(colormap) == 1`? with another case for `len(colormap) == 0`?

plotly/tools.py
```diff
@@ -1503,35 +1503,53 @@ def _unconvert_from_RGB_255(colors):
...
+            color = FigureFactory._convert_to_RGB_255(t_color)
+            color = FigureFactory._label_rgb(color)
+        else:
+            # account for colormaps with more than one color
+            incr = 1./(len(colormap) - 1)
```
theengineear: 🐄 `incr`?

plotly/tools.py
choldgraf: I think I see now what @Kully was getting at, though it took a fair bit of digging for me to figure it out. I wanted to ask - in terms of interpolating between points of a list of multiple colormaps, why not use `colorlover`? You could just do `cl.interp(list_of_colors, 10000)[int(t * 10000)]` right?

plotly/tools.py
Kully: > Sure, looking now.

Thanks. I'll fix this all up.

plotly/tools.py
```diff
@@ -1503,35 +1503,53 @@ def _unconvert_from_RGB_255(colors):
...
-        t_colors = FigureFactory._find_intermediate_color(colormap[0],
-                                                          colormap[1],
-                                                          t)
-        t_colors = t_colors * 255.
-        labelled_colors = ['rgb(%s, %s, %s)' % (i, j, k)
-                           for i, j, k in t_colors.T]
-        return labelled_colors
+        # find the normalized distance t of a triangle face between vmin and
+        #vmax where the distance is normalized between 0 and 1
+        t = (face - vmin) / float((vmax - vmin))
```
theengineear: 🐄 Can you move this check *below* the `if len(colormap) <= 1`? You're trying to say "if there is no colormap or the colormap only has one value, just return the same color, always", right? It might make it easier to understand what's going on in this function.

plotly/tools.py
```diff
@@ -1503,35 +1503,53 @@ def _unconvert_from_RGB_255(colors):
...
+            t_color = colormap[0]
+            color = FigureFactory._convert_to_RGB_255(t_color)
+            color = FigureFactory._label_rgb(color)
+        else:
+            # account for colormaps with more than one color
+            incr = 1./(len(colormap) - 1)
+            low_color_index = int(t/incr)
+
+            if t == 1:
+                t_color = colormap[low_color_index]
```
theengineear: Similar to https://github.com/plotly/plotly.py/pull/493/files#r65412272. Why not just do `colormap[-1]` here? It's much easier to understand. You're just doing `colormap[1 / (1 / (len(colormap) - 1))]`. 

image

theengineear:
Sorry to drag this one out longer on you with this review @Kully, ping me directly when you want another review and I can hop on it quickly.

plotly/tools.py
choldgraf: so, `colormap[np.exp(np.pi * np.sqrt(-1))]`? ;)

plotly/tools.py
choldgraf: If you guys are curious, I threw together a quick proof of concept here:

https://github.com/choldgraf/ecogtools/blob/master/ecogtools/colors.py

Right now it depends on `matplotlib`, `seaborn`, and `colorlover`, because seaborn does interpolation between colors more effectively than `colorlover` does (and because I think it's more natural to represent colors as `matplotlib` maps than lists of rgb strings).

choldgraf: Actually, it's in its own repository now, plus here's an example: 

https://github.com/choldgraf/colorbabel/blob/master/examples/demo.ipynb

plotly/tools.py
Kully: > Why not just do colormap[-1]

That gets me the last element in the list. How is that helpful here?

choldgraf: here's where I thought it might be easy to just use `interp`. I think it'd be a clearer way of doing the same thing. so `t_color = cl.interp(colormap, 255)[int(t * 255)]`.

Kully: > here's where I thought it might be easy to just use interp. I think it'd be a clearer way of doing the same thing. so t_color = cl.interp(colormap, 255)[int(t * 255)].

We're still doing that in another PR right?

I just got Andrew's easier way of computing t_colors working.

choldgraf: ah sure - just lemme know when this is merged. Might be a little bit of time though, because I'm heading out of town for a few days tomorrow

Kully: > ah sure - just lemme know when this is merged. Might be a little bit of time though, because I'm heading out of town for a few days tomorrow

Okay no problem.

Kully:
@theengineear @choldgraf

plotly/tools.py
```diff
@@ -1518,33 +1518,32 @@ def _map_face2color(face, colormap, vmin, vmax):
...
                     colormap[low_color_index + 1],
-                    (t - low_color_index * incr) / incr)
-                color = FigureFactory._convert_to_RGB_255(t_color)
-                color = FigureFactory._label_rgb(color)
+                    t * (len(colormap) - 1) - low_color_index)
+                face_color = FigureFactory._convert_to_RGB_255(face_color)
+                face_color = FigureFactory._label_rgb(face_color)
 
-        return color
+        return face_color
```
theengineear: 💥 so much more readable! thanks!

plotly/tools.py
```diff
@@ -1518,33 +1518,32 @@ def _map_face2color(face, colormap, vmin, vmax):
...
+            face_color = FigureFactory._convert_to_RGB_255(face_color)
+            face_color = FigureFactory._label_rgb(face_color)
         else:
-            # account for colormaps with more than one color
-            incr = 1./(len(colormap) - 1)
-            low_color_index = int(t/incr)
+            # find the normalized distance t of a triangle face between vmin
+            # and vmax where the distance is normalized between 0 and 1
+            t = (face - vmin) / float((vmax - vmin))
+            low_color_index = int(t / (1./(len(colormap) - 1)))
```
theengineear: 🐄 Sorry if I'm just missing something, but isn't:

`1 / ( 1 / val ) = val`

And isn't `iterable[len(iterable) - 1] = iterable[-1]`

So

`iterable[1 / ( 1 / (len(iterable) - 1)] = iterable[-1]`

I feel like `iterable[-1]` is more understandable than `iterable[1 / ( 1 / (len(iterable) - 1)]`

This means that you don't need to calculate `low_color_index = int(t / (1./(len(colormap) - 1)))` until you're inside the `else` statement, which also improves readability here.

theengineear:
@Kully I'm happy with it now :) 💃 from me.

@choldgraf unless there is something jarringly wrong about this still, can we merge it? That way, you could open up a new PR to improve computation speed again (with these new test constraints).

choldgraf:
I say go for it!

Kully:
> @Kully I'm happy with it now :) 💃 from me.

I don't think we can do that. Or maybe I'm not reading this right but...
`int(t / (1./(len(colormap) - 1)))` isn't the same as  `int(1 / (1./(len(colormap) - 1)))`

I think the `t` does make the difference.

Kully:
@theengineear

plotly/tools.py
```diff
@@ -1525,17 +1525,17 @@ def _map_face2color(face, colormap, vmin, vmax):
...
-            # find the normalized distance t of a triangle face between vmin
-            # and vmax where the distance is normalized between 0 and 1
-            t = (face - vmin) / float((vmax - vmin))
-            low_color_index = int(t / (1./(len(colormap) - 1)))
-
-            if t == 1:
-                face_color = colormap[low_color_index]
+            if face == vmax:
+                # pick last color in colormap
+                face_color = colormap[-1]
```
theengineear: 🎉 🔔 🎆

plotly/tools.py
```diff
@@ -1525,17 +1525,17 @@ def _map_face2color(face, colormap, vmin, vmax):
...
             face_color = FigureFactory._label_rgb(face_color)
         else:
-            # find the normalized distance t of a triangle face between vmin
-            # and vmax where the distance is normalized between 0 and 1
-            t = (face - vmin) / float((vmax - vmin))
-            low_color_index = int(t / (1./(len(colormap) - 1)))
-
-            if t == 1:
-                face_color = colormap[low_color_index]
+            if face == vmax:
```
theengineear: This is *even clearer* than what I was suggesting. thanks :)

theengineear:
💃 💃 👯 !

PR #494, opened 2016-06-02
added option to download an image of the most recent plot

yankev:
- This change is for the offline notebook mode
- This is the first attempt at it, there may be better solutions that I plan to look at
- Need to add docstring


yankev:
Made the inline download method a part of the `iplot` call. 
You can initiate it by setting the parameter ` download_image` to `True`, and providing details about the format, height, width and filename. 
Issue right now is that the plots will download each time the notebook is restarted using this method. 

The previous method allowed you to download the plot that was most recently plotted and this wouldn't run when the notebook was rerun.

yankev:
Demo of the functionality:

- Inline Methods (offline.plot)
plot_download
- Inline Methods (offline.iplot)
iplot_download

- This method takes downloads an image of the most recent plot displayed in your ipython notebook
(offline.download_notebook_image)
notebook_image_download

yankev:
You can also find some notebooks to test the functionality here: 
https://github.com/yankev/testing/tree/master/offline_download_image

Note: This will only wok in Firefox, the fix for Chrome will come when [this fix](https://github.com/plotly/plotly.js/pull/604) get's released.

jackparmer:
Hey @yankev - 
Nice! This looks cool!
I like the explicitness of the first option. Its a single step to get an image for a plot. And I don't have to remember the name of a new function. 
I'd also suggest:
- Change `download_image` kwarg to `image`
- Don't necessarily have to include a `filetype` (defaults to PNG) or `filename` (defaults to `plot_image.xyz`)
- Once this works, add a `Save images offline` section here https://plot.ly/python/static-image-export/ and comb back through the StackOverflow and community.plot.ly Q's about this

jackparmer:
We may want to print a message also in Jupyter notebook:

> For higher resolution images and more export options, try Plotly's online image export server.

I'm worried we're going to get a lot of complaints that the client-side image export is not sufficiently fully featured, because folks don't realize we have also have a full-blown server just for image export.

yankev:
@cldougl @theengineear @tarzzz When you guys get the chance, could take a look at this? Thanks!

theengineear:
^^ yah, I'll try to get to it tonight.

plotly/offline/offline.py
```diff
@@ -210,12 +211,22 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
                                has become outdated with your version of
                                graph_reference.json or if you need to include
                                extra, unnecessary keys in your figure.
+    image (default=False) -- If True, the image of the current plot
+        will be downloaded.
+    format (default='png') -- Specifies the format of the image to be
+        downloaded if `downlowad_image` is True.
```
theengineear: 🐄 Can we just do `image=None|'png'|...`? That feels just an understandable to me, but it only requires a single kwarg. Also, it makes it such that `format` doesn't do *nothing* if `image=False`.

plotly/offline/offline.py
```diff
@@ -236,13 +247,39 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+                  '}};'
+                  'if(document.readyState == \'complete\') {{'
+                  'downloadimage(\'{format}\', {height}, {width}, '
+                  '\'{filename}\');}}'
+                  '</script>'
+                  ).format(format=format, width=_width, height=_height,
+                           filename=filename, plot_id=plotdivid)
+        # allow time for the plot to draw
+        import time
+        time.sleep(1)
```
theengineear: Will this work for plots that take > 1 second to download?

plotly/offline/offline.py
```diff
@@ -236,13 +247,39 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
 
     display(HTML(plot_html))
 
+    if image:
+        warnings.warn('For higher resolution images and more export '
+                      'options, consider making requests to the '
+                      'Plotly image servers. Type help(\'py.image\') '
+                      'for more details.', UserWarning)
```
theengineear: 🐄 I don't think you should use a warning here. It's a good idea in theory, but in notebooks, it often makes it look like we have broken code or something.

plotly/offline/offline.py
```diff
@@ -325,13 +373,38 @@ def plot(figure_or_data,
...
+                          'Plotly.downloadImage(p, {{format: format, height: '
+                          'height, width: width, filename: filename}});'
+                          '}};'
+                          'if(confirm(\'Do you want to save this image as '
+                          '{filename}.{format}?\')) {{'
+                          'downloadimage(\'{format}\', {height}, {width}, '
+                          '\'{filename}\');}}'
+                          '</script>'
+                          ).format(format=format, width=_width, height=_height,
+                                   filename=image_filename, plot_id=plotdivid)
```
theengineear: Is this script the same as above? Can we factor this out to be a little more DRY 🐫.

plotly/offline/offline.py
theengineear: Can we display it some other way?

plotly/offline/offline.py
```diff
@@ -236,13 +247,39 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
 
 def plot(figure_or_data,
          show_link=True, link_text='Export to plot.ly',
          validate=True, output_type='file',
          include_plotlyjs=True,
-         filename='temp-plot.html',
-         auto_open=True):
+         filename='temp-plot.html', auto_open=True,
+         image=False, image_filename='plot_image', format='png',
+         _width=800, _height=600):
```
theengineear: Why the `_`? if it's supposed to be private, we should not offer it as a kwarg here, right?

plotly/offline/offline.py
```diff
@@ -325,13 +373,38 @@ def plot(figure_or_data,
...
+
+                # write the download script:
+                script = ('<script>'
+                          'function downloadimage(format, height, width,'
+                          ' filename) {{'
+                          'var p = document.getElementById(\'{plot_id}\');'
+                          'Plotly.downloadImage(p, {{format: format, height: '
+                          'height, width: width, filename: filename}});'
+                          '}};'
+                          'if(confirm(\'Do you want to save this image as '
```
theengineear: Sorry, I think I'm missing something. Why do we only `confirm` for the `plot` (not `iplot`) case?

theengineear:
@yankev I can't think of a better way to do this with the offline restriction. This is basically just taking a snapshot of the graph, which is really the only tool you'd have offline, right?

Not quite ready to pass off on it though, ping me when you've considered some of the above comments.

plotly/offline/offline.py
yankev: Nice, I like this idea.

plotly/offline/offline.py
yankev: For the most part it's the same. Although for the iplot version I added a check to see if the document was completely loaded. This was to ensure that images don't get downloaded again on page reloads. I could definitely put this into a function though, with two options.

plotly/offline/offline.py
yankev: Ah sorry, I'm just abusing notation. I needed a different variable name for height and width because `height` and `width` are returned from the call to `_plot_html()` later on, and would overwrite the values set as parameters in the original plot call.

plotly/offline/offline.py
yankev: Right, this is just a bit inconsistent on my part. Do you think the confirm is a good feature to have?

plotly/offline/offline.py
yankev: Right so this is a bit of a concern. The most complicated plot I tried was one with 300 boxplots, I also dropped `time.sleep` to `0.1` and there was no issue. I'm not really sure of a way to guarantee this success though.

plotly/offline/offline.py
yankev: I had the same reservation. We could use the `confirm` to provide a long winded message about this.

plotly/offline/offline.py
theengineear: hmmm, actually, it prevents a *whole* notebook from executing right? this might not be the right way to go. you could potentially end up clicking through confirm after confirm.

plotly/offline/offline.py
theengineear: See my flip floppiness in https://github.com/plotly/plotly.py/pull/494/files/f5baa3e6060e9f36676c93797063ea3eb9a4ce91#r66386919

thoughts?

plotly/offline/offline.py
yankev: I have a condition `if(document.readyState == 'complete')` so that the download prompts don't occur on page reloads. As long as that holds up, I think it should avoid the issue you're talking about.

yankev:
@theengineear 
Changes:
I moved the warnings into the `confirm prompt`. Also created a function to return the script strings. Also using the single image parameter vs both image and format. 

Addressing some concerns:
I think the one second sleep time and now the confirmation box should allow for enough time for the plot to render on screen, and thus `downloadImage` shouldn't return a blank image. I've tested on some fairly "complicated" plots, and it seems to work. Though there is no guarantee on this one.

I have the condition to check whether the page is loaded before injecting the download script, so on page reloads/re-opening the notebook, the confirm and download prompts shouldn't show up.

theengineear:
>I have the condition to check whether the page is loaded before injecting the download script, so on page reloads/re-opening the notebook, the confirm and download prompts shouldn't show up.

Sweet!

Taking a peek right now.

plotly/offline/offline.py
```diff
@@ -46,6 +48,37 @@ def get_plotlyjs():
...
     return plotlyjs
 
+def image_download_script(type):
+
+    if type == 'iplot':
+        check_start = 'if(document.readyState == \'complete\') {{'
+        check_end = '}}'
+    elif type == 'plot':
+        check_start = ''
+        check_end = ''
```
theengineear: 🐄 

```
else:
    raise SomeException( .. )
```

?

plotly/offline/offline.py
```diff
@@ -46,6 +48,37 @@ def get_plotlyjs():
     plotlyjs = resource_string('plotly', path).decode('utf-8')
     return plotlyjs
 
+def image_download_script(type):
```
theengineear: 🎎 can you add a docstring for this that explains the `type` parameter?

plotly/offline/offline.py
```diff
@@ -210,12 +243,23 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+    image_height (default=600) -- Specifies the height of the image in `px`.
+    image_width (default=800) -- Specifies the width of the image in `px`.
 
     Example:
     ```
     from plotly.offline import init_notebook_mode, iplot
     init_notebook_mode()
     iplot([{'x': [1, 2, 3], 'y': [5, 2, 7]}])
+    # We can also download an image of the plot by setting the image to the
+    format you want ie: `image='png'`
```
theengineear: :goat: this is an example, not an explanation so `ie:` --> `e.g.,` Also, `i.e.,` not `ie:`. ;)

plotly/offline/offline.py
```diff
@@ -236,13 +280,32 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
 
     display(HTML(plot_html))
 
+    if image:
+        if image not in __IMAGE_FORMATS:
+            raise ValueError('The image parameter takes only the '
+                             'following format types: `png`, `jpeg`, '
+                             '`webp`, `svg`')
+        # if the check passes then download script injection will commence.
+        # Write script to download image of the plot
```
theengineear: :dolls: This comment could be cleaner. Also, you might not really need the second line here, it's clear from the code.

plotly/offline/offline.py
theengineear: 🐄 `get_image_download_script`?

plotly/offline/offline.py
```diff
@@ -236,13 +280,32 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+        script = image_download_script('iplot').format(format=image,
+                                                       width=image_width,
+                                                       height=image_height,
+                                                       filename=filename,
+                                                       plot_id=plotdivid)
+        # allow time for the plot to draw
+        import time
```
theengineear: 🐄 This import should be at the top of the module. Let me know if you need a better explanation for why this is the case.

plotly/offline/offline.py
```diff
@@ -236,13 +280,32 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+        display(HTML(script))
+
 
 def plot(figure_or_data,
          show_link=True, link_text='Export to plot.ly',
          validate=True, output_type='file',
          include_plotlyjs=True,
-         filename='temp-plot.html',
-         auto_open=True):
+         filename='temp-plot.html', auto_open=True,
```
theengineear: This won't flake if you don't supply a filename *twice*, right? Will we end up with `temp-plot.html (1)` or something? Or, does the OS just handle it?

theengineear: Ah, nvm, looks like this was here before your changes. 👍

plotly/offline/offline.py
```diff
@@ -325,13 +400,30 @@ def plot(figure_or_data,
             else:
                 plotly_js_script = ''
 
+            if image:
+                if image not in __IMAGE_FORMATS:
+                    raise ValueError('The image parameter takes only the '
+                                     'following format types: `png`, `jpeg`, '
+                                     '`webp`, `svg`')
```
theengineear: 🐫! Use the actual values of `__IMAGE_FORMATS` here.

plotly/offline/offline.py
```diff
@@ -325,13 +400,30 @@ def plot(figure_or_data,
...
+                # if the check passes then download script is injected.
+                # write the download script:
```
theengineear: See https://github.com/plotly/plotly.py/pull/494/files#r66824586

theengineear:
Great! Let's do it! 💃 after you've considered my comments.

plotly/offline/offline.py
yankev: Hmm, this could potentially get called a lot, so it might be a time saver to load it at the top. Let me know if that's why, or if you have additional reasons.

yankev:
@theengineear k, should be good. I put in a neat check to find the the name of the caller of `get_image_download_script`.

plotly/offline/offline.py
theengineear: Yah, exactly it's faster to put it at the top, it's also better Python style. Here's a quick note on it:

plotly/offline/offline.py
```diff
@@ -48,14 +49,26 @@ def get_plotlyjs():
     plotlyjs = resource_string('plotly', path).decode('utf-8')
     return plotlyjs
 
-def image_download_script(type):
+def get_image_download_script(caller):
+    '''
```
theengineear: 🐄 We typically use `"""`, not `'''` for docstrings.

plotly/offline/offline.py
```diff
@@ -48,14 +49,26 @@ def get_plotlyjs():
...
+    This function will return a script that will download an image of a Plotly
+    plot.
 
-    if type == 'iplot':
+    Keyword Arguments:
+    caller ('plot', 'iplot') -- specifies which function made the call for the
+        download script. If `iplot`, then an extra condition is added into the
+        download script to ensure that download prompts aren't iniitated on
+        page reloads.
```
theengineear: Nice 🎎!

plotly/offline/offline.py
```diff
@@ -48,14 +49,26 @@ def get_plotlyjs():
...
+def get_image_download_script(caller):
+    '''
+    This function will return a script that will download an image of a Plotly
+    plot.
 
-    if type == 'iplot':
+    Keyword Arguments:
+    caller ('plot', 'iplot') -- specifies which function made the call for the
+        download script. If `iplot`, then an extra condition is added into the
+        download script to ensure that download prompts aren't iniitated on
```
theengineear: :goat: `iniitated` --> `initiated`.

plotly/offline/offline.py
```diff
@@ -282,18 +295,20 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+            img_n = len(__IMAGE_FORMATS)
             raise ValueError('The image parameter takes only the '
-                             'following format types: `png`, `jpeg`, '
-                             '`webp`, `svg`')
-        # if the check passes then download script injection will commence.
-        # Write script to download image of the plot
+                             'following format types: `{}`, `{}`, '
+                             '`{}`, `{}`'.format(*[__IMAGE_FORMATS[i]
+                                                 for i in range(img_n)]
+                                                 )
```
theengineear: ;__; this will just break if you don't have enough `{}`, right?

Can you:

```
'This image parameter must be one of {}.'.format(__IMAGE_FORMATS)
```

plotly/offline/offline.py
```diff
@@ -49,17 +50,13 @@ def get_plotlyjs():
...
-    Keyword Arguments:
-    caller ('plot', 'iplot') -- specifies which function made the call for the
-        download script. If `iplot`, then an extra condition is added into the
-        download script to ensure that download prompts aren't iniitated on
-        page reloads.
+    This function will return a script that will download an image of the
+    most recently rendered Plotly plot
     '''
+    # get the name of the function that called `get_image_download_script`
+    caller = sys._getframe(1).f_code.co_name
```
theengineear: 😸 I totally see how this is nifty/sneaky, but the other approach was really straightforward/simple. I'm definitely not immune to trying to do things in a *clever* way, but these things can come back to bite you. For example, it's unclear to me whether this would work correctly under the following conditions:

1. we want to mock this function in a test
2. we refactor later to use a `@decorator`

Also, if you look at the [docs for `_getframe`](https://docs.python.org/2/library/sys.html#sys._getframe), they say:

>This function should be used for internal and specialized purposes only. It is not guaranteed to exist in all implementations of Python.

Also, developers will typically use the `_*` naming convention to suggest that the function is not part of the public api!

theengineear:
@yankev sorry to be a party 💩-er, but I think you should stick with the more straight-forward implementation for inferring who called the function.

If it makes you feel better, you could use `no_reload_confirm=True` or something instead of `iplot` vs `plot`. I think your `caller='iplot'` implementation was just fine though 😉

plotly/offline/offline.py
yankev: Hmm I see. It's cool I don't mind changing it back.

plotly/offline/offline.py
yankev: ahh I see. I think I could put the `{}` in a list comprehension to fix this. But I think I'll just implement your suggestion as it appears to be the cleaner option.

theengineear: >Simple is better than complex.

[Zen of Python](https://www.python.org/dev/peps/pep-0020/)

yankev:
@theengineear k seems to be in working shape. Let me know if there's anything else.

theengineear:
Beep, boop, 💃. Nice work!

yankev:
@theengineear thanks man! This was a super messy one, so many avoidable errors. Thanks for reviewing it!

PR #499, opened 2016-06-08
Trisurf and color optimizations

choldgraf:
This is a quick extension of #493. There was a lot of talk about the best way to interpolate between multiple colors in a readable, quick way. This PR mostly does 2 things:

1. Replaces the custom interpolation code with a matplotlib `LinearSegmentedColormap`. This is a way of creating a custom cmap from a list of rgb colors. Then you can pass it any value between 0 and 1 and it'll return the RGB value. Should have the exact same functionality, is vectorized, and means that some extra code can be removed.
2. Re-introduces vectorization to the facecolor creation. New speed is down to ~1.8 seconds from ~3.6 seconds. Not a gigantic increase but is helpful if you're re-plotting trisurfs many many times.

Only drawback I can see is that it introduces another dependency on matplotlib, but it seems like this is already depended on elsewhere, and besides it's likely that the majority of python plotly users will have this installed. LMKWYT.

### DIFF
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -3022,47 +3022,60 @@ class FigureFactory(object):
         return un_rgb_color
 
     @staticmethod
-    def _map_face2color(face, colormap, vmin, vmax):
+    def _map_faces2color(faces, colormap):
         """
-        Normalize facecolor values by vmin/vmax and return rgb-color strings
+        Normalize facecolors by their min/max and return rgb-color strings.
 
-        This function takes a tuple color along with a colormap and a minimum
-        (vmin) and maximum (vmax) range of possible mean distances for the
-        given parametrized surface. It returns an rgb color based on the mean
-        distance between vmin and vmax
-
+        This function takes a tuple color along with a colormap.
+        It returns an rgb color based on the mean distance between the
+        minimum and maximum value of faces.
         """
-        if vmin >= vmax:
-            raise exceptions.PlotlyError("Incorrect relation between vmin "
-                                         "and vmax. The vmin value cannot be "
-                                         "bigger than or equal to the value "
-                                         "of vmax.")
-
-        if len(colormap) == 1:
+        colormap = np.atleast_2d(colormap)
+        if colormap.shape[0] == 1:
             # color each triangle face with the same color in colormap
-            face_color = colormap[0]
-            face_color = FigureFactory._convert_to_RGB_255(face_color)
-            face_color = FigureFactory._label_rgb(face_color)
+            face_colors = colormap
         else:
-            if face == vmax:
-                # pick last color in colormap
-                face_color = colormap[-1]
-                face_color = FigureFactory._convert_to_RGB_255(face_color)
-                face_color = FigureFactory._label_rgb(face_color)
-            else:
-                # find the normalized distance t of a triangle face between
-                # vmin and vmax where the distance is between 0 and 1
-                t = (face - vmin) / float((vmax - vmin))
-                low_color_index = int(t / (1./(len(colormap) - 1)))
+            # Convert face values to between 0 and 1
+            vmin = faces.min()
+            vmax = faces.max()
+            if vmin >= vmax:
+                raise exceptions.PlotlyError("Incorrect relation between vmin"
+                                             " and vmax. The vmin value cannot"
+                                             " be bigger than or equal to the"
+                                             " value of vmax.")
+            # Scale t to between 0 and 1
+            t = (faces - vmin) / float((vmax - vmin))
+            t_ixs = np.round(t * 255).astype(int)
 
-                face_color = FigureFactory._find_intermediate_color(
-                    colormap[low_color_index],
-                    colormap[low_color_index + 1],
-                    t * (len(colormap) - 1) - low_color_index)
-                face_color = FigureFactory._convert_to_RGB_255(face_color)
-                face_color = FigureFactory._label_rgb(face_color)
+            # If a list of colors is given, interpolate between them.
+            color_range = FigureFactory._blend_colors(colormap)
+            face_colors = color_range[t_ixs]
 
-        return face_color
+        # Convert to 255 scale, and round to nearest integer
+        face_colors = np.round(face_colors * 255., 0)
+        face_colors = FigureFactory._label_rgb(face_colors)
+        return face_colors
+
+    @staticmethod
+    def _blend_colors(colormap, n_colors=255.):
+        if len(colormap) == 1:
+            raise ValueError('Cannot blend a colormap with only one color')
+        # Figure out how many splits we need
+        n_split = np.floor(n_colors / (len(colormap) - 1)).astype(int)
+        n_remain = np.mod(n_colors, len(colormap))
+
+        # Iterate through pairs of colors
+        color_range = []
+        for ii in range(len(colormap) - 1):
+            # For each channel (r, g, b)
+            this_interp = []
+            for cstt, cstp in zip(colormap[ii], colormap[ii + 1]):
+                # If it's not an even split, add req'd amount on first iter
+                n_interp = n_split + n_remain if ii == 0 else n_split
+                this_interp.append(np.linspace(cstt, cstp, n_interp))
+            color_range.append(np.vstack(this_interp).T)
+        color_range = np.vstack(color_range)
+        return color_range
 
     @staticmethod
     def _trisurf(x, y, z, simplices, colormap=None, color_func=None,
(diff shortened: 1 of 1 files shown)


choldgraf:
@Kully @theengineear I'd be curious to hear your thoughts!

Kully:
> @Kully @theengineear I'd be curious to hear your thoughts!

Well it's awesome that it saves even more time once again. I'm trying it out now.

I think your drawback is mine as well. I feel like it would be ideal for matplotlib to not be dependant in this way. I'm wondering if we can do something like check `if matplotlib_imported` or something like that. Then, if it is installed by the user, their plotting experience would be faster.

In addition, we could provide a helpful Plotly message upon plotting that could recommend installing matplotlib to speed up the computation time, etc.

choldgraf:
I think that's a fair point. Do you guys have an idea for how many users don't also use matplotlib? My guess is that it'd be a really tiny fraction, but I'm probably not privy to the kinds of userbases that plotly covers.

I think that both usecases could be supported, it'd just require a lot of extra lines of code to duplicate the same functionality. I'm generally a fan of changes that reduce the number of lines of code, but understand if you all don't want a dependency on matplotlib.

If you guys do prefer custom code over matplotlib, I think it's worth cleaning up the color interpolation so that it A. accepts arrays instead of looping over lists, and B. is a little bit more readable (I'm still not sure what all the `1 / (1 / ...)` stuff is doing. LMK and I can try to implement this, but in that case I'll probably just drop the matplotlib dependency.

Kully:
> I'm still not sure what all the 1 / (1 / ...) stuff is doing

Andrew and I (mostly Andrew) made this part of the code more clear.

It looks like `t = (face - vmin) / float((vmax - vmin))`
                   `low_color_index = int(t / (1./(len(colormap) - 1)))`
now.

choldgraf:
yeah, I was taking a look at that before deciding to just use matplotlib colormaps. That's still confusing me a bit. `t` is a value between 0 and 1, right? So in that case why wouldn't it just be `low_color_index = np.floor(t * len(colormap) - 1)`? Maybe I'm missing something.

Kully:
> yeah, I was taking a look at that before deciding to just use matplotlib colormaps. That's still confusing me a bit. t is a value between 0 and 1, right? So in that case why wouldn't it just be low_color_index = np.floor(t * len(colormap) - 1)? Maybe I'm missing something.

Wow, the glorious floor function! I didn't think about that!

Although I don't think it would work for some small neighborhood around t=0, right?

choldgraf:
I think the bigger edge case problem would be around t=1, because if t ==
1, then doing floor would return the `Nth` index, instead of the `N-1th`
index. We'd have to think about how to deal w/ that case, but I think it's
doable...the challenge is how to vectorize the lookup between color pairs

Kully:
> I think the bigger edge case problem would be around t=1, because if t ==
> 1, then doing floor would return the `Nth` index, instead of the `N-1th`
> index. We'd have to think about how to deal w/ that case, but I think it's
> doable...the challenge is how to vectorize the lookup between color pairs

With my code, I just had a simple check for t==1: `if face==vmax'. I'm just checking if our triangle in question is at the maximum height. It forces the last element in the list to be picked.

choldgraf:
ya, I think that's a good way to handle it, the question is how to vectorize the process. Maybe it'd be possible to use `np.linspace` to do this quickly. If you guys don't want matplotlib in there, I can try to throw something together (though it would be re-implementing the wheel a bit, since matplotlib already does this quite nicely)

choldgraf:
Hey folks - I updated this so that it doesn't depend on matplotlib. Basically just wrote our own color blending function (maybe it'll be useful elsewhere?) What do you all think? (sorry for the slow update on this...was off backpacking in yellowstone!)

theengineear:
@choldgraf looks useful to me! @Kully do you think we could integrate this? Might be nice to have a `colors.py` file at this point and move all our color-related functionality (and constants) out of `tools.py`.

choldgraf:
Let me know if you want me to try moving color stuff into its own module...that might be a bigger PR than this one though.

Kully:
> Let me know if you want me to try moving color stuff into its own module...that might be a bigger PR than this one though.

Do you think it's necessary right now? I mean, it's forseeable that we'll eventually need it as we keep adding to the color capabilities. For now it's pretty manageable in `tools.py` imo

choldgraf:
IMO I don't think that it's a necessary change. Though I bet there are points in the codebase where things are being duplicated because it wasn't clear that the functionality is already there. E.g. that `_map_face2color` code is basically just a way to linearly interpolation between a list of colors given a 1d array. Surely the trisurf code isn't the only place this happens in plotly.py, no?

Kully:
> Surely the trisurf code isn't the only place this happens in plotly.py, no?

That's a good point. But that would be _such_ a big project then, to replace all the linear interpolation code; Make a color class and methods in that...

choldgraf:
yeah I agree that it'd be a bigger project. FWIW I wrote a class that does
something like this. Basically is just a way to quickly move between lots
of different color representations because I was tired of looking up the
syntax for the various things that are out there. Lemme know if you think
it'd be useful: https://github.com/choldgraf/colorbabel

Kully:
> yeah I agree that it'd be a bigger project. FWIW I wrote a class that does
> something like this. Basically is just a way to quickly move between lots
> of different color representations because I was tired of looking up the
> syntax for the various things that are out there. Lemme know if you think
> it'd be useful: https://github.com/choldgraf/colorbabel

This looks really good Chris. I'd be interested in pulling this repo and playing around with it to see its capabilities. Let me check it out today and I can get back to you. But it looks really clean! 😄

choldgraf:
Sounds good - it's nothing super fancy...basically just turning a lot of different color representations into a LinearSegmentedColormap and then letting you slice/dice/bin it quickly. Lemme know if it's useful!

Anything else for me to do on this PR by the way?

Kully:
> Sounds good - it's nothing super fancy...basically just turning a lot of different color representations into a LinearSegmentedColormap and then letting you slice/dice/bin it quickly. Lemme know if it's useful!
> 
> Anything else for me to do on this PR by the way?

I think that's fine. To restate what's going on, is your plan to add this color file to this PR and nothing else? I suppose rewriting the FF code to use your module is another PR

choldgraf:
Well, my guess is that plotly would want a pared down version of that class
because as it is now, it'd add a bunch of extra dependencies you probably
don't want in plotly. But something like it could be pretty easy to make.

I think a much bigger project would be going through the code and figuring
out what could be replaced etc. So maybe it makes more sense to keep all of
it for a different PR. WDYT?

Kully:
> WDYT?

I agree. I didn't know your thing has dependancies. `matplotlib` is one of them. RIght now, I'm leaning towards not having `matplotlib` as a dependency for tools.py (vis a vis my comment above), but if we were to optimize these FigureFactories, we, as you are doing, write a module that doesn't have any noncore dependancies for the color stuff.

choldgraf:
Sounds good - with the `blend_colors` function that's in this PR, it would
be very easy to just add support for either RGB arrays, or RGB lists of
strings.

Kully:
@chriddyp @theengineear  Hey again Chris.

I've been working on rewriting a lot of `tools.py` for a few reasons. 1) I am not using the correct Plotly_Scales 2) I want to give users the options to apply their own colorscales (not necessarily linearly interpolated ones) and a few other things.

I am going to try writing a `colors.py` file like Andrew suggested before, but I want to try to do it without matplotlib. Not completely deciding not to use your color module but just letting you know that I am going to give it a try and then hopefully I can start rewriting `tools.py` so that nothing breaks in the process.

I'll let you know how this goes when I get somewhere.

choldgraf:
No worries - it's been a while on this PR, but I believe the latest commit removes all the matplotlib dependencies. It just has its own interpolating functions under the hood. We could get this pulled and then you can iterate off of that if you like, LMK what you'd prefer

Kully:
Oh, I didn't realize/remember that. I _would_ prefer to get my hands wet with my own `colors.py` just so I can figure out exactly what I need out of the colors module.

choldgraf:
I'm pretty sure this PR just adds an extra `_blend_colors` method:

https://github.com/plotly/plotly.py/pull/499/files#diff-2431a3f5483919319eff181a09618fc7R3060

would you prefer to just take a look at that code w/o finishing this PR?

Kully:
You mean without merging this PR? Sure, that makes sense, I can do that.

choldgraf:
I'm happy to clean up this PR and get it merged, I think it's pretty much ready to go, just whatever you'd prefer to do :)

choldgraf:
Is there anything in this PR that seems like it should be changed? maybe it's actually ready to go right now?

Kully:
So I was just looking and was curious if any tests would be failing. So I did a lot of changes to the way the colors are parsed, but looking at your changes it all seems contained within the _trisurf backend functions. So yeah, I guess we should merge

Kully:
Ok, let's do it!

Kully:
I have to run now, but feel free to do it. Thanks for all the help!

choldgraf:
I don't think I have privileges to do any merging, so it'll have to be a plotly person

Kully:
@theengineear Andrew, can I get the green light on doing this?

theengineear:
Ya, merge away guys! I was offline for a few days, sorry for the delay! 💃

cldougl:
@Kully probably better to make a pr so we can check out why the `test_trisurf_all_args` test is failing

PR #501, opened 2016-06-09
Reduce plot size

fu:
Hi,

I added a kwarg _outsource_plotly_, that allows the plotly.js to be separated from the actual plot. The code just adds a separate file containing the plotly.min.js into the plot folder. While it is not so convenient for a single plot (two files vs one file), it is very good if one creates many plots in one folder, since all plots share the same js.library and thus the overall plotting size is reduced significantly.

Cheers

.c

### DIFF
--- a/plotly/offline/__init__.py
+++ b/plotly/offline/__init__.py
@@ -10,5 +10,6 @@ from . offline import (
     iplot,
     iplot_mpl,
     plot,
-    plot_mpl
+    plot_mpl,
+    serve_plotlyjs_from_directory
 )
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -42,10 +42,12 @@ def download_plotlyjs(download_url):
     pass
 
 
-def get_plotlyjs():
-    # path = os.path.join('offline', 'plotly.min.js')
-    # plotlyjs = resource_string('plotly', path).decode('utf-8')
-    plotlyjs = '</script><script src="plotly.min.js">'
+def get_plotlyjs( outsource_plotly = False ):
+    if outsource_plotly:
+        plotlyjs = '</script><script src="plotly.min.js">'
+    else:
+        path = os.path.join('offline', 'plotly.min.js')
+        plotlyjs = resource_string('plotly', path).decode('utf-8')
     return plotlyjs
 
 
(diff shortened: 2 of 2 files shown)


theengineear:
@chriddyp , did you ever take a peek at this one? I think @fu makes a good point here. Right now we're just copying in the entire script where we could just be linking it as suggested. Is there a use-case that I'm not considering?

This is more complex (as was suggested, we need to make sure the `src` in the script points to the proper place), but I'd be willing to consider if it provides a large reduction in file size.

One alternate route would be to allow users to use a flag like:

`plotly.offline.outsource_plotly()`

That can set a module-level flag to change the behavior of `get_plotlyjs()`. Then, the default behavior is not changed for now.

chriddyp:
> `plotly.offline.outsource_plotly()`

Yeah, that makes sense. Perhaps something like `plotly.offline.serve_plotlyjs_from_directory()`?

theengineear:
^^ yah, definitely something more specific than `outsource_plotly`, I was just using the name from the line changes.

The other thing I don't want to do is have a *million* hard-to-discover functions. We could alternatively keep this in config?

```
{
    "offline_serve_plotlyjs_from_directory": True
}
```

We can name-space `offline` config with `offline_*`. Then we'd allow users to keep using:

`session.sign_in`

... which should really have an alias like:

```
def update_session(**kwargs):
    # blah blah
```

I like the idea of having a central place where all the current options can be sorted out. Really I'm ok with either option though since we can just deprecate it later...

@fu, can you change this so that the user has to *manually* call a function or set some config to get this to happen?

fu:
Hi Guys,

thanks for picking this one up! I would happily go back and adapt my pull request to have it in the config or use a function call. Personally, I would go for the config solution, would that be ok?

Cheers

.c

theengineear:
>Personally, I would go for the config solution, would that be ok?

Yup that's fine by me! Let me know if you have any questions. Thanks @fu!

fu:
Hi again,

second thought, setting the config file seems not straight forward in the current implementation, or maybe I could not find the obvious way :) 

Form my understanding, I would have to adjust set_config_file to accept my "offline_serve_plotlyjs_from_directory", and for all following parts this would need to be adjusted as well. Adding just a **kwargs might be a way but sort of defeats all your checking in set_config_file.

Therefore, I opted for the second option: adding serve_plotlyjs_from_directory() that sets a global variable.

Pull request on its way

Cheers and thank you for the great plotly lib - keep up the good work!

.c

theengineear:
>second thought, setting the config file seems not straight forward in the current implementation

Yah, it's definitely not straightforward because there are three levels of settings that you can set:

0. file-level
1. session-level
2. function-call-level

The function call is fine for now, can always just change the function implementation later such that it updates the `session` config down the road.

Thanks for the update :)

PR #510, opened 2016-06-21
introduce new check to see if page is loaded

yankev:
@theengineear Unfortunately the check I was using before the prevent pop ups on page reloads doesn't work for Chrome. Previously I was only able to test it on the Firefox because there was a plotly.js bug that  caused an issue with Chrome. I've tried the current method in both browsers, and it seems to be working.

### DIFF
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -62,7 +62,9 @@ def get_image_download_script(caller):
     """
 
     if caller == 'iplot':
-        check_start = 'if(window.is_loaded==true){{'
+        # note: 5500/5.5s is a handpicked delay. The script will not run again
+        # once this time has elapsed after the timestamp is instantiated.
+        check_start = 'var d=(new Date()).getTime();if(!(d>{utime}+5500)){{'
         check_end = '}}'
     elif caller == 'plot':
         check_start = ''
@@ -129,8 +131,6 @@ def init_notebook_mode(connected=False):
         script_inject = (
             ''
             '<script>'
-            'window.is_loaded = false;'
-            'setTimeout(function(){{window.is_loaded=true}},100);'
             'requirejs.config({'
             'paths: { '
             # Note we omit the extension .js because require will include it.
@@ -147,8 +147,6 @@ def init_notebook_mode(connected=False):
         script_inject = (
             ''
             '<script type=\'text/javascript\'>'
-            'window.is_loaded = false;'
-            'setTimeout(function(){{window.is_loaded=true}},100);'
             'if(!window.Plotly){{'
             'define(\'plotly\', function(require, exports, module) {{'
             '{script}'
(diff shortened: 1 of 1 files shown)


plotly/offline/offline.py
```diff
@@ -129,6 +129,8 @@ def init_notebook_mode(connected=False):
         script_inject = (
             ''
             '<script>'
+            'window.is_loaded = false;'
+            'setTimeout(function(){{window.is_loaded=true}},100);'
```
theengineear: I'd rather we *didn't* inject something into the global namespace to do this (`is_loaded`). Also, JS variables are typically in `camelCase` --> `isLoaded`.

If we *do* need to do this, can it be something like `plotlyIsLoaded`?

yankev: Yeah I could change that if we decide to stick with it. Another possible option could be to encode the time in the script. This way, we could put a condition such that the script will not run if some amount of time `c` has passed. Therefore on page refreshes the embedded js shouldn't run until the next time the cell is ran.

theengineear: ^^ That sounds better than adding a property onto `window`. Want to give that a shot?

yankev:
@theengineear checked on both Chrome and Firefox.

plotly/offline/offline.py
```diff
@@ -62,7 +62,9 @@ def get_image_download_script(caller):
     """
 
     if caller == 'iplot':
-        check_start = 'if(document.readyState == \'complete\') {{'
+        # note: 5500/5.5s is a handpicked delay. The script will not run again
+        # once this time has elapsed after the timestamp is instantiated.
+        check_start = 'var d=(new Date()).getTime();if(!(d>{utime}+5500)){{'
```
theengineear: Man, sorry I keep dragging you through this, but `var d = ...` is the same as `window.d = ...`. If anything I think it's just a little more confusing. Can we do:

`check_start = 'if(!(new Date()).getTime()>{utime}+5500){{'`

yankev: Oh I see, yep I can make that change.

theengineear:
Great! 💃 after you confirm that it works on Chrome and Firefox still.

yankev:
@theengineear ahh bummer.. I might have to add more time for the delay. Really heavy graphs are causing some delays around 10s in Firefox, which results in no download prompt. I wonder if I should bump the delay to something like 20 seconds. With that choice, I'm assuming that users don't refresh/re-open their notebooks that frequently.

theengineear:
@yankev is there something in plotlyjs that we can use to do an async wait? I recall that there is a `Promise` in there that we should have access to. All you'd need to wait for would be for (1) that `Promise` to show up and (2) then to just `.then()` the Promise.

This delay solution is getting a little long and it *still* doesn't guarantee success.

//cc @etpinard and/or @mdtusz for a refresher on this one?

mdtusz:
Correct - plots return a promise.

```js
Plotly.plot(/*plot thingz*/)
    .then(function(gd){ return Plotly.downloadImage(gd, { format: 'png' }); })
    .then(function(){ console.log('Image should be downloaded now!'); });
```

yankev:
@theengineear @mdtusz Yup, I can make that change. Definitely seems like the best approach.

yankev:
Moved this to a new PR: #513

theengineear:
Looks like this got moved --> *re*moving.

PR #513, opened 2016-06-29
Offline image download Redo

yankev:
In order to use the `.then` method with the promises, I had to move the code into `_plot_html`. However even with this method, the issue with the delay persists (if we use the timestamp, which I believe to be the best method that we know of at the moment). Thus I've made a parameter `image_delay` available to the user, such that if their plot doesn't render in the amount of time provided (10secs), they can make the change themselves to guarantee that their image is saved.

Lastly, I tried to add a test that will see if the image gets downloaded.
It searches for the existence of the file, starting from some base/parent directory. I've gotten it to work locally on `OS X`, however I'm not sure what the file structure is like on the test machine, so the test seems to fail. I set the base dir to `/home/ubuntu/`. Let me know if you have any ideas on this, or an alternative test I could possibly write. 

Thanks! @theengineear

### DIFF
--- a/plotly/graph_reference/default-schema.json
+++ b/plotly/graph_reference/default-schema.json
@@ -253,16 +253,36 @@
                             "valType": "number"
                         },
                         "ax": {
-                            "description": "Sets the x component of the arrow tail about the arrow head. A positive (negative) component corresponds to an arrow pointing from right to left (left to right)",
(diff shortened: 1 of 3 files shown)


theengineear:
Hrm. This one got stuck, it seems. I'm going to close this down, but feel free to reopen if someone has cycles to dig in and rejuvenate.

PR #521, opened 2016-07-21
changelog typo

cldougl:


Kully:
@cldougl I think you need to just change the title in Layout from `Love` to `2D Density Plot`

Should I push this change from the master or let you do it in this PR?

PR #531, opened 2016-08-01
Trisurf colorbar

Kully:


Kully:
@theengineear

plotly/tools.py
```diff
@@ -1477,6 +1477,45 @@ class FigureFactory(object):
...
+        on the colors in sequential order. If 'scale' is left empty, a linear-
+        interpolated colorscale will be generated. If 'scale' is a specificed
+        list, it must be the same legnth as colors and must contain all floats
+        For documentation regarding to the form of the output, see
+        https://plot.ly/python/reference/#mesh3d-colorscale
+        """
+        colorscale = []
+
+        if not scale:
+            for j in range(len(colors)):
```
theengineear: 🐄 `for i, color in enumerate(colors):`

No big deal, it just gets around the `colors[i]` below.

plotly/tools.py
```diff
@@ -1477,6 +1477,45 @@ class FigureFactory(object):
...
+                colorscale.append([j * 1./(len(colors) - 1), colors[j]])
+            return colorscale
+
+        else:
+            colorscale = [[scale[j], colors[j]] for j in range(len(colors))]
```
theengineear: 🐄 you could use `zip` here. It's *almost* the same as what you're doing. No big deal, just a 🐄-tip.

`colorscale = [list(tup) for tup in zip(scale, colors)]`

theengineear: I only say that because the `scale[j]` when `j` is related to `colors` here reads a bit funny to me.

plotly/tools.py
```diff
@@ -1477,6 +1477,45 @@ class FigureFactory(object):
...
+            return colorscale
+
+    @staticmethod
+    def _convert_colorscale_to_rgb(colorscale):
+        """
+        Converts the colors in a colorscale to rgb colors
+        """
+        for index in range(len(colorscale)):
```
theengineear: 🐄 ditto on using `enumerate` here.

plotly/tools.py
```diff
@@ -1477,6 +1477,45 @@ class FigureFactory(object):
...
+            colorscale[index][1] = FigureFactory._convert_to_RGB_255(
```
theengineear: 🎎, the unfamiliar reader won't know that `colorscale` is an array of arrays, each with a `color` as the first item and another numeric value as the second.

plotly/tools.py
```diff
@@ -3226,9 +3265,9 @@ def _map_face2color(face, colormap, vmin, vmax):
         return face_color
 
     @staticmethod
-    def _trisurf(x, y, z, simplices, colormap=None, color_func=None,
-                 plot_edges=False, x_edge=None, y_edge=None, z_edge=None,
-                 facecolor=None):
+    def _trisurf(x, y, z, simplices, show_colorbar, colormap=None,
```
theengineear: Should we allow `show_colorbar` to be defaulted to `True` or `False`? Seems reasonable to me.

theengineear: nvm, i see that the user-facing function does have it defaulted 👍

theengineear:
💃 👯 💃 after you've considered my comments.

PR #538, opened 2016-08-09
update pjs (1.16.2) for offline

cldougl:

PR #545, opened 2016-08-17
Make temporary fix to plotly.min.js so that it loads in offline module

etpinard:
As first discovered in this forum [post](http://community.plot.ly/t/offline-mode-in-jupyter-notebook-shows-blank-graphs/1860/), `offline` is broken ever since https://github.com/plotly/plotly.py/pull/538 where plotly.js got bumped to from the `1.14.x` series to the `1.16.2`.

More precisely, `init_notebook_mode()` (with the default `connected=False`) fails to load plotly.js as a require.js module. From my observations, the CDN link (with `connected=True`) appears to load fine.

Something is up with a `mapbox-gl` dependency called `webworkify` in a minified bundle. See https://github.com/substack/webworkify/issues/29 for more info. Ideally, this issue should be fixed upstream.

But for now, I propose a uber hacky fix where I manually fix the minified bundle. 

For those interested in non-nonsensical JS drama, I had to replace 

image

by 

image

to get `offline` working again.


cldougl:
tested! 💃

cldougl:
@etpinard do you mind bumping the version # here too so I can update the package on pip w/ this fix

armanw96:
hey mine still won't work and i don't undertand how you fix it so easily please help @ etpinard

Issue #487, opened 2016-05-28
memory leak after plotting?


choldgraf:
I've noticed that after leaving interactive 3d plots open for a long time, the browser starts to run really, really slowly.

This might be on the JS side, but is there a function in python for clearing the cache / javascript memory / etc? Usually my browser will crash after I make about 10 trisurf plots, and I have to restart it as well as my python kernel.

Anybody else experience this?

choldgraf:
note - this is plotting in offline mode, with pretty large 3d surfaces (60,000 triangles or so). It works fine for the first few plots, however.

jackparmer:
FYI `plotly.js` WebGL folks @mikolalysenko @etpinard @monfera ^^^

@choldgraf We've seen slow memory leaks from WebGL plots in the past, though not in a while. Thanks for reporting!

choldgraf:
hey folks - was anybody able to replicate this problem?

monfera:
@choldgraf Hello Chris, have you a minimal example, e.g. specific `type` of plot? Also, how long does this issue generally take to manifest, and is it the specific browser tab that gets slow (slow to rotate/pan I guess?) or the entire browser, i.e. all its tabs? Which browser are you using, recent Chrome? Lastly, what hardware is it sitting on, especially, is it integrated graphics or dedicated video, and possibly, the size of the GPU memory?

choldgraf:
Sorry @monfera , I should have been more explicit in my original post! I'm getting this problem when making trisurf plots, generally which is calling Mesh3d under the hood, I believe. I'm plotting a large amount of triangles, ~60K or so. I can try to put together some simple code to re-create the problem, though it's not going to be super clear-cut since there's not an error or anything.

monfera:
@choldgraf thank you for the information! I've been wrapping up a functional addition to the WebGL based plotting. As soon as it's done, fixing memory leaks is planned to be the next item.

choldgraf:
Sounds great, I appreciate the help. I'd open a PR if I was efficient at all with javascript, sorry to ask you to do it.

In a sort-of related question, do you know how optimized the webgl plotting scripts are? I'm trying to think if there are any low-hanging fruit for speeding up some of the heavier 3D visualizations.

monfera:
[...]

PR #550, opened 2016-08-19
Debug the trisurf code

choldgraf:
see #487

### DIFF
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -3269,7 +3269,7 @@ class FigureFactory(object):
         return face_color
 
     @staticmethod
-    def _trisurf(x, y, z, simplices, show_colorbar, colormap=None,
+    def _trisurf(x, y, z, simplices, show_colorbar=True, colormap=None,
                  color_func=None, plot_edges=False, x_edge=None, y_edge=None,
                  z_edge=None, facecolor=None):
         """


monfera:
Just to have your [comment](https://github.com/plotly/plotly.py/issues/487#issuecomment-241105653) here as well:

```
looks like a change in the trisurf code broke my brain plotting code. Out of curiosity why was show_colorbar added as a parameter, and not a keyword? Seems like it'd make more sense to just default to False rather than forcing it in the call to _trisurf.
```

choldgraf:
Actually hold up a sec on this one, looking through the code I think I see a few more bugs...I'll comment on them in this PR

choldgraf:
It looks like a bunch of bugs got introduced whenever somebody added the colorbar functionality. E.g., right now if you don't give a colormap, then it throws an error. Moreover, it looks like the output of the `_make_colorscale` function is not always compatible with the `_convert_colorscale_to_rgb` function...

choldgraf:
Also I'm not sure why but the current version re-converted `facecolor` to a list (around L3337)...IMO we should never use lists unless absolutely necessary. A numpy array is almost always a *lot* faster.

choldgraf:
maybe @Kully or @theengineear can comment on this one?

choldgraf:
(on the plus side, it does seem like the memory leak is not happening. It's a bit hard to tell because my original code isn't working due to the above changes, but seems promising)

theengineear:
Re https://github.com/plotly/plotly.py/pull/550#issuecomment-241157670

This was done as a simple fix to make *tests* work between Python 2 and Python 3. That's really the only important rationale for it. I figured that it wouldn't be too much of a performance knock if it happened once at the end, which is why I suggested it to @Kully.

That said, I haven't looked at the actual implications and there's no *other* reason to have a list over a numpy array, so we can definitely change that back.

choldgraf:
Cool - I seem to remember arrays being a lot more efficient than lists in the JSON encoding, but I just tried it out and didn't see a difference:

image

Maybe there were optimizations made to the JSONEncoder class? Either way, that was why I was pushing for objects to be made into arrays before being put in the plotly dictionaries. Maybe the performance differences only come up when they're really large lists/arrays or something.

theengineear:
^^ Yah, that `PlotlyJSONEncoder` is actually pretty dim...

A couple notes:

1. It is going to be transforming np.arrays into lists there *anyhow*.
2. It does some stupid things to prevent extended JSON from being sent to Plotly https://github.com/plotly/plotly.py/blob/master/plotly/utils.py#L147-L163

There's likely a better way, right now this is basically, encoding, unencoding, and then encoding... It had to do with either Plotly backend or plotlyjs code not being able to work with the `extended` JSON that Python takes the liberty of using. At least that's what I *remember* the problem being. Conditions may have changed though, it may not even be a concern. This would be a bigger speedup than the np.array vs list in the output of the color functions, imo.

Feel like digging in 😛?

monfera:
Might not be relevant for this processing step, but one of the render speedup tests we've been evaluating (with positive results so far) is the use of JavaScript typed arrays for input. This can avoid expensive array copies, which matter a lot when a model is large. It usually requires arrays of different shape compared to the current API (e.g. `[vert00, vert01, vert02, .., vertn0, vertn1, vertn2]` for vertex indices, but it's possible to use e.g. [`scijs/ndarray`](https://github.com/scijs/ndarray) by our @mikolalysenko to make it look like friendly multidimensional arrays) and we're still in the prototyping stage, initially with the upcoming 2D `pointcloud` plot. 

As a result, we see that the largest single use of the initial render time is taken by the actual CPU->GPU memory copy (we want to cut down on it too e.g. [by using `gl.TRIANGLE_STRIP`](http://www.learnopengles.com/tag/triangle-strips/)). All in all, we expect great speedup in JavaScript+WebGL, perhaps OOM 100ms for a brain mesh update and probably a small multiple of that for an initial render, if the arrays are passed on in the typed array structure.

A conclusion is that due to faster JavaScript processing, any array preprocessing in Python, and the (presumable) array copy, or serialization/parsing from Python to Javascript will likely become a significant chunk of the overall processing time, assuming larger meshes, from hundreds of thousands of triangles.

These details partly belong to the render speedup issue, @choldgraf please consider opening one. Details e.g. baseline benchmarks can be added later.

monfera:
There is some old thread [on efficiently moving `numpy` arrays to JavaScript typed arrays](https://mail.scipy.org/pipermail/ipython-user/2014-October/013859.html), I don't see the specific link in the answer but perhaps interesting

monfera:
It looks like a fairly direct approach would be to use `webSocketConnection.binaryType = "arraybuffer";` because it should yield an `event.data instanceof ArrayBuffer` which is directly viewable as a JavaScript typed array of the appropriate type (e.g. `Float32Array`) and requires no further copy or transformation. I think IPython notebooks already use WebSockets for binary communication with JavaScript so maybe the JSON serializing and parsing can be avoided. Sorry if I misunderstood something.

monfera:
`np.ndarray` seems to [directly support](http://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.tobytes.html) binary serialization of buffers, and a nice benefit is that Python users could use nice, 2D arrays for trimesh vertices and indices, and they would be directly usable on the other end of the WebSocket as a flat, 1D JS typed arrays, directly usable for the WebGL `gl.bufferData`. (JS users could use the above mentioned `scijs/ndarray` for a similar 2D view effect.)

theengineear:
@monfera super interesting conversation. I agree that this serialization step (when operating in offline mode, or some sort of local mode) is just causing delays. I wonder if this is the appropriate thread to talk about it in though. I have a feeling that this requires a non-trivial refact to get Python/JS to talk in a different way, no?

Just to be clear, I don't see JSON serialization to save things to Plotly's db / AWS instances going away any time soon. But that said, I think you know a lot more than I do on this topic and I'd happily discuss further. Again, maybe best in another thread though?

Thanks for the thorough technical discussion! This is great!

monfera:
@theengineear you're right, I wasn't entirely ontopic, mentioning one significant element (typed JS arrays in `plotly.js`) of a future speedup in case it has relevance to the Python -> JS communication (one comment above by @choldgraf mentioned converting np.arrays to lists but it seems avoidable). Having been asked to work on initial render speedup in `plotly.js` it looks like a serial thing so even if the JS part is very fast, it can still be slow end-to-end.

The thought of replacing JSON for database storage didn't occur to me as I only had the Python -> JS direct communication in mind (I assume, perhaps wrongly, that in order to plot something from Python, there's no interim db storage of the data).

choldgraf:
hey all - I'm gonna close this because I'm pretty sure this PR is pretty stale :-/  if anybody wants to take it over then feel free

monfera:
@choldgraf thanks for your reminder. There are a variety of things here, which might be why it never got assigned (or simply, limited capacity). To take an inventory:

1. addition of `colorbar` causes [some error](https://github.com/plotly/plotly.py/pull/550#issuecomment-241110195)
2. PR commit: turning the `show_colorbar` parameter [into a flag](https://github.com/plotly/plotly.py/pull/550/commits/7b62f54eb05f6a184e0bce16a143c1afafc6aaae) but this PR conversation doesn't tell about it; the linked [issue comment](https://github.com/plotly/plotly.py/issues/487#issuecomment-241105653) points to benefits, I'm not sure if it's clear to those folks working on the Python binding 
3. [performance](https://github.com/plotly/plotly.py/pull/550#issuecomment-241215760) increase [ideas](https://github.com/plotly/plotly.py/pull/550#issuecomment-241228681) for array transfer between Python and JS
4. original memory leak seems to be [addressed](https://github.com/plotly/plotly.py/pull/550#issuecomment-241158492)

Where we're at with each is, afaik:

1. I'm just guessing, but from the description it looks like it might be a `plotly.js` problem, so an issue in the `plotly.js` repo with a minimal example would be fantastic, including stating the expectation. You seem to have detailed info as you tied the appearance of the issue to a feature addition, offending version/commit etc. helps triage
2. I've no info on this, the [suggestion](https://github.com/plotly/plotly.py/issues/487#issuecomment-241106620) was to make it into an item of its own (needs, benefits, critical or nice to have, what it involves)
3. the separate plotly.js [issue](https://github.com/plotly/plotly.js/issues/1784#issuecomment-310854625) covers this topic and has some alternatives collected, but not implemented; looping in @chriddyp as well
4. no action needed, except perhaps being on the outlook for leaks after item 1 is solved

Also, there's a linked [issue](https://github.com/plotly/plotly.py/issues/487) referred above, the conversation ended [here](https://github.com/plotly/plotly.py/issues/487#issuecomment-241107776).

It helps triage and assignment if issues/PRs are isolated and granular, in the repo where it's likely to need the change, and with an issue/PR title that gives the gist of the contents. Whether it's one of the Plotly employed contributors or community contributors, it's easier to take and clear specific items with minimal offending tests and expectations.

choldgraf:
for sure - I just don't have time to work on it ATM and I'm cleaning out my old PR cruft. That's helpful info if somebody else wants to take it over though!

jackparmer:
@choldgraf JSON serialization slowness aside, we have something exciting in the works for brain volume visualization:

http://fhtr.org/gl-volume3d-demo/

image

Let us know if you have any suggestions for this demo on this based on your experience with pysurfer. This volume viz should land in an official plotly.js release early next year.

PR #551, opened 2016-08-19
added appropriate colorbar max and min values, edges_color to param for line coloring, and making code more clear

Kully:
@theengineear


plotly/tools.py
```diff
@@ -3305,7 +3305,7 @@ def _trisurf(x, y, z, simplices, show_colorbar, colormap=None,
                         color_func[index] = FigureFactory._label_rgb(foo)
 
             mean_dists = np.asarray(color_func)
-        else:
+        elif hasattr(color_func, '__call__'):
```
theengineear: 🐄 Why not just `elif color_func is not None`?

plotly/tools.py
```diff
@@ -3333,30 +3333,39 @@ def _trisurf(x, y, z, simplices, show_colorbar, colormap=None,
...
-                                     colorscale=colorscale,
-                                     intensity=[0, 1],
-                                     showscale=True)
+        if not isinstance(mean_dists[0], str) and show_colorbar is True:
+            # make a colorscale from the colors
+            colorscale = FigureFactory._make_colorscale(colormap)
+            colorscale = FigureFactory._convert_colorscale_to_rgb(colorscale)
+
+            colorbar = graph_objs.Scatter3d(
+                x=x[:2],
```
theengineear: Can you comment on why we ditch the first couple values here? I can't remember why we'd want to do this.

theengineear: Edit `only keep` not `ditch`, sorry.

theengineear: Just looking for some documentation on it 😸

plotly/tools.py
```diff
@@ -3333,30 +3333,39 @@ def _trisurf(x, y, z, simplices, show_colorbar, colormap=None,
...
+                    colorscale=colorscale,
+                    showscale=True),
+                hoverinfo='None',
+                showlegend=False
+            )
 
         # the triangle sides are not plotted
-        if plot_edges is not True:
-            if show_colorbar is True:
+        if plot_edges is False:
```
theengineear: 😛 👍

plotly/tools.py
```diff
@@ -3333,30 +3333,39 @@ def _trisurf(x, y, z, simplices, show_colorbar, colormap=None,
...
+            if not isinstance(mean_dists[0], str) and show_colorbar is True:
```
theengineear: 🐄 you use this `not isinstance(mean_dists[0], str)` twice, can you assign it to a meaningful variable name to make the code easier to understand?

plotly/tools.py
```diff
@@ -3622,6 +3637,7 @@ def dist_origin(x, y, z):
             x, y, z, simplices,
             color_func=colors,
             show_colorbar=True,
+            edges_color='rgb(2,85,180)',
```
theengineear: Any reason not to be consistent about the spacing? `rgb(2, 85, 180)`?

theengineear:
Cool, ping me when you get that test passing again for a final review?

plotly/tools.py
Kully: > Why not just elif color_func is not None?

Your way makes sense upon closer inspection. Actually, I think `else` would work as well, since if `color_func` was a function, it would therefore not be `None` nor a `list/array`

Kully:
@theengineear I made all your edits and additionally switched `x[:2]` to just `x[0]` for `x`, `y` and `z` as plotting one invisible point also works for the colorbar's functionality. I also increased version number and will update in `CHANGELOG` after we merge.

Whoop!

theengineear:
@Kully did you mean to edit `plotly/tests/test_optional/temp-plot.html`?

theengineear:
Great! 💃 after you respond to https://github.com/plotly/plotly.py/pull/551#issuecomment-241483950.

Kully:
> did you mean to edit plotly/tests/test_optional/temp-plot.html?

So I accedentally pushed that temp-chart one fateful day ago. Chelsea didn't think it was a big deal iirc but everytime I rune `nosetests` in the `test_optional` dir, it shows that file as modified. Can I actually just delete what I pushed? Or do it after the PR?

theengineear:
Nah! Let's revert it! It's never a good idea to commit things you don't mean to! You can either:

1. make a new commit.
2. refact that commit using an interactive rebase.

Let me know if you need help with either.

Kully:
> make a new commit.

Let's do this.

empet:
I don't think that you can draw a Scatter3d trace  defined only by one point.

By [https://plot.ly/python/reference/#scatter3d-x](https://plot.ly/python/reference/#scatter3d-x)
x, y, z must be lists, numpy arrays or pandas timeseries.
Moreover, the length of x, y, z should be equal to the length of the list color.

In this colorbar definition the above conditions are violated:

    colorbar = graph_objs.Scatter3d(
                x=x[0],
                y=y[0],
                z=z[0],
                mode='markers',
                marker=dict(
                    size=0.1,
                    color=[min_mean_dists, max_mean_dists],
                    colorscale=colorscale,
                    showscale=True),
                hoverinfo='None',
                showlegend=False
            )

I defined a very simple Plotly plot defining a Scatter3d instance like this one and no colorbar is plotted.

Kully:
> I don't think that you can draw a Scatter3d trace defined only by one point.
> 
> By https://plot.ly/python/reference/#scatter3d-x
> x, y, z must be lists, numpy arrays or pandas timeseries.
> Moreover, the length of x, y, z should be equal to the length of the list color.
> 
> In this colorbar definition the above conditions are violated:
> 
> colorbar = graph_objs.Scatter3d(
>             x=x[0],
>             y=y[0],
>             z=z[0],
>             mode='markers',
>             marker=dict(
>                 size=0.1,
>                 color=[min_mean_dists, max_mean_dists],
>                 colorscale=colorscale,
>                 showscale=True),
>             hoverinfo='None',
>             showlegend=False
>         )
> I defined a very simple Plotly plot defining a Scatter3d instance like this one and no colorbar is plotted.

I think the difference is that the `.create_trisurf` function is not just using `Scatter3d` but also creating the trisurf, which may explain why the colorbar is visible for me.

empet:
@Kully I tested with a trisurf too, and the colorbar is not displayed.

Kully:
> @Kully I tested with a trisurf too, and the colorbar is not displayed.

Let's wait till the pip package gets updated (I updated the version number) and we'll see what's up with it then.

Issue #535, opened 2016-08-09
Violin plots print subplot grid information by default


pjbull:
1. Create a violin plot, e.g.:
   https://plot.ly/python/violin-plot/#multiple-violins

Result:
 Grid information is printed to stdout and can't be turned off:

```
This is the format of your plot grid:
[ (1,1) x1,y1 ]  [ (1,2) x2,y1 ]  [ (1,3) x3,y1 ]  [ (1,4) x4,y1 ]  [ (1,5) x5,y1 ]
```

Expected:
 No grid information to print, or a setting to control verbosity.

Possible fix:
- Pass `print=False` explicitly to `make_subplots`
- Make `print=False` the default parameter to `make_subplots`
- Switch to logging rather than print statements so output can easily be controlled/redirected.

Here are the violin plot calls where `print=True`, though `True` is also the default parameter"
https://github.com/plotly/plotly.py/blob/master/plotly/tools.py#L2730
https://github.com/plotly/plotly.py/blob/master/plotly/tools.py#L2798
https://github.com/plotly/plotly.py/blob/master/plotly/tools.py#L2899

Kully:
This is issue has been fixed (along with the subplot grid info printing that occurs in `FF.create_scatterplotmatrix` and will be available in the next pip package of plotly

PR #557, opened 2016-08-29
add print_grid param to create_violin

Kully:


Kully:
@cldougl @theengineear 
A response to this issue: https://github.com/plotly/plotly.py/issues/535. Do you guys think it _would_ be a good idea to provide this `print_grid` param for all FF's? This certainly solves the issue, but I think being consistent with this parameter would be a must afterwards. Thoughts?

cldougl:
@Kully I think the violins were the only ones that used `make_subplots()` which was why the grid was printed.

Kully:
> @Kully I think the violins were the only ones that used make_subplots() which was why the grid was printed.

Scatterplot Matrix does as well, iirc.

Kully:
> Scatterplot Matrix does as well, iirc.

Shall I add the param to `.create_scatterplotmatrix` then?

theengineear:
@Kully I'd suggest *not* exposing that to users. I can't see why there'd be much of a use-case for it. Just implicitly turn them off in the methods that use `make_subplots`. The FF methods are meant to be *helper* functions that reduce the cognitive load of constructing more complicated Plotly plots. Seem OK?

Kully:
> @Kully I'd suggest not exposing that to users. I can't see why there'd be much of a use-case for it. Just implicitly turn them off in the methods that use make_subplots. The FF methods are meant to be helper functions that reduce the cognitive load of constructing more complicated Plotly plots. Seem OK?

Yes it does. I can make them implicit.

Kully:
@theengineear Good to merge? I can make a PR for the Scatterplot Matrices FF and do the same thing...

theengineear:
Yep, 💃!

PR #562, opened 2016-09-02
first pass at nteract support

chriddyp:


chriddyp:
Here's a first pass at offline interact support. Here's a comment in the commit for some details:

```
Display 2 versions of the figure:
 - The raw HTML is the "old-school" style - we just throw in a JS script
   inside the cell. This works for a lot of environments but some JS is
   stripped out in some rendering environments / platforms like
   GitHub, nteract, and eventually Plotly
 - The custom `mime-types` is the new-school way. nteract supports
   rendering plotly graphs this way (see https://github.com/nteract/nteract/pull/662)
   and others will hopefully follow suite.
 Both of these bundles will be saved as part of the notebook. The renderer
 will choose which one to render. So 'text/html' is like a fallback for
 any environment that doesn't have the 'application/json+plotly.v1'
 renderer available.
 The bummer is that we're injecting the figure JSON in here twice, so
 notebooks that are pretty big are about to get a lot bigger.
 The upside is that we're dropping any compatability for older notebook
 versions
```

chriddyp:
Screenshot from nteract: 
image

Associated nteract PR: https://github.com/nteract/nteract/pull/662

chriddyp:
Works across nteract and nbviewer, however opening these notebooks inside my existing jupyter notebook causes a validation error. It's harmless I think, but a pretty scary message to see.

image

chriddyp:
CC @theengineear @rgbkrk @jackparmer @cldougl @charleyferrari @cldougl

plotly/offline/offline.py
```diff
@@ -292,7 +292,33 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+    #   inside the cell. This works for a lot of environments but some JS is
+    #   stripped out in some rendering environments / platforms like
+    #   GitHub, nteract, and eventually Plotly
+    # - The custom `mime-types` is the new-school way. nteract supports
+    #   rendering plotly graphs this way (see https://github.com/nteract/nteract/pull/662)
+    #   and others will hopefully follow suite.
+    # Both of these bundles will be saved as part of the notebook. The renderer
+    # will choose which one to render. So 'text/html' is like a fallback for
+    # any environment that doesn't have the 'application/json+plotly.v1'
+    # renderer available.
```
theengineear: Cool, yah, this makes sense to me. Agreed that notebooks getting bigger is a bummer... I'm not sure how much of an edge case that is.

rgbkrk:
That's some hardcore validation. I'll bet if our key isn't `application/json+` it would let it be.

screen shot 2016-09-02 at 2 33 28 pm

@minrk, @fperez does this fall out of spec?

rgbkrk:
Tried out `application/vnd.plotly.v1+json` in case it was just a hard check on format within `application/json`. No go. This is going to affect the custom mimetype for the widgets in JupyterLab too.

rgbkrk:
Ah, here we go. Pretty [explicit in the schema here](https://github.com/rgbkrk/nbschema/blob/master/v4/nbformat.v4.schema.json#L348-L353) if not the intention. If it's a custom mimetype it expects an array of strings or a string.

rgbkrk:
I suppose we'd need the notebook schema to also have (for us):

```
"patternProperties": {
  "^application/json\+[a-zA-Z0-9]+/[a-zA-Z0-9\\-\\+\\.]+$": {
    "type": "object"
  }
}
```

for JupyterLab/ipywidgets:

```
"patternProperties": {
  "^application/vnd\.[a-zA-Z0-9]+/[a-zA-Z0-9\\-\\+\\.]+$": {
    "type": "object"
  }
}
```

rgbkrk:
Current proposal for registering custom mimetypes in nteract: https://github.com/nteract/nteract/issues/670

fperez:
Not quite sure where your thinking is right now, are you going in the direction of extending the nb schema?

rgbkrk:
I'd like to pick the right future direction while not crippling usage of the classic notebook. There are quite a few options here, so I was outlining some thoughts on them.

fperez:
Got it.  I guess once this is a bit more fleshed out, would be good to run it by the Jupyter list for broader thoughts.

rgbkrk:
Yeah, I'll make an outline in an email on the jupyter list. It's working quite nicely in nteract, and I'm working on vega support, soon bokeh as well.

rgbkrk:
Beyond that though, it seems like the only way for us to do custom extensions is to make them be double encoded JSON in the current schema. Only the pure field `application/json` allows regular objects.

rgbkrk:
Ok, definitely thinking that plotly's mimetype should be `application/vnd.plotly.v1+json`.

rgbkrk:
Since https://github.com/jupyter/nbformat/issues/42 has been merged we can now send over the raw JSON on the next nbformat release.

However, since users of old versions of the notebook should allow this to still work, we can send the serialized JSON as a string directly, which was handled for backwards compatibility in nteract here: https://github.com/nteract/nteract/pull/675/files#diff-18d3d49b31452599ff46a67c571aa4f8R7

One other thing that still needs to be changed is that the mimetype should now be `application/vnd.plotly.v1+json`

theengineear:
@chriddyp @rgbkrk just tried this and it seemed to work fine?

image

(this is just the current plotly release that I used)

Has something just changed in `nteract`?

chriddyp:
@theengineear - online mode works fine, this would just be for offline support.
I'm all for supporting this PR. the only drawback is that we will be including the bundle and the Json twice: once for the interact minetype and another for Jupiter notebooks that don't support the mimetype enhancement yet. while we could detect which notebook type the author is using, we don't know which notebook type the viewer might be using, so we need to include both. big notebooks will go from 15MB to 30MB. 

it doesn't seem like there is any getting around this, so it seems like it's something that we should just add. thoughts?

jackparmer:
+1 from me. We can't really promote _nteract_ without offline mode working.

plotly/offline/offline.py
```diff
@@ -292,7 +292,33 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+    # The bummer is that we're injecting the figure JSON in here twice, so
+    # notebooks that are pretty big are about to get a lot bigger.
+    # The upside is that we're dropping any compatability for older notebook
+    # versions
+    display_bundle = {
+        'application/json+plotly.v1': {
```
theengineear: @chriddyp I believe this needs to be updated now (as per https://github.com/plotly/plotly.py/pull/562#issuecomment-244552719) to `application/vnd.plotly.v1+json`.

plotly/offline/offline.py
```diff
@@ -292,7 +292,33 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+            'data': json.loads(json.dumps(figure['data'], cls=plotly.utils.PlotlyJSONEncoder)),
+            'layout': json.loads(json.dumps(figure['layout'], cls=plotly.utils.PlotlyJSONEncoder))
```
theengineear: @chriddyp @rgbkrk plotly.js now has a new way to supply information to a `Plotly.plot` call. https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_api.js#L49-L63

You can just dump a data *object* in now:

`Plotly.plot(plotDiv, {data: [], layout: {}})`. This is going to be less fragile since the number of arguments doesn't need to be adjusted in future releases of `plotly.py` and `nteract` (`transforms/plotly.js`) for things like `frames` support, etc, etc. This requires a change to the `nteract` transform. I'm happy to make a PR to nteract to change this. Let me know what you think.

plotly/offline/offline.py
```diff
@@ -292,7 +292,33 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+        },
+        'text/html': plot_html
+    }
+
+    display(display_bundle, raw=True)
```
theengineear: It's be a minute since I've thought about how notebooks choose display formats. If it's an easy answer, can someone explain how we'll end up selecting `application/vnd.plotly.v1+json` *over* `text/html`? Else, I can read up on this again.

theengineear: I see the comment in the code above (btw), but I'm just wondering how the renderer decides *when* to fallback to `text/html`.

plotly/offline/offline.py
```diff
@@ -292,7 +292,33 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+    #   GitHub, nteract, and eventually Plotly
+    # - The custom `mime-types` is the new-school way. nteract supports
+    #   rendering plotly graphs this way (see https://github.com/nteract/nteract/pull/662)
+    #   and others will hopefully follow suite.
+    # Both of these bundles will be saved as part of the notebook. The renderer
+    # will choose which one to render. So 'text/html' is like a fallback for
+    # any environment that doesn't have the 'application/json+plotly.v1'
+    # renderer available.
+    # The bummer is that we're injecting the figure JSON in here twice, so
+    # notebooks that are pretty big are about to get a lot bigger.
```
theengineear: If folks with *yuuge* notebooks end up wanting to slim down their documents, we could offer a `set_plotly_display_environment('nteract')`, right? And then we can set the fallback `text/html` to something like `'This plot is only setup to display inside the `nteract` environment`.

Again, if it becomes an actual issue. I'd rather things just *worked* as a first pass, for sure.

theengineear:
@chriddyp, can we test for the `application/vnd.plotly.v1+json` prop in the display bundle using the new notebook testing framework?

Let me know if you have cycles to rebase, fixup, test this. If you don't, I can jump in here.

plotly/offline/offline.py
rgbkrk: Happy to take a PR on nteract's transform for that since it's a direct transfer of the overall payload.

plotly/offline/offline.py
rgbkrk: A jupyter frontend falls back to a lighter media type when it doesn't support the one specified. In the case of nteract, it has an ordered list called `displayOrder`. The default displayOrder is this:

```
  'application/json',
  'application/javascript',
  'text/html',
  'image/svg+xml',
  'text/markdown',
  'text/latex',
  'image/svg+xml',
  'image/gif',
  'image/png',
  'image/jpeg',
  'application/pdf',
  'text/plain',
```

when the plotly transform is part of our app, the plotly mimetype is one of the highest priority mimetypes along with geojson and vega. If that plotly transform+mimetype combo is not available, the next richest is `text/html`.

Jupyter notebook and jupyterlab do something similar, as does nbviewer and the qtconsole.

plotly/offline/offline.py
theengineear: OK, great! Thanks, @rgbkrk!

chriddyp:
> Let me know if you have cycles to rebase, fixup, test this. If you don't, I can jump in here.

Could you jump in here? I don't have the cycles this week. Happy to review though!

theengineear:
I probably should *not* have rebased to fix conflicts / add changes. I have no self-restraint.

Just wanted to give everyone a heads up that might have pulled this down.

theengineear:
^^ weird, `sign_in` seemed to fail **all over the place** in our tests... rebuilding...

theengineear:
hrm... I think there's something off with how other tests are getting `sign_in`'d to. I.e., the setup for the tests is *not* sandboxed... which is problematic. I'm going to take the time to get this *right*.

We should have our `tox` (or CI?) setup *always* start the build off by deleting `~/.plotly/`. Then, the more difficult change is to ensure that we use the `PlotlyTestCase` whenever we write a test that requires changing config/credentials in *any* way.

This will likely take a little while since a lot of our tests are still written in `module-style` (i.e., not class-based), so we need to first refact these tests so that we can properly subclass `PlotlyTestCase` 😿. Maw.

☕️ ☕️ ☕️ It's worth getting this correct *now*, we keep running into these issues.

**Really**, fewer of our tests should even be hitting a `Plotly` server, that that's a problem for another day...

Ideally, we'd just totally mock out the Python API and setup a suite of API tests *inside the `streambed` repo*. Again, definitely for another day, but it would *greatly* speed up our tests and *completely decouple* our tests from any specific plotly server.

theengineear:
@chriddyp @rgbkrk I got tests passing and AFAIK offline mode is working as expected:

image

Anything else to be done here? I'm a little booked up on other tasks and don't want to let this linger much longer.

//cc @jackparmer

theengineear:
^^ That's [this nteract version](https://github.com/nteract/nteract/releases/tag/v0.0.15)

theengineear:
Sorry to keep pinging, but are we all good with this @chriddyp? Anything else you wanted to add?

theengineear:
It's got my 💃 :)

plotly/offline/offline.py
```diff
@@ -273,7 +273,7 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
           validate=True, image=None, filename='plot_image', image_width=800,
           image_height=600):
     """
-    Draw plotly graphs inside an IPython notebook without
+    Draw plotly graphs inside an IPython or Jupter notebook without
```
rgbkrk: s/Jupter/Jupyter/

plotly/offline/offline.py
```diff
@@ -335,7 +335,33 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
     )
 
-    ipython_display.display(ipython_display.HTML(plot_html))
+    figure = tools.return_figure_from_figure_or_data(figure_or_data, validate)
+
+    # Display 2 versions of the figure:
+    # - The raw HTML is the "old-school" style - we just throw in a JS script
+    #   inside the cell. This works for a lot of environments but some JS is
+    #   stripped out in some rendering environments / platforms like
+    #   GitHub, nteract, and eventually Plotly
```
rgbkrk: JavaScript is not stripped out in nteract - the core differences are that:

* `require` is commonjs require rather than requirejs' require in the desktop app
* we don't have the same JavaScript API as the Jupyter notebook

Neither of those should be relied on for HTML outputs regardless (or at least some primitive feature detection and fallbacks need to happen).

plotly/offline/offline.py
```diff
@@ -335,7 +335,33 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+    # - The custom `mime-types` is the new-school way. nteract supports
+    #   rendering plotly graphs this way (see https://github.com/nteract/nteract/pull/662)
+    #   and others will hopefully follow suite.
```
rgbkrk: s/follow suite/follow suit/

plotly/offline/offline.py
theengineear: ^^ I ended up just 🔪-ing a lot of that comment out, I don't think it needs to get committed here.

chriddyp:
looks great to me, this is awesome! 💃

Madhu94:
> while we could detect which notebook type the author is using, we don't know which notebook type the viewer might be using, so we need to include both. big notebooks will go from 15MB to 30MB.

@chriddyp Did you find a way around this ? Or, do notebooks with multiple mime types just have to be larger ?

rgbkrk:
They only have to be larger if you wish to support old versions of the notebooks.

Madhu94:
Thanks for the quick response @rgbkrk.  Sorry, rephrasing my question, "Notebooks with custom mime-types which also have to be used with nbviewer just have to be larger? "

rgbkrk:
Someone would have to add support to nbconvert and nbviewer, that's for sure.

Madhu94:
I'll take a shot.  

Ipywidgets render in nbviewer by fetching the widget models and views from unpkg, and i _think_ even this can now be [configured](https://github.com/jupyter/nbviewer/pull/738). Maybe a similar strategy for fetching mime-renderers from somewhere..? 

Anyway, I'll open a ticket for discussion. Thanks a lot. :)

PR #567, opened 2016-09-07
implicitly made print_grid=False for scatterplot matrix in FF

Kully:


Kully:
@theengineear Same thing here.

theengineear:
💃 !

PR #574, opened 2016-09-27
Update default schema

etpinard:
... get `master` to pass on CircleCI again.

@cldougl @Kully


Kully:
@etpinard Is this version 1.12.10? Was the pip package updated?

PR #577, opened 2016-10-05
Allow the use of custom linkage/distance functions for creating dendrograms.

grst:
[Scipy hierarchy](http://docs.scipy.org/doc/scipy-0.14.0/reference/cluster.hierarchy.html)
allows multiple linkage modes for creating dendrograms. So far it
has not been possible to create dendrograms using plotly with
a different linkage mode than 'complete'.

This commit allows the user to specify arbitrary linkage and distance
function which can come either from the scipy package or user-created
functions.


jackparmer:
Thanks @grst !

PR #549, opened 2016-08-18
Add JS tests for Jupyter NBs

etpinard:
**Behold:** JS tests in Jupyter Notebooks are coming to plotly.py

My strategy was the following:
1. Have a few `ipynb` fixture files with plotly.py code cells but no outputs
2. Execute them using the jupyter [execute api](http://nbconvert.readthedocs.io/en/latest/execute_api.html)
3. Convert the resulting `ipynb` JSON to HTML using [nbconver api](http://nbconvert.readthedocs.io/en/latest/nbconvert_library.html)
4. Add JS test bundle to the resulting HTML file
5. Run the JS tests in a browser!

Step 5 was a little harder than I thought. There's a ton of test runners that take a JS bundle, shim it into an HTML page and run the tests. But, I couldn't find any test runner that takes an already made 
HTML page and runs the test there. A few runners allow you to add HTML to their default test page, but I couldn't get the results I wanted, so I decided to build my little test runner see `lib/server.js`

TODO:
- [x] add more test fixtures / cases
- [x] get this to work on CircleCI 
- [x] make sure this works in all python versions we support

@chriddyp @theengineear @cldougl @mdtusz

PR #584, opened 2016-10-18
Animations/Frames in Python API

Kully:
Relevant Issue: https://github.com/plotly/streambed/issues/8129


Kully:
fyi, this is my PR for issue [8129](plotly/streambed#8129)

So far, `iplot` in offline can properly create a chart with animations in the notebook,
`plot` is next...

cc @theengineear @chriddyp @etpinard @rrusser

Kully:
@theengineear `plot` and `iplot` both work online. I'm tackling online mode now and probably end up making that beta method...

Should I write tests now as I go for offline or just continue?

theengineear:
We're in a bit of a time-crunch, and my bet is that writing tests for this will be difficult. I'd suggest moving on and getting it to work online next.

Nice job so far, btw!

//cc @chriddyp

plotly/offline/offline.py
```diff
@@ -204,11 +213,30 @@ def _plot_html(figure_or_data, show_link, link_text, validate,
...
-        layout=jlayout,
-        config=jconfig)
+    if 'frames' in figure_or_data:
+        script = '''
+        Plotly.plot(
+            '{id}',
+            {data},
+            {layout},
+            {config}
+        ).then(function () {add_frames}).then(function(){animate})
```
rreusser: See: https://github.com/plotly/plotly.js/pull/1054

That PR has been merged and makes the syntax somewhat simpler. I'd elaborate here, but the PR description explains it fairly clearly. Calling `addFrames` separately will remain perfectly fine, but ideally will become unnecessary.

etpinard:
> Should I write tests now as I go for offline or just continue?

@Kully if you feel like figuring out why https://github.com/plotly/plotly.py/pull/549 is still failing on CircleCI, be my guest.

Kully:
> @Kully if you feel like figuring out why #549 is still failing on CircleCI, be my guest.

Looks like you just rebuilt it eh?

plotly/offline/offline.py
Kully: Okay, fantastic. I think Chris mentioned that I'd have to do some rewriting when that gets added.

I'll leave it for now, but does that mean `frames` is available as a "valid attribute" for Python?

rreusser: Yeah, should just be a simplification. Long story short, given `data`, `layout`, `frames`:

- `frames[i].data` follows the same validation as `data` itself
- `frames[i].layout` follows the same validation as `layout`

I'm not 100% sure what that means for validation, but on a strictly abstract level, it's conceptually simple.

rreusser: Oh, sorry, unless you were referring to python arguments instead of data validation… That PR just changes the call signature. Not 100% sure the other concerns, but feel free to ask whatever questions you need and I'll do my best to get you what you need.

Kully: Yeah,  I'm referring to this `py.iplot(fig)` call where `fig` contains `data`, `layout` and `frames`. Right now, I'm still getting that:

```
'frames' is not allowed in 'figure'

Path To Error: ['frames']

Valid attributes for 'figure' at path [] under parents []:

    ['layout', 'data']

```

Kully: I am still receiving the error. Anything I can do to fix? I `sudo pip upgraded` just now and it reported no new changes.

rreusser: Hmm… not sure. It seems like that's a whitelisting issue on the plotly side.

Kully: Okay, so after playing `follow-that-code` for a while, I've traced it to the [v2 plot schema](https://api.plot.ly/v2/plot-schema?sha1=22619e9597093bddcc7aad1158615f7607068cd3) Can we add `frames` to this?

Kully:
Just threw in code from https://github.com/plotly/plotly.py/pull/586 without merging it formally

Kully:
@theengineear @rreusser 

So finally I have a `create_animations` which deals with a `fig` with `frames` in its body. Plotting works okay, just two things:
1) I don't think (know) if animations are live on plotly. @chriddyp Do you know if they work now?
2) `frames` is not showing up [here](https://plot.ly/~AdamKulidjian/1669#code) so I don't know if they are completely passing through.

Additionally a GRID as well as a PLOT is made in the user's account each time `frames` is included, as v2 is being used. Also just realized that all GRIDS are hard-coded as `world_readable`. I can change this later but just noting it here.

Andrew, do you mind looking at my function and giving some PEP-8/style/structure/all-that-fun-stuff comments if you have any? :)

plotly/graph_objs/graph_objs.py
```diff
@@ -836,6 +836,20 @@ def __init__(self, *args, **kwargs):
                                                   _parent_key='data')
     figure_class.__init__ = __init__
 
+    def __setitem__(self, key, value, **kwargs):
+        if key == 'frames':
```
theengineear: Let's add a comment about this one. Maybe tie it to a new issue to better-integrate `frames` into `Figure`? As always, pls put the issue numer (or link) in the `TODO` comment.

plotly/graph_objs/graph_objs.py
```diff
@@ -836,6 +836,20 @@ def __init__(self, *args, **kwargs):
...
+            super(PlotlyDict, self).__setitem__(key, value)
+        else:
+            super(figure_class, self).__setitem__(key, value, **kwargs)
+    figure_class.__setitem__ = __setitem__
+
+    def _get_valid_attributes(self):
+        super(figure_class, self)._get_valid_attributes()
+        if 'frames' not in self._valid_attributes:
```
theengineear: Ditto hear, link it to that same issue ^^

plotly/graph_reference.py
```diff
@@ -38,6 +38,7 @@
     'ErrorZ': {'object_name': 'error_z', 'base_type': dict},
     'Figure': {'object_name': 'figure', 'base_type': dict},
     'Font': {'object_name': 'font', 'base_type': dict},
+    'Frames': {'object_name': 'frames', 'base_type': dict},
```
theengineear: Is this necessary? Did I accidentally leave this in or something ;P?

plotly/offline/offline.py
```diff
@@ -162,8 +162,15 @@ def init_notebook_mode(connected=False):
 
 def _plot_html(figure_or_data, config, validate, default_width,
                default_height, global_requirejs):
-
-    figure = tools.return_figure_from_figure_or_data(figure_or_data, validate)
+    # force no validation if frames is in the call
```
theengineear: `TODO` this and link to some issue pls :)

plotly/plotly/plotly.py
```diff
@@ -1400,9 +1400,6 @@ def add_share_key_to_url(plot_url, attempt=0):
 
 
 def _send_to_plotly(figure, **plot_options):
-    """
-
-    """
```
theengineear: ⚡️

plotly/plotly/plotly.py
```diff
@@ -1414,8 +1411,7 @@ def _send_to_plotly(figure, **plot_options):
                              fileopt=plot_options['fileopt'],
                              world_readable=plot_options['world_readable'],
                              sharing=plot_options['sharing'],
-                             layout=fig['layout'] if 'layout' in fig
-                             else {}),
+                             layout=fig['layout'] if 'layout' in fig else {}),
```
theengineear: Cool. This is nice for consistency.

plotly/plotly/plotly.py
```diff
@@ -1454,6 +1453,129 @@ def _send_to_plotly(figure, **plot_options):
     return r
 
 
+def create_animations(fig, kwargs, payload):
+    """
+    Makes a post to GRIDS and PLOTS if frames is in the figure.
+
+    This function bypasses the current '/clientresp' route  of making a POST
```
theengineear: Hmm. I don't know that you need to go into the details of `/clientresp`. This doc will likely just become outdated/confusing if you're *too* specific here. Not a big deal.

plotly/plotly/plotly.py
```diff
@@ -1454,6 +1453,129 @@ def _send_to_plotly(figure, **plot_options):
...
+    request to return back a url. Instead, the V2 REST API is used if frames
```
theengineear: ditto on `V2`

plotly/plotly/plotly.py
```diff
@@ -1454,6 +1453,129 @@ def _send_to_plotly(figure, **plot_options):
...
+    is part of the figure's keys. Currently, 'error', 'message' and 'warning'
+    are hard codded to the empty string.
+    """
+    url_v2_plot = "https://api.plot.ly/v2/plots"
+    url_v2_grid = "https://api.plot.ly/v2/grids"
```
theengineear: 💀 you need to get these urls from the `config` object. There should be examples of that somewhere. Here's one: https://github.com/plotly/plotly.py/blob/master/plotly/plotly/plotly.py#L685 (something like that should do).

In general, *never* hard-code the `plot.ly` domain into this code.

plotly/plotly/plotly.py
```diff
@@ -1454,6 +1453,129 @@ def _send_to_plotly(figure, **plot_options):
...
+    auth = HTTPBasicAuth(str(payload['un']), str(payload['key']))
+    headers = {'Plotly-Client-Platform': 'python'}
```
theengineear: I think the `_api_v2` thing handles headers as well: https://github.com/plotly/plotly.py/blob/master/plotly/plotly/plotly.py#L674

plotly/plotly/plotly.py
```diff
@@ -1454,6 +1453,129 @@ def _send_to_plotly(figure, **plot_options):
...
+
+    # add layout if not in fig
+    if 'layout' not in fig:
+        fig['layout'] = {}
```
theengineear: Unnecessary comment and this can be a clean one-liner as you've done before.

plotly/plotly/plotly.py
```diff
@@ -1454,6 +1453,129 @@ def _send_to_plotly(figure, **plot_options):
...
+
+    # make a copy of fig
+    fig_with_uids = copy.deepcopy(fig)
```
theengineear: Another unnecessary comment. Just do this and the above layout thing in two lines.

plotly/plotly/plotly.py
```diff
@@ -1454,6 +1453,129 @@ def _send_to_plotly(figure, **plot_options):
...
+
+    # make grid
```
theengineear: This, on the other hand, is a nice little comment, imo.

plotly/plotly/plotly.py
```diff
@@ -1454,6 +1453,129 @@ def _send_to_plotly(figure, **plot_options):
...
+    frames_cols_dict = {}
+    counter = 0
+    trace_num = 0
+    for trace in fig['data']:
+        for var in ['x', 'y']:
+            if 'name' in trace:
+                cols_dict["{name}, {x_or_y}".format(name=trace['name'],
+                                                    x_or_y=var)] = {
+                    "data": list(trace[var]), "order": counter
+                }
```
theengineear: 🐄  if it's going to take up 4 lines anyhow, I'd suggest something like?

```
key = '{}, {}'.format(trace['name'], var)
data = list(trace[var])
cols[key] = {'data': data, 'order': counter}
```

plotly/plotly/plotly.py
```diff
@@ -1457,6 +1448,76 @@ def _send_to_plotly(figure, **plot_options):
     return r
 
 
+def create_animations(fig, kwargs, payload):
```
theengineear: 🚨 💀 😨 I think you should *not* do figure splitting (i.e., exchanging raw data for referenced data). Please take my word on it. It's not worth it. It's *extremely* complicated (Please take my word on it).

^^ Seriously, take my word on it.

Users should be able to do something like:

```
grid = Grid()
grid.add_column('time_data', [1, 2, 3, 4])  # something like this...
grid.add_column('voltage', [120, 119, 122, 121])  # something like this...
# somehow, someway, add data to this grid and make it easy for users to pull out the reference.
# try looking in https://github.com/plotly/plotly.py/blob/master/plotly/grid_objs/grid_objs.py

figure.data[0].x = grid.time_data
figure.frames[0]['data'][0] = grid.time_data  # Not sure if this is right, just for example

create_animations(figure, ...)
```

AFAIK, you don't need to worry *too* much about checking for raw data in the figure. The api will catch you and the request will just fail.

Let's fully shift the responsibility to the user. If we get into the game of guessing, we'll just end up guessing wrong and creating bugs.

plotly/graph_reference.py
Kully: No, I added it in. I don't know if it's actually being used though.

plotly/plotly/plotly.py
Kully: This has been an empty comment for apparently one year. Should I attempt writing something in it and get it corrected?

theengineear: oop, sorry. Weird. I just saw the change and questioned it. sure 🔪  it :)

Kully:
Here's the gameplan. I'll remake the `create_animations` function to be less 💀 . 😜 

According to @etpinard, animations aren't supported in `shareplot`, `embed` yet.

@theengineear, would you know anything about the .json endpoint? This seems like a later-down-the-line task but right now, `frames` is not throwing an error (because of your workaround) but it may be getting lost somewhere in the `streambed` code.

So a recap:
1) Rewrite `create_animations` to put more responsibility on the user
2) Wait for `frames` to be supported in `embed`, `shareplot`
3) ensure that .json is being updated properly [here](https://github.com/plotly/streambed/blob/master/shelly/code_translator/translator.py#L304-L318)

Please advise different approaches if you think of any

plotly/offline/offline.py
```diff
@@ -163,6 +163,7 @@ def init_notebook_mode(connected=False):
 def _plot_html(figure_or_data, config, validate, default_width,
                default_height, global_requirejs):
     # force no validation if frames is in the call
+    # TODO - add validation for frames in call - #
```
theengineear: Thanks for linking issues :heart:! Looks like this one may have been missed though.

Kully: LOL, whoops! That's https://github.com/plotly/plotly.py/issues/605

plotly/plotly/plotly.py
```diff
@@ -1576,6 +1575,39 @@ def bad_create_animations(fig, kwargs, payload):
     return r_dict
 
 
+def get_uid_by_col_name(grid_url, col_name):
```
theengineear: The `grid_url` arg should be a little more flexible. Look at  https://github.com/plotly/plotly.py/blob/master/plotly/plotly/plotly.py#L318. I'm OK with this as is. But I could see it being frustrating to be able to reference files one way in one place and another way in another place.

theengineear: Also. I sort of imagined that this would have been part of our `Grid` class? More like:

```
1. Get `Grid` from `grid_url` (symmetric to `get_figure`), this is a network operation
2. Get `uid` from `Grid` (which is a simple, local operation)
```

plotly/offline/offline.py
```diff
@@ -163,7 +163,7 @@ def init_notebook_mode(connected=False):
 def _plot_html(figure_or_data, config, validate, default_width,
                default_height, global_requirejs):
     # force no validation if frames is in the call
-    # TODO - add validation for frames in call - #
+    # TODO - add validation for frames in call - #605
```
Kully: Fixed now 👍

plotly/plotly/plotly.py
```diff
@@ -33,6 +33,7 @@
 from plotly.session import (sign_in, update_session_plot_options,
                             get_session_plot_options, get_session_credentials,
                             get_session_config)
+from plotly.grid_objs import grid_objs
```
Kully: This import is what I was asking you about. I don't know if `plotly.plotly` can depend on the `grid_objs` module, but that's the direction I took for getting `get_grid` to output an actual GRID.

plotly/plotly/plotly.py
```diff
@@ -1452,160 +1454,44 @@ def _send_to_plotly(figure, **plot_options):
...
+    else:
+        fid_in_url = grid_url[tilda_index + 1: len(grid_url)]
+        fid_in_url = fid_in_url.replace('/', ':')
+    get_url = upload_url + '/' + fid_in_url + '/content'
+
+    r = requests.get(get_url, headers=headers)
+    json_res = json.loads(r.text)
+    if raw is False:
+        # create a new grid
+        new_grid = grid_objs.Grid(
```
Kully: I am a little worried about this one. It's not nearly as bad as the figure splitting I was doing before 💀  but do you think this is an okay approach?

plotly/plotly/plotly.py
Kully: What do you mean by cyclic import? I'm honestly not too concerned about having it in there at all and just keeping it as a standard import, just wondering if folks didn't want that module to be relying on `grid_objs`

theengineear: http://stackoverflow.com/questions/744373/circular-or-cyclic-imports-in-python

plotly/plotly/plotly.py
```diff
@@ -1454,6 +1454,46 @@ def _send_to_plotly(figure, **plot_options):
...
+               'plotly-platform': 'python'}
+    upload_url = _api_v2.api_url('grids')
+
+    tilda_index = grid_url.find('~')
+    if grid_url[-1] == '/':
+        fid_in_url = grid_url[tilda_index + 1: -1].replace('/', ':')
+    else:
+        fid_in_url = grid_url[tilda_index + 1: len(grid_url)]
+        fid_in_url = fid_in_url.replace('/', ':')
+    get_url = upload_url + '/' + fid_in_url + '/content'
```
Kully: What tools are you referring to? Should I be copying how `get_figure` does it?

plotly/plotly/plotly.py
```diff
@@ -1454,6 +1454,46 @@ def _send_to_plotly(figure, **plot_options):
...
+
+    r = requests.get(get_url, headers=headers)
+    json_res = json.loads(r.text)
+    if raw is False:
+        # create a new grid
+        new_grid = grid_objs.Grid(
+            [grid_objs.Column(json_res['cols'][column]['data'], column)
+             for column in json_res['cols']]
```
Kully: :+1:

plotly/plotly/plotly.py
theengineear: https://docs.python.org/2/library/urlparse.html#module-urlparse

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,63 @@ class Grid(MutableSequence):
...
         ```
+        Example from json grid
+        ```
+        grid_json = {
+            'cols': {
+                'time': {'data': [1, 2, 3], 'order': 0, 'uid': '4cd7fc'},
+                'voltage': {'data': [4, 2, 5], 'order': 1, 'uid': u'2744be'}
+            }
+        }
+        grid = Grid(grid_json)
```
theengineear: 💥 🎎 :)

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,63 @@ class Grid(MutableSequence):
...
+            column_names = [column.name for column in columns_or_json]
+            duplicate_name = utils.get_first_duplicate(column_names)
+            if duplicate_name:
+                err = exceptions.NON_UNIQUE_COLUMN_MESSAGE.format(duplicate_name)
+                raise exceptions.InputError(err)
+
+            self._columns = list(columns_or_json)
+            self.id = ''
+        elif isinstance(columns_or_json, dict):
+            # check if 'cols' is only root key
```
theengineear: 'meta' is a valid key! is there a reason you're only allowing 'cols'? I believe `cols` and `meta` are the only two though.

theengineear: 'meta' is a valid key! is there a reason you're only allowing 'cols'? I believe `cols` and `meta` are the only two though.

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,63 @@ class Grid(MutableSequence):
...
+            # check if 'data', 'order' and 'uid' are not in columns
+            grid_col_keys = ['data', 'order', 'uid']
+
+            for column_name in columns_or_json['cols']:
+                for key in grid_col_keys:
+                    if key not in columns_or_json['cols'][column_name]:
+                        raise exceptions.PlotlyError(
+                            "Each column name of your dictionary must have "
+                            "'data', 'order' and 'uid' as keys."
+                        )
```
theengineear: 👍 cool.

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,63 @@ class Grid(MutableSequence):
...
+
+            self._columns = [Column(columns_or_json['cols'][column_name]['data'], column_name)
+                             for column_name in columns_or_json['cols']]
```
theengineear: 💀 doesn't this destroy column order? that will break 2d references to the grid.

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,63 @@ class Grid(MutableSequence):
...
+            # fill in uids
+            for column in self:
+                column.id = columns_or_json['cols'][column.name]['uid']
```
theengineear: I'm just not familiar, `column.id` is supposed to be the `uid` here, right? I know that we often refer to a column id as the full reference, so `fid:uid`, instead of just `uid`. I trust you, but I just want to double-check ;)

plotly/plotly/plotly.py
```diff
@@ -1470,26 +1470,17 @@ def get_grid(grid_url, raw=False):
...
-        # create a new grid
-        new_grid = grid_objs.Grid(
-            [grid_objs.Column(json_res['cols'][column]['data'], column)
-             for column in json_res['cols']]
-        )
-        # fill in uids
-        for column in new_grid:
-            column.id = json_res['cols'][column.name]['uid']
-        return new_grid
+        return Grid(json_res)
```
theengineear: \o/

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,63 @@ class Grid(MutableSequence):
...
+        if isinstance(columns_or_json, (list, tuple)):
+            column_names = [column.name for column in columns_or_json]
+            duplicate_name = utils.get_first_duplicate(column_names)
+            if duplicate_name:
+                err = exceptions.NON_UNIQUE_COLUMN_MESSAGE.format(duplicate_name)
+                raise exceptions.InputError(err)
+
+            self._columns = list(columns_or_json)
+            self.id = ''
+        elif isinstance(columns_or_json, dict):
```
Kully: I like it. Makes more sense in that order.

plotly/grid_objs/grid_objs.py
Kully: I think you may be right. Investigating now. I suppose it would be better to stick to the same order as the original grid, even if this functionality would be just for pulling out the `uid`s (which shouldn't fail)

plotly/grid_objs/grid_objs.py
```diff
@@ -143,20 +143,11 @@ def __init__(self, columns_or_json):
...
-            self._columns = list(columns_or_json)
-            self.id = ''
-        elif isinstance(columns_or_json, dict):
-            # check if 'cols' is only root key
+        if isinstance(columns_or_json, dict):
+            # check if 'cols' is a root key
             if 'cols' not in columns_or_json:
                 raise exceptions.PlotlyError(
-                    "'cols' must be the one and only key in your json grid."
+                    "'cols' must be a root key in your json grid."
```
theengineear: 👍

plotly/grid_objs/grid_objs.py
```diff
@@ -169,12 +160,31 @@ def __init__(self, columns_or_json):
                             "Each column name of your dictionary must have "
                             "'data', 'order' and 'uid' as keys."
                         )
+            # order columns in a list before putting inside the grid
+            ordered_columns = []
+            for order in range(len(columns_or_json['cols'])):
```
theengineear: Ha. So close! Don't depend on order being continuous. The orders might be `2, 5, 6, 19` (AFAIK). We don't depend on that elsewhere. For example, in the front end of the Plotly app, we have a workaround for this:

https://github.com/plotly/streambed/blob/master/shelly/filewell/static/filewell/src/actions/WorkspaceActions.js#L1575-L1579

Can we do something like that here (no opinion on implementation)?

plotly/grid_objs/grid_objs.py
```diff
@@ -169,12 +160,31 @@ def __init__(self, columns_or_json):
...
-            self._columns = [Column(columns_or_json['cols'][column_name]['data'], column_name)
-                             for column_name in columns_or_json['cols']]
             # fill in uids
             for column in self:
                 column.id = columns_or_json['cols'][column.name]['uid']
+        else:
+            column_names = [column.name for column in columns_or_json]
+            duplicate_name = utils.get_first_duplicate(column_names)
+            if duplicate_name:
+                err = exceptions.NON_UNIQUE_COLUMN_MESSAGE.format(duplicate_name)
```
theengineear: 👍

plotly/grid_objs/grid_objs.py
```diff
@@ -160,9 +160,16 @@ def __init__(self, columns_or_json):
                             "Each column name of your dictionary must have "
                             "'data', 'order' and 'uid' as keys."
                         )
-            # order columns in a list before putting inside the grid
+            # collect and sort all orders in case orders do not start
+            # at zero or there are jump discontinuities between them
+            all_orders = []
```
Kully: @theengineear This method works. :+1:

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,80 @@ class Grid(MutableSequence):
...
+                for key in grid_col_keys:
+                    if key not in columns_or_json['cols'][column_name]:
+                        raise exceptions.PlotlyError(
+                            "Each column name of your dictionary must have "
+                            "'data', 'order' and 'uid' as keys."
+                        )
+            # collect and sort all orders in case orders do not start
+            # at zero or there are jump discontinuities between them
+            all_orders = []
+            ordered_columns = []
```
theengineear: 🐄 non-blocking, but in the future (for readability) it helps to define things like `ordered_columns` *right* before they get used (so before the next loop).

Kully: Okay, sounds good. I'll fix that.

chriddyp:
FWIW the `.json` endpoint on plot.ly actually uses _this_ package for validation. This PR should prevent stripping `frames` from a figure JSON and once this is merged we'll be able to start seeing `frames` in the `.json` endpoint. For now, you should be able to see the JSON of the chart by visiting the API at e.g. https://api.plot.ly/v2/plots/chris:17779/content#response (replace `chris:17779` with your username and ID))

mdtusz:
Any progress on this one? Somewhat 🚫 for me.

Kully:
> Any progress on this one? Somewhat 🚫 for me.

Yup. Trying to wrap up now...

The final goal is going to be adding a `get_fid_uid()` type method to grids so that users can easily fill in their `xsrc` and `ysrc` in `data`, `frames`.

chriddyp:
@mdtusz - If validation is the only thing that's blocking, I'm _think_ that it has been implemented in this branch already. You could work off of this branch in your environment with something like:
```
$ pip install git+https://github.com/plotly/plotly.py.git@animations-for-iqt2
```

Kully:
> I'm think that it has been implemented in this branch already

And it has!

Kully:
@chriddyp @jackparmer @mdtusz 

Okay, everything should be ready. Just need a review:

Firstly, this is how a user can create a grid, retrieve the `fid:uid` combination and fill in a `payload` variable to make a POST to create the plot:

```
import plotly.plotly as py
from plotly.grid_objs import grid_objs

column_1 = grid_objs.Column([3, 1, 4], 'apple')
column_2 = grid_objs.Column([1, 5, 9], 'pies')
column_3 = grid_objs.Column([1, 1, 1], 'for')
column_4 = grid_objs.Column([6, 6, 6], 'sale')
grid = grid_objs.Grid([column_1, column_2, column_3, column_4])

py.grid_ops.upload(grid, 'bake_sale', auto_open=False)
```

then grab a the `fid:uid` by column name using `grid.get_fid_uid('apple')` for instance. Last step is to create the plot:

```
import json
import requests
from requests.auth import HTTPBasicAuth
 
auth = HTTPBasicAuth('AdamKulidjian', 'nlpvm9aack')
headers = {'Plotly-Client-Platform': 'python'}
 
payload = {
    'figure': {
        'data': [{'xsrc': grid.get_fid_uid('apple'), 'ysrc': grid.get_fid_uid('pies')}],
        'layout': {'title': 'Apples are Cool.'},
        'frames': [ {'data': [{'xsrc': grid.get_fid_uid('for'), 'ysrc': grid.get_fid_uid('sale')}],
                     'layout': {'title': 'They Are Nice'}}]
    },
    'world_readable': 'true'
}

r = requests.post('https://api.plot.ly/v2/plots', auth=auth, headers=headers, json=payload)
```

You can find the result at the .json endpoint [here](https://api.plot.ly/v2/plots/AdamKulidjian:1881/content#response)

A few notes about the PR:
1) It doesn't seem mandatory at this point, but I had written a `py.get_grid()` function which allows a user to grab a preexisting grid from the cloud in either JSON form or Grid (cls) form. It's handy if the grids have already been created and we need to plot from them.
2) Vis a Vis the functionality in 1), users actually have the ability to input a `grid_id` as a param into `Grid()`. Instead of setting the grid and their column id's to empty strings, if a string is entered then it will set that to the current id. Is there a way to make this variable unreachable by the user?

chriddyp:
Re 
> Last step is to create the plot:
    ```
    import json
    import requests
    from requests.auth import HTTPBasicAuth
     
    auth = HTTPBasicAuth('AdamKulidjian', 'nlpvm9aack')
    headers = {'Plotly-Client-Platform': 'python'}
     
    payload = {
        'figure': {
            'data': [{'xsrc': grid.get_fid_uid('apple'), 'ysrc': grid.get_fid_uid('pies')}],
            'layout': {'title': 'Apples are Cool.'},
            'frames': [ {'data': [{'xsrc': grid.get_fid_uid('for'), 'ysrc': grid.get_fid_uid('sale')}],
                         'layout': {'title': 'They Are Nice'}}]
        },
        'world_readable': 'true'
    }

    r = requests.post('https://api.plot.ly/v2/plots', auth=auth, headers=headers, json=payload)
    ```

Does this PR also include a beta function that wraps this up nicely? From the previous discussion,  I proposed something like:

> Create an animation on the server: `py.create_animation(figure, filename, sharing)`

Kully:
> Does this PR also include a beta function that wraps this up nicely?

Nope, but I'll write one up now!

plotly/plotly/plotly.py
```diff
@@ -1493,6 +1493,42 @@ def get_grid(grid_url, raw=False):
...
+    """
+    credentials = get_credentials()
+    validate_credentials(credentials)
+    username, api_key = credentials['username'], credentials['api_key']
+    auth = HTTPBasicAuth(str(username), str(api_key))
+    headers = {'Plotly-Client-Platform': 'python'}
+
+    json = {
+        'figure': figure,
+        'world_readable': 'true'
```
chriddyp: If we're JSON-ifying everything correctly in the request, then this can simply be `True`. You may need to add `'content-type': 'application/json'` to the headers of the request

plotly/plotly/plotly.py
```diff
@@ -1493,6 +1493,42 @@ def get_grid(grid_url, raw=False):
...
+    # set sharing
+    if sharing == 'public':
+        json['world_readable'] = 'true'
+    elif sharing == 'private':
+        json['world_readable'] = 'false'
+
+    api_url = _api_v2.api_url('plots')
+    r = requests.post(api_url, auth=auth, headers=headers, json=json)
+
+    json_r = json.loads(r.text)
```
chriddyp: this can fail if the response wasn't JSON. we might want to do
```
try:
    parsed_response = r.json()
except:
    parsed_response = r.content
```

plotly/plotly/plotly.py
```diff
@@ -1493,6 +1493,42 @@ def get_grid(grid_url, raw=False):
...
+        json['filename'] = filename
+
+    # set sharing
+    if sharing == 'public':
+        json['world_readable'] = 'true'
+    elif sharing == 'private':
+        json['world_readable'] = 'false'
+
+    api_url = _api_v2.api_url('plots')
+    r = requests.post(api_url, auth=auth, headers=headers, json=json)
```
chriddyp: What happens when the request fails? It can fail for a number of reasons:
- API key wasn't supplied
- API key wasn't correct
- User was offline
- Request was throttled
- Figure was malformed
- Duplicate filename
- Permission denied (can't save the file as `private`)
- Server was down
- And more

Errors can be detected by checking for non-`200` level status codes or by calling `r.raise_for_status()`. Plotly returns the error in a standard format (something like `{error: {message: 'error message'}}` but I don't recall exactly) and we should parse the error message and then rethrow an error with that message for the user.

All of those errors should have a similar error response but you should try to hit those errors manually yourself to verify that the error handling works correctly and is helpful. Bonus points for including all of those types of errors inside actual tests so that we're sure that our error handling doesn't change without us knowing.

plotly/plotly/plotly.py
```diff
@@ -1493,6 +1493,42 @@ def get_grid(grid_url, raw=False):
...
+    headers = {'Plotly-Client-Platform': 'python'}
+
+    json = {
+        'figure': figure,
+        'world_readable': 'true'
+    }
+
+    # set filename if specified
+    if filename:
+        json['filename'] = filename
```
chriddyp: In `py.plot`, if the user specifies a `/` in the filename then we'll automatically make folders for them. That won't happen here. We might want to check for `/` in the filename and then issue a warning to the user that this beta function won't create folders automatically for them.

plotly/plotly/plotly.py
```diff
@@ -1493,6 +1493,42 @@ def get_grid(grid_url, raw=False):
...
+
+    # set sharing
+    if sharing == 'public':
+        json['world_readable'] = 'true'
+    elif sharing == 'private':
+        json['world_readable'] = 'false'
```
chriddyp: we should add `'secret'` in here. we should also throw an error if `sharing` isn't one of `public`, `private`, or `secret`

plotly/plotly/plotly.py
```diff
@@ -1493,6 +1493,42 @@ def get_grid(grid_url, raw=False):
         return json_res
 
 
+def create_animations(figure, filename=None, sharing='public'):
+    """
+    Put description here.
```
chriddyp: ⚡️

plotly/plotly/plotly.py
```diff
@@ -1493,6 +1493,42 @@ def get_grid(grid_url, raw=False):
...
+
+    For parameter descriptions, see the doc string for `py.plot()`.
```
chriddyp: instead of `py.plot`, `plotly.plotly.plot` since `py` is just an alias.

plotly/plotly/plotly.py
```diff
@@ -1493,6 +1493,42 @@ def get_grid(grid_url, raw=False):
...
+    `private` is not supported currently for param 'sharing'. Returns the
+    url for the plot if response is ok.
+    """
```
chriddyp: We should include some basic examples of animations in here.

plotly/plotly/plotly.py
```diff
@@ -1493,6 +1493,42 @@ def get_grid(grid_url, raw=False):
         return json_res
 
 
+def create_animations(figure, filename=None, sharing='public'):
+    """
```
chriddyp: We should tell the user that this is a `beta` endpoint that is subject to deprecation in the future.

chriddyp: And we should describe how this is similar and dissimilar to `plotly.plotly.plot`. Includes things like:
- Folders aren't supported
- Overwrite isn't supported
- Animations via frames are supported
- There might be some other things that I'm forgetting about.

plotly/plotly/plotly.py
```diff
@@ -1493,6 +1493,42 @@ def get_grid(grid_url, raw=False):
         return json_res
 
 
+def create_animations(figure, filename=None, sharing='public'):
```
chriddyp: We might want to create an `i` version of this that works in ipython notebooks like `icreate_animations` or else add it as like an output method.

plotly/plotly/plotly.py
```diff
@@ -1449,6 +1454,81 @@ def _send_to_plotly(figure, **plot_options):
...
+    if sharing == 'public':
+        json['world_readable'] = 'true'
+    elif sharing == 'private':
+        json['world_readable'] = 'false'
+
+    api_url = _api_v2.api_url('plots')
+    r = requests.post(api_url, auth=auth, headers=headers, json=json)
+
+    json_r = json.loads(r.text)
+    return json_r['file']['web_url']
```
chriddyp: I forget if this URL includes the title text at the end of the URL (e.g. `https://plot.ly/~chris/15/my-title-of-my-graph`) or not. We should make sure that we're returning the minimal version of the url like `https://plot.ly/~chris/15` just like how `py.plot` does

plotly/plotly/plotly.py
```diff
@@ -1449,6 +1454,81 @@ def _send_to_plotly(figure, **plot_options):
...
+
+
```
chriddyp: We should add tests for this function including the possible types of errors that can occur

plotly/plotly/plotly.py
Kully: I think it _is_ minimal but will check

plotly/plotly/plotly.py
jackparmer: 👍

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,96 @@ class Grid(MutableSequence):
...
+            # check that grid_id is entered
+            if grid_id is None:
+                raise exceptions.PlotlyError(
+                    "If you are manually converting a raw json/dict grid "
+                    "into a Grid instance, you must ensure that make "
+                    "'grid_id' is set to your file ID. This looks like "
+                    "'username:187'."
+                )
+            # TODO: verify that grid_id is a correct fid if a string
+            self.id = grid_id
```
theengineear: For consistency, can we call this `self.fid`?

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,96 @@ class Grid(MutableSequence):
     py.plot([trace], filename='graph from grid')
     ```
     """
-    def __init__(self, iterable_of_columns):
+    def __init__(self, columns_or_json, grid_id=None):
```
theengineear: For consistency, can we call this `fid`?

plotly/grid_objs/grid_objs.py
```diff
@@ -187,3 +254,17 @@ def get_column(self, column_name):
         for column in self._columns:
             if column.name == column_name:
                 return column
+
+    def get_fid_uid(self, column_name):
```
theengineear: Why is it called `get_fid_uid`? Maybe just `get_uid`? Also, are users supposed to use this? They shouldn't need to know about `uid`s or `src`s. If it's something users are playing with, I'd suggest `get_column_reference` or something (which should return the `fid:uid` pair (and could potentially handle `2d` references).

plotly/grid_objs/grid_objs.py
```diff
@@ -187,3 +254,17 @@ def get_column(self, column_name):
...
+
+        Returns an empty string if either the column name does not exist in
+        the grid or if the id of the specified column has the empty string id.
+        """
+        uid = ''
+        for column in self._columns:
+            if column.name == column_name:
+                uid = column.id
+                break
+        return uid
```
theengineear: Should this raise if it can't find a uid? Is there a case where we would want to accept the empty string if a column uid cannot be found?

plotly/plotly/plotly.py
```diff
@@ -1449,6 +1454,194 @@ def _send_to_plotly(figure, **plot_options):
     return r
 
 
+def get_grid(grid_url, raw=False):
+    """
+    Returns a JSON figure representation for the specified grid.
```
theengineear: 🎎 This feels a little misleading. Aren't we returning either JSON or a custom object. And, in the JSON case, it's a dict, right.

plotly/plotly/plotly.py
```diff
@@ -1449,6 +1454,194 @@ def _send_to_plotly(figure, **plot_options):
...
+    being retrieved. If True, raw JSON will be returned.
+    """
+    credentials = get_credentials()
+    validate_credentials(credentials)
+    username, api_key = credentials['username'], credentials['api_key']
+    headers = {'plotly-username': username,
+               'plotly-apikey': api_key,
+               'plotly-version': version.__version__,
+               'plotly-platform': 'python'}
+    upload_url = _api_v2.api_url('grids')
```
theengineear: Did the `_api_v2` thingy not have a `get_headers` method?

plotly/plotly/plotly.py
```diff
@@ -1449,6 +1454,194 @@ def _send_to_plotly(figure, **plot_options):
...
+def create_animations(figure, filename=None, sharing='public', auto_open=True):
+    """
+    BETA function that creates plots with animations via `frames`.
+
+    Creates an animated plot using 'frames' alongside 'data' and 'layout'.
+    This BETA endpoint is subject to deprecation in the future. In relation
+    to `plotly.plotly.plot`, folder-creation and overwriting are not supported
+    but creating a plot with or without animations via frames is supported.
+
+    :param (str) filename: if set to 'None', an automatically generated plot
```
theengineear: :goat: `automatically-generated` I think.

plotly/plotly/plotly.py
```diff
@@ -1449,6 +1454,194 @@ def _send_to_plotly(figure, **plot_options):
...
+
+    py.create_animations(figure, 'new_plot_with_animations')
+    ```
+    """
+    credentials = get_credentials()
+    validate_credentials(credentials)
+    username, api_key = credentials['username'], credentials['api_key']
+    auth = HTTPBasicAuth(str(username), str(api_key))
+    headers = {'Plotly-Client-Platform': 'python',
+               'content-type': 'application/json'}
```
theengineear: Again, I think we should have a `get_headers` function or something to reuse code a bit here.

plotly/plotly/plotly.py
```diff
@@ -1449,6 +1454,194 @@ def _send_to_plotly(figure, **plot_options):
...
+
+    json = {
+        'figure': figure,
+        'world_readable': True
```
theengineear: Hm. @chriddyp not sure if you weighed in on this, but we typically default to `world_readable: false`, right?

chriddyp: I think that `py.plot` actually defaults to `True`. Moving forward, defaulting to `False` would probably be the right move.

chriddyp: It's 1 for Community users (similar to GitHub) - See https://plot.ly/products/cloud/

plotly/grid_objs/grid_objs.py
Kully: Yes, users are supposed to use this. And it _does_ return the `fid:uid` pair

plotly/grid_objs/grid_objs.py
Kully: You're right, it should. It's possible for the column_reference to be the empty string `''` if a grid hasn't been uploaded yet, for instance.

plotly/plotly/plotly.py
Kully: Correct again! A oversight on my part.

plotly/grid_objs/grid_objs.py
Kully: grids are instantiated with `id`s so I don't think that's a risk here. For now, I think it's fine to have it just return the `id` of that column, whatever it happens to be.

plotly/plotly/plotly.py
Kully: There's this `_api_v2.headers()` which returns for me

```
{'authorization': u'Basic QWRhbUt1bGlkamlhbjpubHB2bTlhYWNr', 'plotly-client-platform': 'python 1.12.10'}
```

The thing is, usually I'm adding more things to `headers` so it wouldn't exactly make it more DRY ya know? @theengineear

plotly/plotly/plotly.py
Kully: I don't think it makes sense to make set `sharing` to `private` as a default if the Community users get only 1. I think that may be confusing and more work if they have to set it to `public` each time they make a plot.

Kully: An error message for `private` makes sense

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,96 @@ class Grid(MutableSequence):
...
+        ```
         """
-
         # TODO: verify that columns are actually columns
+        if isinstance(columns_or_json, dict):
+            # check that fid is entered
+            if fid is None:
+                raise exceptions.PlotlyError(
+                    "If you are manually converting a raw json/dict grid "
+                    "into a Grid instance, you must ensure that make "
```
theengineear: 🐐 `ensure that make`

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,96 @@ class Grid(MutableSequence):
...
+                    "'fid' is set to your file ID. This looks like "
+                    "'username:187'."
+                )
+            # TODO: verify that fid is a correct fid if a string
```
theengineear: You mean validate it? It'd be a whole extra api call to actually check. I think you don't need to worry about it.

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,96 @@ class Grid(MutableSequence):
...
+            self.id = fid
```
theengineear: 🐄 non-blocking, but I'd go with `self.fid` here.

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,96 @@ class Grid(MutableSequence):
...
+                    if columns_or_json['cols'][column_name]['order'] == order:
+                        break
+
+                ordered_columns.append(Column(
+                    columns_or_json['cols'][column_name]['data'],
+                    column_name)
+                )
+            self._columns = ordered_columns
+
+            # fill in uids
```
theengineear: :dolls: `uids` --> `column_ids`.

plotly/grid_objs/grid_objs.py
```diff
@@ -118,29 +118,96 @@ class Grid(MutableSequence):
...
+            for column in self:
+                column.id = self.id + ':' + columns_or_json['cols'][column.name]['uid']
```
theengineear: 🐄 we typically use `format` i think. non-blocking

plotly/grid_objs/grid_objs.py
```diff
@@ -187,3 +254,27 @@ def get_column(self, column_name):
...
+    def get_column_reference(self, column_name):
+        """
+        Returns the column reference of given column in the grid by its name.
+
+        Raises an error if the column name is not in the grid. Otherwise,
+        returns the fid:uid pair, which may be the empty string.
+        """
+        col_names = []
+        for column in self._columns:
+            col_names.append(column.name)
```
theengineear: 🐄 this is confusing placement of this little code block. Can you move it into the `if column_id is None` block so that:

1. it only needs to be executed on failure
2. it's clear that it's only used in that error message.

plotly/grid_objs/grid_objs.py
```diff
@@ -187,3 +254,27 @@ def get_column(self, column_name):
...
+        for column in self._columns:
+            if column.name == column_name:
+                column_id = column.id
+                break
+
+        if column_id is None:
+            raise exceptions.PlotlyError(
+                "Whoops, that column name doesn't match any of the column "
+                "names in your grid. You must pick from {cols}".format(col_names)
+            )
```
theengineear: 👍

plotly/plotly/plotly.py
theengineear: But you could improve the `_api_v2.headers()` method to allow for more things! 🐫! Centralizing functionality like this is almost always a good thing!

plotly/plotly/plotly.py
```diff
@@ -1449,6 +1454,195 @@ def _send_to_plotly(figure, **plot_options):
...
+            "or 'secret'."
+        )
+
+    api_url = _api_v2.api_url('plots')
+    r = requests.post(api_url, auth=auth, headers=headers, json=json)
+    r.raise_for_status()
+
+    try:
+        parsed_response = r.json()
+    except:
```
theengineear: Any way to narrow this exception clause? It's too broad. If you just copied this code, I'm OK with leaving it as-is. Just sets off 🚨 when I see it.

plotly/plotly/plotly.py
```diff
@@ -1449,6 +1454,195 @@ def _send_to_plotly(figure, **plot_options):
...
+        parsed_response = r.content
+
+    if 'error' in r and r['error'] != '':
+        raise exceptions.PlotlyError(r['message'])
```
theengineear: huh? Wouldn't this want to print out `r['error']` If that's not the case can you add a quick code comment? This is a little confusing to me.

plotly/plotly/plotly.py
```diff
@@ -1449,6 +1454,195 @@ def _send_to_plotly(figure, **plot_options):
...
+
+
+def icreate_animations(figure, filename=None, sharing='public', auto_open=False):
+    """
+    Create a unique url for this animated plot in Plotly and open in IPython.
+
+    This function is based off `plotly.plotly.iplot`. See `plotly.plotly.
+    create_animations` Doc String for param descriptions.
+    """
+    # TODO - create a wrapper for iplot and icreate_animations
```
theengineear: Still need doing?

plotly/grid_objs/grid_objs.py
Kully: Actually, if you call `grid.id` you'll get the `fid` back. That was in place before I started working on this PR. so id === fid for the grid.

PR #586, opened 2016-10-20
Patch `Figure` to allow `frames`.

theengineear:
We need frames for animations, this is a hack since they don’t exist in
the graph reference.

### DIFF
--- a/plotly/graph_objs/graph_objs.py
+++ b/plotly/graph_objs/graph_objs.py
@@ -836,6 +836,20 @@ def _patch_figure_class(figure_class):
                                                   _parent_key='data')
     figure_class.__init__ = __init__
 
+    def __setitem__(self, key, value, **kwargs):
+        if key == 'frames':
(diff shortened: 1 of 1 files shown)


theengineear:
@Kully merge this into yr branch if you want :) I didn't want to commit anything directly to yr branch just so you have full control of this.

Can you test this out and see if it will work for you? If you just want to mess around with it locally without merging, you can cherry-pick it like so:

```
git fetch origin
git cherry-pick d53c7ef715b1f0390435ad22147ae49a64d2e9cb
git reset HEAD^
``` 

That does the following:

1. fetch the new things from this repo
2. add this commit to your PR locally
3. move your HEAD from that commit to the one before, moving those changes into your staged files

Be careful about `git reset HEAD^`, that undoes *one* (`^`) commit from the current commit yr pointing to in git.

Feel free to copy/paste if that feel simpler :)

Kully:
Closing since the added code lines here have been manually entered into the referenced PR #584

Issue #593, opened 2016-10-25
Slow trisurf + Bug in displaying with iplot


choldgraf:
Two quick things:

1. Something in the `create_trisurf` code has changed, and now the code takes an order of magnitude longer to run. (now at about 45 seconds when it used to take 3-4 seconds). I'm not sure if somebody un-did the optimizations that I added a few months back, or if something new has been added that's slow, but just FYI.
2. There seems to be a bug in the `iplot` function for offline plotting. E.g.:

![image](https://cloud.githubusercontent.com/assets/1839645/19688084/18f21774-9a96-11e6-8bc6-dcc263f428f6.png)

choldgraf:
it looks like it this is because of the `colors.py` module, which loops through triplets of colots (in a for loop) and multiplies each one by 255. This is the same issue that I addressed in the previous PR optimizing the trisurf code.

Looping through a list of triplets and individually multiplying each number by a scalar is a *really* inefficient way of doing this, and makes the code unusable for any decent-sized amount of data.

jackparmer:
Hey @choldgraf I'm happy to review and merge if you want to make a quick PR. Thanks for the heads-up on this

choldgraf:
I can make a quick PR on this one, but to be honest I'm a little hesitant to do this, since I spent a while improving the `map_face2color` and `trisurf` code, and I opened this issue because somebody basically undid all of that...

jackparmer:
Oh, yikes, sorry about that! Would the PR be essentially the same changes? If so we can look at those commits and try to cherrypick them back. I don't remember why that work was undone, I'll try to be more vigilant this time around!

choldgraf:
No, I think there was some refactoring that happened, but as a part of the refactoring it reverted to the old way of doing the multiplication. IIRC we had a back and forth on another PR and some folks seemed like they were trying to minimize the use of numpy to reduce dependencies. IMO I think Numpy should be treated as a core part of Python in this field, so I don't see it as a dependency, but some folks disagree and I think that's why the current state of code isn't vectorized w/ numpy (and thus takes like 10 times longer)

choldgraf:
maybe I will have some time to work on this over the weekend or next week,
if I do then I could make some additions, though I'd be curious what the
other python devs think since we've had versions of this conversation
already...
--

theengineear:
@choldgraf ugh. Sorry this happened. I wonder if there is a way we could add some tests to keep this locked down?

[...]

PR #599, opened 2016-11-01
Fix iplot offline

Kully:
Re iplot issue brought up in https://github.com/plotly/plotly.py/issues/593#issuecomment-257392993

Fixed by just removing the unnecessary `show_link` and `link_text` from the `_plot_html()` call in [this line](https://github.com/plotly/plotly.py/blob/master/plotly/offline/offline.py#L312).

`plot()` looks fine as it has `show_link` and `link_text` in the signature.

### DIFF
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -210,7 +210,6 @@ def plot(figure_or_data, validate=True, **plot_options):
 
     """
     figure = tools.return_figure_from_figure_or_data(figure_or_data, validate)
-
     for entry in figure['data']:
         if ('type' in entry) and (entry['type'] == 'scattergl'):
             continue
@@ -499,7 +498,7 @@ class Stream:
                 "cannot write to a closed connection. "
                 "Call `open()` on the stream to open the stream."
             )
-   
+
     @property
     def connected(self):
         if self._stream is None:
@@ -1400,10 +1399,14 @@ def add_share_key_to_url(plot_url, attempt=0):
 
 
 def _send_to_plotly(figure, **plot_options):
-    """
-
-    """
     fig = tools._replace_newline(figure)  # does not mutate figure
+
+    # replace '/' with ':' for errorless response
+    if 'data' in fig:
+        for item in fig['data']:
+            for key in ['xsrc', 'ysrc']:
+                item[key] = item[key].replace('/', ':')
+
     data = json.dumps(fig['data'] if 'data' in fig else [],
                       cls=utils.PlotlyJSONEncoder)
     credentials = get_credentials()
@@ -1414,8 +1417,7 @@ def _send_to_plotly(figure, **plot_options):
                              fileopt=plot_options['fileopt'],
                              world_readable=plot_options['world_readable'],
                              sharing=plot_options['sharing'],
-                             layout=fig['layout'] if 'layout' in fig
-                             else {}),
+                             layout=fig['layout'] if 'layout' in fig else {}),
                         cls=utils.PlotlyJSONEncoder)
 
     # TODO: It'd be cool to expose the platform for RaspPi and others
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -868,7 +868,7 @@ class grid_ops:
         for req_col in request_columns:
             for resp_col in response_columns:
                 if resp_col['name'] == req_col.name:
-                    req_col.id = '{0}/{1}'.format(grid_id, resp_col['uid'])
+                    req_col.id = '{0}:{1}'.format(grid_id, resp_col['uid'])
                     response_columns.remove(resp_col)
 
     @classmethod
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -1400,13 +1400,6 @@ def add_share_key_to_url(plot_url, attempt=0):
 
 def _send_to_plotly(figure, **plot_options):
     fig = tools._replace_newline(figure)  # does not mutate figure
-
-    # replace '/' with ':' for errorless response
-    #if 'data' in fig:
-    #    for item in fig['data']:
-    #        for key in ['xsrc', 'ysrc']:
-    #            item[key] = item[key].replace('/', ':')
-
     data = json.dumps(fig['data'] if 'data' in fig else [],
                       cls=utils.PlotlyJSONEncoder)
     credentials = get_credentials()
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -1402,10 +1402,10 @@ def _send_to_plotly(figure, **plot_options):
     fig = tools._replace_newline(figure)  # does not mutate figure
 
     # replace '/' with ':' for errorless response
-    if 'data' in fig:
-        for item in fig['data']:
-            for key in ['xsrc', 'ysrc']:
-                item[key] = item[key].replace('/', ':')
+    #if 'data' in fig:
+    #    for item in fig['data']:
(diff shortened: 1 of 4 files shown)


Kully:
Tests passed locally but getting strange encoding error on circle. It looks like its in `site-packages` which I think is core Python. Any thoughts @theengineear ?

cc: @choldgraf

Kully:
The last bit of the error message:

```
Cleaning up...
Command /home/ubuntu/.pyenv/versions/2.7.8/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip_build_ubuntu/pandas/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-DxAgzf-record/install-record.txt --single-version-externally-managed --compile failed with error code 1 in /tmp/pip_build_ubuntu/pandas
  File "/home/ubuntu/.pyenv/versions/2.7.8/bin/pip", line 11, in <module>
    sys.exit(main())
  File "/home/ubuntu/.pyenv/versions/2.7.8/lib/python2.7/site-packages/pip/__init__.py", line 185, in main
    return command.main(cmd_args)
  File "/home/ubuntu/.pyenv/versions/2.7.8/lib/python2.7/site-packages/pip/basecommand.py", line 161, in main
    text = '\n'.join(complete_log)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 72: ordinal not in range(128)
☁☀☂ setup.sh: ☁☀☂ 39: can't install optional for Python 2.7.8.


bash circle/setup.sh returned exit code 1

Action failed: bash circle/setup.sh
```

theengineear:
Hmm. I think the tests need some attention in general. I'll come back to this later in the day and try to fix up our test suite. That is an old `pip` error I've seen before (not super unicode-friendly). I *thought* it was related to http://stackoverflow.com/questions/25036897/pip-install-unicodedecodeerror, but i don't know why it would have been in a directory with a unicode-y name.

At any rate, I'll take a look later today.

theengineear:
Changes look good! I'll try to get a patch in to fix the test suite and we can get this in soon!

choldgraf:
hey all - shall I take a look at this now to see if it works?

Kully:
> hey all - shall I take a look at this now to see if it works?

Go ahead! I just rebuilt the test and it failed again.

choldgraf:
ok, will try to spend some time on this over the weekend
--

choldgraf:
I can confirm that iplot now works on this branch. Still slow as heck but it does plot :)

Kully:
> I can confirm that iplot now works on this branch. Still slow as heck but it does plot :)

Do you know which PR solved it? I was working on one but  the tests were not passing for some reason...

choldgraf:
I just pulled the code on this PR and tried it out

Kully:
Oh wait, I'm ridiculous. _This_ PR was the PR I was talking about.

Yeah, it works but the tests are still failing for some reason... 😕

Kully:
@choldgraf @theengineear Working on this now.

Finally getting test failures to occur locally - - helpful!

Kully:
Getting there. Cleared up this error which https://github.com/plotly/plotly.py/pull/614 was getting:

screen shot 2016-11-21 at 6 29 20 pm

Kully:
@chriddyp Can I get a 💃 ? Finally we have success.

Who else should I ping in the future? Is one reliable person enough? 😄

chriddyp:
> Who else should I ping in the future? 

Me or @theengineear !

Kully:
> Me or @theengineear !

K, sounds good. Let me know if you approve here.

plotly/plotly/plotly.py
```diff
@@ -869,7 +868,7 @@ def _fill_in_response_column_ids(cls, request_columns,
         for req_col in request_columns:
             for resp_col in response_columns:
                 if resp_col['name'] == req_col.name:
-                    req_col.id = '{0}/{1}'.format(grid_id, resp_col['uid'])
+                    req_col.id = '{0}:{1}'.format(grid_id, resp_col['uid'])
```
chriddyp: So did making plots based off of grid columns just not work before? Can you show me an example where this didn't work before but now, with this fix, it works?

plotly/tools.py
```diff
@@ -6208,6 +6207,9 @@ def create_dendrogram(X, orientation="bottom", labels=None,
         if len(s) != 2:
             exceptions.PlotlyError("X should be 2-dimensional array.")
 
+        if distfun is None:
+            distfun = scs.distance.pdist
+
```
chriddyp: what is the motivation behind this change? It seems like by removing it from the call signature it is less informative to users about what `distfun` is.

chriddyp:
Just left a couple of questions that I'd like to understand better before we merge. Otherwise looks good!

plotly/plotly/plotly.py
Kully: Yeah, it wasn't working until I made the switch. Here's one example that didn't work before the change that worked after:
```
c1 = Column([6, 6, 6, 5], 'pillow')
c2 = Column(['a', 'b', 'c', 'd'], 'couch2')
g = Grid([c1, c2])

py.grid_ops.upload(g, 'pillow_and_couch_grid222', auto_open=False)
url = py.plot([Scatter(xsrc=g[0], ysrc=g[1])], auto_open=False, filename='222pillow222')
```

I'm curious about something though...when plotly makes the request to V2 for the upload of the grid, prior to that, the local grid has blank `id`s (eg. `grid[0].id` returns `""` ) and after the request, your local grid has the proper filled in id eg. "AdamKulidjian:1782:u3fg1"

How is my local grid getting updated if only a url after `py.upload()` is run?

chriddyp: Got it. That's weird, because this definitely worked at some point. Here is the section and example in our docs: 
image

chriddyp: > How is my local grid getting updated if only a url after py.upload() is run?

`py.grid_ops.upload(g)` updates the columns in `g`. Those columns in `g` have the same reference as `c1` and `c2`, so changing them in `g` changes `c1` and c2` too. See http://nedbatchelder.com/text/names.html for some more info.

plotly/tools.py
chriddyp: Got it. should `scs` by `scp` in this case? Also, what happens if a user runs this function but they don't have `scipy` installed? It's not a required dependency of the `plotly` package (and it shouldn't be) but folks might be surprised if running the function fails due to a missing module. 

What do we do in other parts of the code that use `scipy`? Do we just let it fail or do we tell users that they need to install `scipy`?

Kully:
> If this is the case, then it seems like maybe the plotly backend API changed and stopped allowing IDs in this format.

That's what I suspected, but wasn't prepared to dig into streambed ;)

plotly/tools.py
chriddyp: Right. It's just that the error won't be "scipy is not defined" because it won't break on the `import scipy` step, it'll break on the `scs.distance.pdist` step which means that the error will probably be `scs is not defined` which isn't very intuitive.

Could you try creating a clean virtualenv without scipy and try calling this function and seeing what the error message is? Scipy may be common but Plotly may be the first package and experience that users have in Python and we want to make it as easy as possible for people to get going. We can fix the issue in a separate PR but it'd be nice to know how this will break for our users.

Kully: Okay sounds good.

chriddyp: Great, thanks for looking into that!

chriddyp:
OK, this looks good to me. 💃

PR #589, opened 2016-10-24
Add support for config options for offline plots

Rikorose:
fixes issue #399

You can now enter in a dictionary for all the options you want using
the parameter ‘config’.
(cherry picked from commit ecfe1fb)

PR #601, opened 2016-11-03
fixing config bug in #589

gte620v:
#589 changed the argument signature of `_plot_html`.  This PR fixes the two calls to that function.

See https://github.com/plotly/plotly.py/issues/399#issuecomment-258153104

### DIFF
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -308,8 +308,9 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
     if not tools._ipython_imported:
         raise ImportError('`iplot` can only run inside an IPython Notebook.')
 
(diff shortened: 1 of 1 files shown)


gte620v:
It looks like https://github.com/plotly/plotly.py/pull/410 might fix this as well.

Kully:
> It looks like #410 might fix this as well.

There are so many open PRs to this it looks like. I made this one https://github.com/plotly/plotly.py/pull/599 yesterday. The same as yours except for an additional PEP-8 thing (I indented the content of some tuple of `config_keys` and _didn't_ replace the `show_link` and `link_text` keys in the `plot` call.

gte620v:
@Kully Yeah, same idea.  Feel free to close this one if you have it under control.

Issue #613, opened 2016-11-21
Downloading malformed `plot-schema.json` can cause `import plotly` to break


chriddyp:
See https://github.com/plotly/plotly.py/issues/612 for the original issue that prompted this discussion. 

To make sure that this never happens again, we should either:

- Extend the try / except around the construction of `graph_objs` off of the downloaded `plot-schema.json` and use the cached version if it does not work
- Remove the auto-downloading `plot-schema.json` on import and perform more frequent (weekly) releases of the python library in sync with the plotly.js version updates.

etpinard:
The more radical solution is implemented in https://github.com/plotly/plotly.py/issues/613

rcscott:
I feel strongly that the auto-downloading should be removed. If someone wants the newest features then they can upgrade their plotly version. It's a strange design decision that the library can change automatically even if I have a specific version in my `requirements.txt`.

theengineear:
Removed in 1.13.0.

PR #614, opened 2016-11-21
Load graph reference from pkg copy of plot schema

etpinard:
fixes https://github.com/plotly/plotly.py/issues/613

### DIFF
--- a/plotly/graph_reference.py
+++ b/plotly/graph_reference.py
@@ -15,9 +15,6 @@ import six
 
 from plotly import files, utils
 
-GRAPH_REFERENCE_PATH = '/v2/plot-schema'
-GRAPH_REFERENCE_DOWNLOAD_TIMEOUT = 5  # seconds
-
 
 # For backwards compat, we keep this list of previously known objects.
 # Moving forward, we only add new trace names.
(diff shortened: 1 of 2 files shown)


etpinard:
### PROS

- `import plotly` will never break because of plot schema bugs e.g.https://github.com/plotly/plotly.js/pull/1182
- `import plotly` will never make a request to plot.ly 

### CONS

- `plotly.plotly` (i.e. users making graph via plot.ly) won't be relying in the _correct_ plot schema for key validation. For example, if we add an attribute to plotly.js and then to plot.ly, `plotly.plotly` users trying to user this attribute will get a validation error until we bump the [`default-schema.json`](https://github.com/plotly/plotly.py/blob/master/plotly/graph_reference/default-schema.json) during `plotly.offline` plotly.js version bumps.

I personally think the item under CONS isn't a big deal because:

- We should update the `plotly.offline` plotly.js frequently
- If a user _really_ want to use bleeding edge keys, they can always turn off validation

etpinard:
cc @theengineear @chriddyp @cldougl @Kully

theengineear:
@etpinard I'm shifting things around wrt our api for a large-ish release. I like the idea of no longer auto-downloading the plot-schema. We have a *very* simple way to update the plot schema and can just do that with minor version bumps if desired.

@chriddyp and @Kully, ya'll OK with stopping the auto-download of `plot-schema` as well?

chriddyp:
I'm OK with it!

theengineear:
The ideas from this have been implemented. Closing.

PR #615, opened 2016-11-23
bump version to 1.12.11

Kully:
@chriddyp or @theengineear


Kully:
@chriddyp The fix in the `CHANGELOG` about the `plotly.offline` thing, I was hesitant to put in cause it doesn't effectively change things for users, AFAIK. I included it to be thorough.

chriddyp:
💃

PR #623, opened 2016-12-01
hide tick labels on boxplots in scatterplotmatrix

Kully:


Kully:
-provided diag options in `scatterplotmatrix` doc string
-made function to hide tick labels in subplots that use box plots
-updated test in `test_figurefactory` to look at a chart output with `diag='box'`

@theengineear  Let me know if this looks good to you. 🐔  No rush.

plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -1141,29 +1141,93 @@ def test_scatter_plot_matrix(self):
...
+        #                           'domain': [0.55, 1.0], 'title': 'Fruit'},
+        #                'yaxis1': {'anchor': 'x1',
+        #                           'domain': [0.575, 1.0], 'title': 'Numbers'},
+        #                'yaxis2': {'anchor': 'x2',
+        #                           'domain': [0.575, 1.0]},
+        #                'yaxis3': {'anchor': 'x3',
+        #                           'domain': [0.0, 0.425], 'title': 'Fruit'},
+        #                'yaxis4': {'anchor': 'x4',
+        #                           'domain': [0.0, 0.425]}}
+        # }
```
theengineear: ⚡️ ?

plotly/tools.py
```diff
@@ -3881,6 +3881,9 @@ def _scatterplot(dataframe, headers, diag, size,
             title=title,
             showlegend=True
         )
+
+        fig = FigureFactory._hide_tick_labels_from_box_subplots(fig)
```
theengineear: Can you rename this to something more generic now? I.e., the `box` in there? Just looks funny to me.

theengineear: nvm. this is new! i'll comment down there :)

plotly/tools.py
```diff
@@ -4559,6 +4563,23 @@ def _scatterplot_theme(dataframe, headers, diag, size, height,
                     return fig
 
     @staticmethod
+    def _hide_tick_labels_from_box_subplots(fig):
+        """
+        Hides tick labels for box plots in scatterplotmatrix subplots.
```
theengineear: 👌

plotly/tools.py
```diff
@@ -4559,6 +4563,23 @@ def _scatterplot_theme(dataframe, headers, diag, size, height,
...
+        """
+        boxplot_xaxes = []
+        for j, plot_data in enumerate(fig['data']):
+            if plot_data['type'] == 'box':
+                boxplot_xaxes.append(
+                    'xaxis{}'.format(plot_data['xaxis'][-1])
```
theengineear: I'm not 100% convinced this is correct. Doesn't `x1` map to `xaxis` in layout? ORRRRR, is it because in plotly.py we don't use `xaxis`, instead using `xaxis1`? Mind throwing a code comment in here?

plotly/tools.py
```diff
@@ -4559,6 +4563,23 @@ def _scatterplot_theme(dataframe, headers, diag, size, height,
                     return fig
 
     @staticmethod
+    def _hide_tick_labels_from_box_subplots(fig):
+        """
+        Hides tick labels for box plots in scatterplotmatrix subplots.
+        """
+        boxplot_xaxes = []
+        for j, plot_data in enumerate(fig['data']):
```
theengineear: Why the `j` and `enumerate`?

theengineear: Also, we usually call `plot_data` --> `trace`.

plotly/tools.py
```diff
@@ -4559,6 +4563,23 @@ def _scatterplot_theme(dataframe, headers, diag, size, height,
...
+            if plot_data['type'] == 'box':
+                boxplot_xaxes.append(
+                    'xaxis{}'.format(plot_data['xaxis'][-1])
+                )
+
+        for xaxis in boxplot_xaxes:
+            fig['layout'][xaxis]['showticklabels'] = False
+
+        return fig
```
theengineear: Personal preference, but I'm not a huge fan of the chaining pattern. You're mutating `fig` in this function and it makes it *super* clear that that's your intention if you don't return anything.

I.e., don't return `fig` to make it clear to callers that this function mutates the `fig` you give it.

plotly/tools.py
```diff
@@ -4806,7 +4827,8 @@ def create_scatterplotmatrix(df, index=None, endpts=None, diag='scatter',
             that defines intervals on the real line. They are used to group
             the entries in an index of numbers into their corresponding
             interval and therefore can be treated as categorical data
-        :param (str) diag: sets the chart type for the main diagonal plots
+        :param (str) diag: sets the chart type for the main diagonal plots.
+            The options are 'scatter', 'histogram' and 'box'.
```
theengineear: 👌

Kully:
> Non-blocking comments, but a little cleanup, documentation would be nice :) Also, is there an open issue that this closes?

Don't think so, but a user complained about it. Plus, I had been trying to figure this out a while ago. Is it better to have an issue to link to to make us look productive? 😛

plotly/tests/test_optional/test_figure_factory.py
Kully: Whoopsi, forgot to erase

plotly/tools.py
Kully: This makes a lot of sense. I `fig`ured that you'd need to assign back to `fig` cause was pointing to another fig and not mutating, but I suppose that's what it's doing.

theengineear:
>Is it better to have an issue to link to to make us look productive? 

Well, it's sorta nice to have a sense of a queue that we're getting work done in. No huge deal though.

plotly/tools.py
theengineear: The 👍 seems to be popular/common.

PR #628, opened 2016-12-06
Better error messages for (i)create_animations

Kully:
Regarding Chris' suggestion here: https://github.com/plotly/plotly.py/pull/584#discussion_r89549480

Haven't tested all of the ways in which a request can fail, mostly because most of the `not done` ones I don't know how to test (i.e. `Request was Throttled`, `Server was Down`) Yes, I said _most_ when we're talking about 2 of the 3 on the list. :slightly_smiling_face: 

- [x] A File already exists with this name
- [x] API key wasn't correct
- [x] API key wasn't supplied
- [x] User was offline
- [x] Permission denied (can't save the file as private)
- [ ]  Request was throttled
- [ ] ~Figure was malformed~
- [ ] Server was down

### DIFF
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -1586,15 +1586,15 @@ def create_animations(figure, filename=None, sharing='public', auto_open=True):
 
     api_url = _api_v2.api_url('plots')
     r = requests.post(api_url, auth=auth, headers=headers, json=json)
-    r.raise_for_status()
 
     try:
         parsed_response = r.json()
     except:
         parsed_response = r.content
 
-    if 'error' in r and r['error'] != '':
-        raise exceptions.PlotlyError(r['error'])
+    # raise error message
+    if r.ok is False:
+        raise exceptions.PlotlyError(parsed_response['errors'][-1]['message'])
 
     if sharing == 'secret':
         web_url = (parsed_response['file']['web_url'][:-1] +
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -1593,8 +1593,17 @@ def create_animations(figure, filename=None, sharing='public', auto_open=True):
         parsed_response = r.content
 
     # raise error message
-    if r.ok is False:
-        raise exceptions.PlotlyError(parsed_response['errors'][-1]['message'])
+    if not r.ok:
+        message = ''
+        if isinstance(parsed_response, dict):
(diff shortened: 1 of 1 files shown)


Kully:
@chriddyp

plotly/plotly/plotly.py
```diff
@@ -1586,15 +1586,15 @@ def create_animations(figure, filename=None, sharing='public', auto_open=True):
...
     try:
         parsed_response = r.json()
     except:
         parsed_response = r.content
 
-    if 'error' in r and r['error'] != '':
-        raise exceptions.PlotlyError(r['error'])
+    # raise error message
+    if r.ok is False:
+        raise exceptions.PlotlyError(parsed_response['errors'][-1]['message'])
```
theengineear: Hmm, I don't understand why we can assume that `['errors'][-1]['message']` will exist? Certainly a `500 Internal Server Error` won't have that, right?

theengineear: Won't this just result in a `KeyError` that will confound and frustrate users? *BEFORE*, at least we got a `HEY, you got a 404!` or something from `r.raise_for_status()`.

theengineear: Can we just do something like:

```
if not r.ok:
    message = ''
    if isinstance(parsed_response, dict):
        errors = parsed_respnonse.get('errors')
        if errors and errors[-1].get('message'):
            message = errors[-1]['message']
    if message:
        raise exceptions.PlotlyError(message)
    else:
        # shucks, we're stuck with a generic error...
        r.raise_for_status()
```

Kully: When would `parsed_response` not be a dict?

chriddyp: 504 gateway timeouts. An HTML response gets returned instead of JSON

chriddyp: Also if the network is down. e.g. try turning off your wifi and then making the request

Kully: Yeah, that worked last time I tried in here...

Should we check if the figure is malformed though?
I tried passing a `figure` by changing the label `data` to `datas` (obviously wrong) and the request went through with all but a title and no plot.

Kully: Okay, I'm good with this

Kully:
@theengineear good to go? 💃 ?

theengineear:
💃 !

PR #630, opened 2016-12-07
return error if get_grid reques is not ok

Kully:


plotly/plotly/plotly.py
```diff
@@ -1488,9 +1488,10 @@ def get_grid(grid_url, raw=False):
     # check if request is ok
     if not r_meta.ok:
         raise exceptions.PlotlyError(
-            "Whoops, you got a 404 response. Your grid url {} may be "
+            "Whoops, you got a 404 response. {} may be "
```
Kully: @theengineear decided to go with a custom message that states it's a `404` because this http message could mean at least 1) grid is private 2) it's not a grid but a plot, etc.

plotly/plotly/plotly.py
```diff
@@ -1484,6 +1484,16 @@ def get_grid(grid_url, raw=False):
...
     r_meta = requests.get(meta_get_url, headers=headers)
+
+    # check if request is ok
+    if not r_meta.ok:
+        raise exceptions.PlotlyError(
+            "Whoops, you got a 404 response. {} may be "
+            "pointing to a plot rather than a grid. If it is a grid, make "
+            "sure your grid is 'public' and not 'private' or "
+            "'secret'.".format(grid_url)
+        )
```
theengineear: 💀 ! How do you know it's a 404?!? Just do `r_meta.raise_for_status()`. Vague is better than wrong 😸. If you really want a gold ⭐️, you can write a function that takes in a `requests.Response` object and raises with the most pertinent information it can get at.

I.e., this can also be shared with the other error-handling code you're writing right now.

theengineear: That said, thanks for jumping on this so quick! It'll be much clearer for users when it's fixed!

Kully:
@theengineear just gonna leave it like this for now. Will _strongly_ consider writing that gold 🌟 function later

theengineear:
Works for me 💃.

PR #639, opened 2016-12-21
Typo contributing.md

alexandresobolevski:


cldougl:
💃 good catch @alexandresobolevski !

PR #640, opened 2016-12-21
Try getting the error message in response_handler

alexandresobolevski:
[WIP, not ready for review...]

The issue tackled here is this https://github.com/plotly/streambed/issues/8801

I decided to give a hand and tackle it because it's part of 2.1 release that we want to deliver before Christmas and Adam is out on vacation.

Now the returned error is 
```
PlotlyError: Sorry, a file already exists with this name.
```

@theengineear I'm failing `test_duplicate_folders` test but in an expected way I guess because I hit the above `existing file` error. I'd love suggestions about changing the test or modifying the PR for a sound result and a review is much appreciated since it's my first time in this repo.

@chriddyp fyi and 👀 ?


chriddyp:
Looks good to me!

chriddyp:
I guess this is technically a backwards incompatible change so we should make a major version bump. Any code that is depending on error handling by subclassing `requests.exceptions.RequestException` will fail with this change as it is currently written. 

Here is the test that is failing:
```
    def test_duplicate_folders(self):
        first_folder = self._random_filename()
        try:
        except requests.exceptions.RequestException as e:
            self.assertTrue(400 <= e.response.status_code < 500)
        else:
            self.fail('Expected this to fail!')
```

I like that the previous code's exception contained `status_code`. We can just modify the message of the `requests.exceptions.RequestException` but nothing else?

plotly/plotly/plotly.py
```diff
@@ -1308,7 +1308,19 @@ def response_handler(cls, response):
...
                 )
             else:
-                raise requests_exception
+                try:
+                    parsed_response = response.json()
+                except:
+                    parsed_response = response.content
+
+                try:
+                    # try get a friendly error message
```
theengineear: 🐐 `try get`

plotly/plotly/plotly.py
```diff
@@ -1308,7 +1308,19 @@ def response_handler(cls, response):
...
+                    errors = parsed_response.get('errors')
+                    message = errors[-1]['message']
+                except:
+                    raise requests_exception
+
+                raise exceptions.PlotlyError(message)
```
theengineear: I feel like we *just* did this:

https://github.com/plotly/plotly.py/blob/master/plotly/plotly/plotly.py#L1704-L1720

Want to try and follow that more closely, OR, extract the common functionality out? It'd be great to just have one error handler.

In particular, the code you wrote assume that `errors` is *not empty*, which we might not want to assume.

theengineear: Thanks for adding this to the central error handler, I actually forgot we even had it!

theengineear:
@chriddyp a couple thoughts. We *have* a `PlotlyRequestException` that is meant to take a `requests.exceptions.RequestException` and return the proper message (though I think it's outdated).

That means we have *3* ways to raise exceptions for requests:

1. Use a generic `PlotlyError`
2. Use a `PlotlyRequestException`
3. Just raise the `requests.exceptions.RequestException`

I'd vote for *not* requiring a user to import `requests` and *always* raising `PlotlyRequestException` (and yes +1 to a version bump since this is backwards incompatible).

plotly/plotly/plotly.py
alexandresobolevski: > Want to try and follow that more closely, OR, extract the common functionality out? It'd be great to just have one error handler.

that's what I was hoping to achieve, I took the main logic and moved it to the response_handler and use it in animations to catch errors instead of having a code block in animations for that.

> In particular, the code you wrote assume that errors is not empty, which we might not want to assume

it's in a `try` bloc, so if my assumption is incorrect then it raises the usual `requests_exception`. I could put two `if` for looking to see if `errors` and `message` exist but would that not do the same?

theengineear: aw jeez. i didn't realize it was still in a try block. that's fine then. But, ya that's what I was talking about. The lines I was linking to were:

```
    try:
        parsed_response = r.json()
    except:
        parsed_response = r.content

    # raise error message
    if not r.ok:
        message = ''
        if isinstance(parsed_response, dict):
            errors = parsed_response.get('errors')
            if errors and errors[-1].get('message'):
                message = errors[-1]['message']
        if message:
            raise exceptions.PlotlyError(message)
        else:
            # shucks, we're stuck with a generic error...
            r.raise_for_status()
```

alexandresobolevski:
I am still confused as to what I should do

> I like that the previous code's exception contained status_code. We can just modify the message or the __str__ of the requests.exceptions.RequestException but nothing else?

Doing this does not bump the version and I am not entirely sure how to change the message of `requests.exceptions.HTTPError`

> I'd vote for not requiring a user to import requests and always raising PlotlyRequestException (and yes +1 to a version bump since this is backwards incompatible).

This bumps the version and I assume we're talking `PlotlyRequestError`, if so Im not sure how to get the status from the API in there.


this is what I have so far

```
                try:
                    # try get a friendly error message
                    errors = parsed_response.get('errors')
                    message = errors[-1]['message']
                except:
                    # otherwise raise a generic error
                    raise requests_exception
                else:
                    requests_exception.message = message
                    raise exceptions.PlotlyRequestError(requests_exception)

```

theengineear:
@alexandresobolevski , i was suggesting *always* raising the `PlotlyRequestError` (i believe that it sets the status code)

So, something more like:

```
                try:
                    # try get a friendly error message
                    errors = parsed_response.get('errors')
                    message = errors[-1]['message']
                except:
                    # otherwise raise a generic error
                    raise exceptions.PlotlyRequestError(requests_exception)
                else:
                    requests_exception.message = message
                    raise exceptions.PlotlyRequestError(requests_exception)
```

theengineear:
@chriddyp , I slacked with @alexandresobolevski a little bit on this. I'm going to hop in here unify how we do error handling. You're right that it requires a version bump to make this change, but I don't want to force users to change their error-handling code *twice*, that's silly.

theengineear:
@chriddyp ^^ here's a long-overdue Christmas miracle :p. I did a full overhaul on organizing our api calls in this repo. It's dead simple and unifies how both request *creation* and response *validation* work in both `v1` and `v2` code paths.

I also defer content parsing to the `Response.json` method in `requests`, which *appears* to by `PY2`/`PY3` agnostic. This should greatly simplify problems with decoding/unicode schtuff. However, I'm going to let our test suite churn for now (may be a lot of brokenness).

And yep, this will be a big version bump. Again we were going to need a version bump as suggested in https://github.com/plotly/plotly.py/pull/640#issuecomment-268701225. I figured we may as well really tackle this error-handling thing if we're going to need to do a version bump anyhow.

theengineear:
Also, we're still only supporting `2.7.8 3.3.3 3.4.1`, we should test `3.5` and `3.6` as well (Python devs are moving fast :p)

theengineear:
^^ 😮 holy $#!T! It passed *already*.

theengineear:
Manually test the following in Python 3:

- [ ] loading graph reference
- [ ] saving / retrieving plots with unicode

theengineear:
Just as a heads up, I tagged this as a WIP for now, I have a lot of tests that need adding in for the new `api` module I created. What I'd *like* to do is ensure that we have *lots* of tests with `requests.request` mocked and then just add a *handful* of tests that are actual integration tests. This test suite is unnecessarily slow right now, and will continue to feel slower as we test on more and more Python versions.

plotly/api/utils.py
```diff
@@ -0,0 +1,41 @@
...
+    :param (str) username: A Plotly username.
+    :param (str) password: The password for the given Plotly username.
+    :returns: (str) An 'authorization' header for use in a request header.
+
+    """
+    if isinstance(username, str):
+        username = username.encode('latin1')
+
+    if isinstance(password, str):
+        password = password.encode('latin1')
```
chriddyp: weird, why `latin1` and not `utf-8`?

plotly/api/v2/files.py
```diff
@@ -0,0 +1,85 @@
...
+    :param (str) path: The '/'-delimited path specifying the file location.
+    :param (int) parent: Parent id, an integer, which the path is relative to.
+    :param (str) user: The username to target files for. Defaults to requestor.
+    :param (bool) exists: If True, don't return the full file, just a flag.
+    :returns: (requests.Response) Returns response directly from requests.
+
+    """
+    url = build_url(RESOURCE, route='lookup')
+    params = make_params(path=path, parent=parent, user=user, exists=exists)
+    return request('get', url, params=params)
```
chriddyp: nice, this all looks really good

plotly/plotly/plotly.py
```diff
@@ -386,28 +416,49 @@ def get_figure(file_owner_or_url, file_id=None, raw=False):
...
+                        if entry['orientation'] == 'h':
+                            _swap_xy_data(entry)
+                        del entry['orientation']
+                    if ('y' in entry) and ('x' not in entry):
+                        if entry['orientation'] == 'v':
+                            _swap_xy_data(entry)
+                        del entry['orientation']
+            figure['data'][index] = entry
+        except KeyError:
+            pass
```
chriddyp: is this to pass validation?

plotly/plotly/plotly.py
```diff
@@ -1738,7 +1554,7 @@ def icreate_animations(figure, filename=None, sharing='public', auto_open=False)
     This function is based off `plotly.plotly.iplot`. See `plotly.plotly.
     create_animations` Doc String for param descriptions.
     """
-    # Still needs doing: create a wrapper for iplot and icreate_animations
+    # TODO: create a wrapper for iplot and icreate_animations
```
chriddyp: I don't think that this comment still applies anymore: https://github.com/plotly/plotly.py/blob/05328c3db09f353e92d4b04785d2b4e763421e8b/plotly/plotly/plotly.py#L1734

chriddyp:
Wow, huge stuff! Looks great to me so far

plotly/api/utils.py
theengineear: You'd have to ask [Kenneth Reitz](https://www.kennethreitz.org/). It's just something I pulled from `requests.HTTPBasicAuth`. My guess is that it doesn't matter, but I'd rather just leave it as was written there.

theengineear:
Closing in favor of #650

PR #645, opened 2016-12-31
Try tox.

theengineear:
I tried using tox *looonog* ago and it didn't seem to fit our needs for CI. However, I just tried it yesterday again and was able to get a solid setup for our core vs. optional testing needs.

We're considering adding back 2.6 support and having the ability to specify a possibly-different set of deps in tox would be the better option, imo.

See https://github.com/plotly/plotly.py/pull/635


theengineear:
@chriddyp @cldougl @Kully, any objections?

theengineear:
^^ still working on ensuring the base python versions line up correctly. not sure why that's not better documented, pretty important!

theengineear:
Man. There's always something... I decided to freeze some deps into our tox tests, which messed up support for `Python 3.3` (`pandas` doesn't seem to like being installed on `Python 3.3`, I should be able to knock it to core-support-only for `Python3.3`...).

Also, for whatever reason, accessing 3.5 on Circle CI via environment variables is super painful... I ended up just being absurdly specific about it so that tox/circle don't do *any* guessing about finding Python interpreters, since they were being bad guessers...

theengineear:
Woo! Aight, I think we're good to go! I'm shocked that our coverage is above 70% (not *too* bad, folks!)

I just added *one* more change to make sure we're explicitly printing out the python version that's being tested (as a sanity check). If that all passes, I'll likely go ahead and merge this in since it only really deals with how tests are run and doesn't change code other than that.

They're not extensive, but I did write up some documentation in `contributing.md` that might be helpful for understanding our `tox` setup. It's actually incredibly flexible and while it wasn't the easiest to set up, I think we'll start to profit as we decide we need/want to support more complicated environments in our test suites 🍻.

//cc @chriddyp @Kully @cldougl

chriddyp:
💃  Looks good!

Issue #642, opened 2016-12-29
2.0.0 Wishlist


theengineear:
@chriddyp @Kully @cldougl 

I haven't had many cycles to play around with the python api library lately, but recently got in a little deep for https://github.com/plotly/plotly.py/pull/640, where we've now got *fully-unified* api modules.

I think enough shifted (mostly additions, but the error handling is definitely not backwards compat) that we should open up a discussion on non-minor changes that we've been looking to add to the plotly module. Here's a list of topics:

- [x] API request/response unification (https://github.com/plotly/plotly.py/pull/651)
- [ ] Fix or drop support for matplotlib
- [x] Add `numpy` as an official dep. and use it to speed up the library. (#667)
- [x]  Related to above, clean up how we handle optional imports (https://github.com/plotly/plotly.py/pull/317)
- [x] Stop auto-loading graph reference and just use packaged graph reference (https://github.com/plotly/plotly.py/pull/614)**
- [x] ~~Add transforms to graph objects (https://github.com/plotly/plotly.py/issues/587)~~ holding off on this until the transforms spec settles in plotly.js.
- [x] Add frames for graph objects (https://github.com/plotly/plotly.py/issues/604)
- [x] Add a sign_in validation #525 (#668)
- [x] ~~Fix up streaming? (this might be more work than it's worth at this point https://github.com/plotly/plotly.py/pull/253)~~ No time for this atm.
- [x] Some module re-organizing. Most importantly, we should move FigureFactory into it's own sub-package and each Figure-generator should be it's own file. The `tools.py` module is enormous and is a constant point of pain for circular imports (https://github.com/plotly/plotly.py/pull/652).

**In fact, if we *do* do this, we can just add some meta programming to auto-create a module full of all the objects from graph-reference. This way we don't have those dump "Scatter is not defined" errors, etc.

Anything else?

theengineear:
I'm still going to be working mostly on frontend code, but I'd definitely be interested in making some non-trivial improvements here. Thoughts?

chriddyp:
>  Add numpy as an official dep. and use it to speed up the library.

I'm OK with using this in all of our `FigureFactory` methods but I'd still like to keep this as a protected import. I think there will still be new users who might have trouble installing it on their machine with their default python distribution.

chriddyp:
For `2.0.0`, I think that we only need to consider backwards incompatible changes if we're going to version as according to SemVer. In other words, it doesn't need to be a big release.

[...]

PR #648, opened 2017-01-05
Local plot schema

theengineear:
This PR stops auto-loading the plot-schema on import. We decided in #642 that this shouldn't need to force a major version bump, so I'll add a minor version bump here.

See #614 for the original PR. I noticed that we actually had a weird PY3 conflict due to a change in how importing works in PY3 (See commit notes).


theengineear:
@Kully can you give this a review? We discussed the no-download-on-import portion of this in #642, but I've also done the following (which we *haven't* discussed wrt Semantic Versioning):

* `.js` and `.json` files --> `plotly/package_data`. I can't imagine any users are using the non-python files directly in scripts.
* auto-object generation. This isn't backwards-incompatible. In fact, it's exactly what we had before, except now it's not generated on import in memory.

@chriddyp, 👍 on the meta programming introduced in 51ef008?

plotly/graph_objs/graph_objs_tools.py
```diff
@@ -37,7 +37,8 @@ def _list_help(object_name, path=(), parent_object_names=()):
     items = graph_reference.ARRAYS[object_name]['items']
     items_classes = [graph_reference.string_to_class_name(item)
                      for item in items]
-    lines = textwrap.wrap(repr(items_classes), width=LINE_SIZE-TAB_SIZE)
+    items_classes.sort()
+    lines = textwrap.wrap(repr(items_classes), width=LINE_SIZE-TAB_SIZE - 1)
```
Kully: Nit Picky, but would `lines = textwrap.wrap(repr(items_classes), width=LINE_SIZE - TAB_SIZE - 1)`
or `lines = textwrap.wrap(repr(items_classes), width=LINE_SIZE-TAB_SIZE-1)` be better for consistency?

plotly/graph_objs/graph_objs_tools.py
```diff
@@ -167,7 +169,7 @@ def _dict_attribute_help(object_name, path, parent_object_names, attribute):
                 if isinstance(val, list) and attribute == 'showline':
                     val = val[0]
 
-                lines = textwrap.wrap(val, width=LINE_SIZE)
+                lines = textwrap.wrap(val, width=LINE_SIZE - 1)
```
Kully: See [previous comment](https://github.com/plotly/plotly.py/pull/648/files#r94804612)

plotly/graph_objs/graph_objs_tools.py
```diff
@@ -54,9 +55,10 @@ def _list_help(object_name, path=(), parent_object_names=()):
 
 def _dict_object_help(object_name, path, parent_object_names):
     """See get_help()."""
-    attributes = graph_reference.get_valid_attributes(object_name,
-                                                      parent_object_names)
-    lines = textwrap.wrap(repr(list(attributes)), width=LINE_SIZE-TAB_SIZE)
+    attributes = list(
+        graph_reference.get_valid_attributes(object_name, parent_object_names))
+    attributes.sort()
+    lines = textwrap.wrap(repr(list(attributes)), width=LINE_SIZE-TAB_SIZE - 1)
```
Kully: See [previous comment](https://github.com/plotly/plotly.py/pull/648/files#r94804612)

update_graph_objs.py
```diff
@@ -0,0 +1,271 @@
...
+FLAG = '# AUTO-GENERATED BELOW. DO NOT EDIT! See makefile.'
+
+
+def get_non_generated_file_lines():
+    """
+    Copy each line up to our special FLAG line and return.
+
+    :raises: (ValueError) If the FLAG isn't found.
+    :return: (list) The lines we're copying.
+
```
Kully: I think the doc strings should not have the extra space between last line and the triple quotes """

update_graph_objs.py
```diff
@@ -0,0 +1,271 @@
...
+    with open('./plotly/graph_objs/graph_objs.py', 'r') as f:
+        for line_to_copy in f:
+            if line_to_copy.startswith(FLAG):
+                flag_found = True
+                break
+            lines_to_copy.append(line_to_copy)
+    if not flag_found:
+        raise ValueError(
+            'Failed to find flag:\n''{}\n''in graph_objs_tools.py.'
+            .format(FLAG)
```
Kully: More a suggestion, but for legibility, what do you think about putting the `format(FLAG)` on the line above? I think it's 80 chars on that line, but would be worth IMO

update_graph_objs.py
```diff
@@ -0,0 +1,271 @@
...
+        )
+    return lines_to_copy
+
+
+def print_figure_patch(f):
+    """Print a patch to our Figure object into the given open file."""
```
Kully: For one-liner doc strings, which is more common to your knowledge:

```one-line-here```

```
one-line-here
```

I usually do the latter. Let me know if you think this is worth worrying/thinking about

update_graph_objs.py
theengineear: Hehe, I disagree on this one. But I'm OK with changing it. You like to do the blank line *after* the `"""` then?

update_graph_objs.py
theengineear: Huh... that string looks wrong to me, the extra `'` characters? I'll fix this up.

update_graph_objs.py
theengineear: One-liners can/should be one-liners:

https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

theengineear:
@Kully alright with this?

theengineear:
@chriddyp , it'd be great to get a high-level 👍 from you on 51ef008 still.

You don't need to dig into the code very much. In general I'm just wondering if you think the approach of moving the meta programming we were *already* doing into auto-generated lines inside the `graph_objs.py` files seems OK.

Specifically, all the work is done in `update_graph_objs.py`, where I literally print new lines into `graph_objs.py` programmatically.

theengineear:
Since I've got a lot of moving parts happening in this repo, I'm going to merge after chatting with @Kully in slack and with @chriddyp's 👍 for hard-coding our graph_objs.

PR #656, opened 2017-01-09
Improve `frames` support in graph_objs.py.

### LINKED ISSUE #604: Better Integrate `frames` into `Figure`
Referring to fix in https://github.com/plotly/plotly.py/pull/584 -> https://github.com/plotly/plotly.py/blob/a156503b275763a3ec7738f5f251b93597824ed0/plotly/graph_objs/graph_objs.py#L839

@theengineear

theengineear:
This does the following:

* Allow object-or-string in frames array
* Refuse `frames` in frame object if nested
* Map frame entry `data` and `layout` objects to `figure` attribute information.
* Tests this functionality

Fixes #604

### DIFF
--- a/plotly/graph_objs/graph_objs.py
+++ b/plotly/graph_objs/graph_objs.py
@@ -787,7 +787,7 @@ class GraphObjectFactory(object):
 
         # We patch Figure and Data, so they actually require the subclass.
         class_name = graph_reference.OBJECT_NAME_TO_CLASS_NAME.get(object_name)
-        if class_name in ['Figure', 'Data']:
+        if class_name in ['Figure', 'Data', 'Frames']:
             return globals()[class_name](*args, **kwargs)
         else:
             kwargs['_name'] = object_name
@@ -1097,7 +1097,7 @@ class Figure(PlotlyDict):
     """
     Valid attributes for 'figure' at path [] under parents ():
     
-        ['data', 'layout']
+        ['data', 'frames', 'layout']
     
     Run `<figure-object>.help('attribute')` on any of the above.
     '<figure-object>' is the object at []
@@ -1108,22 +1108,7 @@ class Figure(PlotlyDict):
     def __init__(self, *args, **kwargs):
         super(Figure, self).__init__(*args, **kwargs)
         if 'data' not in self:
-            self.data = GraphObjectFactory.create('data', _parent=self,
-                                                  _parent_key='data')
-
-    # TODO better integrate frames into Figure - #604
-    def __setitem__(self, key, value, **kwargs):
-        if key == 'frames':
-            super(PlotlyDict, self).__setitem__(key, value)
-        else:
-            super(Figure, self).__setitem__(key, value, **kwargs)
-
-    def _get_valid_attributes(self):
-        super(Figure, self)._get_valid_attributes()
-        # TODO better integrate frames into Figure - #604
-        if 'frames' not in self._valid_attributes:
-            self._valid_attributes.add('frames')
-        return self._valid_attributes
+            self.data = Data(_parent=self, _parent_key='data')
 
     def get_data(self, flatten=False):
         """
@@ -1241,8 +1226,45 @@ class Font(PlotlyDict):
     _name = 'font'
 
 
-class Frames(dict):
-    pass
+class Frames(PlotlyList):
+    """
+    Valid items for 'frames' at path [] under parents ():
+        ['dict']
+
+    """
+    _name = 'frames'
+
+    def _value_to_graph_object(self, index, value, _raise=True):
+        if isinstance(value, six.string_types):
+            return value
+        return super(Frames, self)._value_to_graph_object(index, value,
+                                                          _raise=_raise)
+
+    def to_string(self, level=0, indent=4, eol='\n',
+                  pretty=True, max_chars=80):
+        """Get formatted string by calling `to_string` on children items."""
+        if not len(self):
+            return "{name}()".format(name=self._get_class_name())
+        string = "{name}([{eol}{indent}".format(
+            name=self._get_class_name(),
+            eol=eol,
+            indent=' ' * indent * (level + 1))
+        for index, entry in enumerate(self):
+            if isinstance(entry, six.string_types):
+                string += repr(entry)
+            else:
+                string += entry.to_string(level=level+1,
+                                          indent=indent,
+                                          eol=eol,
+                                          pretty=pretty,
+                                          max_chars=max_chars)
+            if index < len(self) - 1:
+                string += ",{eol}{indent}".format(
+                    eol=eol,
+                    indent=' ' * indent * (level + 1))
+        string += (
+            "{eol}{indent}])").format(eol=eol, indent=' ' * indent * level)
+        return string
 
 
 class Heatmap(PlotlyDict):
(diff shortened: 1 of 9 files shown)


theengineear:
@Kully and I hacked in frames support when we were in a rush. This improves how `frames` is integrated into the `Figure` object. Since `frames` is basically considered an array of `Figure` objects.

Still working on this one, since I think saying that an entry in a `Frames` array is a `Figure` is misleading, imo...

theengineear:
Can't add `Frame` since it conflicts with `animation.frame` and the name doesn't matter since we're not exposing it :). I chose `frame_entry`, which does *not* become a public class because there's no need to patch the object. Note how it ends up as `dict( .. )` in the `to_string()` call.

plotly/graph_objs/graph_objs.py
```diff
@@ -1241,8 +1226,45 @@ class Font(PlotlyDict):
...
+    """
+    Valid items for 'frames' at path [] under parents ():
+        ['dict']
+
+    """
+    _name = 'frames'
+
+    def _value_to_graph_object(self, index, value, _raise=True):
+        if isinstance(value, six.string_types):
+            return value
```
theengineear: This is a patch to allow a `string` in the `frames` array.

plotly/graph_objs/graph_objs.py
```diff
@@ -1241,8 +1226,45 @@ class Font(PlotlyDict):
...
+        """Get formatted string by calling `to_string` on children items."""
+        if not len(self):
+            return "{name}()".format(name=self._get_class_name())
+        string = "{name}([{eol}{indent}".format(
+            name=self._get_class_name(),
+            eol=eol,
+            indent=' ' * indent * (level + 1))
+        for index, entry in enumerate(self):
+            if isinstance(entry, six.string_types):
+                string += repr(entry)
```
theengineear: Another patch to allow a string in the `frames` array. This sorta sucks since it's not possible to `super`... but I'm not terribly concerned.

plotly/graph_objs/graph_objs_tools.py
```diff
@@ -35,8 +35,14 @@ def get_help(object_name, path=(), parent_object_names=(), attribute=None):
...
-    items_classes = [graph_reference.string_to_class_name(item)
-                     for item in items]
+    items_classes = set()
+    for item in items:
+        if item in graph_reference.OBJECT_NAME_TO_CLASS_NAME:
+            items_classes.add(graph_reference.string_to_class_name(item))
+        else:
+            # There are no lists objects which can contain list entries.
+            items_classes.add('dict')
+    items_classes = list(items_classes)
```
theengineear: This allows us to print `dict` in a list's `to_string()` return if the item isn't a publicly-available `class`.

plotly/graph_reference.py
```diff
@@ -410,12 +410,25 @@ def _patch_objects():
...
     OBJECTS['figure'] = {'meta_paths': [],
                          'attribute_paths': [],
                          'additional_attributes': figure_attributes}
 
+    frame_entry_attributes = {
+        'data': {'role': 'object', '_isLinkedToArray': True},
+        'group': {'role': 'info'},
+        'layout': {'role': 'object'},
+        'name': {'role': 'info'}
+    }
```
theengineear: @etpinard or @rreusser Is this information available anywhere? I don't really want to hard-code these. If it's *not* available, mind helping me to flesh out this list of valid attributes for a `figure.frames[*]` entry?

rreusser: Unfortunately there's not a dedicated doc page since this doesn't currently live on the plot scheme. In lieu of that, the attributes are defined [here](https://github.com/plotly/plotly.js/blob/master/src/plots/frame_attributes.js). @etpinard is there some way I can add these to the plot schema? The way animation config and frame attributes live outside of the rest of plotly is driving me a little crazy.

theengineear: Awesome! That'll do for now!

etpinard: Those are now part of the plot schema:

image

theengineear: oh damn boom!

theengineear: @etpinard , trying to grok `datasrc`, that one's a little confusing for me. How do you `src` an array of *objects*?

Also, `traces` are meant to be stored in a `Grid`? I.e., it's definitely not `info`, right? (just to confirm).

etpinard: Oh `datasrc` shouldn't be in there. Same for `tracessrc`.

theengineear:
@etpinard , OK, this would work well from the Python side of things for the `frames` object:

```
    "frames": {
        "items": {
            "frames_entry": {
                "baseframe": {
                    "description": "The name of the frame into which this frame's properties are merged before applying. This is used to unify properties and avoid needing to specify the same values for the same properties in multiple frames.",
                    "role": "info",
                    "valType": "string"
                },
                "data": {
                    "description": "A list of traces this frame modifies. The format is identical to the normal trace definition.",
                    "role": "object",
                    "valType": "any"
                },
                "group": {
                    "description": "An identifier that specifies the group to which the frame belongs, used by animate to select a subset of frames.",
                    "role": "info",
                    "valType": "string"
                },
                "layout": {
                    "role": "object",
                    "description": "Layout properties which this frame modifies. The format is identical to the normal layout definition.",
                    "valType": "any"
                },
                "name": {
                    "description": "A label by which to identify the frame",
                    "role": "info",
                    "valType": "string"
                },
                "traces": {
                    "description": "A list of trace indices that identify the respective traces in the data attribute",
                    "role": "info",
                    "valType": "info_array"
                },
                "role": "object"
            }
        },
        "role": "object"
    },
```

The name `frames_entry` is up for discussion but it would greatly help me if it's unique in the plot-schema, fwiw... i.e., not `frame`. This is because it's tricky to map `data` and `layout` attrs properly inside `frames_entry` objects.

Note that `FrameEntry` will *not* be a name that is exposed to users in the Python API, it's merely helpful for mapping attrs.

theengineear:
OK, I "updated" the default schema with what I'd *like* to be in there. Obviously not ready to merge since this isn't a real reflection of the plot-schema, I just figured It'd be helpful.

theengineear:
Oh right, we have a test on the schema anyhow so I *can't* merge it 😊

etpinard:
@theengineear done (I think) in https://github.com/plotly/plotly.js/pull/1292

plotly/graph_reference.py
```diff
@@ -68,9 +68,62 @@ def get_graph_reference():
     """
     path = os.path.join('package_data', 'default-schema.json')
     s = resource_string('plotly', path).decode('utf-8')
-    graph_reference = _json.loads(s)
+    graph_reference = utils.decode_unicode(_json.loads(s))
+
+    # TODO: Patch in frames info until it hits streambed. See #659
+    graph_reference['frames'] = {
```
theengineear: Just patching in what I'm expecting to see here. This means that we don't need to wait for any changes to land on Plotly prod and can get this into `1.X` *now*. I'll remove this once that changes.

theengineear:
@Kully, this look OK to you? Mind giving it a review?

Kully:
> @Kully, this look OK to you? Mind giving it a review?

Yeah, will do today.

Kully:
> Yeah, will do today.

@theengineear 💃  Looks like a lot of work you had to do! 😅

theengineear:
Yah, the fact that you can have `data` and `layout` inside the `frame` was tricky. We *could* have just fixed this in `plot-schema`, but that would have taken a *whole* lot more work I think.

theengineear:
Cool, I'm able to play around and make frames and such using the python api lib still. Gonna update changelog and merge.

PR #658, opened 2017-01-11
2.0.0

### LINKED ISSUE #642: 2.0.0 Wishlist

theengineear:
Aight, we're set to merge this. I'm going to ask for a final round of review and then merge and setup a `pre-release` for this before I push anything new up to PyPI.

Closes #642


theengineear:
@chriddyp @Kully @cldougl is there any reason for us to support `1.X` anymore after we do a major version bump? Bigger projects often support (and continue to update minor versions on) all major versions. (see the `stable/X` branches in [django](https://github.com/django/django) for an example). *I don't see a need for us to do this* and we should always be able to go back in time, check out a previous version and cherry-pick in a change if necessary.

In other words, I'd like to keep our versioning linear if possible, each new version should be a higher number than the last. I suppose we can always change this later by pulling old commits into new branches, etc, etc.

 Again, I'm suggesting that we merge `2.0.0` directly into master and not manage `1.X` and `2.X` in separate branches or something.

Kully:
I feel like it would be pretty annoying to maintain both branches. I recall a lot of organization stuff (FigureFactory) which won't affect the users' experience and more additions which make 2.0.0 sounds better.

theengineear:
>I feel like it would be pretty annoying to maintain both branches.

Yup, it would be. Not supporting multiple, concurrent versions is definitely my vote, but I just want to get consensus on our approach.

cldougl:
+1 for not continuing to support version `1.X`
*We'll have to update the docs as well.

theengineear:
Great! Consensus-enough! Then I'm going to merge `2.0.0` to master today and get it up on PyPI.

theengineear:
Okie dokie. Heres' the state of `2.0.0`:

* ensure that layout gets passed in along to `clientresp`: #666
* require `numpy` in FF methods: #667
* validate on `py.sign_in`: #668

I'll version-bump when all that's through ;)

theengineear:
Aight, final tests are happening now!

@chriddyp @cldougl @Kully , any final reservations? I'll create a pre-release tag for this and may ask for a bit more manual testing after this hits `master`.

cldougl:
no reservations!
and I'm happy to help manual test too

theengineear:
OK, done waiting, I think the next step needs to be manual testing. Merge time!

PR #661, opened 2017-01-13
user-controllable plotly config directory

### LINKED ISSUE #533: Allow user to set plotly directory as environment variable
It would be very convenient if there were a way to set the plotly directory in the PLOTLY_DIR environment variable, or default to ~/.plotly if the variable is not set. It can be easily done by changing line 4 here: https://github.com/plotly/plotly.py/blob/master/plotly/files.py
to the following:

`PLOTLY_DIR = os.environ.get("PLOTLY_DIR", os.path.join(os.path.expanduser("~"), ".plotly"))`

Then, to specify a custom plotly directory when running a script which uses the plotly module, we can do `PLOTLY_DIR=/path/to/plotly_dir python script_that_uses_plotly.py`

itsb:
Ran into a concurrency bug wherein some processes would fail when there were too many parallel plotly tasks hammering the same config files in ~/.plotly.
Adding user-controllable config directory resolved the issue when each parallel task used a unique PLOTLY_DIR environment variable.
Also fixes #533.


theengineear:
@itsb , I'm happy to add support for environment variables, but I'm not really sure how this is helping?

I sorta feel like it would make more sense to just set environment variables and then use those?

I think the desired order of precedence should be (from lowest to highest precedence):

1. actual files
2. environment variables
3. session
4. call args

What do you think about just allowing folks to set the following environment variables?

```
PLOTLY_USERNAME
PLOTLY_API_KEY
PLOTLY_DOMAIN
PLOTLY_API_DOMAIN
...
```

Additionally, would it help to use the `session` module to set the credentials/config you want?

itsb:
I made a docker repo to clearly demo the issue: https://hub.docker.com/r/itsb/plotly/
You can also demo the patch solution: https://github.com/itsb/plotly/tree/master/patch
We can also use this approach to validate all proposed workarounds.
I'm not sure the session module would help, because files.py is still being invoked on plotly module init which leads to the concurrency issue.
Thanks,

theengineear:
>because files.py is still being invoked on plotly module init which leads to the concurrency issue.

Noted. Yep, I see what you mean now. Thanks for following up!

ladhani:
Also a big headacke getting plotly running with Microsoft IIS on Windows Server using CGI binding with python.exe. As home directory will be pointing to some subfolder under windows system folder. So a no-no for writing .plotly and .testing subfolders!!!

I made the following ugly change in files.py:

======================
import os

# file structure
#PLOTLY_DIR = os.path.join(os.path.expanduser("~"), ".plotly")
#CREDENTIALS_FILE = os.path.join(PLOTLY_DIR, ".credentials")
#CONFIG_FILE = os.path.join(PLOTLY_DIR, ".config")

#TEST_DIR = os.path.join(os.path.expanduser("~"), ".test")
#TEST_FILE = os.path.join(PLOTLY_DIR, ".permission_test")

#Changed by Al-Noor

PLOTLY_DIR = ".plotly"
CREDENTIALS_FILE = os.path.join(PLOTLY_DIR, ".credentials")
CONFIG_FILE = os.path.join(PLOTLY_DIR, ".config")

TEST_DIR = ".plotly_test"
TEST_FILE = os.path.join(PLOTLY_DIR, ".permission_test")
==============



So now ,plotly folders are made in the Web folder where IIS User has permission to write/read.

And now I can use plotly on Microsoft 2016 server with IIS and Python via CGI.

But it would be better to have a more clean solution, I think!
(local accounts now get the ,plotly folders in each of their local directory where they run py with plotly.

I will submit this issue / feature request to plotly.

mungojam:
We want to use plotly offline in windows, but our `~` is on a network drive. We want to avoid that altogether so a way to override the local config file storage area would be great.

jonmmease:
Thanks for chiming in @mungojam. I definitely agree this is important, and I'll try to take a look at this for 3.3 (we just released 3.2 and no ETA for 3.3 yet).

jonmmease:
This PR became the starting pont of #1195.  Thanks for the contribution @itsb!

PR #667, opened 2017-01-18
FF requires numpy

theengineear:
Note that we don’t necessarily *require* `numpy` for in all the FF
methods. We’re just looking to make the compatibility-breaking
change so that we can depend on it in `2.0.0`.

Also note that I’d forgotten to include `figure_factory` in setup.py
before ;__;.


theengineear:
@chriddyp for review. Seem ok?

//cc @choldgraf 😸 just some setup towards a faster figure factory methods.

theengineear:
^^ huh, rebuilt, seems like an intermittent npm fail?

choldgraf:
sweet! thx for keeping me in the loop

PR #668, opened 2017-01-18
Validate signin

theengineear:
Simply raises a `PlotlyError` if we can't hit `/v2/users/current` successfully given the session credentials.

I decided to let the session creds/config get set and not revert on fail. It felt simpler to me and I didn't know why it would be beneficial to do anything otherwise.

### DIFF
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -48,7 +48,15 @@ tools.ensure_local_plotly_files()
 
 
 # don't break backwards compatibility
-sign_in = session.sign_in
+def sign_in(username, api_key, **kwargs):
+    session.sign_in(username, api_key, **kwargs)
+    try:
+        # The only way this can succeed is if the user can be authenticated
+        # with the given, username, api_key, and plotly_api_domain.
+        v2.users.current()
+    except exceptions.PlotlyRequestError:
+        raise exceptions.PlotlyError('Sign in failed.')
+
 update_plot_options = session.update_session_plot_options
 
 
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -24,11 +24,9 @@ import six
 import six.moves
 from requests.compat import json as _json
 
-from plotly import exceptions, tools, utils, files
+from plotly import exceptions, files, session, tools, utils
 from plotly.api import v1, v2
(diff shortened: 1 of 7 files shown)


theengineear:
@Kully review pls :)

//cc @chriddyp seem OK?

plotly/plotly/plotly.py
```diff
@@ -24,11 +24,9 @@
 import six.moves
 from requests.compat import json as _json
 
-from plotly import exceptions, tools, utils, files
+from plotly import exceptions, files, session, tools, utils
```
Kully: I'm just curious, why did you switch up the order like this? Is it because `tools` is moving into a different location for `2.0.0.` and therefore its priority is being hierarchically diminished? (I don't know the proper jargon for what I'm describing...)

plotly/tests/test_core/test_plotly/test_credentials.py
```diff
@@ -1,39 +1,43 @@
...
+        file_creds = tls.get_credentials_file()
+        self.assertEqual(creds, file_creds)
 
-
-class TestSignIn(TestCase):
+    def test_sign_in(self):
+        un = 'anyone'
+        ak = 'something'
+        # TODO, add this!
+        # si = ['this', 'and-this']
```
Kully: What does this `TODO` mean exactly? I'm guessing `si` means something here?

Kully:
> @Kully review pls :)

All good, just a few questions I threw in there. 💃

plotly/plotly/plotly.py
chriddyp: looks like it's just alphabetical :sunglasses:

chriddyp:
good to me!

plotly/tests/test_core/test_plotly/test_credentials.py
theengineear: That's just old code. I just needed to add it to the `TestCase` class. I think it's for *stream ids* probably? I'm not going to futz with it in this PR though.

theengineear:
Uggs. GH/CI stuff has been weird the last day or so. These tests all passed in the most recent build:

https://circleci.com/gh/plotly/plotly.py/1883

Gonna merge!

Issue #603, opened 2016-11-10
Expose rugplot parameter in `_violinplot` to `create_violin`?


colinfang:
Can we config `rugplot` in `create_violin`?

    _violinplot(vals, fillcolor='#1f77b4', rugplot=True)

https://github.com/plotly/plotly.py/blob/86291ffc593048becceb7d430464524f811598a3/plotly/tools.py#L2755

Kully:
Do you just want `rugplot` exposed in the function signature?

colinfang:
Yep. `rugplot` causes too much performance overhead when the dataset is large. Currently, I have to manually strip those traces away in json.

PR #670, opened 2017-01-20
add rugplot + changelog

Kully:
Fixes this issue: https://github.com/plotly/plotly.py/issues/603

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+### Updated
+- `plotly.figure_factory.create_violin()` now has a `rugplot` parameter which determines whether or not a rugplot is draw beside each violin plot.
+
 ## [2.0.0]
 
 ### Changed
--- a/plotly/figure_factory/_violin.py
+++ b/plotly/figure_factory/_violin.py
@@ -194,8 +194,8 @@ def violinplot(vals, fillcolor='#1f77b4', rugplot=True):
 
 
 def violin_no_colorscale(data, data_header, group_header, colors,
-                          use_colorscale, group_stats,
-                          height, width, title):
+                         use_colorscale, group_stats, rugplot,
+                         height, width, title):
     """
     Refer to FigureFactory.create_violin() for docstring.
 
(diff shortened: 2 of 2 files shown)


Kully:
@theengineear ready for a quick 👀

theengineear:
Huh, weird intermittent sign-in failure... If it keeps happing, you could mock the `users.current` request so that it just skips the validation part. Up to you. My guess is that a rebuild will work for now though.

Kully:
> Huh, weird intermittent sign-in failure... If it keeps happing, you could mock the users.current request so that it just skips the validation part. Up to you. My guess is that a rebuild will work for now though.

I tried rebuilding once and the same error occurred

theengineear:
LGTM, *but*, is there an existing test that you can improve to cover this? The `violin` coverage is only at 62%, which isn't great. https://1894-14579099-gh.circle-artifacts.com/0/tmp/circle-artifacts.mLgBx3K/coverage-reports/py27-optional/plotly_figure_factory__violin_py.html

💃 as is, but it'd be great to add some test coverage 😸

theengineear:
>I tried rebuilding once and the same error occurred

Did it occur for the same Python version in the same test? I ask because it's a little odd.

Kully:
> Did it occur for the same Python version in the same test? I ask because it's a little odd.

It looks like one was Python 3.4.3 and the other Python 3.3.3.

Kully:
> 💃 as is, but it'd be great to add some test coverage 😸

I didn't know about this test coverage page, sooo cool! Sure, I'll some before merging then ping you.

theengineear:
>I didn't know about this test coverage page

ya! definitely good to check in with every now and again!

Kully:
And scatterplotmatrix is at 41% - ouch!

I think a lot of the coverage is coming from not checking all types of violins or scatterplot matrices.

Kully:
Change of plans, I'm just going to merge this guy.

I don't really understand how the coverage page works...it says https://1894-14579099-gh.circle-artifacts.com/0/tmp/circle-artifacts.mLgBx3K/coverage-reports/py27-optional/plotly_figure_factory__violin_py.html#n556 is not covered for instance, even though there _is_ a test for it. And if I decided to test `violin_colorscale` then I don't know if it would increase the coverage %.

Can you illuminate how this works?

theengineear:
@Kully the link you pasted suggests that the following condition is not met:

`if pd and isinstance(data, pd.core.frame.DataFrame):`

Therefore, none of the code inside that is ever touched. That means that either `pd` is not imported or `data` is not an instance of `pd.core.frame.DataFrame`.

PR #678, opened 2017-01-30
Speed up graph objs

### LINKED ISSUE #497: Graph_obj.py slow performance when .append_trace on large amount of subplots
I just started using Plotly with Python and I think I have found a significant speed up improvement in the graph_objs.py module.

Here are some details:
- I am creating a plot with 24 subplots that has 60 traces in each subplot
  - Each trace contains 6 points.
- To generate this plot it took 1945 seconds, which seems like way too long
  - I had a print state for each subplot and it progressively takes longer for each subplot
- I ran “kernprof” on the code looking specifically at the subroutine that I use to loop through the data to .append_trace to the figure
  - I noticed that all the time was spent on the .append_trace function that is called out of graph_objs.py
- As I dug through graph_objs.py and found that line 936 contained “self[‘data’] += [trace]”
  - In my limited understanding of python, I always heard this was a very inefficient way to do large string appends
- I changed line 936 to “self[‘data’].append(trace)”, which I believe is equivalent in function to “self[‘data’] += [trace]” but I think I have read that this is a much more efficient way to append
  - This sped up the script to 8 seconds! A 243x improvement in speed.

From what I can tell, I think it is valid to change “self[‘data’] += [trace]” to “self[‘data’].append(trace)” but I am not sure of all the ramifications to other functionality. Is this a valid change to make and if so can graph_objs.py be updated with this?

Thanks!

theengineear:
Introduces a utility decorator to memoize functions along with some other performance improvements surrounding `graph_reference.

This was something I tinkered on over the weekend. I spent a bit more time today formalizing it so that the work didn't go to waste.

Closes #497

^^ I'm closing that issue, but it's likely that there are other improvements to be had. My suggestion would be to separate `make_subplots` so that users can *choose* whether they'd like to use `graph_objs` or native Python objects to create a plot. That can/should be handled as a different issue though.

### DIFF
--- a/plotly/tests/test_core/test_utils/test_utils.py
+++ b/plotly/tests/test_core/test_utils/test_utils.py
@@ -1,10 +1,12 @@
 from __future__ import absolute_import
 
+from inspect import getargspec
 from unittest import TestCase
 
 from requests.compat import json as _json
 
-from plotly.utils import PlotlyJSONEncoder, get_by_path, node_generator
+from plotly.utils import (PlotlyJSONEncoder, get_by_path, memoize,
+                          node_generator)
 
 
 class TestJSONEncoder(TestCase):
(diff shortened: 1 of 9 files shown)


theengineear:
@Kully, wanna put some 👀 on this?

//cc @depettit, thanks for the clear issue documentation. This targets the `make_subplots` portion. It's quite possible that the `append_trace` is sill going to be problematic for you. I'll likely ask for your help in manually testing this after this PR is merged into master.

Kully:
> @Kully, wanna put some 👀 on this?
Sure, tomorrow most likely :+1:

dhirschfeld:
FWIW I always use the [`decorator`](http://pythonhosted.org/decorator/documentation.html) module to implement signature/docstring preserving decorators. If the decorated functions aren't user-facing then it's possibly not a big issue however I've found it useful when debugging my own code to have the decorated functions have meaningful signatures and docstrings.

The added utility does have to be weighed up against another dependency but these days with conda I don't see that as such a big issue and it's already a dep for sever other widely used packages.

theengineear:
Oops, thanks @dhirschfeld , I meant to preserve the docstrings but forgot 😓.

I typically use something like [wraps](https://docs.python.org/2/library/functools.html#functools.wraps). What benefits beyond something like `wraps`, does the `decorator` package provide?

theengineear:
FWIW, these decorated functions aren't really meant to be user-facing, *but*, it was my intention to make a more general-purpose decorator that would preserve names/docs.

theengineear:
OK, just read through the docs for decorate. Looks great! I'll try to
rewrite using this, seems like the dep is pretty small.

dhirschfeld:
The decorator module will preserve docstring and signature on both py2/3 *and* will ensure `inspect.getargspec` also works correctly.

From a quick test, much of this is mitigated on Py3 where wraps also preserves the signature and where `getargspec` is deprecated in favour of `inspect.signature` which does work with wrapped functions.

My mantra of always using the `decorator`  module comes from a long history of Py2 usage :stuck_out_tongue: . Still, if you have to support both it's probably worth considering...

theengineear:
@dhirschfeld , I took you up on the suggestion. Thanks!

@chriddyp the `decorator` module is a [dep for ipython](https://github.com/ipython/ipython/blob/master/setup.py#L204), so it's likely that folks will have it already. It's fairly small and provides some nice Python-version-independent functionality. Given it's size, I'm a proponent of adding it as a dep, but I figured I'd ping you on it.

theengineear:
Additionally, I should note that instead of letting callers specify a `key_function`, I decided it was better form to nudge developers to just create functions that are *naturally* memoizable (all args/kwargs hashable).

theengineear:
@Kully unless there's pushback on the new `decorator` dep, I'm done futzing with this PR.

plotly/tools.py
```diff
@@ -1000,8 +1000,7 @@ def _get_anchors(r, c, x_cnt, y_cnt, shared_xaxes, shared_yaxes):
     # Function pasting x/y domains in layout object (2d case)
     def _add_domain(layout, x_or_y, label, domain, anchor, position):
         name = label[0] + 'axis' + label[1:]
-        graph_obj = '{X_or_Y}Axis'.format(X_or_Y=x_or_y.upper())
-        axis = getattr(graph_objs, graph_obj)(domain=domain)
+        axis = {'domain': domain}
```
Kully: I suppose you did this because `axis` will never be equal to anything other than `{'domain': domain}`, and so it's like a trivial case of memoization? :)

plotly/tests/test_core/test_utils/test_utils.py
```diff
@@ -50,3 +52,98 @@ def test_node_generator(self):
...
+        @memoize()
+        def add(a, b):
+            name_space.call_count += 1
+            return a + b
+
+        tests = [[(1, 1), 2], [(2, 3), 5], [(3, -3), 0]]
+
+        self.assertEqual(name_space.call_count, 0)
+        for i, (inputs, result) in enumerate(tests, 1):
+            for _ in range(10):
```
Kully: Out of curiosity, is `_` the canonical (or a typical) variable used for loops where the looping variable is not used inside?

plotly/tools.py
theengineear: this isn't about memoization as much as it's about performance. the line `graph_obj = '{X_or_Y}Axis'.format(X_or_Y=x_or_y.upper())` was eating up a ton of cpu time because it was instantiating that `graph_object` and then immediately setting it inside `layout`, which effectively was instantiating it again.

plotly/utils.py
```diff
@@ -444,3 +446,47 @@ def set_sharing_and_world_readable(option_set):
...
+
+def memoize(maxsize=128):
+    """
+    Memoize a function by its arguments. Note, if the wrapped function returns
+    a mutable result, the caller is responsible for *not* mutating the result
+    as it will mutate the cache itself.
+
+    :param (int|None) maxsize: Limit the number of cached results. This is a
+                               simple way to prevent memory leaks. Setting this
+                               to `None` will remember *all* calls. The 128
```
theengineear: Yah, this mimics [lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache). I'd have tried to use this, but it's not available in Python < 3.2.

plotly/tests/test_core/test_utils/test_utils.py
theengineear: yup, `_` is the common placeholder when you don't intent to use the value.

theengineear:
@Kully , what's the word, 🐦? Can this 💃?

@chriddyp , can I get a confirmation that adding the `decorator` dep is OK? Again, IPython uses it, it's PY2/3 agnostic, and it's quite small (nudge, nudge, nudge).

chriddyp:
> can I get a confirmation that adding the decorator dep is OK? Again, IPython uses it, it's PY2/3 agnostic, and it's quite small (nudge, nudge, nudge).

yup!

Kully:
> Can this 💃?

Yup! 💃  :man_dancing:  👯

PR #679, opened 2017-01-31
refactor the files.py

srinivasreddy:


theengineear:
@srinivasreddy , can you comment on why this is necessary, looks like you've just made the context that the file is opened in more explicit?

srinivasreddy:
The purpose of the refactor is to group related statements to function. And introducing better coding standards too.

theengineear:
@srinivasreddy ok, looks fine by me. Thanks 😸

PR #680, opened 2017-02-01
Enable heartbeat in Stream.write

AJamesPhillips:
The [plot.ly streaming documentation](https://plot.ly/streaming/) under the
`Disconnections` section states that:

> A connection can be maintained by writing a heartbeat within the 60 second
> window, a heartbeat is simply a newline

If an empty string (and maybe an empty dictionary too, untested) is passed to
Stream's `write` method however it results in an empty row being created in the plot's
data rather than only keeping the connection awake.  This commit aims to support
using the write method for sending a heartbeat.

### DIFF
--- a/plotly/plotly/chunked_requests/chunked_request.py
+++ b/plotly/plotly/chunked_requests/chunked_request.py
@@ -69,8 +69,12 @@ class Stream:
             msglen = format(len(msg), 'x')  # msg length in hex
             # Send the message in chunk-encoded form
             self._conn.sock.setblocking(1)
-            self._conn.send('{msglen}\r\n{msg}\r\n'
-                            .format(msglen=msglen, msg=msg).encode('utf-8'))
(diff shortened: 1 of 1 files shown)


AJamesPhillips:
Happy to write a test for it if it's something that is desired to be added to the repo and if I have the implementation details correct.  I haven't tested it, I'm going solely by the documentation.

theengineear:
@AJamesPhillips , I think this is *supposed* to be solved with https://github.com/plotly/plotly.py/blob/master/plotly/plotly/plotly.py#L544. I.e., it's solved at a higher level than in the `chunked_request` module. Does this not work, or is it not part of your workflow?

AJamesPhillips:
@theengineear good spot.  Thank you very much.  I'll check it out on Friday next week when I get back to that project.

jonmmease:
Closing as stale. Feel free to open a new issue if you're still having trouble with this. Thanks!

PR #682, opened 2017-02-01
remove support email from error messages

cldougl:
since the [support app](https://support.plot.ly) has been released the support@plot.ly email is no longer in use. 
cc @theengineear @Kully


theengineear:
💃 , thanks @cldougl 😸

Issue #515, opened 2016-06-30
MathJax not loaded in plotly offline


rreusser:
[`plotly.offline.plot`](https://github.com/plotly/plotly.py/blob/master/plotly/offline/offline.py#L314) fails (shows an unstyled plot after bailing out early due to errors) if latex is used since MathJax is not loaded in the default html. Short of getting fancy, it seems like `plotly.offline.plot` might need a manual option to inject MathJax from a cdn into the head tag.

yankev:
Thanks Ricky, I'll check this out.

physinet:
#105 addresses this as well. When using `plotly.offline.iplot`, the plot looks all right, except that LaTeX is not rendered in any of the text labels.

mmlanger:
This is actually the only reason that stops me using plotly for my publication graphics.
No LaTeX in offline plotly is a serious issue, is there any news on this?
#444 and #445also adresses latex problems with plotly.

szsiddiq:
FYI, sometimes after executing iplot, the LaTeX I have else where in my notebook (outside of plotly) breaks and stops rendering. I'm having trouble replicating this on sample data that I can share, but will update with code if I'm able to do so.

jjaraalm:
This is not really an issue with plotly.py per-se, but is the result of issues upstream at plotly.js.

plotly.js does not encapsulate mathjax but instead relies on it being loaded globally and configured in a specific way. The problem is that there are multiple different output renderers for mathjax. Juypter globally loads mathjax with HTML output, but plotly.js requires SVG output. A quick and easy fix is just to change the output renderer for mathjax globally as this does not affect Juypter's markdown rendering. 

In order to address #445, mathjax should also be configured to set displayAlign = center since plotly.js explicitly sets this to left align. As far as I can tell in simple testing, the display alignment actually has no bearing on the output rendered by plotly.

Right now the code block I'm using to fix all these issues is:

```python

# Startup plotly
init_notebook_mode(connected=True)

# The polling here is to ensure that plotly.js has already been loaded before
# setting display alignment in order to avoid a race condition.
display(HTML(
    '<script>'
        'var waitForPlotly = setInterval( function() {'
            'if( typeof(window.Plotly) !== "undefined" ){'
                'MathJax.Hub.Config({ SVG: { font: "STIX-Web" }, displayAlign: "center" });'
                'MathJax.Hub.Queue(["setRenderer", MathJax.Hub, "SVG"]);'
[...]

PR #683, opened 2017-02-02
first pass at adding mathjax

chriddyp:
Using the workaround as suggested here:
https://github.com/plotly/plotly.py/issues/515#issuecomment-263097892

### DIFF
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -140,7 +140,20 @@ def init_notebook_mode(connected=False):
             '</script>'
             '').format(script=get_plotlyjs())
 
+    mathjax_script = (
+        '<script>'
+        'var waitForPlotly = setInterval( function() {'
+        '   if( typeof(window.Plotly) !== "undefined" ){'
+        '       MathJax.Hub.Config('
+        '           { SVG: { font: "STIX-Web" }, displayAlign: "center" });'
+        '       MathJax.Hub.Queue(["setRenderer", MathJax.Hub, "SVG"]);'
+        '       clearInterval(waitForPlotly);'
+        '   }}, 250 );'
+        '</script>'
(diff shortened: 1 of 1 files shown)


chriddyp:
The codeblock as mentioned by @jjaraalm doesn't actually work for me. Perhaps that patch only works for a certain version of Jupyter?

image

The error from mathjax is pretty cryptic, I'm not very familar with how this works: https://github.com/mathjax/mathjax-docs/wiki/'Can't-make-callback-from-given-data'-error-if-resetEquationNumbers-is-called-when-no-math-is-typeset

jjaraalm:
Wow. Completely forgot about this ... was going to make a PR at one time but just never did. I'll check this out tonight and see if I can help figure out what's going on.

chriddyp:
@jjaraalm - that would be great, thank you!

chriddyp:
Alternatively, sometimes I get this message
image

jjaraalm:
I'm not having any issues on my end, so I'm going to just try and take a stab at what's going on for you. To me it looks like a configuration issue with the way Jupyter is setup or installed. In the first screenshot, you'll get that error if MathJax hasn't been started (by calling `MathJax.Hub.Configured()`), but both Jupyter and plotly.js should be calling that. For the second screenshot, it's failing to load the SVG config file `jax/output/SVG/config.js` for some reason. Can you check to make sure you have the file? If MathJax can't load the render config file, it looks like it just keeps trying and enters into an infinite loop.

Also, looks like you're running an older version of Jupyter. Current Jupyter ships with MathJax 2.6, but you have 2.5.3. Not sure if this is important though.

jonmmease:
superseded by work in https://github.com/plotly/plotly.py/pull/1169

PR #688, opened 2017-02-06
Reduce size V3

fu:
Hi,

updated pull request. See #501 for more details.

Cheers

.c


jonmmease:
This functionality is being added in #1234 using a slightly different approach.  With this PR, the `include_plotlyjs` argument to `plotly.offline.plot` accepts a few options in addition to `True` and `False`.  In particular, setting `include_plotlyjs='directory'` results in the behavior of this PR (a copy of plotly.min.js is copied to the output directory if it does not already exist, and the script tag references this local copy).

I'd like to avoid having this be a one-off global flag for the time being. Eventually I hope we'll be able to revamp our settings system a bit to make it possible to register default values for options like `include_plotlyjs`.

Thanks for your work on this @fu, the `include_plotlyjs='directory'` option wouldn't have been included in #1234 without it!

PR #693, opened 2017-02-13
Fix distplot color looping

Kully:
Re: a user's issue. Looks like if you have more than 10 traces in a `distplot`, your colors won't loop around and start on the first color. Just did `mod len(colors)`.


Kully:
@cldougl @theengineear <- Whoever can get to it first. :)

cldougl:
this looks good to me @Kully - just wondering if this is an issue with any other figurefactory traces as well?

cldougl:
Looks like this was the only one 
💃

Kully:
> Looks like this was the only one
> 💃

fantastic!

Issue #646, opened 2016-12-31
API proposal for create_dashboard() and update_dashboard()


jackparmer:
Starting a discussion for a wrapper proposal for V2 dashboard creating and updating:

https://api.plot.ly/v2/dashboards#create

I'm not aware of any formal dashboard JSON spec documentation, but this dashboard is almost exhaustive in its options usage:

https://plot.ly/dashboard/jackluo:430/view
JSON: https://plot.ly/dashboard/jackluo:430/view.json

The only thing that it is missing is a table (an embedded plotly grid), which this dashboard has:

https://plot.ly/dashboard/jackp:16818
https://plot.ly/dashboard/jackp:16818.json

All of the options for styling embedded plotly grids (through URL query parameters) are here:
http://help.plot.ly/add-a-table-to-spectacle-editor/

## Creating a dashboard

`create_dashboard()`

This call to `create_dashboard` would create the first 2 rows of this dashboard, plus a 3rd row with a table and a markdown cell:

https://plot.ly/dashboard/jackluo:430/view

```
# First row has 1 graph
row1 = list( dict( type='plot', url='https://plot.ly/~jackluo/400' ) )

# Second row has 2 graphs
# These dashboard cells have titles
row2 = list(
    dict( type='plot', url='https://plot.ly/~jackluo/402',
        title='Average rotor diameter by manufacturer' ), 
    dict( type='plot', url='https://plot.ly/~jackluo/404', 
        title='Number of turbines by manufacturer' ) )

markdown_text = ""## Jack Luo | Plotly\n\nDownload original dataset here:     \nhttp://www.nature.com/articles/sdata201560\n"

table = dict(
    type = 'table', 
    url = 'https://plot.ly/~datasets/2798',
    show_row_numbers = False,
    text_transform = 'uppercase',
    header_font_weight = 300,
    header_background = '%23ab63fa'
)

# Third row has 1 table and 1 text box
# All table styling options are here: http://help.plot.ly/add-a-table-to-spectacle-editor/
row3 = list(
    table,
    dict( type='text', text=markdown_text ), 
)

dashboard = dict(
    rows = list( row1, row2, row3 ),
    foreground_color = "#cccccc", 
    box_background_color = "#020202", 
    links = [ { "url": "http://www.nature.com/articles/sdata201560", 
                "title": "| Download dataset here" } ], 
    title = "US Wind Turbine Dataset", 
    box_border_color = "#020202", 
    header_foreground_color = "#cccccc", 
[...]

PR #694, opened 2017-02-15
Dashboard wrapper

Kully:


Kully:
Dashboard discussion thread: https://github.com/plotly/plotly.py/issues/646

A few notes:
-I need to add HTML preview representation. Current idea is to have a `dashboard_html` string that just represents the HTML for a rectangle. I want to add a couple methods that add lines/numbers to this representation and put them in the `Dashboard` class (it would just append to `dashboard_html`). Since we have to be able to pull dashboards from online, we would need to update the HTML preview in one-shot and _not_ iteratively where it only updates each time you insert. Clearly if we are getting a JSON form of the dashboard online in one piece, we should be able to construct the HTML when we need it.

-hidden in the Dashboard `dict` is a `box_ids_dict` (I should probably rename to `box_id_paths` to be clear, but wanted to get this PR started :+1:) and that is what keeps track of the number associated to each box. So a single item in there looks like `{0: ['first', 'first', 'second']}`.

- strings docs aren't really finished. I'll worry about them once I know for sure I'm on the right track and get the thumbs up.

- no tests for the same reason above.

- need to add a few more manipulation methods (i.e. swap, rotate, resize). Shouldn't be too strenuous.

cc @theengineear

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,302 @@
...
+        self['first'] = box_1
+        self['second'] = box_2
+
+
+class Dashboard(dict):
+    def __init__(self, dashboard_json=None, backgroundColor='#FFFFFF',
+                 boxBackgroundColor='#ffffff', boxBorderColor='#d8d8d8',
+                 boxHeaderBackgroundColor='#f8f8f8', foregroundColor='#333333',
+                 headerBackgroundColor='#2E3A46', headerForegroundColor='#FFFFFF',
+                 links=[], logoUrl='', title='Untitled Dashboard'):
```
theengineear: The more I look at all these, the more I wonder whether we need non-native objects at all? We're just using these as dicts and the user isn't meant to interact with anything here...

What do you think of just having a top-level *interface* to an underlying dict? I.e., the `Dashboard` is just an interface to make it easier to manipulate a dashboard dict?

Initialization would look something like this:

```
class Dashboard(object):

    def __init__(self, content=None):
        if content is None:
            content = {}
        if not isinstance(content, dict):
            raise TypeError('Content must be a dict.')

        # Ensure top-level attributes exist.
        content['version'] = 2
        if content.get('settings') is None:
            content['settings'] = {}
        if content.get('layout') is None:
            content['layout'] = {}

        # Defer initialization of layout to general method.
        self._initialize_split(content['layout'])

        self.content = content

    def _initialize_box(self, content):
        # Ensure top-level attributes exist.
        content['type'] = 'box'

        # yadda, yadda, more logic to decide what *kind*
        # of box this is and make sure to initialize it right.

    def _initialize_split(self, content):
        # Ensure top-level attributes exist.
        content['type'] = 'split'
        if content.get('first') is None:
            content['first'] = {}
        if content.get('second') is None:
            content['second'] = {}

        # yadda, yadda, more logic to decide what else we
        # need to properly initialize the top-level attributes of
        # the split

        # Defer initialization to general methods.
        if content['first'].get('type') == 'split':
            self._initialize_split(content['first']):
        else:
            self._initialize_box(content['first'])

        if content['second'].get('type') == 'split':
            self._initialize_split(content['second']):
        else:
            self._initialize_box(content['second'])
```

Pros:

* No need to inherit from `dict`, which means we have a handy separation-of-concerns. I.e., users create a `Dashboard` instance so that they can more-easily manipulate the `content`, which is a simple `dict` (no frills at all).
* No confusing call signatures. The less-specific we are, the better here. We want as many details as possible to be handled by the backend API *not* by the api wrapper here.
* Only one new object for users to grok.

theengineear: ^^ what do you think about this? Attempting to add too many frills has been a problem for us in the past. I don't want to repeat old mistakes if there's no reason to.

The important thing to note here is that the backend API should remain the ultimate source of truth. Dashboards are small JSON blobs; let's just throw whatever the user puts in there at our API and report back success/failure.

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,302 @@
...
+                    finished_iteration = True
+
+            for node in all_nodes:
+                if (node[1] != () and node[0]['type'] == 'box' and
+                    node[0]['boxType'] != 'empty'):
+                    try:
+                        max_id = max(self.box_ids_dict.keys())
+                    except ValueError:
+                        max_id = 0
+                    self.box_ids_dict[max_id + 1] = list(node[1])
```
theengineear: I don't think it makes sense to *ever* cache this information. `box_ids_dict` should be calculated *every time* it is required (or... you'd need to figure out a smart way to do the caching...).

Consider the following steps:

1. instantiate a dashboard (so you set `self.box_ids_dict` during instantiation)
2. manually change the dashboard in some way (which should be allowable for simplicity)
3. hrm, was the `Dashboard` instance smart enough to update the box_ids? Yes? No? Shrug?

Imo, you should do the dumb thing (which also turns out to be the more-testable thing) and just calculate these box ids on the fly each time.

Thoughts?

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,302 @@
...
+
+    def _insert(self, box_or_container, array_of_paths):
+        """Performs user-unfriendly box and container manipulations."""
+        if any(path not in ['first', 'second'] for path in array_of_paths):
+            raise exceptions.PlotlyError(
+                "Invalid path. Your 'array_of_paths' list must only contain "
+                "the strings 'first' and 'second'."
+            )
```
theengineear: 🐄 this is a private function, do we really need to protect it from bad call arguments? if you feel like it's important to protect it, that's OK, but it may not be necessary. Just a thought :)

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,302 @@
...
+            for node in all_nodes:
+                if (node[1] != () and node[0]['type'] == 'box' and
+                    node[0]['boxType'] != 'empty'):
+                    try:
+                        max_id = max(self.box_ids_dict.keys())
+                    except ValueError:
+                        max_id = 0
+                    self.box_ids_dict[max_id + 1] = list(node[1])
+
+    def _insert(self, box_or_container, array_of_paths):
```
theengineear: 🐄 I think `path` would be a better name than `array_of_paths`.

plotly/dashboard_objs/dashboard_objs.py
Kully: That's probably a better idea since that's exactly what I'm doing anyways when I pull a dashboard from online. I have code in here for dealing with that so reusing the same code would probably be best. So a big :+1: from me

plotly/dashboard_objs/dashboard_objs.py
Kully: Yeah, just wanted to have it in there to show my current reasoning and to show what I want to prevent. Will have to look more closely and see if I can reorganize/remove it from the private function.

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,302 @@
...
+                )
+            elif side == 'right':
+                old_box = self._get_box(box_id)
+                self._insert(
+                    Container(old_box, box, direction='horizontal'),
+                    self.box_ids_dict[box_id]
+                )
+
+
+def upload_dashboard(dashboard_object, filename, world_readable,
```
Kully: This is just a semantic thing right, since a dashboard is JSON and a `Dashboard` object is also JSON (a `dict`)? The renaming is for clarity in what can be passed, correct?

theengineear: I'm thinking of `Dashboard` being the class and a `dashboard` being either a *dict` or a `Dashboard` here. So:

`upload_dashboard(Dashboard(), ..)`

OR

`upload_dashboard({}, ..)`

^^ is that clearer?

Kully: And by class you mean the instance of a class correct?  I think I get that, and it makes sense.

Kully:
@theengineear @chriddyp @jackparmer 
Alright. So here's what's going on.

-You can now use `.insert(box, side=,box_id)` on your dashboard and feed it a box you define (as a dict)
-the `.get_preview()` works well

Need to add/improve:
-dashboard_obj.upload() needs to be able to update existing filenames. I need to use `update` or `partial_update` which are already in v2 request functions, but it seemed to error out for me. Any ideas?
-the `size` of each `_container()` is fixed at the display_height of the HTML representation (400 rn) but this needs to programatically change. In particular, if a user is making a dashboard on the fly, they don't wanna have to specify that your size is decreasing as you are shoving boxes into the layout. If you don't touch the size, the dashboard will have a bunch of plots that only display titles (they are collapsed). Need to implement a way of dealing with this.
-need to add a resize method to alter the sizes of the boxes in the dashboard layout, if they want to be changed
-need to update the HTML representation to reflect a change in container sizes.
-all doc strings should be properly written out
-tests

Alright, let me know your thoughts and please pound me with as many questions as you want to ask!

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+        for node in all_nodes:
+            if (node[1] != () and node[0]['type'] == 'box'
+                    and node[0]['boxType'] != 'empty'):
+                try:
+                    max_id = max(self.box_ids_to_path.keys())
+                except ValueError:
+                    max_id = 0
+                self.box_ids_to_path[max_id + 1] = list(node[1])
+
+    def _insert(self, box_or_container, path):
```
theengineear: 🐄 let's try to nail down our language around `boxes`, `containers` and `splits`?

I think the `defs` in the schema are a little misleading (that's my fault, but it's not something we can't change)... anyhow what do you think about the following statements?

1. A `box` is a `container` with `type: 'box'`
2. A `split` is a `container` with `type: 'split'`
3. A `plotBox` is a `box` with `boxType: 'plot'`
4. An `emptyBox`...
5. etc etc

Would that be helpful? `box_or_container` feels weird to me... anyways sort of just something for discussion, not critical.

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+
+        if 'first' in self['layout']:
+            loc_in_dashboard = self['layout']
+            for index, first_second in enumerate(path):
+                if index != len(path) - 1:
+                    loc_in_dashboard = loc_in_dashboard[first_second]
+                else:
+                    loc_in_dashboard[first_second] = box_or_container
+
+        else:
```
theengineear: I don't get it, why can't `path` be `('second', 'second')`? or something?

theengineear: 🎎 ?

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+            self['layout'] = box_or_container
+
+    def get_box(self, box_id):
```
theengineear: Is this required as part of the api for users? I.e., why would a user use this?

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+        self._assign_boxes_to_ids()
+
+        loc_in_dashboard = self['layout']
+        for first_second in self.box_ids_to_path[box_id]:
+            loc_in_dashboard = loc_in_dashboard[first_second]
+        return loc_in_dashboard
+
+    def _path_to_box(self, path):
+        """Returns box from specified path."""
+        self._assign_boxes_to_ids()
```
theengineear: ah, here's a good example. if you had `_assign_boxes_to_ids()` just *return* the computation, then you could just change the call signatures of functions like these to:

```
def _path_to_box(self, box_ids, path)
```

or somesuch...

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+        while not finished_iteration:
+            try:
+                all_nodes.append(node_gen.next())
+            except StopIteration:
+                finished_iteration = True
+
+        all_paths = []
+        for node in all_nodes:
+            all_paths.append(list(node[1]))
+        if ['second'] in all_paths:
```
theengineear: hmmm. this is fragile. i thought your paths were `tuples`, no? is there a better way to do this?

theengineear: also, what's the deal with `['second']` is it always just an empty box? non-existent? not allowed?

theengineear: ^^ i.e., 🎎

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+            all_paths.remove(['second'])
+
+        max_path_len = max(len(path) for path in all_paths)
+        # search all paths of the same length
```
theengineear: 🎎 , why?

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+                        current_box_specs['top_left_x'],
+                        current_box_specs['top_left_y'],
+                        current_box_specs['box_w'],
+                        current_box_specs['box_h'],
+                        direction=self._path_to_box(path)['direction']
+                    )
+
+                    # determine the specs for resulting two boxes from split
+                    is_horizontal = (
+                        self._path_to_box(path)['direction'] == 'horizontal'
```
theengineear: Not that it probably matters much, but you end up re-computing box ids as you have this written right now. See my suggestion in https://github.com/plotly/plotly.py/pull/694/files#r103069681

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+                    )
+                    x = current_box_specs['top_left_x']
+                    y = current_box_specs['top_left_y']
+                    box_w = current_box_specs['box_w']
+                    box_h = current_box_specs['box_h']
+
+                    new_box_w = box_w*(1 - is_horizontal*0.5)
```
theengineear: pep:8ball: space around `*`

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+                    new_box_h = box_h*(1 - (not is_horizontal)*0.5)
```
theengineear: 🐄 Also, `boolean` * `float`? C'mon. ;)

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+
+                    box_1_specs = {
+                        'top_left_x': x,
+                        'top_left_y': y,
+                        'box_w': new_box_w,
+                        'box_h': new_box_h
+                    }
+                    box_2_specs = {
+                        'top_left_x': (x + is_horizontal*0.5*box_w),
+                        'top_left_y': (y + (not is_horizontal)*0.5*box_h),
```
theengineear: Again, this is pretty confusing code.

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+        # display HTML representation
+        if ipython:
+            return IPython.display.HTML(html_figure)
+
+    def insert(self, box, side='above', box_id=None):
+        """
+        The user-friendly method for inserting boxes into the Dashboard.
+
+        box: the box you are inserting into the dashboard.
+        box_id: pre-existing box you use as a reference point.
```
theengineear: 🎎  Pls use our standard docstring format for these args.

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+            'boxType': 'plot',
+            'fileId': '',
+            'shareKey': None,
+            'title': ''
+        }
+
+        # force box to have all valid box keys
+        for key in init_box.keys():
+            if key not in box.keys():
+                box[key] = init_box[key]
```
theengineear: 🐫, I have a feeling you'll want to factor this out and generalize it.

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+
+    def insert(self, box, side='above', box_id=None):
+        """
+        The user-friendly method for inserting boxes into the Dashboard.
+
+        box: the box you are inserting into the dashboard.
+        box_id: pre-existing box you use as a reference point.
+        """
+        self._assign_boxes_to_ids()
+        init_box = {
```
theengineear: Shouldn't this initializer be based on the given `box['type']`?

theengineear:
I sorta stopped part-way through since I'm being pretty vocal. I don't want to be *too* overwhelming all-at-once :)

plotly/api/v2/dashboards.py
```diff
@@ -0,0 +1,47 @@
...
+def retrieve(fid, share_key=None):
+    """Retrieve a dashboard from Plotly."""
+    url = build_url(RESOURCE, id=fid)
+    return request('get', url)
+
+
+def update(fid):
+    """Completely update the writable."""
+    url = build_url(RESOURCE, id=fid)
+    return request('put', url)
```
Kully: How so? I briefly looked at https://api.plot.ly/v2/dashboards#update but yeah, I was going to attempt fixing the overwrite-filename limitation on upload, but just threw update in there to start

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+from plotly import exceptions, optional_imports
+from plotly.utils import node_generator
+
+IPython = optional_imports.get_module('IPython')
+
+# default variables
+master_width = 400
+master_height = 400
+container_size = master_height
+font_size = 10
```
Kully: okay, np

plotly/api/v2/dashboards.py
theengineear: 😛 because the call signature doesn't give you any way to make an update.

theengineear: `def update(fid, update)`? or `def update(fid, content)`, or something?

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+
+dashboard_html = ("""
+<!DOCTYPE HTML>
+<html>
+  <head>
+    <style>
+      body {{
+        margin: 0px;
+        padding: 0px;
+      /}}
```
Kully: I dunno, I copied a template for the html bit...I can test without

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+            self['settings'] = {}
+        else:
+            self['layout'] = content['layout']
+            self['version'] = content['version']
+            self['settings'] = content['settings']
+
+            self._assign_boxes_to_ids()
+
+    def _assign_boxes_to_ids(self):
+        self.box_ids_to_path = {}
```
Kully: What would passing it around my functions look like?

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+        all_nodes = []
+        node_gen = node_generator(self['layout'])
+
+        finished_iteration = False
+        while not finished_iteration:
+            try:
+                all_nodes.append(node_gen.next())
+            except StopIteration:
+                finished_iteration = True
```
Kully: Yeah, so running `.next()` on the `node_gen` object will actually mutate, specifically remove a set of nodes. Eventually it will be empty and `node_gen` will result in an error, so in order to grab all the nodes and remember them, I store them all in a list `all_nodes`

plotly/api/v2/dashboards.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def partial_update(fid):
+    """Partially update the writable."""
+    url = build_url(RESOURCE, id=fid)
+    return request('patch', url)
```
Kully: > I.e., why can't we just do for node in node_generator(self['layout'])

I'll give this a try right now and see where we get... 👍

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+        'boxType': 'plot',
+        'fileId': fileId,
+        'shareKey': shareKey,
+        'title': title
+    }
+    return box
+
+
+def _container(box_1=_empty_box(), box_2=_empty_box(), size=container_size,
+               sizeUnit='px', direction='vertical'):
```
Kully: Anyway to fix that easily? It needs to be a `dict` so I can stick it in the dashboard.

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+        node_gen = node_generator(self['layout'])
+
+        finished_iteration = False
+        while not finished_iteration:
+            try:
+                all_nodes.append(node_gen.next())
+            except StopIteration:
+                finished_iteration = True
+
+        for node in all_nodes:
```
Kully: I'll give it a try

plotly/dashboard_objs/dashboard_objs.py
Kully: 1. and 2. is _really_ confusing and non-intuitive for me. It's funny how two people can be on both sides of this where the intuitive thing for one is not intuitive for the other, but anyways...

I don't like how in 1. we're saying "A `box` is a `CONTAINER`" because the object with `type: 'split'` that holds a `first` and `second` has the duck-like (am I using that right?) to something "that contains". Calling a box a container just violates my idea of what a container should be and what properties, namely having children in `first` and `second`, a container should have.

I get the logic in 3. and 4.; `plotBox` and `emptyBox` are subsets of a `box`, specifying the type.

What do you think about this? Alternatively, I could switch up the terminology from `container` to `split` (so `box_or_container` becomes `box_or_split`)

plotly/dashboard_objs/dashboard_objs.py
Kully: Am I being unclear with `first_second`? That variable more accurately means `first_or_second`, so it can take values `first` or `second`.

On another or related note, `dashboard['layout']['second']` is *always* an emptyBox, so no path will ever look like `('second', 'second', 'first', ...)`

plotly/dashboard_objs/dashboard_objs.py
Kully: Ahh, good question! It _is_ intended and it's meant to view the box by specifying the box_id, vis a vis looking at the HTML representation. If you swap box_ids 1 and 2 for example, you can verify where the resulting boxes will go.

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+            loc_in_dashboard = loc_in_dashboard[first_second]
+        return loc_in_dashboard
+
+    def get_preview(self):
+        """Returns JSON and HTML respresentation of the dashboard."""
+        # assign box_ids
+        self._assign_boxes_to_ids()
+
+        # print JSON
+        pprint.pprint(self)
```
Kully: Do we not want to do the pretty print if IPython is None?

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+
+        # construct HTML dashboard
+        x = 0
+        y = 0
+        box_w = master_width
+        box_h = master_height
+        html_figure = copy.deepcopy(dashboard_html)
```
Kully: Okay so just `html_figure = dashboard_html`?

plotly/dashboard_objs/dashboard_objs.py
Kully: exactly that's the deal. always an empty box, as commented above.

plotly/dashboard_objs/dashboard_objs.py
Kully: Why did I write that?

I guess I don't need that comment. I could get rid of it, as I'm just going through all the paths and iteratively picking the paths with a given length (length 1, length 2, ...)

plotly/dashboard_objs/dashboard_objs.py
Kully: Wait, why not? Is it just a little clunky? I don't have to rewrite code and I like the simplicity.

plotly/dashboard_objs/dashboard_objs.py
Kully: yeah, agreed, but I like the compactness. It reminded me of a simpler time...(math at uni)

plotly/dashboard_objs/dashboard_objs.py
Kully: haha, yeah I half assed it. I'll fix it all up

plotly/dashboard_objs/dashboard_objs.py
theengineear: The same as usual :)

```
def _container(box_1=None, ...):
   if box_1 is None:
        box_1 = _empty_box()
```

plotly/dashboard_objs/dashboard_objs.py
theengineear:
```
def _compute_box_ids(self):
    # compute, compute compute
    return box_ids_to_path

def _foo(self, box_ids_to_path):
    # i don't need to compute box_ids_to_path...
    # do foo

def _bar(self, box_ids_to_path):
    # i don't need to compute box_ids_to_path...
    # do bar

def user_facing_function(self):
    box_ids_to_path = self._compute_box_ids()
    self._foo(box_ids_to_path)
    self._bar(box_ids_to_path)
```

Does that make sense / is that helpful? See how we only need to compute it once and then `_foo` and `_bar` can be more *pure* this way?

plotly/dashboard_objs/dashboard_objs.py
theengineear: wat. `object will actually mutate`? I think you're maybe misunderstanding what a generator is doing?

Also, can't you just `list(node_gen)` if all you want to do is go from a `generator` to a `list`?

plotly/dashboard_objs/dashboard_objs.py
theengineear: >I could switch up the terminology from ...

👍 cool sounds good! ya, go with what makes the most sense to you terminology-wise :).

plotly/dashboard_objs/dashboard_objs.py
theengineear: >On another or related note...

Yup, that's what I was getting at. OK :)

plotly/dashboard_objs/dashboard_objs.py
theengineear: >Do we not want to do the pretty print if IPython is None?

In general, if I can have a pretty html display, nope, I don't want the pretty-print.

plotly/dashboard_objs/dashboard_objs.py
theengineear: I could be less vague :).

What I'm curious about is a description of the algorithm you're implementing to create the html. Can you describe it in some documentation? It'd be helpful to review your code if I knew the high-level description of what's going on.

plotly/dashboard_objs/dashboard_objs.py
theengineear: It's pretty confusing to understand that, imo.

```
if is_horizontal:
    new_box_w = box_w / 2
    new_box_h = box_h
else:
    new_box_h = box_w
    new_box_h = box_h / 2
```

OR

```
new_box_w = box_w / 2 if is_horizontal else box_w
new_box_h = box_h if is_horizontal else box_h / 2
```

Pick either, but it's clearer to use a `bool` as a `bool` instead of implicitly as an `int`.

plotly/dashboard_objs/dashboard_objs.py
theengineear: Yah, but when have you ever said to yourself:

>wow, that was difficult to understand, I'm glad he didn't use another couple lines to make it easier to understand!

😸

theengineear:
@Kully, okie dokie, responded. I'll be on this train for about 4 hours, after that, I probably will be hard to reach for a while 😄.

Kully:
> @Kully, okie dokie, responded. I'll be on this train for about 4 hours, after that, I probably will be hard to reach for a while 😄.

thanks buddy. I'll be sure to review.

plotly/dashboard_objs/dashboard_objs.py
Kully: removed

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,391 @@
...
+      context.lineWidth = 2;
+      context.strokeStyle = 'black';
+      context.stroke();
+      </script>
+  </body>
+</html>
+""".format(width=master_width, height=master_height))
+
+
+def draw_line_through_box(dashboard_html, top_left_x, top_left_y, box_w,
```
Kully: I `_`ed the other function like that too

plotly/dashboard_objs/dashboard_objs.py
Kully: for now, I'm gonna leave the terminology as is. This may change before my next commit, most a mental note to self.

Kully:
@theengineear 
A note about sizing of the containers/splits. I think we should hold off on giving the users' the ability to change the size for this first pass. I think it's something that they can toy around with in the online editor when they upload it. I was thinking about something like `resize(box_id, percentage=60)` where `percentage` will determine how much of the container the box corresponding to the `box_id` will take up, but even that's a little difficult to understand, since they would need to know about the `direction` of the particular container they are manipulating.

I think for now, I'm just going to make sure that the `size` of each `container` is programmatically set to values that match to the HTML rep. Thoughts?

jackparmer:
Hey @Kully - To make sure I'm understanding correctly, would layouts like this still be possible?

https://files.slack.com/files-pri/T06LPNGUD-F4B1SGXPC/screen_shot_2017-02-27_at_1.26.25_pm.png

How would the widths of the cells in this HTML preview be set?

Kully:
> How would the widths of the cells in this HTML preview be set?

Yes they would. The way I'm doing it now is that I'm checking how "deep" each  `container` is, meaning how long the path is in the dashboard, and then assigning it a `size` which decreases as the path length increases. The equation looks like `MASTER_SIZE /  (2^path_len)`

jackparmer:
Nice. The above sounds OK to me. We can revisit later if exact width sizing becomes a big issue.

Kully:
@theengineear @chriddyp If one of you want to take a final 👀  for this PR.

chriddyp:
taking a first peek at this now

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,441 @@
+"""
+dashboard_objs
+==========
+
+A module for creating and manipulating dashboard content. You can create
+a Dashboard object, insert boxes, swap boxes, remove a box and get an HTML
+preview of the Dashboard.
```
chriddyp: It might be nice to include a simple example of usage here

plotly/plotly/plotly.py
```diff
@@ -1342,6 +1345,114 @@ def get_grid(grid_url, raw=False):
     return Grid(parsed_content, fid)
 
 
+class dashboard_ops:
+    """
+    Interface to Plotly's Dashboards API.
+    Plotly Dashboards are JSON blobs. They are made up by a bunch of
+    containers which contain either empty boxes or boxes with file urls.
+    """
```
chriddyp: It'd be nice to include one or two simple getting started examples.

Ideally, the user has enough info when calling `help()` on all of these methods to get started. Here's what `help` looks like now:

image

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,441 @@
+"""
```
chriddyp: should we include a 
```
from dashboard_objs import dashboard_objs
```

inside `__init__.py` that way you just have to do:

```
plotly.dashboard_objs.Dashboard
```

instead of 
```
plotly.dashboard_objs.dashboard_objs.Dashboard

```

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,441 @@
...
+               box_w=box_w, box_h=box_h, font_size=FONT_SIZE)
+
+    index_to_add_text = dashboard_html.find('</script>') - 1
+    dashboard_html = (dashboard_html[:index_to_add_text] + html_text +
+                      dashboard_html[index_to_add_text:])
+    return dashboard_html
+
+
+class Dashboard(dict):
+    def __init__(self, content=None):
```
chriddyp: A docstring with a simple example here would be great. That way, a user can just call help on this object and know what to do:

image

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,441 @@
...
+        """
+        Insert a box into your dashboard layout.
+
+        :param (dict) box: the box you are inserting into the dashboard.
+        :param (str) side: specifies where your new box is going to be placed
+            relative to the given 'box_id'. Valid values are 'above', 'below',
+            'left', and 'right'.
+        :param (int) box_id: the box id which is used as the reference box for
+            the insertion of the box.
+        """
```
chriddyp: similarly here - a couple simple examples of usage go a long way

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,441 @@
...
+                raise exceptions.PlotlyError(
+                    "If there is at least one box in your dashboard, you "
+                    "must specify a valid side value. You must choose from "
+                    "'above', 'below', 'left', and 'right'."
+                )
+
+        self._set_container_sizes()
+
+    def remove(self, box_id):
+        """Remove a box from the dashboard by its box_id."""
```
chriddyp: and here!

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -0,0 +1,441 @@
...
+            adjacent_box = container_for_box_id[adjacent_path]
+
+            self._insert(adjacent_box, path[:-1])
+        else:
+            self['layout'] = _empty_box()
+
+        self._set_container_sizes()
+
+    def swap(self, box_id_1, box_id_2):
+        """Swap two boxes with their specified ids."""
```
chriddyp: here too :)

plotly/plotly/plotly.py
```diff
@@ -1342,6 +1345,114 @@ def get_grid(grid_url, raw=False):
...
+            your Plotly account. Will overwrite a dashboard of the same
+            name if it already exists in your files.
+        :param (str) sharing: can be set to either 'public', 'private'
+            or 'secret'. If 'public', your dashboard will be viewable by
+            all other users. If 'secret', only you can see your dashboard.
+            If 'secret', the url will be returned with a sharekey appended
+            to the url. Anyone with the url may view the dashboard.
+        :param (bool) auto_open: automatically opens the dashboard in the
+            browser.
+        """
```
chriddyp: a simple example here too would be pretty helpful

plotly/plotly/plotly.py
```diff
@@ -1342,6 +1345,114 @@ def get_grid(grid_url, raw=False):
...
+        ).json()
+        if only_content:
+            dashboard_json = json.loads(dashboard['content'])
+            return dashboard_json
+        else:
+            return dashboard
+
+    @classmethod
+    def get_dashboard(cls, dashboard_name):
+        """Returns a Dashboard object from a dashboard name."""
```
chriddyp: For some reason these docstrings don't appear in `help`
image

plotly/plotly/plotly.py
```diff
@@ -1342,6 +1345,114 @@ def get_grid(grid_url, raw=False):
...
+
+        :param (dict) dashboard: the JSON dashboard to be uploaded. Use
+            plotly.dashboard_objs.dashboard_objs to create a Dashboard
+            object.
+        :param (str) filename: the name of the dashboard to be saved in
+            your Plotly account. Will overwrite a dashboard of the same
+            name if it already exists in your files.
+        :param (str) sharing: can be set to either 'public', 'private'
+            or 'secret'. If 'public', your dashboard will be viewable by
+            all other users. If 'secret', only you can see your dashboard.
```
chriddyp: "If 'private'", not "If 'secret'"

plotly/plotly/plotly.py
```diff
@@ -1342,6 +1345,114 @@ def get_grid(grid_url, raw=False):
...
+
+            if matching_dashboard['filetype'] == 'dashboard':
+                old_fid = matching_dashboard['fid']
+                res = v2.dashboards.update(old_fid, data)
+
+        except exceptions.PlotlyRequestError:
+            res = v2.dashboards.create(data)
+        res.raise_for_status()
+
+        url = res.json()['web_url']
```
chriddyp: we should also append the share key when the user sets the `sharing` to `secret`. Otherwise, they will probably mistakeningly try to share the URL to a coworker who won't be able to see it

plotly/plotly/plotly.py
```diff
@@ -1342,6 +1345,114 @@ def get_grid(grid_url, raw=False):
...
+        try:
+            v2.files.lookup(filename)
+            matching_dashboard = cls._get_dashboard_json(
+                filename, False
+            )
+
+            if matching_dashboard['filetype'] == 'dashboard':
+                old_fid = matching_dashboard['fid']
+                res = v2.dashboards.update(old_fid, data)
+
```
chriddyp: missing an `else` here. 

I'm assuming that `v2.dashboards.update` will fail if the filename already exists as a different type like a `plot`?

Perhaps we can just `throw exceptions.PlotlyRequestError('{filename} is already a {filetype} in your account. While you can overwrite dashboards with the same name, you can't change overwrite files with a different type.\nTry deleting '{filename}' in your account or changing the filename'.format(filename=filename, filetype=matching_dashboard['filetype']')`

plotly/plotly/plotly.py
Kully: `help(py.meta_ops.upload)` also results in the same message.

I think it's fine since `help(py.dashboard_ops)` will display the doc strings for all the other methods in the class

plotly/plotly/plotly.py
Kully: It's taken care of here:

```
        if sharing == 'secret':
            url = add_share_key_to_url(url)
```

Kully:
> Looking good to me! I was able to create a few different dashboards. Great stuff!
> 
> I just want to see some more docstrings to help newcomers out.

Alrighty, all done! Waiting for tests to pass, then would love to merge!

Kully:
@chriddyp Can I get a 💃 ?

plotly/plotly/plotly.py
chriddyp: Got it. It looks like the URL gets returned with the sharekey but the browser doesn't open up with the sharekey: 
shareky

chriddyp: NVM, just saw that you updated this logic in https://github.com/plotly/plotly.py/pull/694/commits/9c4b2357114bc10cdce268597e17805986bcf984#diff-07784ffd70c058caed46ede342e6dc61R1459

chriddyp:
💃

Issue #684, opened 2017-02-02
offline.plot() causes console error 


cldougl:
import plotly.offline as offline

offline.plot({'data': [{'y': [4, 2, 3, 4]}],
                    'layout': {'title': 'Test Plot', 
                                  'font': dict(size=16)}})

<img width="842" alt="screen shot 2017-02-02 at 4 11 06 pm" src="https://cloud.githubusercontent.com/assets/2789078/22568594/634f50f0-e962-11e6-95c2-bd04b61ef4ec.png">

PR #698, opened 2017-02-20
offline plot event listener edit

cldougl:
fixes https://github.com/plotly/plotly.py/issues/684


cldougl:
need a small review @chriddyp @theengineear @Kully

Kully:
> need a small review @chriddyp @theengineear @Kully

looking at now

plotly/offline/offline.py
```diff
@@ -454,7 +454,6 @@ def plot(figure_or_data, show_link=True, link_text='Export to plot.ly',
         resize_script = (
             ''
             '<script type="text/javascript">'
-            'window.removeEventListener("resize");'
```
chriddyp: just for some historical context:

I'm pretty sure that this line was in here for `offline.iplot`: when you would re-run `iplot` over and over, we wanted to remove the previously added `eventListener` that might've been added on the first `iplot` run.

From refactors of `plotly/offline/offline.py`, I think that this got copied and pasted around and is now in `plot` even though it doesn't need to be. `plot` doesn't need to call `removeEventListener` because the following `window.addEventListener` will only ever be added once in the `plot.html` context.

chriddyp:
💃  from me

Kully:
💃 from me.

Note: Don't know if this is appropriate to comment on here, but in [this section](https://github.com/plotly/plotly.py/blob/4b16b329da5a9913a1b8ac141dd5964971f5ffec/CHANGELOG.md#201---2017-02-07) of the CHANGELOG, there's two `Added` sections we should reduce to one. Unless it's intention to promote significance to the first bullet point.

cldougl:
yeah @Kully that's not from this pr but you can reduce it into 1

PR #704, opened 2017-03-05
Add json output type to offline plot.

syamajala:
I added a json output type to the offline plot function. It returns a (data, layout, config) tuple of json strings.


jonmmease:
Thanks for working on this @syamajala , this functionality is going to be provided by the `to_json` and `write_json` functions in the `plotly.io` module. See #1188 for progress.

PR #713, opened 2017-03-15
Add Plotly custom mimetype to NB outputs.

tarzzz:
Azure notebooks sanitize HTML when the user requests to 'Download as HTML'. This leads to Plotly Charts being rendered in plain text. To tackle the same, we send the Plotly chart data as a custom mimetype: `text/vnd.plotly.v1+html`.

[PlotlyHTMLExporter](https://github.com/tarzzz/plotlyhtmlexporter) (Plotly friendly, Jupyter Notebook HTML Exporter) recognizes this 
mimetype, and renders the charts in their interactive version when notebook is converted to HTML. 

`PlotlyHTMLExporter` sanitizes all html content but leaves Plotly Charts intact.


tarzzz:
@Kully Please review..
cc: @chriddyp

plotly/offline/offline.py
```diff
@@ -140,7 +140,11 @@ def init_notebook_mode(connected=False):
             '</script>'
             '').format(script=get_plotlyjs())
 
-    ipython_display.display(ipython_display.HTML(script_inject))
+    display_bundle = {
+        'text/html': script_inject,
+        'text/vnd.plotly.v1+html': script_inject
+    }
+    ipython_display.display(display_bundle, raw=True)
```
Kully: Why `raw=True`?

Kully:
@tarzzz Okay Tarun, it's clearer now how this is supposed to work based on the code. Just have that one comment but the rest looks fine to me.

plotly/offline/offline.py
tarzzz: @Kully : 

From IPython Docs:

> raw : bool
Are the data objects raw data or Python objects that need to be formatted before display? [default: False]

We dont want IPython to do any further processing on this, so we use `raw=True`.

tarzzz:
@Kully : We will also need to release a new version for this change, so that the users can do a pip upgrade and be able to use it with `plotlyhtmlexporter`. How should I do that?

cldougl:
@tarzzz you can update the [changelog](https://github.com/plotly/plotly.py/blob/master/CHANGELOG.md) and [version](https://github.com/plotly/plotly.py/blob/master/plotly/version.py) and we'll release on pip

tarzzz:
@cldougl : Updated Changelog and version .. 👍

cldougl:
lgtm 💃 after tests pass

tarzzz:
Great, merging!! 🎉

PR #714, opened 2017-03-16
Make offline.plot/iplot accept a plot config dict

kernc:
Make `plotly.offline.plot()` (`iplot()`) accept a `config` dictionary so the various [configuration options](https://plot.ly/javascript/configuration-options/) can be passed. Includes unit tests.

Addresses https://community.plot.ly/t/removing-hover-toolbar-and-other-configuration-options-with-the-python-api/374.

Also took liberty to slightly improve surrounding `offline` tests.

If you find the proposed feature something acceptable, I'd be happy to also amend the docs.

### DIFF
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -275,7 +275,7 @@ def _plot_html(figure_or_data, config, validate, default_width,
 
 def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
           validate=True, image=None, filename='plot_image', image_width=800,
-          image_height=600):
+          image_height=600, config=None):
     """
     Draw plotly graphs inside an IPython or Jupyter notebook without
     connecting to an external server.
@@ -308,6 +308,9 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
         will be saved to. The extension should not be included.
     image_height (default=600) -- Specifies the height of the image in `px`.
     image_width (default=800) -- Specifies the width of the image in `px`.
+    config (default=None) -- Plot view options dictionary. Keyword arguments
+        `show_link` and `link_text` set the associated options in this
+        dictionary if it doesn't contain them already.
 
     Example:
     ```
@@ -331,9 +334,9 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
     if not ipython:
         raise ImportError('`iplot` can only run inside an IPython Notebook.')
 
-    config = {}
-    config['showLink'] = show_link
-    config['linkText'] = link_text
+    config = dict(config) if config else {}
+    config.setdefault('showLink', show_link)
+    config.setdefault('linkText', link_text)
 
     plot_html, plotdivid, width, height = _plot_html(
         figure_or_data, config, validate, '100%', 525, True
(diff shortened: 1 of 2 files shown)


Kully:
@kernc taking a look right now

plotly/offline/offline.py
```diff
@@ -327,9 +330,9 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
     if not ipython:
         raise ImportError('`iplot` can only run inside an IPython Notebook.')
 
-    config = {}
-    config['showLink'] = show_link
-    config['linkText'] = link_text
+    config = dict(config) if config else {}
+    config.setdefault('showLink', show_link)
```
Kully: Why not just write `config['showLink] = show_link`? It's clearer to me.

Same goes with `config.setdefault('linkText', link_text)`

plotly/offline/offline.py
```diff
@@ -441,9 +448,9 @@ def plot(figure_or_data, show_link=True, link_text='Export to plot.ly',
             "Adding .html to the end of your file.")
         filename += '.html'
 
-    config = {}
-    config['showLink'] = show_link
-    config['linkText'] = link_text
+    config = dict(config) if config else {}
+    config.setdefault('showLink', show_link)
```
Kully: Same as above

plotly/tests/test_core/test_offline/test_offline.py
```diff
@@ -56,22 +56,23 @@ def test_default_plot_generates_expected_html(self):
 
         # I don't really want to test the entire script output, so
         # instead just make sure a few of the parts are in here?
-        self.assertTrue('Plotly.newPlot' in html) # plot command is in there
-        self.assertTrue(data_json in html)        # data is in there
-        self.assertTrue(layout_json in html)      # so is layout
-        self.assertTrue(PLOTLYJS in html)         # and the source code
+        self.assertIn('Plotly.newPlot', html)  # plot command is in there
```
Kully: Fantastic edit! Simplifies the function of the test with the `assertIn` method

plotly/tests/test_core/test_offline/test_offline.py
```diff
@@ -56,22 +56,23 @@ def test_default_plot_generates_expected_html(self):
...
+        self.assertIn(layout_json, html)       # so is layout
+        self.assertIn(PLOTLYJS, html)          # and the source code
         # and it's an <html> doc
         self.assertTrue(html.startswith('<html>') and html.endswith('</html>'))
 
     def test_including_plotlyjs(self):
         html = self._read_html(plotly.offline.plot(fig, include_plotlyjs=False,
                                                    auto_open=False))
-        self.assertTrue(PLOTLYJS not in html)
+        self.assertNotIn(PLOTLYJS, html)
```
Kully: :+1:

Kully:
@kernc I've reviewed and just have a couple suggestions for improving clarity of code. It's a 💃 from me after that.

plotly/offline/offline.py
kernc: The reason for [`dict.setdefault`](https://docs.python.org/2/library/stdtypes.html#dict.setdefault) is that it doesn't set the value if it is already set. This gives user the option to construct the _whole_ view configuration dict beforehand, not needing to worry about having `showLink` overwritten or needing to pass it separately as `show_link`. I'd rather go with:
```py
if 'linkText' not in config:
    config['linkText'] = link_text
```
but I think `.setdefault()` is pretty standard. Do you insist? The behavior of not overriding the option if it's already set is documented.

Kully: Nope, don't insist. I get the reasoning now.

Kully:
💃 from me. (Reminder for myself to update CHANGELOG 🐱  )

PR #726, opened 2017-04-06
check that share key is enabled

cldougl:


plotly/plotly/plotly.py
```diff
@@ -1308,7 +1308,14 @@ def add_share_key_to_url(plot_url):
     body = {'share_key_enabled': True, 'world_readable': False}
     response = v2.files.update(fid, body)
 
-    return plot_url + '?share_key=' + response.json()['share_key']
+    if not v2.files.retrieve(fid).json()['share_key_enabled']:
```
chriddyp: I'd add a comment here about why we're trying multiple times

plotly/plotly/plotly.py
```diff
@@ -1308,7 +1308,14 @@ def add_share_key_to_url(plot_url):
...
+        attempt += 1
+        if attempt == 10:
```
chriddyp: I'd be OK with bumping this up to 50 even

plotly/plotly/plotly.py
```diff
@@ -1308,7 +1308,14 @@ def add_share_key_to_url(plot_url):
...
+            raise Exception
```
chriddyp: We're trying to wrap everything in plotly exceptions so that users can catch them. So just: `raise exceptions.PlotlyException` (or something)

chriddyp:
💃  once you consider my comments

plotly/plotly/plotly.py
cldougl: 👍 bumped to 50

PR #729, opened 2017-04-11
update permissions warning to point users to support site

cldougl:


cldougl:
@chriddyp & @Kully -- quick warning message update to point users to support app rather than upgrading for support.

Kully:
💃 no typos, all good

PR #731, opened 2017-04-12
Facet grid FigureFactory

Kully:
will be making a notebook for a proper showcase of these charts


plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+
+    show_legend = False
+
+    if facet_row:
+        num_of_rows = len(data[facet_row].unique())
+
+    if facet_col:
+        num_of_cols = len(data[facet_col].unique())
```
chriddyp: Thoughts about adding binning for numerical columns?

plotly/figure_factory/__init__.py
```diff
@@ -16,3 +16,4 @@
 from plotly.figure_factory._table import create_table
 from plotly.figure_factory._trisurf import create_trisurf
 from plotly.figure_factory._violin import create_violin
+from plotly.figure_factory._facet_grid import create_facet_grid
```
cldougl: imports in alphabetical order

plotly/figure_factory/_facet_grid.py
Kully: I wasn't thinking about it but I can after I get something working. I'm not finished with the code here, so I can ping you when I do

jackparmer:
How 1-1 is this figure factory with the ggplot2 facet geom? Would be great to see some screenshot output examples when this gets closer.

Kully:
> How 1-1 is this figure factory with the ggplot2 facet geom?
pretty close, just missing some variables right now. Yeah, I'm gonna put together either screenshots or a notebook presentation or both.

Kully:
Progress:
screen shot 2017-04-16 at 8 43 50 pm
screen shot 2017-04-16 at 8 49 53 pm

Kully:
@jackparmer Here's my progress so far.
@cldougl Ready for a review.
@chriddyp Here's a [Jupyter Notebook](https://plot.ly/~AdamKulidjian/3383) of some examples!

Things that we could add:
- binning for numerical columns (Chris' suggestion and exists for scatterplot_matrix FigureFactory)

jackparmer:
@Kully This looks very cool! Can you please make a Jupyter notebook that uses these dataframes and recreates these examples using your FigureFactory?

http://ggplot2.tidyverse.org/reference/facet_grid.html

I think this will be a nice way to be sure we're covering canonical faceting usage.

Kully:
> Can you please make a Jupyter notebook that uses these dataframes and recreates these examples using your FigureFactory?

For sure!

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,787 @@
...
+        yref='paper',
+        text=text,
+        font=dict(
+            size=13,
+            color=AXIS_TITLE_COLOR
+        )
+    )
+
+    return annotation_dict
+
```
cldougl: you can add 2 new lines between `def`s

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,787 @@
...
+                                 num_of_rows - row_count,
+                                 num_of_rows,
+                                 row_col='row')
+                )
+
+    # add annotations
+    fig['layout']['annotations'] = annotations
+
+    return fig
+
```
cldougl: 2 spaces between defs

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,787 @@
...
+def create_facet_grid(data, x, y, facet_row=None, facet_col=None, color=None,
+                      colorscale=None, color_dict=None, title='facet grid',
+                      height=600, width=600, **kwargs):
+    """
+    Returns data for a facet grid.
```
cldougl: this returns a figure right? not just the data object

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,787 @@
...
+        color='column c',
+        colorscale='Viridis'
+    )
+
+    py.iplot(my_grid, filename='facet_grid_numerical_color')
+    ```
+    """
+    if not pd:
+        raise exceptions.ImportError(
+            "'pandas' must be imported for this FigureFactory."
```
cldougl: since we changed the syntax I would probably say figure_factory or figure type or chart type

cldougl:
@Kully looks like a weird thing is going on when turning off the traces: 
tracebg

plotly/figure_factory/_facet_grid.py
Kully: changed 'data' -> 'figure'

Kully:
> @Kully looks like a weird thing is going on when toggling the traces:

Yeah, it's pretty wack.

For some reason when a subplot is empty, the background just disappears. I think it has to do with all the customization I'm using.

Are you a fan of having all those duplicate trace icons, or is one per category enough.

Kully:
@cldougl @chriddyp Can one of you guys do a review for me after tests are passing? Will also need this on `pip` as soon as possible after successfully merging.

Tests still failing...

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+LEGEND_COLOR = '#efefef'
+PLOT_BGCOLOR = '#ededed'
+ANNOT_RECT_COLOR = '#d0d0d0'
+SUBPLOT_SPACING = 0.015
+LEGEND_BORDER_WIDTH = 1
+LEGEND_ANNOT_X = 1.05
+LEGEND_ANNOT_Y = 0.5
+MAX_TICKS_PER_AXIS = 10
+THRES_FOR_FLIPPED_FACET_TITLES = 10
+GRID_WIDTH = 1
```
chriddyp: are these configurable by the user?

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+                         'layer': 'below',
+                         'line': {'color': annot_rect_color, 'width': 1},
+                         'type': 'rect',
+                         'x0': 1.005,
+                         'x1': 1.05,
+                         'xref': 'paper',
+                         'y0': fig['layout'][key]['domain'][0],
+                         'y1': fig['layout'][key]['domain'][1],
+                         'yref': 'paper'}
+                    )
```
chriddyp: This could be drier. It's hard to immediately see which attributes are different for each of these conditonal statements. Something like this might be better:
```
shape = {
    'fillcolor': annot_rect_color,
    'layer': 'below',
    'line': {'color': annot_color}, 
     # etc - the rest of the shared attributes
}
if 'xaxis' in key::
    shape['x0'] = fig['layout'][key]['domain'][0]
    shape['x1'] = fig['layout'][key]['domain'][0]
    shape['y0'] = 1.005
    if flipped_cols:
        shape['y1'] = 1.05 + 0.5
    else:
        shape['y1'] = 1.05
elif 'yaxis' in key:
    shape['x0'] = 1.005
    shape['y0'] = fig['layout'][key]['domain'][0]
    shape['y1'] = fig['layout'][key]['domain'][0]
    if flipped_rows:
        shape['x1'] = 1.05 + 1
     else:
         shape['x1'] = 1.05
```

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+
+    return fig
+
+
+def _facet_grid_color_numerical(df, x, y, facet_row, facet_col, color_name,
+                                colormap, height, width, num_of_rows,
+                                num_of_cols, facet_row_labels,
+                                facet_col_labels, trace_type, size,
+                                flipped_rows, flipped_cols, show_boxes,
+                                kwargs, kwargs_marker):
```
chriddyp: It seems like most of the code is shared between this function and `_facet_grid_color_categorical`. Is there any way to make this more dry and unify the functions?

chriddyp: It could just be like:
```
def _facet_grid_color(..., type='categorical')

# all the shared code is the same
[...]

# but then when the code differs, just an if statement:
if type == 'categorical':
    # do the categorical thing
elif type == 'numerical':
    # do the numerical thing
```

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+
+    fig = make_subplots(rows=num_of_rows, cols=num_of_cols,
+                        shared_xaxes=True, shared_yaxes=True,
+                        horizontal_spacing=SUBPLOT_SPACING,
+                        vertical_spacing=SUBPLOT_SPACING, print_grid=False)
+
+    annotations = []
+    if not facet_row and not facet_col:
+        trace = graph_objs.Scatter(
```
chriddyp: The categorical one creates new traces for every unique item (similar to how it facets) in this case. Should the numeric one do the same for consistency?

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+            x=df[x].tolist(),
+            y=df[y].tolist(),
```
chriddyp: do we need this `tolist()`? our `PlotlyJSONEncoder` handles Pandas Series: https://github.com/plotly/plotly.py/blob/master/plotly/utils.py#L218

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+            mode='markers',
+            type=trace_type,
+            marker=dict(
+                color=df[color_name],
+                size=size,
+                colorscale=colormap,
+                showscale=True,
+                **kwargs_marker
```
chriddyp: nice way to make these configurable!

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+            ),
+            **kwargs
```
chriddyp: could we change this to `kwargs_trace` to match `kwargs_marker`?

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+        )
+        fig.append_trace(trace, 1, 1)
+
+    elif facet_row and not facet_col:
+        groups_by_facet_row = list(df.groupby(facet_row))
```
chriddyp: for numerical columns, this is going to create a unique entry for every unique value right? I feel like we should group by ranges of values instead, and maybe 5 ranges by default (but configurable). So something like:
```
groups_by_facet_row = df.groupby(pd.cut(df[facet_row], np.linspace(df[facet_row].min(), df[facet_row].max(), N_ROWS)).groups
```

chriddyp: We could even allow users to pass in their own "`cut`" arrays, so that they could make custom groups (for example, they might want to groups like `[[0, 1], [1, 10], [10, 100], [100, 1000]]`

chriddyp: here's what that looks like with evenly spaced groups:
```
>>> df.groupby(pd.cut(df['a'], np.linspace(df['a'].min(), df['a'].max(), 5))).groups

{'(1, 3.25]': [1, 2, 13],
 '(3.25, 5.5]': [3, 4, 11, 14],
 '(5.5, 7.75]': [5, 6],
 '(7.75, 10]': [7, 8, 9, 12, 15]}
```

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+        for j, group in enumerate(groups_by_facet_row):
+            trace = graph_objs.Scatter(
+                x=group[1][x].tolist(),
+                y=group[1][y].tolist(),
+                mode='markers',
+                type=trace_type,
+                marker=dict(
+                    color=df[color_name].tolist(),
```
chriddyp: it looks like `color_name` is required here. is that true? I think that it should be optional

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+                **kwargs
+            )
+            fig.append_trace(trace, 1, j + 1)
+
+            label = _return_label(group[0], facet_col_labels, facet_col)
+
+            annotations.append(
+                _annotation_dict(label, j + 1, num_of_cols, row_col='col',
+                                 flipped=flipped_cols)
+            )
```
chriddyp: these two blocks could be a little drier, but it's not a big deal. Something like this:
```
if (facet_row and not facet_col) or (not facet_row and facet_col):
    groups = list(df.groupby(facet_row if facet_row else facet_col))
    for j, group in enumerate(groups):
        ...
        fig.append_trace(trace, j + 1 if facet_row else 1, 1 if facet_row else j +1)
    ... 
```

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+                        vertical_spacing=SUBPLOT_SPACING, print_grid=False)
+    annotations = []
+    if not facet_row and not facet_col:
+        trace = graph_objs.Scatter(
+            x=df[x].tolist(),
+            y=df[y].tolist(),
+            mode='markers',
+            type=trace_type,
+            marker=dict(
+                color=DEFUALT_MARKER_COLOR,
```
chriddyp: spelling: `DEFAULT` not `DEFUALT`

plotly/figure_factory/_facet_grid.py
chriddyp: oh i see - `_facet_grid` deals with the case when color isn't specified.

plotly/figure_factory/_facet_grid.py
Kully: they're not

plotly/figure_factory/_facet_grid.py
Kully: It's not as easy to separate the code between the two functions as I recall. But I'll see what I can do

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+    # add annotations
+    fig['layout']['annotations'] = annotations
+
+    return fig
+
+
+def _facet_grid(df, x, y, facet_row, facet_col, height, width, num_of_rows,
+                num_of_cols, facet_row_labels, facet_col_labels,
+                trace_type, size, flipped_rows, flipped_cols, show_boxes,
+                kwargs, kwargs_marker):
```
chriddyp: this could be a lot DRYer. As far as I can tell, `_facet_grid` is the same as `_facet_grid_color_numerical` which is pretty similar to `_facet_grid_color_categorical`. all of these functions can be merged into one function with a few `if` statements I think. 

For example, 
```
color=df[color_name] if type == 'numerical' else colormap[color_val] if type === 'categorical' else DEFAULT_MARKER_COLOR`
```
or if you prefer:
```
color=(
    df[color_name] if type == 'numerical'
    else colormap[color_val] if type === 'categorical' 
    else DEFAULT_MARKER_COLOR`
)
```

or alternatively, if there are multiple values that depend on the type:

```python
trace = graph_objs.Scatter(
    x=df[x],
    y=df[y],
    mode='markers',
    type=trace_type,
    marker=dict(
        size=size,
        **kwargs_marker
    )
)

if type == 'categorical':
   trace['marker']['color'] = colormap[color_val]
elif type == 'numerical':
   trace['marker']['color'] = df[color_name]
else:
   trace['marker']['color'] = DEFAULT_MARKER_COLOR
```

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+    """
+    Returns figure for facet grid.
+
+    :param (pd.DataFrame) df: the dataframe of columns for the facet grid.
+    :param (str) x: the key of the dataframe to be used as the x axis df.
+    :param (str) y: the key of the dataframe to be used as the y axis df.
+    :param (str) facet_row: the key for row filter column for the facet grid.
+    :param (str) facet_col: the key for the column filter column for the facet
+        grid.
+    :param (str) color_name: the name of your dataframe column that will
```
chriddyp: Should this be: "the key of your dataframe column" instead of "the name of your dataframe column"? The rest of the descriptions say "key" instead of "name"

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+        function as the colormap variable.
+    :param (str|list|dict) colormap: the param that determines how the
+        color_name column colors the data. If the dataframe contains numeric
+        data, then a dictionary of colors will group the data categorically
```
chriddyp: it's not "If the dataframe contains numeric data" it's more like "If `color_name` was set to a numerical column, then a continuous colorscale will be applied to the data. This colorscale can be customized by setting `colormap` to a Plotly colorscale name or through a custom colorscale. If `color_name` was set to a categorical column, then each unique value will be mapped a unique color. These colors can be customized by passing a dictionary that has unique values names as keys and color strings as variables. To learn more about colors and the type of a colorscale, run `help(plotly.colors)`."

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+        while a Plotly Colorscale name or a custom colorscale will treat it
+        numerically. To learn more about colors and types of colormap, run
+        `help(plotly.colors)`.
+    :param (bool) color_is_cat: determines whether a numerical column for the
+        colormap will be treated as categorical (True) or sequential (False).
+            Default = False.
+    :param (bool) widen_frame: if set to True, all points in each subplot
+        are strickly contained in the region of the subplot by increasing the
```
chriddyp: spelling: `strictly`

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+        maximum and minimum range values by 1. Setting to False doesn't do
+        anything.
```
chriddyp: maybe "If `False`, then points may be plotted on the edge of the frame."

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+        Default = False
+    :param (str|dict) facet_row_labels: set to either 'name' or a dictionary
+        of all the values in the facetting row mapped to some text to show up
```
chriddyp: spelling: `faceting`

chriddyp: and it's all of the "unique values" not just "values", right?

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+        in the label annotations. If None, labelling works like usual.
```
chriddyp: spelling: `labeling`

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+    :param (str|dict) facet_col_labels: set to either 'name' or a dictionary
+        of all the values in the facetting row mapped to some text to show up
+        in the label annotations. If None, labelling works like usual.
+    :param (int) height: the height of the facet grid figure.
+    :param (int) width: the width of the facet grid figure.
+    :param (int) size: point size in pixels.
+    :param (str) trace_type: decides the type of plot to appear in the
+        facet grid. The options are 'scatter' and 'scattergl'.
```
chriddyp: although presumably you could pass in something like `histogram` and a set of `kwargs` to customize that histogram right?

chriddyp: we could even skip applying the `marker` and `mode` options `if type is not in ['scatter', 'scattergl']`, allowing users to create valid facet_grids with heatmaps, histograms, 2dhistograms, etc

chriddyp: In order for histogram to work, you'd need to also make `x` and `y` optional too

chriddyp:
Some other thoughts as I play around with this:
- I'm curious if it can be generalized a little bit to be able to plot any chart type, not just scatter. I find myself frequently wanting to make a faceted histogram, so this could be a big win. It seems like it could with a few small changes:
    - making `x` and `y` optional: in a histogram you would only supply `x` (or, if you wanted horizontal histogram, `y`)
    - skipping the `marker` assignments if the type isn't `scatter` or `scattergl`
- Sizing I think that we should make the height a multiple of the number of rows and the width a mulitple of the number of columns. Something like `height=min(600, 175 * n_rows`). That way, if you have 20 unique values, your graph isn't 15px tall. If you have only two unique values, then we keep the default height (like 600 or w/e it is)
- Text - We should add an optional text `column_name` that adds `text` to the grouped values

chriddyp:
Is there a requirement that the graphs need to look like `ggplot2` by default? I would prefer that we keep the styling more aligned to the plotly defaults to be consistent with the rest of the library. There are also a few things that bother me about the ggplot styling:
- The points are pure black which is really stark. Without opacity, it's hard to tell if they overlap - here's an article about avoiding pure blacks in design: https://ianstormtaylor.com/design-tip-never-use-black/
- The points are really small.
- I personally find the grey background and grey background shapes kind of old looking
- I think the tick placement much heavier than it needs to be

Some people like the theme however, and so we can keep it in with maybe a `theme=ggplot2` argument.

Here is a little redesign:
image
Compared with:
image

Here is the code I'm using to convert the `facet_grid` styles:
```
gg = ff.create_facet_grid(mpg, 'cty', 'hwy', 'class')
p = copy.deepcopy(gg)
for trace in p['data']:
    # Overwrite marker styles: 
    # - slightly bigger than usual points
    # - opacity in the points
    # - a small border around the points
    trace['marker'] = {
        'color': 'rgba(31, 119, 180, 0.5)',
        'size': 8,
        'line': {'color': 'darkgrey', 'width': 1}
    }

# removing the dark grey background
del p['layout']['shapes']

# making the plot height porportional to the number of rows
p['layout']['height'] = len(p['data']) * 150


for k, v in p['layout'].iteritems():
    if 'axis' in k:
        # Reverting to default grid styles
        del v['gridcolor']
        del v['gridwidth']
        del v['tickfont']
        del v['tickwidth']
        del v['dtick']
        # Except the ticks: removing the ticks
        v['ticklen'] = 0

# Removing the grey background
del p['layout']['plot_bgcolor']

# Adding a slightly off-white margin background
# This makes it easier to distinguish one subplot from the other
p['layout']['paper_bgcolor'] = 'rgb(251, 251, 251)'

# Update hovering mode for scatter plots
p['layout']['hovermode'] = 'closest'
```

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+    if 'marker' in kwargs:
+        kwargs_marker = kwargs['marker']
+    else:
+        kwargs_marker = {}
+    for param in ['size', 'color']:
+        kwargs_marker.pop(param, None)
+    kwargs.pop('marker', None)
+
+    # make sure dataframe index starts at 0
+    df.index = range(len(df))
```
chriddyp: I'm not sure we want to be doing this: this is going to be modifying the user's dataframe without them knowing.

For example:
image

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+    flipped_cols = False
+    if facet_row:
+        num_of_rows = len(df[facet_row].unique())
+        flipped_rows = _is_flipped(num_of_rows)
+        if isinstance(facet_row_labels, dict):
+            for key in df[facet_row].unique():
+                if key not in facet_row_labels.keys():
+                    unique_keys = df[facet_row].unique().tolist()
+                    raise exceptions.PlotlyError(
+                        "If you are using a dictioanry for custom labels for "
```
chriddyp: spelling: "dictionary"

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+                    )
+    if facet_col:
+        num_of_cols = len(df[facet_col].unique())
+        flipped_cols = _is_flipped(num_of_cols)
+        if isinstance(facet_col_labels, dict):
+            for key in df[facet_col].unique():
+                if key not in facet_col_labels.keys():
+                    unique_keys = df[facet_col].unique().tolist()
+                    raise exceptions.PlotlyError(
+                        "If you are using a dictioanry for custom labels for "
```
chriddyp: spelling: "dictionary".

plotly/figure_factory/_facet_grid.py
Kully: Damn, you're right. I can fix that with some rewriting.

plotly/figure_factory/_facet_grid.py
```diff
@@ -0,0 +1,1073 @@
...
+        range_are_numbers = (isinstance(min_range, Number) and
+                             isinstance(max_range, Number))
+
+        if range_are_numbers:
+            # modify ranges
+            min_range = math.floor(min_range)
+            max_range = math.ceil(max_range)
+            if widen_frame:
+                min_range -= 1
+                max_range += 1
```
chriddyp: Good call in fixing the range across all subplots so that it's easy to compare values.

I think we can be a bit smarter about setting the autorange though and then make it the default and remove this `widen_frame` argument:
- Adding/subtracting `1` like this won't work if the numerical ranges are really big or if the numerical ranges are within a small range like between `0.5` and `0.6` (in which case it'll end up making the range too large)
- Instead, why don't we just make it say 5% bigger than the absolute range. I believe this is how plotly.js does it. Something like: `range = [min - (max - min) * 0.05, max + (max - min) * 0.05]`. You could ask the #plotly_js folks exactly what value they use

plotly/figure_factory/utils.py
```diff
@@ -10,23 +10,27 @@
...
+    'Blues': ['rgb(5,10,172)', 'rgb(220,220,220)'],
+    'Picnic': ['rgb(0,0,255)', 'rgb(255,0,0)'],
+    'Rainbow': ['rgb(150,0,90)', 'rgb(255,0,0)'],
+    'Portland': ['rgb(12,51,131)', 'rgb(217,30,30)'],
+    'Jet': ['rgb(0,0,131)', 'rgb(128,0,0)'],
+    'Hot': ['rgb(0,0,0)', 'rgb(255,255,255)'],
+    'Blackbody': ['rgb(0,0,0)', 'rgb(160,200,255)'],
+    'Earth': ['rgb(0,0,130)', 'rgb(255,255,255)'],
+    'Electric': ['rgb(0,0,0)', 'rgb(255,250,220)'],
+    'Viridis': ['#440154', '#fde725']
```
chriddyp: These colorscales seem a little off - aren't there supposed to be more than 2 colors in them?

plotly/figure_factory/_facet_grid.py
Kully: Great idea Chris

plotly/figure_factory/utils.py
Kully: Yeah that colorscale list is not correct. It's being used by `_scatterplot.py` for colorscales, so I eventually want to make a PR to replace the colorscales in utils with `plotly.colors.PLOTLY_SCALES` (the proper one) and rewrite the way scatterplot handles colorscales.

Kully: I'll just do it in this PR 👍

plotly/figure_factory/_facet_grid.py
Kully: correct

plotly/figure_factory/_facet_grid.py
Kully: For this pass, can we not include trace types besides `scatter` and `scattergl`? `mode` for example is not a valid key for a bar chart, and would result in an error.

plotly/figure_factory/_facet_grid.py
Kully: > Sizing I think that we should make the height a multiple of the number of rows and the width a mulitple of the number of columns

I like this idea. Going to add now

plotly/figure_factory/utils.py
Kully: > The points are pure black which is really stark. Without opacity, it's hard to tell if they overlap - here's an article about avoiding pure blacks in design: https://ianstormtaylor.com/design-tip-never-use-black/

Good read. It's funny because the "black" that they pit against the examples of nearly-black in app layouts are actually not `#000000` either. The website may be swiping a filter over the whole page

Kully:
> The points are pure black which is really stark. Without opacity, it's hard to tell if they overlap - here's an article about avoiding pure blacks in design: https://ianstormtaylor.com/design-tip-never-use-black/

Good read. It's funny because the "black" that they pit against the examples of nearly-black in app layouts are actually not `#000000` either. The website may be swiping a filter over the whole page

Kully:
@jackparmer @chriddyp @cldougl 
A few words about this PR. I want to get this thing merged ASAP (today if possible).
1. Can one of you do a quick review of the facet grid? CHANGELOG and version number are already bumped the next pip package.
2. I want to move the remaining things/ideas for this figure factory - numerical binning/custom cuts for binning/other trace-type support - to another PR, as they are not canonical features.

Sounds good?

jackparmer:
> I want to move the remaining things/ideas for this figure factory - numerical binning/custom cuts for binning/other trace-type support - to another PR, as they are not canonical features.

SGTM!

Kully:
@jackparmer @chriddyp 💃 ?

chriddyp:
> - numerical binning/custom cuts for binning to another PR

Yeah, that sounds good. In the meantime, we can document a workflow where the user creates their own bins in their dataframe and facets off of that. For example:
```
>> df = pd.DataFrame({'x': [1, 2, 3, 4, 1, 2, 3, 4], 'y': ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h']})
>> df['group'] = pd.cut(df['x'], 5, ['first group', 'second group', 'third group', 'fourth group', 'fifth group'])
>> df
   x  y         group
0  1  a  (0.997, 1.6]
1  2  b    (1.6, 2.2]
2  3  c    (2.8, 3.4]
3  4  d      (3.4, 4]
4  1  e  (0.997, 1.6]
5  2  f    (1.6, 2.2]
6  3  g    (2.8, 3.4]
7  4  h      (3.4, 4]
````
or, with custom labels:
```
>> df['group'] = pd.cut(df['x'], 5, labels=['first group', 'second group', 'third group', 'fourth group', 'fifth group'])
>> df
   x  y         group
0  1  a   first group
1  2  b  second group
2  3  c  fourth group
3  4  d   fifth group
4  1  e   first group
5  2  f  second group
6  3  g  fourth group
7  4  h   fifth group
````

chriddyp:
thanks for making all of those changes. this feature is really great now. 💃 !

PR #734, opened 2017-04-13
v2.0.8 only write config file if changed

daliati:
We are having an issue of plotly credentials occasionally deleted for some of the users. I was not able to produce the bug consistently. It seems that the credentials are overwritten once concurrency comes in place. 

The .credentials and .config files are open many times for writing (+w), every time the ensure_local_plotly_files (https://github.com/plotly/plotly.py/blob/master/plotly/tools.py#L77) is called, and it is called every time plotly is used. 

This fix will only overwrite (or open for writing) the configuration files, if the configuration keys change, but not every time when plotly is called. Once the correct config is present, those files won't be open for writing anymore, and hence won't be overwritten due to exceptions caused by  system file locks and such, which can happen in case of concurrency.


Kully:
@daliati Thanks for this! I'm going to take a look today.

Kully:
💃 This looks good to me. Thanks @daliati for taking the time to do this.

@cldougl shall we leave the version num change as is here (2.0.8)? You said you were going to update the pip package soon.

cldougl:
yep @Kully that's ok

daliati:
Thanks a lot!

PR #741, opened 2017-04-25
update plotly min

cldougl:
updated plotly.js to v 1.26.0 for offline

cc @Kully @chriddyp


chriddyp:
💃  when tests pass

Kully:
same as ^💃

PR #749, opened 2017-05-05
BUG FIX: init_notebook_mode(True) not work

wang-shun:
Due to the wrong plotly-latest.min.js source url, the init_notebook_mode(connected=True) didn't work correctly. Changing the js url from https://cdn.plot.ly/plotly-latest.min to https://cdn.plot.ly/plotly-latest.min.js solves this issue


wang-shun:
my mistake

PR #778, opened 2017-06-21
update pjs to 1.28.2

cldougl:
headsup @chriddyp @Kully 

- [x] test manually


Kully:
@cldougl it's a 💃 from me.

I tried a few offline examples and they all work fine except for one (gapminder ex in https://plot.ly/python/offline/#generating-offline-graphs-within-jupyter-notebook) but it's an issue with the example itself.

Issue #600, opened 2016-11-01
Plotly needs a jupyterlab extension to work with Jupyter Lab


aggFTW:
As @tc0312 mentioned in https://github.com/jupyterlab/jupyterlab/issues/1188:

Plotly uses AMD to inject its javascript library to the browser from the kernel: [code](https://github.com/plotly/plotly.py/blob/master/plotly/offline/offline.py#L159)
This's not working in JupyterLab right now because lab.html doesn't load require.js.

Discussion in that issue points to the fact that Jupyter Lab will require packages to load front end code via the lab extensions framework.

cc. @mdtusz and @yankev

jasongrout:
I'm not familiar with the plotly code, but perhaps a good model to follow is the [vega plugin](https://github.com/altair-viz/jupyterlab_vega)

rgbkrk:
As noted in jupyterlab/jupyterlab#1188:

nteract has a [plotly renderer](https://github.com/nteract/nteract/blob/0f14082c6e98bbc86a54ea1d3c33d655b1f252f2/src/notebook/components/transforms/plotly.js) that relies on the [Plotly JSON Chart Schema](http://help.plot.ly/json-chart-schema/) with mimetype `application/vnd.plotly.v1+json`

@chriddyp has initial work for that support in https://github.com/plotly/plotly.py/pull/562 (which would translate to jupyterlab as well with the same mimetype).

jasongrout:
nice, thanks for pointing that out, @rgbkrk.

rgbkrk:
JupyterLab extension for Plotly courtesy @gnestor: https://github.com/gnestor/jupyterlab_plotly

sudburyrob:
cc @cldougl

gnestor:
FYI https://github.com/gnestor/jupyterlab_plotly is moving over to https://github.com/jupyterlab/jupyter-renderers. Classic notebook extension coming soon!

Juanlu001:
@gnestor Is there a place where we can track the progress of the classic notebook extension? Thanks!

gnestor:
We're tracking it here: https://github.com/jupyterlab/jupyter-renderers/issues/3

gnestor:
This can be closed with the addition of @jupyterlab/plotly-extension and plotlywidget 👍

PR #793, opened 2017-07-08
Make using offline.init_notebook_mode() optional

lgeiger:
Running `init_notebook_mode()` isn't required in frontends with support for the `application/vnd.plotly.v1+json` mime type (like [nteract](https://github.com/nteract/nteract) and [Hydrogen](https://github.com/nteract/hydrogen)):
plotly

This PR makes `offline.init_notebook_mode()` optional when using `offline.iplot()`

/cc @rgbkrk

### DIFF
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -322,25 +322,12 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
     iplot([{'x': [1, 2, 3], 'y': [5, 2, 7]}], image='png')
     ```
     """
-    if not __PLOTLY_OFFLINE_INITIALIZED:
-        raise PlotlyError('\n'.join([
-            'Plotly Offline mode has not been initialized in this notebook. '
-            'Run: ',
-            '',
-            'import plotly',
-            'plotly.offline.init_notebook_mode() '
-            '# run at the start of every ipython notebook',
-        ]))
     if not ipython:
         raise ImportError('`iplot` can only run inside an IPython Notebook.')
 
     config = dict(config) if config else {}
     config.setdefault('showLink', show_link)
     config.setdefault('linkText', link_text)
-
-    plot_html, plotdivid, width, height = _plot_html(
-        figure_or_data, config, validate, '100%', 525, True
-    )
 
     figure = tools.return_figure_from_figure_or_data(figure_or_data, validate)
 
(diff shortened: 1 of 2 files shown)


rgbkrk:
Hooray!

rgbkrk:
Oops, I thought this was a Hydrogen PR for some reason. At any rate, I'm a big fan and will help maintain on the frontend side for this.

chriddyp:
While `init_notebook_mode` isn't necessary for `nteract`, doesn't removing this prevent notebooks created with `nteract` being shared in other environments that don't support `application/vnd.plotly.v1+json`? For example:
1 - User uses `nteract` and creates a plotly graph
2 - User uploads that notebook to nbviewer / shares it with someone who isn't running `nteract` but is running an older `jupyter notebook`
3 - Graphs don't work?

rgbkrk:
The main reasoning is for Hydrogen users and others to have a way to plot immediately without the `init_notebook_mode` ritual, since plotly is integrated into the nteract frontends directly. JupyterLab is shipping with vega built in by default now.

/cc @ellisonbg

chriddyp:
Got it. It definitely makes sense for the nteract and hydrogen users who don't need it. I'm still concerned about the rest of the users that are using Jupyter notebook:
- nteract user's notebooks won't be able to be shared with Jupyter notebook users (until JupyterLab supports `application/vnd.plotly.v1+json` and users upgrade to that) or on nbviewer or plotly
- Jupyter notebook users who forget to do `init_notebook_mode` won't be warned anymore, encoutering some confusion when seeing a blank plot

Maybe what we can do instead is detect for `Plotly` and display an error message (in JS) if it's not found telling them to run `init_notebook_mode`. This "script" wouldn't run in nteract or hydrogen but it would run in the raw Jupyter Notebook HTML Output.

Would something like that work?

rgbkrk:
> Maybe what we can do instead is detect for Plotly and display an error message (in JS) if it's not found telling them to run init_notebook_mode. This "script" wouldn't run in nteract or hydrogen but it would run in the raw Jupyter Notebook HTML Output.

Yeah that makes sense. We won't have `Plotly` defined globally but we'll be taking precedence on the `application/vnd.plotly.v1+json` output anyways. 💯

Juanlu001:
If an extension for the classic notebook is being developed (see https://github.com/plotly/plotly.py/issues/600#issuecomment-323556106), I guess the warnings or even the `init_notebook_mode` won't be needed anymore?

PR #806, opened 2017-08-02
Add missing rugplot property passing inside violin_no_colorscale

n17r4m:
When trying to plot my violin, I noticed rugplot=False was not being respected. This is a proposed fix.
As a point of discussion, should the rugplot=True|False property name be normalized between the figure factories? For example, in _distplot.py the corresponding property name is "show_rug" instead of "rugplot".

### DIFF
--- a/plotly/figure_factory/_violin.py
+++ b/plotly/figure_factory/_violin.py
@@ -224,7 +224,8 @@ def violin_no_colorscale(data, data_header, group_header, colors,
         if color_index >= len(colors):
             color_index = 0
         plot_data, plot_xrange = violinplot(vals,
-                                            fillcolor=colors[color_index])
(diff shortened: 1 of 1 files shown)


cldougl:
Hey @n17r4m could you let me know what issue you're having exactly? `rugplot=False` seems to be working for me but maybe you're expecting a different result. Here's an example: https://plot.ly/~chelsea_lyn/15199

n17r4m:
I'm very very sorry, I should have mentioned that the issue is only experienced when drawing multiple violin plots (which, as I found out, requires the use of pandas DataFrames, which as a related issue, sending in arrays of data lists is not supported by the current violin implementation)

cldougl:
@n17r4m got it- thanks for the fix!

PR #807, opened 2017-08-02
Add annotation parameter for dendrogram

parulsethi:
Hi,

I've added an additional parameter which could enable users to define text annotations for complete hierarchy (on hover). See the example usage in this [notebook](http://nbviewer.jupyter.org/github/parulsethi/gensim/blob/c1e2b9be39cb242a190a1357e7a566bb198f98a3/docs/notebooks/Topic_dendrogram.ipynb).
`annotation` would be a list containing the list of text strings for every single trace which makes up the dendrogram.

Let me know if there should be any modifications to make it adaptable for more general use for plotly users.


cldougl:
Hey @parulsethi -- thanks for this pr! Typically we use annotations to describe text that is static on the chart and hovertext to describe the text visible on hover, so it may be better to rename that attribute.

parulsethi:
Thanks for the suggestion, updated the attribute name

PR #810, opened 2017-08-03
edit stage test account and update plot schema

cldougl:


cldougl:
cc @chriddyp tests were failing because one of the test accounts was removed (from the stage db). I've created a new account and mentioned the accounts used in our tests in the contributing doc.

chriddyp:
💃

Issue #805, opened 2017-08-01
Gantt chart doesn't have layout attribute. 


lwileczek:
I created a gantt chart from Figure Factory and plot some 500 bars. The Task list on the y-axis looks bad because of this. I was hoping to be able to hide it. I tried:

  `  figure = ff.create_gantt(df,index_col='Resource', group_tasks=True,
 title ='Time at each gate by Plane',width=1400,height=600)
    figure.layout.yaxis1.update({'showticklabels': False})`
 but I get the error
    

>  figure.layout.yaxis1.update({'showticklabels': False})
AttributeError: 'dict' object has no attribute 'layout'

Is there any way around this? or will there be an update adding 'layout' attribute to gantt charts?

![gantt_error](https://user-images.githubusercontent.com/30635219/28842253-ec5a9ce2-76ca-11e7-9cde-aa8cdc738d14.PNG)

on the plotly webpage you can load a gantt chart into the gooey interface you can select style>layout>Title and fonts> color and select white to hide this problem. Is there a way I can do this through the python code I am writing?
![delete](https://user-images.githubusercontent.com/30635219/28933543-6af4373a-784b-11e7-84a2-48e5a34d5f14.PNG)

crgboud:
I am still able to reproduce this error on plotly 3.1.0 with the following code:

```python
import plotly.figure_factory as ff

df = [dict(Task="Job-1", Start='2017-01-01', Finish='2017-02-02', Resource='Complete'),
      dict(Task="Job-1", Start='2017-02-15', Finish='2017-03-15', Resource='Incomplete'),
      dict(Task="Job-2", Start='2017-01-17', Finish='2017-02-17', Resource='Not Started'),
      dict(Task="Job-2", Start='2017-01-17', Finish='2017-02-17', Resource='Complete'),
      dict(Task="Job-3", Start='2017-03-10', Finish='2017-03-20', Resource='Not Started'),
      dict(Task="Job-3", Start='2017-04-01', Finish='2017-04-20', Resource='Not Started'),
      dict(Task="Job-3", Start='2017-05-18', Finish='2017-06-18', Resource='Not Started'),
      dict(Task="Job-4", Start='2017-01-14', Finish='2017-03-14', Resource='Complete')]

colors = {'Not Started': 'rgb(220, 0, 0)',
          'Incomplete': (1, 0.9, 0.16),
          'Complete': 'rgb(0, 255, 100)'}

[...]

PR #811, opened 2017-08-04
return gantt and dendrogram figure

cldougl:
cc @chriddyp 
return plotly figure for gantt and dendrograms so they are consistent with other figure factories

fixes https://github.com/plotly/plotly.py/issues/805


chriddyp:
:dancer:

Issue #820, opened 2017-08-19
Change axis widths


willgdjones:
Hi there - I am creating a Gantt chart and would like to widen the axis because my tasks are longer than the default:

<img width="169" alt="screen shot 2017-08-19 at 06 23 29" src="https://user-images.githubusercontent.com/1719848/29486981-0e97046e-84a7-11e7-9c87-5c3bcf3b3ce8.png">

What's the best way to do this?

Thanks

Kully:
@willgdjones 
Hello there.

There seems to be a few problems with your output: you only have one date showing on the chart, so I suspect you are putting an invalid date for one of your bounds.

Also, I don't exactly understand your question. You want all of your tasks to show because yours are longer correct? The window should adjust accordingly.

Can you include your source code so I can better help you? Thank you.

willgdjones:
Hi there - thanks for the response. All I would like to do actually is expand the width of the chart for the y-ticks so that the labels do not yet cut off.

I have created a screen grab of the left hand side of my plot, and you can see that the labels have been cropped! Just want to increase that space so that it doesn't crop a word.

Thanks a lot.

Kully:
So I was going to create a fix but there is an easy thing to remember. You can just adjust the margin of your chart by doing the following thing:

```
fig = gant_chart(...)
fig['layout']['margin'] = {'l': 200}
```
The parameter is documented here: 

https://plot.ly/python/reference/#layout-margin-l

willgdjones:
So I didn't find that parameter! Thanks for pointing it out.

PR #832, opened 2017-09-15
Adds a 'lmargin' param to allow users to adjust left margin

Kully:
addresses this issue: https://github.com/plotly/plotly.py/issues/820#issuecomment-329724739 i.e. job titles getting cut off


chriddyp:
Instead of adding parameters for every `layout` attribute, can we just tell users how to modify the figure afterwards?
```
fig = gant_chart(...)
fig['layout']['margin'] = {'l': 80}
```

Kully:
> Instead of adding parameters for every layout attribute, can we just tell users how to modify the figure afterwards?

That sounds a lot easier. I'm onboard. Shall I do that then?

PR #837, opened 2017-09-21
Remove offline tests

Kully:
@cldougl attempting to resolve the intermittent fails on circle due to tests signing into stage


cldougl:
@Kully it would be helpful to remove the presentations file from this pr

cldougl:
looks like those are valid dendrogram failures now so you may want to investigate if those tests are manually working as expected

cldougl:
thanks for taking a look at these tests @Kully !!
💃

PR #850, opened 2017-10-05
v1.31 plotly.js for offline

cldougl:
- [x] test manually

cc @chriddyp @Kully


chriddyp:
💃  

Here's the related Dash PR: https://github.com/plotly/dash-core-components/pull/87 (`v0.13.0`)

PR #852, opened 2017-10-09
update graph_objs.py

cldougl:
update for https://github.com/plotly/streambed/pull/10150
cc @jackparmer @chriddyp @Kully


plotly/version.py
```diff
@@ -1 +1 @@
-__version__ = '2.0.16'
+__version__ = '2.0.17'
```
chriddyp: I think we've actually been doing semver wrong. `Table` is a new feature, so I believe that this should be `2.1.0` (http://semver.org/)

chriddyp:
was this generated by a script or done by hand?

plotly/version.py
cldougl: oh I thought we were only bumping if it was a new plotly.py specific feature

cldougl:
by script: `make update_default_schema`
https://github.com/plotly/plotly.py/blob/master/makefile#L19

jackparmer:
💃  looks like it will work to me. Once it's on pip I'm happy to take it for a spin on `streambed`

PR #862, opened 2017-10-20
Corrects typo in docstring in graph_objs.py

ketigid:
Corrects a typo in docstring in graph_objs.py.


cldougl:
thanks @ketigid !

PR #864, opened 2017-10-25
Fix issue with concurrent import of plotly module and checking permissions

alexnirvan:
Issue raises when _permissions() is called concurrently (Ex.: in case of concurrent import plotly from two different processes or scripts):
- open TEST_FILE on a second call would raise an exception and _files_permissions would potentially wrongfully set to False.

### DIFF
--- a/plotly/files.py
+++ b/plotly/files.py
@@ -1,11 +1,10 @@
 import os
+from stat import ST_MODE, S_IWUSR, S_IRUSR
 
 # file structure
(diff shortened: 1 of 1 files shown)


jonmmease:
Closing as this issue has been resolved by https://github.com/plotly/plotly.py/pull/1195 and released in version 3.3.0. Feel free to open a new issue if you're still having trouble. Thanks!

PR #866, opened 2017-10-26
Presentation setup.py fix

Kully:
is adding `presentation_objs` file to setup.

Do I need to update version number to 2.2.1?

cc @cldougl @chriddyp


cldougl:
yep we'll have to bump the version number to resend to pip

cldougl:
the fix looks good to me 👍  
@kully can you update the version & changelog in this pr?

Kully:
> the fix looks good to me 👍
> @Kully can you update the version & changelog in this pr?

doing now

cldougl:
💃  after tests pass

PR #867, opened 2017-10-26
Fix reversed color scale in annotated heatmap

inguar:
Currently, `reversescale` does not get included into the `trace` dictionary. Therefore, color scale doesn't change when setting `reversescale=True` (although the color of annotations gets inverted).

Here is how to reproduce the error:
```python
iplot(ff.create_annotated_heatmap(z=[[1, 30], [20, 60]], reversescale=False))
```
plot1

```python
iplot(ff.create_annotated_heatmap(z=[[1, 30], [20, 60]], reversescale=True))
```
plot2

This pull request fixes the problem.


jonmmease:
Superseded by #1251. Thanks for getting this started @inguar!

PR #872, opened 2017-11-07
Bullet Chart FF

Kully:
NEEDED: Tests

update: tests added


Kully:
Screenshots:
screen shot 2017-11-07 at 3 24 16 pm

With Colorhunt theme:
screen shot 2017-11-07 at 3 28 38 pm
content.com/10369095/32516254-ba2ea526-c3d0-11e7-8aac-b5e3b8ec71f4.png)

Kully:
@chriddyp ready for a review

chriddyp:
It looks like we're using `shapes` to create these bars. Have you looked into using bar chart traces with `width`, `offset`, and `base` instead? Here's the PR that added it https://github.com/plotly/plotly.js/pull/1075 and here is an example that I was making a couple of weeks ago for a separate project: https://codepen.io/chriddyp/pen/mBQVEy

chriddyp:
The advantage of using bar charts with width and offset are:
- Easier to load and style inside chart studio (https://plot.ly/create)
- Easier for user to update their figure programatically 
- Hover labels

Kully:
@chriddyp ready for a review now. The code has been reworked for bar traces 📊

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,257 @@
...
+                  measure_colors=None, subplot_spacing=None,
+                  title='Bullet Chart', height=600, width=1000):
+    """
+    Returns figure for bullet chart.
+
+    :param (pd.DataFrame | list) df: either a JSON list of dicts or a pandas
+        DataFrame. All keys must be one of 'title', 'subtitle', 'ranges',
+        'measures', and 'markers'.
+    :param (bool) as_rows: if True, the bars are placed horizontally as rows.
+        If False, the bars are placed vertically in the chart.
```
chriddyp: Would `orientation='v' | 'h'` be more consistent with the rest of plotly here?

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,257 @@
...
+def create_bullet(df, as_rows=True, marker_size=16,
+                  marker_symbol='diamond-tall', range_colors=None,
+                  measure_colors=None, subplot_spacing=None,
+                  title='Bullet Chart', height=600, width=1000):
+    """
+    Returns figure for bullet chart.
+
+    :param (pd.DataFrame | list) df: either a JSON list of dicts or a pandas
+        DataFrame. All keys must be one of 'title', 'subtitle', 'ranges',
+        'measures', and 'markers'.
```
chriddyp: This seems unlike the rest of our figure factories, but I might be mistaken. It seems odd to require users to reformat their dataframes. Would it be better if it was something like:
```
create_bullet(df, title='column a', measures='column b', markers='column c')
```
? 

In other words, let users pass in whatever dataframe they want but then have them specify the column mapping

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,257 @@
...
+        )
+        fig['layout']['annotations'].append(annot)
+
+    return fig
+
+
+def create_bullet(df, as_rows=True, marker_size=16,
+                  marker_symbol='diamond-tall', range_colors=None,
+                  measure_colors=None, subplot_spacing=None,
+                  title='Bullet Chart', height=600, width=1000):
```
chriddyp: Do we include these types of properties in the other figure factories? I personally would prefer a more generic solution like requiring users to modify the layout directly or passing in `**layout_options` or `layout=None`
```
fig = create_bullet(...)
fig['layout'].update({
    'title': 'My Chart',
    'height': 600,
    'width': 1000,
    'annotations': [{...}]
})
```

or (`**layout_options`)

```
fig = create_bullet(..., title='my chart', width=500, annotations=[{...}])
```

or (`layout_options`)
```
fig = create_bullet(..., layout_options={'title': 'my chart', 'width': 500, 'annotations': [{...}]})
```

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,257 @@
...
+    Returns figure for bullet chart.
+
+    :param (pd.DataFrame | list) df: either a JSON list of dicts or a pandas
+        DataFrame. All keys must be one of 'title', 'subtitle', 'ranges',
+        'measures', and 'markers'.
+    :param (bool) as_rows: if True, the bars are placed horizontally as rows.
+        If False, the bars are placed vertically in the chart.
+    :param (int) marker_size: sets the size of the markers in the chart.
+    :param (str | int) marker_symbol: the symbol of the markers in the chart.
+        Default='diamond-tall'
```
chriddyp: similarly here, can we make this more generic with something like `scatter_options=None` that get recursively merged into the trace:

```
fig = create_bullet(df, scatter_options={'marker': {'size': 12, 'symbol': 'diamond-tall'}, 'name': 'Bullet'})
```

By making it more abstract, we allow the user to pass in whatever they want and we don't have to keep adding new attributes when they are requestd (e.g. what if the user wants a line around their markers, or wants to change the name of the trace, or wants to change the color of the marker - instead of having `marker_line_color=None, name=None, marker_color=None`, we could just let the user pass in an object that we recursively merge onto the appropriate trace)

chriddyp: by recursively merge i mean deep merge

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,257 @@
...
+        the rectangles for the range are drawn. These rectangles are meant to
+        be qualitative indicators against which the marker and measure bars
+        are compared.
+        Default=['rgb(198, 198, 198)', 'rgb(248, 248, 248)']
+    :param (list) measure_colors: a list of two colors which is used to color
+        the thin quantitative bars in the bullet chart.
+        Default=['rgb(31, 119, 180)', 'rgb(176, 196, 221)']
+    :param (float) subplot_spacing: set the distance between each bar chart.
+        If not specified an automatic spacing is assigned based on the number
+        of bars to be plotted.
```
chriddyp: What units are these in and what's the range of these numbers?

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,257 @@
...
+        )
+
+    # add necessary columns if missing
+    for key in VALID_KEYS:
+        if key not in df:
+            if key in ['title', 'subtitle']:
+                element = ''
+            else:
+                element = []
+            df[key] = [element for _ in range(len(df))]
```
chriddyp: ⛔️  - We should never mutate the user's data - these functions should have no side-affects

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,257 @@
...
+
+    # make sure ranges and measures are not NAN or NONE
+    for needed_key in ['ranges', 'measures']:
+        for idx, r in enumerate(df[needed_key]):
+            try:
+                r_is_nan = math.isnan(r)
+                if r_is_nan or r is None:
+                    df[needed_key][idx] = []
```
chriddyp: similarly here. if you need to change data, make a copy of their data and then mutate the copy

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,257 @@
...
+
+    fig = _bullet(
+        df, as_rows, marker_size, marker_symbol, range_colors, measure_colors,
+        subplot_spacing
+    )
+
+    fig['layout'].update(
+        title=title,
+        height=height,
+        width=width,
```
chriddyp: if this was `layout_options` then you could just pass in `fig['layout'].update(layout_options)`

plotly/tests/test_optional/test_figure_factory.py
```diff
@@ -2184,3 +2183,503 @@ def test_valid_facet_grid_fig(self):
...
+            range_colors=range_colors
+        )
+
+        self.assertRaisesRegexp(
+            PlotlyError, pattern, ff.create_bullet, df,
+            measure_colors=measure_colors
+        )
+
+    def test_full_bullet(self):
+        df = pd.read_json('https://cdn.rawgit.com/plotly/datasets/master/BulletData.json')
```
chriddyp: Can we include this data in the repo itself? That way these tests don't break if that dataset is removed or if `rawgit` goes down.

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,257 @@
...
+            measure_colors, subplot_spacing):
+    num_of_lanes = len(df)
+    num_of_rows = num_of_lanes if as_rows else 1
+    num_of_cols = 1 if as_rows else num_of_lanes
+    if not subplot_spacing:
+        subplot_spacing = 1./num_of_lanes
+    fig = plotly.tools.make_subplots(
+        num_of_rows, num_of_cols, print_grid=False,
+        horizontal_spacing=subplot_spacing,
+        vertical_spacing=subplot_spacing
```
chriddyp: does it make sense to provide `horizontal_spacing` and `vertical_spacing` as input arguments to `ff.make_bullet` instead? that way it's consistent with `make_subplots`

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,257 @@
...
+    :param (list) measure_colors: a list of two colors which is used to color
+        the thin quantitative bars in the bullet chart.
+        Default=['rgb(31, 119, 180)', 'rgb(176, 196, 221)']
+    :param (float) subplot_spacing: set the distance between each bar chart.
+        If not specified an automatic spacing is assigned based on the number
+        of bars to be plotted.
+    :param (str) title: title of the bullet chart.
+    :param (float) height: height of the chart.
+    :param (float) width width of the chart.
+    """
```
chriddyp: Could we include a couple of simple examples in here?

plotly/colors.py
```diff
@@ -442,21 +442,37 @@ def make_colorscale(colors, scale=None):
...
-            lowcolor[2] + intermed * diff_2)
+    inter_med_tuple = (
+        lowcolor[0] + intermed * diff_0,
+        lowcolor[1] + intermed * diff_1,
+        lowcolor[2] + intermed * diff_2
+    )
+
+    if colortype == 'rgb':
+        # covert back to rgb
+        inter_med_rgb = label_rgb(inter_med_tuple)
```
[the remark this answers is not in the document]
Kully: Yes that's what I mean. More clearer your way.

plotly/colors.py
```diff
@@ -498,14 +514,20 @@ def convert_to_RGB_255(colors):
...
-def n_colors(lowcolor, highcolor, n_colors):
+def n_colors(lowcolor, highcolor, n_colors, colortype='tuple'):
     """
     Splits a low and high color into a list of n_colors colors in it
 
     Accepts two color tuples and returns a list of n_colors colors
     which form the intermediate colors between lowcolor and highcolor
-    from linearly interpolating through RGB space
+    from linearly interpolating through RGB space. If colortype is 'rgb'
+    the function will return a list of colors in the same form.
```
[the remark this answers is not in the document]
Kully: I can see how there's confusion.

If `colortype='tuple'`, `lowcolor` and `highcolor` must be what I have called tuple(-like) colors i.e. tuples of the form (a,b,c) where a, b, c are between 0 and 1, where (0,0,0) == rgb(0,0,0), etc.

If the input is`lowcolor`=(0,0,0)  `highcolor`=(1,1,1), `n_colors=3` then the output is:

`[(0,0,0), (0.5,0.5,0.5), (1,1,1)]`

If `colortype='rgb`, then the list above will contain the rgb equivalent colors in a list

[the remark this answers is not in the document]
Kully: Also this function will error if you input tuple-colors and colortype is set to `rgb`. Another thing to fix.

plotly/colors.py
```diff
@@ -520,6 +542,10 @@ def n_colors(lowcolor, highcolor, n_colors):
                      lowcolor[2] + (index * incr_2))
         color_tuples.append(new_tuple)
 
+    if colortype == 'rgb':
+        # convert back to rgb
+        color_tuples = color_parser(color_tuples, label_rgb)
```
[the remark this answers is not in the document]
Kully: It should, but I kept it like that so that `color_tuples` is a common variable in both paths the function can take. Shall I change to `list_of_colors` for an apt name?

plotly/figure_factory/_bullet.py
Kully: Yes it does!

plotly/figure_factory/_bullet.py
Kully: Haha, if there are they are probably ones I wrote. This makes way more sense, great suggestion.

plotly/figure_factory/_bullet.py
Kully: Yes again. :+1:

plotly/figure_factory/_bullet.py
Kully: 0 to 1.

Kully:
@chriddyp bumped ver num and CHANGELOG update

ready for a final 💃

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,342 @@
...
+            marker=scatter_options['marker'],
+            name='markers',
+            hoverinfo='x' if orientation == 'h' else 'y',
+            xaxis='x{}'.format(row + 1),
+            yaxis='y{}'.format(row + 1)
+        )
+
+        for k in scatter_options:
+            if k != 'marker':
+                markers[k] = scatter_options[k]
```
chriddyp: Could this be written just like:
```
go.Scatter(
    x=x,
    y=y,
    name='markers',
    hoverinfo='...',
    xaxis='...',
    yaxis='...',
    **scatter_options
)
```
?

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,342 @@
...
+        ranges of performance, usually a 3-item list [bad, okay, good]. They
+        correspond to the grey bars in the background of each chart.
+    :param (str) subtitles: the column name or dictionary key for the subtitle
+        of each subplot chart. The subplots are displayed right underneath
+        each title.
+    :param (str) titles: the column name or dictionary key for the main label
+        of each subplot chart.
+    :param (bool) orientation: if 'h', the bars are placed horizontally as
+        rows. If 'v' the bars are placed vertically in the chart.
+    :param (int) marker_size: sets the size of the markers in the chart.
```
chriddyp: this is outdated

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,342 @@
...
+    :param (str | int) marker_symbol: the symbol of the markers in the chart.
+        Default='diamond-tall'
```
chriddyp: this too

plotly/figure_factory/_bullet.py
Kully: nice catch

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,342 @@
...
+    :param (list) range_colors: a list of two colors between which all
+        the rectangles for the range are drawn. These rectangles are meant to
+        be qualitative indicators against which the marker and measure bars
+        are compared.
+        Default=['rgb(198, 198, 198)', 'rgb(248, 248, 248)']
```
chriddyp: I don't see `rgb(248, 248, 248)` anywhere else in the code.

Would it make sense to supply this value directly in the function instead of `None`? i.e.
```
def create_bullet(data, markers=None, measures=None, ranges=None,
                   subtitles=None, titles=None, orientation='h',
                   range_colors=('rgb(198, 198, 198)', 'rgb(248, 248, 248)',),
                   measure_colors=('rgb(31, 119, 280)', 'rgb(176, 196, 221)', ),
                   horizontal_spacing=None, vertical_spacing=None,
                   scatter_options={}, **layout_options):
```
(note that I'm using a tuple instead of a list because mutable default arguments aren't safe: http://docs.python-guide.org/en/latest/writing/gotchas/)

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,342 @@
...
+        scatter_options={'marker': {'symbol': 'circle'}}, width=700
+
+    )
+    py.iplot(fig)
+    ```
+    """
+    # validate df
+    if not pd:
+        raise exceptions.ImportError(
+            "'pandas' must be imported for this figure_factory."
```
chriddyp: "imported" - it's really installed, not imported, right? The user doesn't need to do a `import pandas` before they call this function do they?
If so, could we correct this typo in any other place that it might appear

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,342 @@
...
+        )
+
+    if isinstance(data, list):
+        if not all(isinstance(item, dict) for item in data):
+            raise exceptions.PlotlyError(
+                'If your data is a list, all entries must be dictionaries.'
```
chriddyp: I think it's confusing to lead with a "If your data is a list" - the user might wonder if there data is indeed a list or not. What about something like "Every entry of the `data` argument (a list) must be a dictionary."

Kully: good point!

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,342 @@
...
+    if isinstance(data, list):
+        df = pd.DataFrame(
+            [
+                [d[titles] for d in data] if titles else [''] * len(data),
+                [d[subtitles] for d in data] if subtitles else [''] * len(data),
+                [d[markers] for d in data] if markers else [[]] * len(data),
+                [d[measures] for d in data] if measures else [[]] * len(data),
+                [d[ranges] for d in data] if ranges else [[]] * len(data),
+            ],
+            index=col_names
```
chriddyp: Rather than looping through every element and reconstructing a list, I believe that you can just pass in your list of dictionaries directly into pandas:
```
In [2]: pd.DataFrame([{'x': 1, 'y': 3}, {'x': 3, 'y': 10}])
Out[2]:
   x   y
0  1   3
1  3  10
```

plotly/figure_factory/_bullet.py
```diff
@@ -0,0 +1,342 @@
...
+      {"e": "Order Size", "d":"US$, average","c": [350, 500, 600],
+       "b": [100,320],"a": [550]},
+      {"e": "New Customers", "d": "count", "c": [1400, 2000, 2500],
+       "b": [1000,1650],"a": [2100]},
+      {"e": "Satisfaction", "d": "out of 5","c": [3.5, 4.25, 5],
+       "b": [3.2,4.7], "a": [4.4]}
+    ]
+
+    fig = ff.create_bullet(
+        data, titles='e', subtitles='d', markers='a', measures='b',
```
chriddyp: could we use more descriptive names than `e`, `d`, `a`, `b`?

chriddyp:
:dancer once my comments are addressed

plotly/figure_factory/_bullet.py
Kully: That's true, but I want to insert empty columns where a column name is not inserted.

Kully: Ex. if the user doesn't assign the `titles` variable to anything, I want a column named `titles` with empty strings in each cell of that column

plotly/figure_factory/_bullet.py
Kully: yes can do. the point was to have titles that are different than the names of the columns that the second dataframe I construct in the figure factory.

Kully:
What do you think also about allowing tuples as input arguments for the data?

There's no reason why not to allow this.

I was thinking of altering the part of code that checks if `isinstance(data, list)` to something like:

`hasattr(data, index)`

plotly/figure_factory/_bullet.py
Kully: that's correct, it just needs to be installed on their system

chriddyp:
> What do you think also about allowing tuples as input arguments for the data?

To an extent, python is a duck-typed language so we don't always need to explicitly check the type as long as the object that is supplied implements the right methods. So lists, tuples, numpy arrays, (and others) might look and feel the same and should all be accepted.

Here is a solution that checks if the object is a `Sequence` which is the base class for things like lists and tuples: https://stackoverflow.com/a/37842328/4142536

plotly/figure_factory/_bullet.py
Kully: How about this:

```
'Every entry of the data argument (a list, tuple, etc) must be a dictionary.'
```

chriddyp: 👍  very nice

plotly/figure_factory/_bullet.py
Kully: Can I leave this the same way it is? I think it will be more straightforward than doing `pd.DataFrame(data)` then adding in the empty columns afterwards.

chriddyp: sure

PR #875, opened 2017-11-10
Add dependency versions

christopherjenness:


setup.py
```diff
@@ -49,7 +49,7 @@ def readme():
       package_data={'plotly': ['package_data/*']},
       install_requires=['decorator>=4.0.6',
                         'nbformat>=4.2',
-                        'pytz',
-                        'requests',
-                        'six'],
```
chriddyp: It looks like these are the _latest_ versions. I believe that this package works with older versions than these (I haven't looked into the earliest versions that we are compatible with) and so I don't think that we should force users to upgrade to the latest version in case that breaks a separate package's dependencies. In other words, our install requires should be as lenient as possible.

christopherjenness: Good point.  Do you know which versions are compatible?

jonmmease:
Closing as stale and we don't intend to restrict the dependency versions more than necessary.

PR #878, opened 2017-11-13
manage plotly/colors.py and plotly/figure_factory/utils.py

Kully:
Currently a couple figure factories are using incorrect plotly colorscales.

A PR to ensure that:
- the correct plotly colorscales are being used by all figure factories
- functions for color manipulating/handling are _only_ in the `colors.py` module
- rewritting some figure_factory code to comply with correct colorscales

### DIFF
--- a/plotly/colors.py
+++ b/plotly/colors.py
@@ -608,3 +608,50 @@ def convert_colorscale_to_rgb(colorscale):
     for color in colorscale:
         color[1] = label_rgb(color[1])
     return colorscale
+
+
+def endpts_to_intervals(endpts):
+    """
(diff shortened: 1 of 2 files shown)


jonmmease:
@Kully Is there anything here that we should pick back up?

Kully:
> Is there anything here that we should pick back up?

yes

so the original plan was to organize utils and colors files in the figure_factory folder. There are some duplicates functions across the two folders and some figure factory modules use `function A` from one file but `function A` from the other.

Kully:
closing in favour of https://github.com/plotly/plotly.py/pull/1279

PR #885, opened 2017-11-24
Subplot proportions

Kully:
Want to add way to specify column proportions and row proportions in the domains of subplots, like `go.Table`.

(Note that this in service of a [Sparkline](https://plotly.slack.com/files/U07MQ8SJV/F7Z4B2M16/image.png) FigureFactory I'm working on)

So far I've added `column_width` as another accepted kwargs for `plotly.tools.make_subplots()`. The idea is that you specify a ratio of the columns (eg. [4, 7, 7, 2]) and the domains are changed proportionally.

cc @etpinard since you originally wrote most of this code
@chriddyp any comments would be helpful

note: I changed a couple of the tests as some of the domain values were off by the last floating point due to changing the number of +'s and x's in computing the main grid. Personally I don't think it's much to worry about.


etpinard:
Looks good to me :dancer:

plotly/tools.py
```diff
@@ -807,7 +807,7 @@ def make_subplots(rows=1, cols=1,
 
     # Throw exception if non-valid kwarg is sent
     VALID_KWARGS = ['horizontal_spacing', 'vertical_spacing',
-                    'specs', 'insets', 'subplot_titles']
+                    'specs', 'insets', 'subplot_titles', 'column_width']
```
etpinard: Oh actually, you should add a docstring about this above.

Kully: > Looks good to me

I also want to add a `row_width` param. I'll do that then ping you after your other comment is addressed

plotly/tools.py
```diff
@@ -907,21 +907,44 @@ def _checks(item, defaults):
...
-    width = (1. - horizontal_spacing * (cols - 1)) / cols
+    # Set height of each subplot cell (excluding padding)
     height = (1. - vertical_spacing * (rows - 1)) / rows
 
+    try:
+        column_width = kwargs['column_width']
+        if not isinstance(column_width, list) or len(column_width) != cols:
+            raise Exception(
+                "Keyword argument 'column_width' must be a list "
+                "containing the same number of floats as the number of cols."
```
chriddyp: or ints? maybe just `numbers`?

Kully:
@etpinard I've added `row_width` and added docs and tests! :taco:

Issue #444, opened 2016-04-19
Converting ipynb to html using latex and plot.ly does not work properly


jaschwer:
I am using jupyter to write a notebook using latex and plotly.
Within the notebook everything is fine and I can see both properly, but after converting the notebook into a html (no matter if I download it via the menu or use a nbconvert in the terminal), I do see the plot, but the equations are shown in the normal markdown code. You can see the result [here](https://github.com/jaschwer/Third_Assignment/blob/master/test2.html).

The notebook can be found [here](https://github.com/jaschwer/Third_Assignment), named test2

I have the latest mathjax version and I did not find any suggestions using google. Not including the plotly packages nor the plot I get the latex equations right.

Can anybody help me?

Problem arised both on mac and windows.

cldougl:
refer to https://github.com/plotly/plotly.py/issues/515 for updates in this issue

PR #888, opened 2017-11-29
Sparklines Figure Factory

Kully:
@jackparmer progress so far

screen shot 2017-11-29 at 3 28 32 pm

you can set column_widths (like in go.Table) with a list [1, 4, 2, ...]
you set the chart types with a list: i.e. the one above is produced with `chart_types=['name', 'bullet', 'line', 'avg', 'bar']`

```
# TODO:
# add  more color variety
# add tests
```
any suggestions?


Kully:
Playing around with the column_width param:
```
fig = ff.create_sparkline(
    df, chart_types=['name', 'bullet', 'bar'],
    column_width=[1, 7, 10], left_aligned=False
)
py.iplot(fig, filename='my-sparkline-obvious-column-widths')
```
https://plot.ly/~AdamKulidjian/5610

Picture:
screen shot 2017-11-29 at 3 35 04 pm

jackparmer:
Looks like they're on the right track! You happy with them?

Can you do a lines example? For example, can you recreate this?
image

I think there's more vertical and horizontal padding needed between the subplots:

image

Can you also try adding an option to alternate row color? As a sort of benchmark, see how close you can get to this:

image

(From Stephen's [blog](http://www.perceptualedge.com/blog/wp-content/uploads/2012/10/dashboard-competition-winner.png))

Kully:
Some of the possibilities:
screen shot 2017-12-08 at 3 30 58 pm

```
df = pd.read_csv('https://raw.githubusercontent.com/plotly/datasets/master/Canada%20Immigration.csv')

fig = ff.create_sparkline(
    df, chart_types=['name', 'line', 'bullet', 'bar'],
    textalign='center', show_titles=True,
    horizontal_spacing=0, alternate_row_color=True,
    lane_colors=('rgba(251, 252, 230, 0.6)',
                 'rgba(255, 245, 219, 0.6)'),
    title=''
)
py.iplot(fig)
```
|
V

screen shot 2017-12-08 at 3 55 48 pm

Kully:
@chriddyp and @cldougl ready for a review when tests pass 😛

chriddyp:
@Kully - do you have a sample notebook or documentation that shows how this works?

jackparmer:
A link to a sample notebook on plotly or nbviewer would be really great

Kully:
> @Kully - do you have a sample notebook or documentation that shows how this works?

I'll make one now. Caught another bug in the FF so I'll deal with that first

Kully:
@jackparmer  @chriddyp 
here's a notebook with nearly all the features of the ff: https://plot.ly/~AdamKulidjian/6128

jackparmer:
This looks really nice. Nice work. A few small suggestions and usage questions that jumped out at me:

- Maybe `label` instead of `name`
- `text_align` instead of `textalign` to match the other kwargs
- Maybe `row` instead of `lane` and `row_colors` instead of `lane_colors`?
- What is `colors`? Maybe rename this `sparkline_colors` or `trace_colors`? 
- Is `column_width` using a grid system (eg `column_width=[1, 13]` and `column_width=[1, 4, 4]`)? I can't figure it out. Should it be absolute pixels instead? 
- The mpg.csv example is great!

jackparmer:
It would nice if `colors` supported varying trace colors by row (such as in the plot below). I'm also thinking layouts like all of the row traces are grey except the 49th one, which is red.

image

Kully:
> Is column_width using a grid system (eg column_width=[1, 13] and column_width=[1, 4, 4])? I can't figure it out. Should it be absolute pixels instead?

`column_width` comes from the `go.Table` param of the same name (actually `columnwidth`) and which is now a new param in `tls.make_subplots`. It's a ratio assigned for each column. So [1, 2] for a 2-column subplot means column 1 gets 1/3 of the width and column 2 gets 2/3 of the total width.

Kully:
> It would nice if colors supported varying trace colors by row (such as in the plot below). I'm also thinking layouts like all of the row traces are grey except the 49th one, which is red.

We could do something simple like alternating through the Plotly colors if the user wants changing trace colors. But we'll run out quickly for large enough figures.

How do you feel about
a) interpolating between existing plotly colors when we go through all 10
b) being able to enter a Plotly Colorscale name and the traces getting their own equally spaced color within the colorspectrum of that colorscale

I have all this machinery to do this stuff easily 🎰

jackparmer:
You already have `colors` which seems to take a list of 2 colors (Is that correct, these 2 colors are primary and secondary trace colors for each row?). What if `colors` could also take a list of list of colors. For example:

#### What you have now:
`colors = ['#e763fa', '#ab63fa']`

#### Same as what you have now, but alternates primary and secondary trace colors by row:
`colors =[['#e763fa', '#ab63fa'], ['#636efa', '#EF553B']]`

#### traces of every 3rd row are red:
`colors =[['#444', '#ccc'], ['#444', '#ccc'], ['#00ffff', '#00fff',]]`

#### 1D lists longer than 2 elements are the primary trace color for each row, so this could be a colorscale
`colors=['#00cc96', '#00cc96', '#00cc96']`
Could set the secondary color as 0.5 x opacity of the primary color, like area chart fills?

> interpolating between existing plotly colors when we go through all 10

I think if given an array, we should cycle instead of interpolate. This is what `table` does and I don't see much use for sequential or diverging color scale rows.

Maybe try this idea out and see if you can replicate something like:

image

Kully:
> Maybe try this idea out and see if you can replicate something like:

I like this whole idea. It covers all the bases, including first 48 rows grey and last one red

Kully:
> 1D lists longer than 2 elements are the primary trace color for each row, so this could be a colorscale

My thoughts on this now are that users will more often than not, not want to specify the second color and will be okay with a light color as half the opacity of the main. Why don't we:
-allow users to input lists like:
```
trace_colors =[['#e763fa', '#ab63fa'], 'rgb(240, 10, 14)', ['#636efa', '#EF553B'], '#636efa']
```
and we
- interpret each nested single color as the primary color
and
-treat the nested lists as the obvious darklight color combo

Kully:
@jackparmer new notebook of charts/features: https://plot.ly/~AdamKulidjian/6215

Kully:
@chriddyp ready for a review (when tests pass)

plotly/figure_factory/_sparkline.py
```diff
@@ -0,0 +1,457 @@
...
+        `help(plotly.tools.make_subplots)` for more info on this subplot param
+    :param (bool) show_titles: determines if title of chart type is displayed
+        above their respective column
+    :param (str) text_align: aligns label and avg cells. Use either 'center',
+        'left', or 'right'. Default='center'.
+    :param (float) horizontal_spacing: Space between subplot columns.
+        Applied to all columns
+    :param (float) vertical_spacing: Space between subplot rows.
+        Applied to all rows
+    :param (float) alternate_row_color: set to True to enable the alternate
```
cldougl: Is this a float or a bool? can the default (True) be included here too?

plotly/figure_factory/_sparkline.py
```diff
@@ -0,0 +1,457 @@
...
+        row coloring of the chart. Uses the trace_colors from param 'row_colors'
+    :param (list) row_colors: a list/tuple of colors that are used to
```
cldougl: maybe include the default here too

Kully: I'll include more of the defaults

chriddyp:
@Kully - Looks really nice!

It looks like each chart in each "row" represents the same data:
image
Is there ever the case that the charts in the same row have different data? For example:
- row 1: `toyata`
- row 2: `tesla`
- row 3: `ford`
- column 1: `power`
- column 2: `torque`
- column 3: `mpg`

where `power` is represented as a bullet, torque is represented as a bar, and mpg is represented as a line?
?

Kully:
> Is there ever the case that the charts in the same row have different data?

No there is not.

chriddyp:
> >    Is there ever the case that the charts in the same row have different data?

> No there is not.

I mean, should there be? In the examples or datasets that you've seen? I suspect that supporting data on a per-cell level, rather than a per-row level, will require a slightly different API design that we would want to do before release.

Here's the tidy format of the data (see http://www.jeannicholashould.com/tidy-data-in-python.html)

| car    | metric | value | sample  |
|--------|--------|-------| -------|
| toyota | mpg    | 30    | a |
| toyota | mpg    | 34    | b |
| toyota | mpg    | 32    | c |
| toyota | power  | 12    | a | 
| toyota | power  | 11    | b |
| toyota | power  | 9    | c |
| toyota | torque | 3     | a |
| toyota | torque | 34     | b |
| toyota | torque | 32     | c |
| tesla  |mpg   | 4       | a | 
| tesla  |mpg   | 42       | b | 
| tesla  |mpg   | 41       | c | 
| tesla  |power   | 3      | a | 
| tesla  |power   | 32      | b | 
| tesla  |power   | 31      | c | 
| tesla  |torque   | 50     | a |
| tesla  |torque   | 53     | b |
| tesla  |torque   | 55     | c |
| ford  |mpg   | 11      | a | 
| ford  |mpg   | 13      | b | 
| ford  |mpg   | 15      | c | 
| ford  |power   | 90      | a |
| ford  |power   | 98      | b |
| ford  |power   | 99      | c |
| ford  |torque   | 82      | a |
| ford  |torque   | 81      | b |
| ford  |torque   | 85     | c |

(in this case, samples is the same across all rows, but I believe that they could be unique in each column. By making them the same, the user can easily share x-axes. If they were different, there would need to be a unique x-axis for each row)

by assuming the data is in that format, the API could be something like this:
```
FF.spark_lines(
   df, 
   x='sample',
   y='value',
   rows='car',
   columns='metric',
   column_chart_types={'mpg': 'histogram', 'power': 'line', 'torque': 'bullet'})
```

does that make sense?

Kully:
> I mean, should there be? In the examples or datasets that you've seen? I suspect that supporting data on a per-cell level, rather than a per-row level, will require a slightly different API design that we would want to do before release.

I was under the impression this figurefactory was going to be very basic to start with. And which dataset are you getting the tidy dafaframe above from? It's not [this one](https://raw.githubusercontent.com/plotly/datasets/master/mpg.csv).

chriddyp:
> And which dataset are you getting the tidy dafaframe above from?

I just created the dataset manually as an example

chriddyp:
> > And which dataset are you getting the tidy dafaframe above from?

> I just created the dataset manually as an example

But by organizing the data in this "tidy" format, it allows one to make an api where the variables (`x`, `y`, `rows`, `columns`) are column names instead of implicitly looping through the user's columns in their dataframe.

Kully:
> But by organizing the data in this "tidy" format, it allows one to make an api where the variables (x, y, rows, columns) are column names instead of implicitly looping through the user's columns in their dataframe.

So regarding my next steps to finish this FigureFactory, should we change a users' `df` in the figure  factory if they select a 'tidy' option, _or_ should we require that users' submit 'tidy' data sets? WDYT?

Kully:
> does that make sense?
@chriddyp 
Yes it does!

Been working on this for the week, and am nearing completion of tidy-data filtering. (don't know the term for this)

Before I continue, I want to run this syntax for chart_types by you:
Instead of a dict
{'mpg': 'histogram', 'power': 'line', 'torque': 'bullet'})
like you had, I am thinking a tuple/list
`chart_types = ({'4': 'area'}, {'f': 'line'}, {'r': 'bar'})`
so that we preserve a desired order of the columns the user specifies.

chriddyp:
> Secondly, I want to keep this figurefactory simple enough, so I suggest going with only x, not x and y

As per our slack conversation:
> i’d be interested in seeing if implementing some type of way to specify chart types by column inside `create_facet_grid` would allow us to recreate all of the sparklines plots, or if there are any other features that are missing. it’s nice to have some convergence in these features and APIs so that our users don’t have to learn a separate syntax and format their data in a different way for the different figure factories

I feel like sparklines should just be an extension of `create_facet_grid`. From slack:
> so then, it seems like it begs the question - what’s different about `spark_lines` and `create_facet_grid`? 
> - One thing is that the sparklines plots seem to support different chart types for different columns.
> - They also seem to have a “tighter”, smaller layout

In other words, `create_facet_grid` with a new `theme` that is called "sparklines" or "tight" (`FF.create_facet_grid(..., theme='sparklines')`), support for a couple more chart types, and a new key for specifying `chart_types`. `create_facet_grid` already supports `x` and `y`, among many other great features, so it seems like we get a lot for free by extending this existing figure factory (not to mention it will be easier for maintenance, documentation, and for our users).

***

For the `chart_types` key, it seems like we have a couple options:
1 - Map unique values from the `facet_col` grouping to trace types
2 - Allow users to pass in an arbitrary column, like the rest of the attributes.

Earlier I suggested (1) as option. Thinking about it again, it seems like (2) is the more generic solution, as you could achieve (1) with (2) by just passing in the same column to `chart_types`. For example:
```
ff.create_facet_grid(
    df,
    x='displ',
    y='cty',
    facet_col='drv',
    trace_type='drv',
    trace_type_attributes=['line', 'histogram', 'bar']
)
```
This would even allow users to map types to different rows if they wanted:
```
ff.create_facet_grid(
    df,
    x='displ',
    y='cty',
    facet_row='cyl',
    trace_type='cyl',
    trace_type_attributes=['line', 'histogram', 'bar']
)
```
or a different variable all together:
```
ff.create_facet_grid(
    df,
    x='displ',
    y='cty',
    trace_type='cyl',
    trace_type_attributes=['line', 'histogram', 'bar']
)
```

In this case, `trace_type_attributes` would be something that would be merged into each set of traces created by the `trace_type` mapping. To start, we could accept just a list and merge it into the `type` attribute, but eventually I could see this being extended to:
- a full set of trace attributes that are merged into the trace, like:
```
ff.create_facet_grid(
    df,
    x='displ',
    y='cty',
    facet_col='drv',
    facet_row='cyl',
    trace_type='cyl',
    trace_type_attributes=[{'type': 'scatter', 'mode': 'lines', 'marker': {'color': 'blue'}, 'name': 'Tesla'}, {'type': 'histogram', 'name': 'Ford'}, {'type': 'box', 'name': 'Honda'}]
)
```

- Or, like how we've done with `colormap`, a mapping between unique values in the `trace_type` mapping to attributes:
```
ff.create_facet_grid(
    df,
    x='displ',
    y='cty',
    facet_col='drv',
    facet_row='cyl',
    trace_type='cyl',
    trace_type_attributes={"Honda": "histogram", "Tesla": "bar"}
)
```
and/or
```
ff.create_facet_grid(
    df,
    x='displ',
    y='cty',
    facet_col='drv',
    facet_row='cyl',
    trace_type='cyl',
    trace_type_attributes={"Honda": {'type': 'histogram', 'marker': {'color': 'blue'}}, "Tesla": {'type': 'box', 'showpoints': 'all'}}
)
```

***

Good point w.r.t order. It seems like this affects the `facet_row` and `facet_col` attributes as well. Some options:
1 - Require the user to re-order their dataframe in the order that they want things plotted. Something like this: https://stackoverflow.com/a/27255567/4142536. This would require no work from us on first pass
2 - Supply a new key like `trace_type_order`. Similarly `facet_col_order` and `facet_row_order`.
3 - Supply order in the `trace_type_attributes` somehow like
```
trace_type_attributes=[{'order': 0, 'value': 'Ford', 'attributes': {'type': 'histogram'}, {'order': 1, 'value': 'Tesla' 'attributes': {'type': 'scatter', 'mode': 'markers', 'marker': {'color': 'blue'}}
```

I think that I would prefer (1) or (2) to start

Kully:
@chriddyp so I like most of your suggestions, but I found something that may pose a problem.

The entire code in `facet_grid` is partitioned in terms of what kind of color you are using: colorscale name (ex `Viridis`), or the column name for color. I was using a new syntax for color in the `_sparkline` FF I was working on, something like:

```
colors =[['#444', '#ccc'], ['#444', '#ccc'], ['#ff0000', '#ff0000',]]
```

I really want to keep this method for coloring (first column gets first list: `[lightcolor, darkcolor]`) What do you think?

Kully:
There's also the issue of the simple approach to sparklines, which is just giving the API a dataframe, and then just mapping each column to a row and entering charts based on a `chart_types` variable. The current structure of facet grids doesn't allow this to be naturally added in, and it's gonna take a lot of extra rewritting to make that work.

For these two reasons, I think it would make more sense to keep sparklines as its own FigureFactory, but perhaps to also change the variable names in sparklines to match those in facet_grid.

But you do make good points about the familiarity of the already existing FF and building on top of it.

Any thoughts?

Kully:
Current Tasks:
- [ ] replace `_annotation_dict()` in `_facet_grid.py` with `annotation_dict_for_label` in `ff.utils.py`
- [ ] `theme='sparklines'` that accurately replicate the old sparklines
- [ ] all the old trace types from sparklines `['line', 'bullet', 'label', 'avg', 'area']`
- [ ] add ability to use non-tidy data in facet_grid
- [ ]

jackparmer:
> How about the max value is the vertical marker, the thin black bar is the mean and we only have one thick bar which is the standard dev?

Hm... toughy. What about leaving it to the user to define what these things mean? They'll have to calculate mean, max, etc on their own, but this would provide the most flexibility and forward compatibility.

Kully:
> Hm... toughy. What about leaving it to the user to define what these things mean?

While it's more robust, I don't like what this will mean for users who just want a quick and dirty bullet chart. What would the syntax for that look like?

jackparmer:
Yeah, I can see that. Let's go with your proposal - I think it's totally reasonable. 👍

jonmmease:
Hi @Kully , I haven't read through this PR thread yet, but is this something you're planning to keep working on?

JeroneGagliano:
Do you have a timeline on when this PR will be merged with the master branch?
Thanks!

JeroneGagliano:
@chriddyp and @jonmmease 
I have pulled Kully's latest commit 8d665ac imported it and had success running his the first few examples after fixing a couple of errors in the sparklines.py file.  As an engineer, and not a real programmer, I am not comfortable finishing sparklines-ff to merge into master. I would be happy to participate in testing and passing along the code errors I found.

nicolaskruchten:
I think this kind of thing is easier to make now with `make_subplots` and `indicator` and `px` and Dash, so I'm going to close out this PR for now.

firmai:
Hi @nicolaskruchten , In finance (currently) at least, you can't build a state of the art dashboard without sparklines see e.g, yahoo. https://finance.yahoo.com/world-indices/

At which point would we be able to build something similar using make_subplots and indicator and px and Dash? Is this currently a wip?

nicolaskruchten:
These kinds of tables are quite common, but IMO a data visualization library such as `plotly` is a poor fit for generating the whole thing. [Dash](https://dash.plotly.com/) is what I would recommend for doing this kind of thing in Python, specifically the  [Dash DataTable](https://dash.plotly.com/datatable). Sparklines are currently supported there via this approach: https://community.plotly.com/t/sparklines-as-fonts-embedding-minimal-sparklines-in-tables-components/39468

firmai:
Thanks, I tried dash, it unfortunately doesn't work, I get about 20% there after 3 days spent. Hopefully at some point a company could sponsor this so that it can be open sourced. @Kully was approaching something interesting, a templatable framework that could be used by ordinary users.

PR #894, opened 2017-12-11
Update _dendrogram.py

444thLiao:
demo

At previous version of dendrogram script, it use easy way to detect a node whether a leaves. But in our daily work, we will meet situation like this graph show. It will make the branch become all 0. Including the parent branches. So it can't use this easy way.

Need to be cautious: It is not the leaves 'U-shape' is all zeros. It also have some parent branches 'U- shape' is all zeros. (zeros I mean is there yvalus)

But I don't have much time to fix it. So I just add a wrapper to fix this. I have checked scipy doc. It is a safe way to build a new tickvals. So I build it in this way.


jonmmease:
Done in https://github.com/plotly/plotly.py/pull/1180

Issue #711, opened 2017-03-13
Box size % to Dashboards API


Kully:
Be able to manually control box percentages and change the size of the dashboard easily

Kully:
Additionally, solve the problem of stacking several plots on top of each other.

Currently with 50%, stacking 8 plots on top of each returns a layout:

![screen shot 2017-03-28 at 4 08 44 pm](https://cloud.githubusercontent.com/assets/10369095/24428256/d4f5edce-13db-11e7-9e93-a9e631ff2a85.png)

alexcthompson:
+1.  I'm trying to create a dashboard programmatically, and I'm ending up reverse engineering the way the splits and sizes are created... tricky.

Kully:
> +1. I'm trying to create a dashboard programmatically, and I'm ending up reverse engineering the way the splits and sizes are created... tricky.

Agreed. It's definitely something I don't think users should need to learn about. And one of the major weak points in the Dashboard API.

iveney:
Is this issue not being fixed?

Kully:
> Is this issue not being fixed?

Thanks for the reminder @iveney . I have other projects that I am working on at the moment so I will be fixing this issue after those.

Kully:
Closing as issue is now fixed and will appear in the next version of plotly.

Kully:
@iveney @alexcthompson 
In case you did not know, this issue is now fixed and dashboards now have an `.insert` feature. See the [CHANGELOG](https://github.com/plotly/plotly.py/blob/master/CHANGELOG.md) for more details on what else is added in the newest version of plotly.
[please +1 this message to confirm I know you received the message]

alexcthompson:
+1 - glad to hear it, good work.

PR #946, opened 2018-02-23
added fill_percent to params for insert, swap and remove

Kully:
This PR accomplishes:
- [X] added fill_percent to params for insert, swap and remove
- [x] fix HTML preview to reflect `fill_percent`
- [x] update changelog
- [x] remove `#TODO`s from `dashboard_objs.py`  

re: https://github.com/plotly/plotly.py/issues/711


Kully:
@cldougl ready for a 👀

cldougl:
hey @Kully this looks good, I'm trying out some examples now! Are these changes intended to be backwards compatible?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2.4.2] - UNRELEASED
```
cldougl: I would say this is more of added functionality than a bug fix (do you agree?) so I think it should be version 2.5.0 based on: https://semver.org/

Kully:
> hey @Kully this looks good, I'm trying out some examples now! Are these changes intended to be backwards compatible?

As far as I know, yes. `fill_percent` is defaulted to 50, so it behaves the same way before the changes.

There are a couple technically non-backwards compatible things:
- The main container in an init dashboard `dashboard.Dashboard()` that holds all the boxes has size in `%`, not `px`. It's a small technical thing.
- the default height and width of the HTML preview are bigger.

CHANGELOG.md
Kully: I agree they are not bug fixes. The question I have now about my message below is if we consider these points as backwards-compatible. Functionally the API is the same and produces dashboards that look the same. I just had to modify a test or two.

plotly/dashboard_objs/dashboard_objs.py
```diff
@@ -440,7 +434,7 @@ def insert(self, box, side='above', box_id=None):
         my_dboard.insert(box_1, 'left', 1)
         my_dboard.insert(box_1, 'below', 2)
         my_dboard.insert(box_1, 'right', 3)
-        my_dboard.insert(box_1, 'above', 4)
+        my_dboard.insert(box_1, 'above', 4, fill_percent=30)
```
cldougl: screen shot 2018-02-28 at 11 29 07 am

does this output for the example look correct to you @Kully ? Not sure `right/left` or `above/below` is working

CHANGELOG.md
cldougl: Is there no way to add this functionality as an option rather than the default so that it remains backwards compatible?

plotly/dashboard_objs/dashboard_objs.py
Kully: Running the exact same code in my branch, I am getting this preview:
screen shot 2018-02-28 at 3 22 48 pm

What could be different about our branches? (you are cloning from the latest commit I assume)

cldougl: @Kully running on your branch in python 2.7 I get the desired output but in python 3.6 I get: 
image
We should make this consistent across python versions.

CHANGELOG.md
Kully: I don't think it's worth it. There was a technical issue with, before, using pixels as the metric for slicing up the boxes (since it was _always_ in half) but the new param uses percentage to insert the boxes next to a pre-existing box

plotly/dashboard_objs/dashboard_objs.py
Kully: wow, that's interesting -
I'll investigate that (and probably need to add better tests)

CHANGELOG.md
cldougl: ok got it 👍

Kully:
@cldougl Now the API auto-sizes very similarly to the dashboard GUI. Ready for I believe a final look.

CHANGELOG.md
cldougl: @Kully did you get a chance to update this to `2.5.0` and update `plotly/version.py` as well?

Kully: no thank you for reminding me

cldougl:
@Kully for python 2.7 and 3.6 I am getting this preview from the example:

screen shot 2018-03-05 at 11 12 14 am

Can you confirm that is correct? That is not what I was expecting from 

```
my_dboard = dashboard.Dashboard()
my_dboard.insert(box_1)
my_dboard.insert(box_1, 'left', 1)
my_dboard.insert(box_1, 'below', 2)
my_dboard.insert(box_1, 'right', 3)
my_dboard.insert(box_1, 'above', 4, fill_percent=30)
```

cldougl:
@kully can you also add an example of the easiest way to add 4 vertical boxes of the same height? 

screen shot 2018-03-05 at 11 20 24 am

Kully:
> @Kully can you also add an example of the easiest way to add 4 vertical boxes of the same height?

```
# 4 equally stacked vectical boxes
import plotly.dashboard_objs as dashboard

box_1 = {
    'type': 'box',
    'boxType': 'plot',
    'fileId': 'username:some#',
    'title': 'box 1'
}

my_dboard = dashboard.Dashboard()
my_dboard.insert(box_1)
my_dboard.insert(box_1, 'below', 1)
my_dboard.insert(box_1, 'below', 1)
my_dboard.insert(box_1, 'below', 3)
```

screen shot 2018-03-05 at 8 36 06 am

Kully:
> Can you confirm that is correct? That is not what I was expecting from

Yes this is correct.

This is perhaps the weakest point in the API currently. `box_id` numbers (the numbers that appear in the HTML preview and correspond to a box in the layout) are generated on-the-fly. This just means that you need to look at where the numbers  so the original `1` marker

Kully:
I am going to edit `plotly/dashboard_objs/dashboard_objs.py` now so that `box_1` is rewritten as `box_a` to avoid confusion. You suggested a similar variable-renaming to the docs a while ago.

cldougl:
💃

PR #954, opened 2018-03-03
Basic stream functionality

aidan-plenert-macdonald:
Working towards basic functionality for offline streams. I tried to base the functionality off of the existing streams, so the API's are similar in Python.

```python
import numpy as np
import plotly.offline as py
from IPython.display import display, HTML, Javascript, display_javascript

py.init_notebook_mode(connected=True)

stream_token = 's0'
py.iplot(
    [
        {
            'x': [0, 1],
            'y': [1, 0],
            'type': 'scatter', 
            'mode': 'lines+markers',
            'marker': {
                'size': 12
            },
            'stream': {
                'token': stream_token, 
            }
        }
    ],
    filename='basic-stream'
)

s = py.Stream(stream_token)
s.open()
s.write({'x': [5.0, 9.0], 'y': [6.0, 7.0]})
s.write({'x': list(np.arange(10, 15)), 'y': list(np.random.randn(5))})
```

This code works as expected, but I was hoping for guidance on what one might like to see coming from an offline stream API. Like a set of unit test, or design guidance.

Are there a set of unit tests or something I should try to get the offline streams to pass?

### DIFF
--- a/plotly/offline/__init__.py
+++ b/plotly/offline/__init__.py
@@ -10,5 +10,6 @@ from . offline import (
     iplot,
     iplot_mpl,
     plot,
-    plot_mpl
+    plot_mpl,
+    Stream
 )
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -351,6 +351,9 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
         plot_html, plotdivid, width, height = _plot_html(
             figure_or_data, config, validate, '100%', 525, True
         )
+        for i, ele in enumerate(data):
+            if 'stream' in ele:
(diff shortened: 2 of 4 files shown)


cldougl:
Thanks @aidan-plenert-macdonald !
To give you an idea, here are the current online streaming tests: https://github.com/plotly/plotly.py/blob/master/plotly/tests/test_core/test_stream/test_stream.py

nicolaskruchten:
Is this PR still something that you're interested in looking at? We've changed a fair bit under the hood since it was opened, so I'm thinking we should probably close it and start again...

aidan-plenert-macdonald:
@nicolaskruchten Hey, I don't really remember writing this, so I'll just close it. It would be a good feature. I think I just used Bokeh Streaming instead https://bokeh.pydata.org/en/latest/docs/user_guide/data.html#streaming

PR #958, opened 2018-03-06
Don't crash if figure_factory is imported but pandas isn't installed

### LINKED ISSUE #941: Not all dependecies are listed
Would it be possible to add pandas (and numpy) to the list of dependencies? They're both imported in various places but not listed. This occasionally causes problems for packages importing plotly, since `pip` may not install all dependencies.

Xref: https://github.com/deeptools/deepTools/issues/669

dpryan79:
This fixed #941 and makes it possible to `import plotly.figure_factory` without having `pandas` installed. The general framework for this seems to be used throughout the figure factory code, so presumably the lack of it in this case was a simple oversight.

### DIFF
--- a/plotly/figure_factory/__init__.py
+++ b/plotly/figure_factory/__init__.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import
 
+from plotly import optional_imports
+
 # Require that numpy exists for figure_factory
 import numpy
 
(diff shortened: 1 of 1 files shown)


chriddyp:
this looks good to me, thanks! @Kully - can you review too?

Kully:
> this looks good to me, thanks! @Kully - can you review too?

Yes! 👍

Kully:
Looks good to me.  💃 

Note: It looks like your commit here is not fully addressing the original issue that you made. - i.e. Adding Pandas to the Required Dependancies - but for fixing the crash upon import I'm all for.

PR #970, opened 2018-03-11
make iplot include resize listener

krakentoa:
'offline.plot' makes plotly resize a plot when the corresponding div is resized. 'offline.iplot' doesn't.
The script is not included if both width and height properties in the layout are set to anything other than "100%"
This pull request makes 'offline.iplot' include the same resizing script under the same conditions that 'offline.plot' does.


jonmmease:
This functionality is being added as part of PR #1234. Thanks for the PR and for bringing this up @krakentoa!

Issue #283, opened 2015-09-03
Remove `graph_objs` dependency from `tools.py`


theengineear:
We'll need to keep some imports protected for now, but perhaps we can deprecate these and eventually remove?

There are a couple validation functions in there that we need to be careful with related to backwards compat... likely not a big deal though.

jonmmease:
No longer applicable to version 3

PR #975, opened 2018-03-15
minor test fixes for ipyplotly_integration

Kully:
- fixes minor test failures in https://github.com/plotly/plotly.py/pull/942
cc. @chriddyp @jmmease 

# Minor Test Fixes - Status
NB Tests that are _not_ minor and require changing something in a fundamental way are still checked if all other minor tests in the module _are_ fixed. A description/error message is provided just underneath the bullet point to specify the error message for the unsolved problem
### Core
- [x] test_tools
  - [x] test_make_subplots <br> `1 F w/ test_two_row_bottom_left`
  - [x] test_get_subplots <br> `same issue as test_make_subplots`
  - [X] test_configuration
- [x] test_plotly
  - [x] test_plot <br> `__init__() got an unexpected keyword argument '_raise'`
  - [x] test_credentials
- [x] test_stream
- [x] test_offline
- [x] test_graph_objs
  - [x] test_annotations <br> `'Annotations' object has no attribute 'validate'`
  - [x] test_append_trace <br> `same issue as 'test_make_subplots'`
  - [x] test_data <br> `'Data' object has no attribute 'validate'`
  - [X] test_error_bars
  - [x] test_figure <br> `__init__() missing 1 required positional argument: 'class_strs_map'`
  - [x] test_frames <br> `AttributeError: 'Frames' object has no attribute 'to_string'`
  - [x] test_get_figure <br> `AttributeError: 'Figure' object has no attribute 'get_data'`
  - [X] test_graph_objs
  - [X] test_graph_objs_tools <br> `ImportError: cannot import name 'graph_objs_tools'`
  - [X] test_plotly_base_classes <br> `from plotly.graph_objs.graph_objs import PlotlyDict, PlotlyList`
  - [X] test_strip_style <br> `AttributeError: 'Figure' object has no attribute 'strip_style'`
  - [x] test_to_string <br> `AttributeError: 'Figure' object has no attribute 'to_string'`
  - [x] test_update <br> `AttributeError: 'Data' object has no attribute 'update'`
- [x] test_grid <br> `The 'xsrc' property is a string and must be specified as: - A string`
- [X] test_get_figure <br> `TypeError: __init__() got an unexpected keyword argument '_raise'`
- [x] test_api

### Optional
- [x] test_tools
  - [X] TestQuiver
  - [x] TestFinanceCharts
  - [x] TestAnnotatedHeatmap
  - [x] TestTable
  - [x] TestGantt
  - [x] Test2D_Density <br> `('ncontours', 'colorscale', 'reversescale', 'showscale') not valid in Scatter`
- [X] test_matplotlylib
  - [x] test_annotations <br> `AssertionError: ['line']['width'] = 1.5 should be 1.0` (I think the issue is just locally due to matplotlib ver ==2.2.2; our tests use matplotlib1.3.1)
  - [x] test_axis_scales <br> `from plotly.tests.test_optional.test_matplotlylib.data.axis_scales import *` causes the error <br>
```
ValueError: 
    Invalid value of type 'builtins.bool' received for the 'tickmode' property of layout.xaxis
        Received value: False

    The 'tickmode' property is an enumeration that may be specified as:
      - One of the following enumeration values:
            ['auto', 'linear', 'array']
```

- [x] test_figure_factory
  - [x] TestDistplot
  - [X] TestStreamline
  - [x] TestDendrogram <br>`X = np.array([[1, 2, 3, 4], [1, 1, 3, 4], [1, 2, 1, 4], [1, 2, 3, 1]]) dendro = ff.create_dendrogram(X)` fails
  - [X] TestTrisurf
  - [X] TestScatterPlotMatrix
  - [X] TestGantt
  - [X] TestViolin (deprecated messages pop up)
  - [x] TestFacetGrid
  - [X] TestBullet
  - [x] TestChoropleth (not installing geopandas, pyshp and shapely)

- - -
# Current Issues/Test failures from April 12, 2018:

```
/Users/adamkulidjian/Desktop/Adam/Plotly/plotly.py/plotly/tests/test_optional/test_figure_factory/test_figure_factory.py
AttributeError: 'Scatter' object has no attribute 'pop'
```
- no `.to_dataframe` (test_optional/test_graph_objs)

```
(venv3.6) adamkulidjian@Adams-Air:~/desktop/adam/plotly/plotly.py/plotly/tests/test_core/test_to$ clenosetests test_make_subplots.py
.......................................................F.
FAIL: test_two_row_bottom_left (plotly.tests.test_core.test_tools.test_make_subplots.TestMakeSubplots)
  File "/Users/adamkulidjian/Desktop/Adam/Plotly/plotly.py/plotly/tests/test_core/test_tools/test_make_subplots.py", line 166, in test_two_row_bottom_left
    self.assertEqual(fig.to_plotly_json(), expected.to_plotly_json())
AssertionError: {'dat[156 chars][0.0, 1.0]}, 'yaxis': {'anchor': 'x', 'domain': [0.0, 0.425]}}} != {'dat[156 chars][0.0, 1.0]}, 'yaxis': {'anchor': 'x', 'domain': [0.575, 1.0]}}}
  {'data': [],
   'layout': {'xaxis': {'anchor': 'y', 'domain': [0.0, 1.0]},
              'xaxis2': {'anchor': 'y2', 'domain': [0.0, 1.0]},
-             'yaxis': {'anchor': 'x', 'domain': [0.0, 0.425]},
?                                                    -------

+             'yaxis': {'anchor': 'x', 'domain': [0.575, 1.0]},
?                                                   +++++++

              'yaxis2': {'anchor': 'x2', 'domain': [0.575, 1.0]}}}
-------------------- >> begin captured stdout << ---------------------
This is the format of your plot grid:
[ (2,1) x2,y2 ]
[ (1,1) x1,y1 ]
```
-removing the to_dataframe to_string, strip_styles, get_data, and validate methods?


Kully:
@jmmease @chriddyp 
Jon, I notice that two of your bullet points in `Failing Test Modes` (https://github.com/plotly/plotly.py/pull/942) are more intertwined than they are presented as individual bullet points and I want to address:

- Validation exceptions not raised on compatibility classes. Some of the current graph_objs.py
...and...
- New datatype objects don't have a validate() method

I noticed that your intention is to keep validation for backwards compatibility but have it do nothing, i.e.  https://github.com/plotly/plotly.py/blob/de95849ef2b03efb2a16dd2f6521bfd9ada48317/plotly/graph_objs/graph_objs.py#L116

But with it not doing anything, PlotlyError messages and such are bound to not get raised. i.e this doesn't work:
```
@raises(PlotlyDictKeyError)
def test_validate_error():
    data = Data()
    data.append({'not-a-key': 'anything'})
    data.validate()
```

My current plan:
- [X] Have `Data`, `BaseFigure`, `BasePlotlyType` classes inherit from [`PlotlyBase`](https://github.com/plotly/plotly.py/blob/e6c6a63f90a1851bf8ecef3fbf6f6c4251f1da1f/plotly/graph_objs/graph_objs.py#L40) which gives `.validate()`, `.to_string()` and other methods for free
- [ ] remove/comment out all tests that expect error using `.validate()` and `.to_graph_objs()` (both kept for back compt.)

This plan will change as I understand your changes more.

For now any thoughts? :+1: to proceed with this plan?

Kully:
https://github.com/plotly/plotly.py/blob/de95849ef2b03efb2a16dd2f6521bfd9ada48317/plotly/graph_objs/graph_objs.py#L115:L117

jmmease:
@Kully The lines you referenced in graph_objs.py (the `validate` and `to_graph_objs` methods) weren't changes that I made.  My assumption is that these are compatibility carryovers from some prior version of plotly.py.  Does this sound right @chriddyp ? Could we go ahead and deprecate them for 3.0?

Kully:
@jmmease Right now when you call a `Layout()`, `Figure()` object for examples, you get the location in memory and not a `dict`-`list`-like object like before:

```
>>>Figure()
Output:
<plotly.datatypes.Figure at 0x1136929e8>
```

Is there plans to change this? The closest we have to getting `{'data':[]}` is the method `.to_dict()` or `.to_plotly_json` that you included.

jmmease:
@Kully These are going to stay object classes, not dicts or lists.  The output you're seeing is the default repr for the class. I think we should eventually override the repr to display something more dict-like when these are output to the screen

Kully:
> I think we should eventually override the repr to display something more dict-like when these are output to the screen

In your PR or another one?

I'm asking because if not this PR, I can adjust some of the tests to get them to pass.

jmmease:
You can go ahead and adjust the tests here, I think we'll eventually merge your testing changes into my branch. I'm not doing anything with the tests right now (I'm focusing on review and cleanup of the new code), so your testing changes should merge in easily.

Kully:
Py3.6 error with at least `pd.read_csv` and `pd.read_table`:
```
import pandas as pd
mpg = pd.read_table('https://raw.githubusercontent.com/plotly/datasets/master/mpg_2017.txt')
```

```
SSLError                                  Traceback (most recent call last)
/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/request.py in do_open(self, http_class, req, **http_conn_args)
   1317                 h.request(req.get_method(), req.selector, req.data, headers,
-> 1318                           encode_chunked=req.has_header('Transfer-encoding'))
   1319             except OSError as err: # timeout error

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/http/client.py in request(self, method, url, body, headers, encode_chunked)
   1238         """Send a complete request to the server."""
-> 1239         self._send_request(method, url, body, headers, encode_chunked)
   1240 

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/http/client.py in _send_request(self, method, url, body, headers, encode_chunked)
   1284             body = _encode(body, 'body')
-> 1285         self.endheaders(body, encode_chunked=encode_chunked)
   1286 

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/http/client.py in endheaders(self, message_body, encode_chunked)
   1233             raise CannotSendHeader()
-> 1234         self._send_output(message_body, encode_chunked=encode_chunked)
   1235 

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/http/client.py in _send_output(self, message_body, encode_chunked)
   1025         del self._buffer[:]
-> 1026         self.send(msg)
   1027 

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/http/client.py in send(self, data)
    963             if self.auto_open:
--> 964                 self.connect()
    965             else:

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/http/client.py in connect(self)
   1399             self.sock = self._context.wrap_socket(self.sock,
-> 1400                                                   server_hostname=server_hostname)
   1401             if not self._context.check_hostname and self._check_hostname:

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/ssl.py in wrap_socket(self, sock, server_side, do_handshake_on_connect, suppress_ragged_eofs, server_hostname, session)
    400                          server_hostname=server_hostname,
--> 401                          _context=self, _session=session)
    402 

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/ssl.py in __init__(self, sock, keyfile, certfile, server_side, cert_reqs, ssl_version, ca_certs, do_handshake_on_connect, family, type, proto, fileno, suppress_ragged_eofs, npn_protocols, ciphers, server_hostname, _context, _session)
    807                         raise ValueError("do_handshake_on_connect should not be specified for non-blocking sockets")
--> 808                     self.do_handshake()
    809 

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/ssl.py in do_handshake(self, block)
   1060                 self.settimeout(None)
-> 1061             self._sslobj.do_handshake()
   1062         finally:

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/ssl.py in do_handshake(self)
    682         """Start the SSL/TLS handshake."""
--> 683         self._sslobj.do_handshake()
    684         if self.context.check_hostname:

SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:749)

During handling of the above exception, another exception occurred:

URLError                                  Traceback (most recent call last)
<ipython-input-74-75578e8bdd0b> in <module>()
      1 import pandas as pd
----> 2 mpg = pd.read_table('https://raw.githubusercontent.com/plotly/datasets/master/mpg_2017.txt')

~/venv3.6/lib/python3.6/site-packages/pandas/io/parsers.py in parser_f(filepath_or_buffer, sep, delimiter, header, names, index_col, usecols, squeeze, prefix, mangle_dupe_cols, dtype, engine, converters, true_values, false_values, skipinitialspace, skiprows, nrows, na_values, keep_default_na, na_filter, verbose, skip_blank_lines, parse_dates, infer_datetime_format, keep_date_col, date_parser, dayfirst, iterator, chunksize, compression, thousands, decimal, lineterminator, quotechar, quoting, escapechar, comment, encoding, dialect, tupleize_cols, error_bad_lines, warn_bad_lines, skipfooter, skip_footer, doublequote, delim_whitespace, as_recarray, compact_ints, use_unsigned, low_memory, buffer_lines, memory_map, float_precision)
    707                     skip_blank_lines=skip_blank_lines)
    708 
--> 709         return _read(filepath_or_buffer, kwds)
    710 
    711     parser_f.__name__ = name

~/venv3.6/lib/python3.6/site-packages/pandas/io/parsers.py in _read(filepath_or_buffer, kwds)
    431     compression = _infer_compression(filepath_or_buffer, compression)
    432     filepath_or_buffer, _, compression = get_filepath_or_buffer(
--> 433         filepath_or_buffer, encoding, compression)
    434     kwds['compression'] = compression
    435 

~/venv3.6/lib/python3.6/site-packages/pandas/io/common.py in get_filepath_or_buffer(filepath_or_buffer, encoding, compression)
    188 
    189     if _is_url(filepath_or_buffer):
--> 190         req = _urlopen(filepath_or_buffer)
    191         content_encoding = req.headers.get('Content-Encoding', None)
    192         if content_encoding == 'gzip':

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/request.py in urlopen(url, data, timeout, cafile, capath, cadefault, context)
    221     else:
    222         opener = _opener
--> 223     return opener.open(url, data, timeout)
    224 
    225 def install_opener(opener):

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/request.py in open(self, fullurl, data, timeout)
    524             req = meth(req)
    525 
--> 526         response = self._open(req, data)
    527 
    528         # post-process response

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/request.py in _open(self, req, data)
    542         protocol = req.type
    543         result = self._call_chain(self.handle_open, protocol, protocol +
--> 544                                   '_open', req)
    545         if result:
    546             return result

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/request.py in _call_chain(self, chain, kind, meth_name, *args)
    502         for handler in handlers:
    503             func = getattr(handler, meth_name)
--> 504             result = func(*args)
    505             if result is not None:
    506                 return result

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/request.py in https_open(self, req)
   1359         def https_open(self, req):
   1360             return self.do_open(http.client.HTTPSConnection, req,
-> 1361                 context=self._context, check_hostname=self._check_hostname)
   1362 
   1363         https_request = AbstractHTTPHandler.do_request_

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/request.py in do_open(self, http_class, req, **http_conn_args)
   1318                           encode_chunked=req.has_header('Transfer-encoding'))
   1319             except OSError as err: # timeout error
-> 1320                 raise URLError(err)
   1321             r = h.getresponse()
   1322         except:

URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:749)>
```

Kully:
@chriddyp Could you give my work a tiny 👀 ? I've finished test in `test_optional/test_figure_factory` and am still going through everything

plotly/figure_factory/_dendrogram.py
```diff
@@ -90,14 +90,15 @@ def create_dendrogram(X, orientation="bottom", labels=None,
...
+    return graph_objs.Figure(data=dendrogram.data,
+                             layout=dendrogram.layout)
 
 
 class _Dendrogram(object):
     """Refer to FigureFactory.create_dendrogram() for docstring."""
 
     def __init__(self, X, orientation='bottom', labels=None, colorscale=None,
-                 width="100%", height="100%", xaxis='xaxis', yaxis='yaxis',
+                 width=np.inf, height=np.inf, xaxis='xaxis', yaxis='yaxis',
```
chriddyp: this seems odd to me: infinite width?

plotly/figure_factory/_facet_grid.py
```diff
@@ -108,7 +106,7 @@ def _annotation_dict(text, lane, num_of_lanes, SUBPLOT_SPACING, row_col='col',
         showarrow=False,
         xref='paper',
         yref='paper',
-        text=text,
+        text=str(text),
```
chriddyp: why was this change necessary?

plotly/figure_factory/_facet_grid.py
```diff
@@ -331,10 +329,7 @@ def _facet_grid_color_categorical(df, x, y, facet_row, facet_col, color_name,
                                  row_col='row', flipped=flipped_rows)
             )
 
-    # add annotations
-    fig['layout']['annotations'] = annotations
-
-    return fig
+    return fig, annotations
```
chriddyp: why was this change made? was it for this? https://github.com/plotly/plotly.py/pull/975/files#diff-83b399f13d8b361d38016264ed1ce589R987 If so, why does `fig['layout']['annotations'].append` not work anymore?

plotly/figure_factory/_trisurf.py
```diff
@@ -160,7 +160,7 @@ def trisurf(x, y, z, simplices, show_colorbar, edges_color, scale,
                 color=[min_mean_dists, max_mean_dists],
                 colorscale=colorscale,
                 showscale=True),
-            hoverinfo='None',
+            hoverinfo='none',
```
chriddyp: Why was this change made?
On another note, the fact that this was `'None'` before seems a little odd to me: why not just `None`?

plotly/graph_objs/graph_objs.py
```diff
@@ -0,0 +1 @@
+from plotly.graph_objs import *
```
chriddyp: this change seems odd to me, seems like something that should exist (if necessary) in https://github.com/plotly/plotly.py/pull/942. What's the rational behind this change?

plotly/graph_reference.py
```diff
@@ -39,6 +39,8 @@
     'Histogram2d': {'object_name': 'histogram2d', 'base_type': dict},
     'Histogram2dContour': {'object_name': 'histogram2dcontour',
                            'base_type': dict},
+    'Histogram2dcontour': {'object_name': 'histogram2dcontour',
+                           'base_type': dict},
```
chriddyp: this change seems odd to me, seems like something that should exist (if necessary) in #942. What's the rational behind this change?

plotly/tests/test_core/test_graph_objs/test_annotations.py
```diff
@@ -25,33 +24,28 @@ def test_trivial():
...
 
 
 def test_dict_instantiation():
     Annotations([{'text': 'annotation text'}])
 
 
-@raises(PlotlyDictKeyError)
 def test_dict_instantiation_key_error():
-    print(Annotations([{'not-a-key': 'anything'}]))
+    assert Annotations([{'not-a-key': 'anything'}]) == [{'not-a-key': 'anything'}]
```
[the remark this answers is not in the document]
chriddyp: OK I see, the new way to do this would be to just use a regular list of `go.layout.Annotation`:
```
[go.layout.Annotation(**{'not-a-key': 'anything'})]
```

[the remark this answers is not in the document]
chriddyp: So, we should probably update this test name as well (we can do that in the end, once we've gone through the rest of the tests)

plotly/tests/test_core/test_graph_objs/test_data.py
```diff
@@ -25,42 +25,42 @@ def test_trivial():
     assert Data() == list()
 
 
-@raises(PlotlyError)
+#@raises(PlotlyError)
 def test_weird_instantiation():  # Python allows this...
-    print(Data({}))
+    assert Data({}) == []
```
chriddyp: 🚨 - Breaking Change

plotly/tests/test_core/test_graph_objs/test_data.py
```diff
@@ -25,42 +25,42 @@ def test_trivial():
...
 
 
 def test_default_scatter():
-    assert Data([{}]) == list([{'type': 'scatter'}])
+    assert Data([{}]) == list([{}])
```
chriddyp: 🚨 - Breaking Change

plotly/tests/test_core/test_graph_objs/test_data.py
```diff
@@ -25,42 +25,42 @@ def test_trivial():
...
 
 def test_dict_instantiation():
     Data([{'type': 'scatter'}])
 
 
-@raises(PlotlyDictKeyError)
+# @raises(PlotlyDictKeyError)
 def test_dict_instantiation_key_error():
-    print(Data([{'not-a-key': 'anything'}]))
+    assert Data([{'not-a-key': 'anything'}]) == [{'not-a-key': 'anything'}]
```
chriddyp: 🚨 - Breaking Change

plotly/tests/test_core/test_graph_objs/test_error_bars.py
```diff
@@ -41,6 +41,5 @@ def test_instantiate_error_y():
            width=5)
 
 
-@raises(PlotlyDictKeyError)
 def test_key_error():
-    ErrorX(value=0.1, typ='percent', color='red')
+    assert ErrorX(value=0.1, typ='percent', color='red') == {'color': 'red', 'typ': 'percent', 'value': 0.1}
```
chriddyp: So, if we're deprecating `go.ErrorX` in favor of `go.scatter.ErrorX`, then we should probably have two tests here:
- One for the deprecated behaviour (errors are no longer thrown)
- Another for the new syntax, which should basically be the same test as before:
```
@raises(PlotlyDictKeyError)
def test_key_error():
    go.scatter.ErrorX(value=0.1, typ='percent', color='red')
```

And we should verify that the error that is raised remains `PlotlyDictKeyError`

chriddyp: 🚨 - Breaking Change

plotly/tests/test_core/test_graph_objs/test_figure.py
```diff
@@ -16,7 +16,7 @@ def test_instantiation(self):
             'frames': []
         }
 
-        Figure(native_figure)
+        Figure(**native_figure)
```
chriddyp: 🚨 - Breaking Change

plotly/tests/test_core/test_graph_objs/test_frames.py
```diff
@@ -56,20 +58,37 @@ def test_deeply_nested_layout_attributes(self):
...
-        frames[0].data = [Bar()]
-        frames[0].data[0].marker.color = 'red'
+        #frames = Frames()
+        #frames.append({})
+        #frames[0].data = [Bar()]
+        #frames[0].data[0].marker.color = 'red'
+
+        frames = Frame
+        frames.data = [Bar()]
+        frames.data[0].marker.color = 'red'
```
chriddyp: This doesn't seem right to me. The new syntax is replacing this with a list of `Frame` objects,
```
help(go.Frames)
        plotly.graph_objs.Frames is deprecated.
Please replace it with a list or tuple of instances of the following types
  - plotly.graph_objs.Frame
```
So, this test should be:
```
frames = []
frames.append(go.Frame())
frames[0].data = [go.Bar()]
frames[0].data[0].marker.color = 'red'
```

chriddyp: Although `go.Frame()` doesn't seem to work:
```
In [11]: go.Frame()
TypeError                                 Traceback (most recent call last)
<ipython-input-11-d2b9d2e3e8b5> in <module>()
----> 1 go.Frame()

~/Repos/plotlypy-jmease/plotly.py/plotly/graph_objs/_frame.py in __init__(self, baseframe, data, group, layout, name, traces, **kwargs)
    181         # ---------------------
    182         self._validators['baseframe'] = v_frame.BaseframeValidator()
--> 183         self._validators['data'] = v_frame.DataValidator()
    184         self._validators['group'] = v_frame.GroupValidator()
    185         self._validators['layout'] = v_frame.LayoutValidator()

~/Repos/plotlypy-jmease/plotly.py/plotly/validators/frame/_data.py in __init__(self, plotly_name, parent_name)
      6     def __init__(self, plotly_name='data', parent_name='frame'):
      7         super().__init__(
----> 8             plotly_name=plotly_name, parent_name=parent_name, role='object'
      9         )

TypeError: __init__() got an unexpected keyword argument 'role'
```
Am I doing something wrong here? cc @jmmease

plotly/tests/test_core/test_graph_objs/test_frames.py
```diff
@@ -56,20 +58,37 @@ def test_deeply_nested_layout_attributes(self):
...
+        matches = []
+        for r in raw_matches:
+            r = r.replace(' ', '')
+            r = r.replace('\n', '')
+            matches.append(r)
 
         # It's OK if this needs to change, but we should check *something*.
         self.assertEqual(
-            frames[0].data[0].marker.line._get_valid_attributes(),
+            set(matches),
```
chriddyp: Instead of a regex, a better way to do this might be to assert on  
```
all_attributes = dir(frames[0].data[0].marker.line)
public_attributes = [d for d in all_attributes if d[0] != '_' and d[:1] != 'on']
assertEqual(set(public_attributes), {'cliponaxis',
 'connectgaps',
 'customdata',
 'customdatasrc',
 'dx',
 'dy',
 'error_x',
 'error_y',
 'figure',
 'fill',
 'fillcolor',
 'hoverinfo',
 'hoverinfosrc',
 'hoverlabel',
 'hoveron',
 'hovertext',
 'hovertextsrc',
 'ids',
 'idssrc',
 'legendgroup',
 'line',
 'marker',
 'mode',
 'name',
 'opacity',
 'parent',
 'plotly_name',
 'r',
 'rsrc',
 'selected',
 'selectedpoints',
 'showlegend',
 'stream',
 't',
 'text',
 'textfont',
 'textposition',
 'textpositionsrc',
 'textsrc',
 'to_plotly_json',
 'tsrc',
 'type',
 'uid',
 'unselected',
 'update',
 'visible',
 'x',
 'x0',
 'xaxis',
 'xcalendar',
 'xsrc',
 'y',
 'y0',
 'yaxis',
 'ycalendar',
 'ysrc'})
```

plotly/tests/test_core/test_graph_objs/test_frames.py
```diff
@@ -56,20 +58,37 @@ def test_deeply_nested_layout_attributes(self):
...
             {'colorsrc', 'autocolorscale', 'cmin', 'colorscale', 'color',
              'reversescale', 'width', 'cauto', 'widthsrc', 'cmax'}
         )
 
     def test_frame_only_attrs(self):
-        frames = Frames()
+        frames = Frame
```
chriddyp: https://github.com/plotly/plotly.py/pull/975/files#r178922227 applies here as well

plotly/tests/test_core/test_graph_objs/test_scatter.py
```diff
@@ -14,8 +14,7 @@
 
 def test_trivial():
     print(Scatter())
-    assert Scatter() == dict(type='scatter')
-
+    assert Scatter().to_plotly_json() == dict(type='scatter')
```
chriddyp: 👍  seems good to me

plotly/tests/test_core/test_graph_objs/test_scatter.py
```diff
@@ -26,8 +25,8 @@ def test_validate():
...
-@raises(PlotlyError)
-def test_validate_error():
-    scatter = Scatter()
-    scatter['invalid'] = 'something'
-    scatter.validate()
+# @raises(PlotlyError)
+# def test_validate_error():
+#     scatter = Scatter()
+#     scatter['invalid'] = 'something'
+#     scatter.validate()
```
chriddyp: 🚨 Breaking Change (immediate validation)
Let's keep this test but remove the `scatter.validate()` as the new expected behaviour is for `scatter['invalid'] = 'something'` to raise an error.

similarly, let's add a new test that uses the dot syntax to add the property:
```
scatter.invalid = 'something'
```

chriddyp: and we'll want a whole set of new validation tests since https://github.com/plotly/plotly.py/pull/942 is much more strict. So, for each of the possible value types (`enumerated, boolean, number, integer, string, color, colorlist, colorscale, angle, subplotid, flaglist, any, info_array`), we'll want to attempt adding a type that is _not_ the correct one and verify that an exception is thrown. We'll also want to verify that all possible _correct_ values are possible. 

For example, here are examples of exceptions that should be thrown:
```
go.Scatter(x=1)
go.scatter.Marker(color=3)
go.scatter.Marker(color='redz')
go.scatter.Marker(size='a')
```
and here are examples of valid ones we should check:

```
valid_colors = [
    # Name
    'hotpink',
    'transparent',

    # Hexadecimal syntax',
    #f09',
    #F09',
    #ff0099',
    #FF0099',

    # Functional syntax',
    'rgb(255,0,153)',
    'rgb(255, 0, 153)',
    'rgb(100%,0%,60%)',
    'rgb(100%, 0%, 60%)',
    'rgb(255 0 153)',

    # Hexadecimal syntax with alpha value',
    #f09f',
    #F09F',
    #ff0099ff',
    #FF0099FF',

    # Functional syntax with alpha value',
    'rgb(255, 0, 153, 1)',
    'rgb(255, 0, 153, 100%)',

    # Whitespace syntax',
    'rgb(255 0 153 / 1)',
    'rgb(255 0 153 / 100%)',

    # Functional syntax with floats value',
    'rgb(255, 0, 153.6, 1)',
    'rgb(1e2, .5e1, .5e0, +.25e2%)',

    # Hexadecimal syntax
    '#3a30',                    #   0% opaque green 
    '#3A3F',                    # full opaque green 
    '#33aa3300',                #   0% opaque green 
    '#33AA3388',                #  50% opaque green 

    # Functional syntax
    'rgba(51, 170, 51, .7)',    #  70% opaque green 
    'rgba(51, 170, 51,  1)',    # full opaque green 

    # Whitespace syntax
    'rgba(51 170 51 / 0.4)',    #  40% opaque green 
    'rgba(51 170 51 / 40%)',    #  40% opaque green 

    # Functional syntax with floats value
    'rgba(255, 0, 153.6, 1)',
    'rgba(1e2, .5e1, .5e0, +.25e2%)',

    # HSL
    'hsl(30, 10, 10)',
]

for color in valid_colors:
    go.scatter.Marker(color=color)
    go.scatter.Marker(color=[color]*5)

go.scatter.Marker(color=1)
go.scatter.Marker(color=[1]*5)
```

plotly/tests/test_core/test_graph_objs/test_update.py
```diff
@@ -28,7 +28,6 @@ def test_update_dict_empty():
     trace2 = Scatter(x=[1, 2, 3], y=[3, 2, 1])
     data = Data([trace1, trace2])
     data.update({})
-    print(data.to_string())
```
chriddyp: 🚨 - Breaking Change

I didn't know we had a `to_string()` method. I suppose that we're dropping it? It doesn't seem that useful to me, as you should be able to just do `str(data)`

plotly/tests/test_core/test_tools/test_make_subplots.py
```diff
@@ -143,18 +143,10 @@ def test_two_row_bottom_left(self):
         expected = Figure(
             data=Data(),
             layout=Layout(
-                xaxis1=XAxis(
-                    domain=[0.0, 1.0],
-                    anchor='y1'
-                ),
```
chriddyp: Why is this removed?

plotly/tests/test_core/test_tools/test_validate.py
```diff
@@ -8,28 +8,28 @@
...
+                'x': [1, 2, 3],
+                'y': [2, 1, 2]
             }
         ]
     }
     tls.validate(fig, 'Figure')
 
 
-@raises(PlotlyError)
+@raises(ValueError)
```
chriddyp: 🚨 Breaking Change

plotly/tests/test_optional/optional_utils.py
```diff
@@ -29,6 +29,25 @@ def _format_path(self, path):
         str_path = [repr(p) for p in path]
         return '[' + ']['.join(sp for sp in str_path) + ']'
 
+    def assert_fig_equal(self, d1, d2, msg=None, ignore=['uid']):
+        """
+        Helper function for asser_dict_equal
```
chriddyp: typo `assert_dict_equal`

plotly/tests/test_optional/optional_utils.py
```diff
@@ -29,6 +29,25 @@ def _format_path(self, path):
...
+
+        By defualt removes uid from d1 and/or d2 if present
+        then calls assert_dict_equal.
+
+        :param (list|tuple) ignore: sequence of key names as
+            strings that are removed from both d1 and d2 if
+            they exist
+        """
+        for key in ignore:
```
chriddyp: We might want to make a copy (`import copy; copy.deepcopy` i think) before deleting, that way there are no surprises from the callers code. Otherwise, the user has something like:
```
my_scatter = {'x': 1, 'uid': 5}
assert_fig_equal(my_scatter, ...)
print(my_scatter) # is now just {'x': 1}
```

plotly/tests/test_optional/test_figure_factory/test_figure_factory.py
```diff
@@ -54,41 +56,38 @@ def test_simple_distplot_prob_density(self):
...
                               'legend': {'traceorder': 'reversed'},
-                              'xaxis1': {'anchor': 'y2', 'domain': [0.0, 1.0], 'zeroline': False},
-                              'yaxis1': {'anchor': 'free', 'domain': [0.35, 1], 'position': 0.0},
-                              'yaxis2': {'anchor': 'x1',
+                              'yaxis2': {'anchor': 'x',
                                          'domain': [0, 0.25],
                                          'dtick': 1,
                                          'showticklabels': False}}
-        self.assertEqual(dp['layout'], expected_dp_layout)
+        self.assert_dict_equal(dp.to_plotly_json()['layout'], expected_dp_layout)
```
chriddyp: we might want to put `to_plotly_json` inside the `assert_fig_equal` method as well, that way we're not repeating this everywhere

plotly/tests/test_optional/test_figure_factory/test_figure_factory.py
```diff
@@ -54,41 +56,38 @@ def test_simple_distplot_prob_density(self):
...
                                  'histnorm': 'probability density',
                                  'legendgroup': 'distplot',
-                                 'marker': {'color': 'rgb(31, 119, 180)'},
+                                 'marker': {'color': 'rgb(31,119,180)'},
                                  'name': 'distplot',
                                  'opacity': 0.7,
                                  'type': 'histogram',
                                  'x': [1, 2, 2, 3],
-                                 'xaxis': 'x1',
+                                 'xaxis': 'x',
```
chriddyp: 🚨 Breaking Change

plotly/tests/test_optional/test_figure_factory/test_figure_factory.py
```diff
@@ -442,9 +433,9 @@ def test_simple_streamline(self):
             'type': 'scatter',
             'mode': 'lines'
         }
-        self.assertListEqual(strln['data'][0]['y'][0:100],
-                             expected_strln_0_100['y'])
-        self.assertListEqual(strln['data'][0]['x'][0:100],
+        self.assertListEqual(list(strln['data'][0]['y'][0:100]),
```
chriddyp: 🚨 Breaking Change

What is it now? A numpy array?

plotly/tests/test_optional/test_figure_factory/test_figure_factory.py
```diff
@@ -2051,9 +2065,24 @@ def test_valid_colorscale_name(self):
                           color_name='c', colormap=colormap)
 
     def test_valid_facet_grid_fig(self):
-        mpg = pd.read_table('https://raw.githubusercontent.com/plotly/datasets/master/mpg_2017.txt')
-        df = mpg.head(10)
+        mpg = [
```
chriddyp: nice - future proof

plotly/tests/test_optional/test_figure_factory/test_figure_factory.py
```diff
@@ -2062,45 +2091,30 @@ def test_valid_facet_grid_fig(self):
...
-               'y':    [16,
-                   18,
-                   18],
-               'yaxis': 'y1'}],
+               'y': [16, 18, 18],
+               'yaxis': 'y'}],
              'layout': {'annotations': [{'font': {'color': '#0f0f0f', 'size': 13},
                 'showarrow': False,
-                'text': 4,
+                'text': '4',
```
chriddyp: 🚨 Breaking Change - `go.Annotation.text` has to be a string now

Although maybe we should update the plotschema for this one as well, since technically I think it can be a number. We made a proposal for updating the schema here: https://github.com/plotly/plotly.js/issues/2519

plotly/utils.py
```diff
@@ -257,19 +262,21 @@ def encode_as_numpy(obj):
...
-            # we'll get a value error if trying to convert with naive datetime
-            pass
-        except TypeError:
-            # pandas throws a typeerror here instead of a value error, it's OK
-            pass
-        except AttributeError:
-            # we'll get an attribute error if astimezone DNE
-            raise NotEncodable
+        # In PY36, isoformat() converts UTC
+        # datetime.datetime objs to UTC T04:00:00
```
chriddyp: This is odd - why did this need to change?

plotly/figure_factory/_dendrogram.py
jmmease: Context: The string '100%' is forbidden by the number `valType` of `width`.

plotly/figure_factory/_facet_grid.py
jmmease: Context: `layout.annotation.text` has a `string` `valType`. I don't currently autoconvert numbers (or any other Python types) to strings. 

Do we want to support any string coercions?

plotly/figure_factory/_facet_grid.py
jmmease: Context: Object array properties like (e.g. `layout.annotations`, `layout.images`, etc.) are now stored as tuples instead of lists so that they are immutable.
So `fig['layout']['annotations'].append(annot)` won't work anymore, but `fig['layout']['annotations'] += (annot,)` will work.

plotly/figure_factory/_trisurf.py
jmmease: Context: here is the schema for `scatter3d.hoverinfo`
```
            "hoverinfo": {
                    "arrayOk": true,
                    "description": "Determines which trace information appear on hover. If `none` or `skip` are set, no information is displayed upon hovering. But, if `none` is set, click and hover events are still fired.",
                    "dflt": "all",
                    "editType": "calc",
                    "extras": [
                        "all",
                        "none",
                        "skip"
                    ],
                    "flags": [
                        "x",
                        "y",
                        "z",
                        "text",
                        "name"
                    ],
                    "role": "info",
                    "valType": "flaglist"
                },
```

The string `'none'` is one of the valid `extras` states. Currently, I require all `flaglist` values to be a case-sensitive match to those in the schema.  Setting `hoverinfo` to the Python `None` would mean to unset any previously set value and let plotly.js choose the default (`all` in this case).

Should we be more generous and coerce case-insensitive matches to the form in the schema?

plotly/graph_objs/graph_objs.py
jmmease: I actually did add this in #942 as well. It is a bit weird, but some tests (and examples I've seen online) have something like `from plotly.graph_objs import graph_objs` (e.g. https://plot.ly/python/scatterplot-matrix/)

It's an easy enough thing to keep backward compatible, but do we want to keep it around or deprecate it?

plotly/graph_reference.py
jmmease: I think we should be able to delete `graph_reference.py` altogether (but maybe it's still used for testing?). I copied this mapping to `codegen/compatibility.py` to use when generating the deprecated classes.

plotly/tests/test_core/test_graph_objs/test_figure.py
jmmease: FYI: `Figure` can now take in another figure, or a graph representing a figure, as the first arg.

plotly/tests/test_core/test_graph_objs/test_frames.py
jmmease: Nope, I just broke it :-) Fixed in 8a7b476dded4d56d94549c4dae2d007095fcb833

plotly/tests/test_core/test_graph_objs/test_scatter.py
jmmease: I actually wrote a pretty extensive set of tests for the new validators. I hadn't brought them over yet because I wrote them using pytest.  I just added them in commit 45ec0846bf7bdb385135cda600d2367701eaeb20 so you can take a look.

Run them with
```
~> pip install pytest
~> pytest _plotly_utils/tests/
```

So, we can either port them out of pytest, or consider adopting pytest.

plotly/tests/test_core/test_graph_objs/test_update.py
jmmease: One thing that we still need to talk about is what we should do for the `__repr__`, and `__str__` methods on `Figure` and `BasePlotlyType` objects.

plotly/tests/test_core/test_tools/test_validate.py
jmmease: FYI. So far I haven't incorporated any of the legacy plotly-specific exceptions types. If we want to do this we'll need a bit of an inventory of which exceptions we want to keep around for what purposes.

plotly/tests/test_optional/test_tools/test_figure_factory.py
```diff
@@ -248,21 +255,33 @@ def test_one_ohlc(self):
                          'data': [{'y': [33.0, 33.0, 33.2, 32.7,
                                          33.1, 33.1, None],
                                    'line': {'width': 1,
-                                            'color': '#3D9970'},
+                                            'color': '#3d9970'},
```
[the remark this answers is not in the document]
jmmease: Context: At the moment I'm normalizing color string during validation: This involves changing them to lowercase and removing spaces (e.g. `'rgb(0, 0, 255)'` -> `'rgb(0,0,255)'`). The uppercase versions, like `'#3D9970'` are accepted as input.

I could undo this normalization if we want.

plotly/tests/test_optional/test_tools/test_figure_factory.py
```diff
@@ -798,64 +834,70 @@ def test_annotated_heatmap_kwargs(self):
                                                                    [1,
                                                                     '#e6005a']]
                                                        )
-        expected_a = {'data': [{'colorscale': [[0, '#ffffff'], [1, '#e6005a']],
+        expected_a = {'data': [{'colorscale': ((0, '#ffffff'), (1, '#e6005a')),
```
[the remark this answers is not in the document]
jmmease: `colorscales` are still accepted as nested lists, but they are coerced into nested tuples during validation.  This is so that they are immutable.  If I left them as lists, people could do things like
```
fig.data[0].marker.colorscale[0][1] = 'green'
```
and then I would know to update the frontend.

plotly/figure_factory/_facet_grid.py
Kully: Thanks Jon for jumping in on Chris' comments. You definitely know the code changes better than I do.

@chriddyp I reworked facet_grid to accommodate annotations being tuples and not lists

plotly/tests/test_optional/test_utils/test_utils.py
```diff
@@ -96,20 +96,23 @@ def test_encode_as_numpy(self):
         res = utils.PlotlyJSONEncoder.encode_as_numpy(np.ma.core.masked)
         self.assertTrue(math.isnan(res))
 
-    def test_encode_as_datetime(self):
+    def test_encode_valid_datetime(self):
 
         # should *fail* without 'utcoffset' and 'isoformat' and '__sub__' attrs
-        non_datetimes = [datetime.date(2013, 10, 1), 'noon', 56, '00:00:00']
+        #non_datetimes = [datetime.date(2013, 10, 1), 'noon', 56, '00:00:00']
+        non_datetimes = [datetime.date(2013, 10, 1)]
```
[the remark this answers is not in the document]
Kully: So Python3.6 handles datetimes differently than the other Python versions - and since we were not testing Py36 before it's resulting in a failing test now

plotly/utils.py
Kully: see above comment

plotly/figure_factory/_dendrogram.py
chriddyp: Got it. I wonder what `width="100%"` was doing before. I don't think that it doesn't anything in plotly.js as `width='100%'` and `height='100%'` should correspond to `autosize=True`.

Maybe the better setting would be to:
- set the arguments as `width=None` and `height=None` 
- if `width` or `height` is not `None`, apply it in the figure
- set `autosize=True`

plotly/figure_factory/_facet_grid.py
chriddyp: This seems like it should be an addition to the plotly.js schema. I'll make an issue

chriddyp: Tracking in https://github.com/plotly/plotly.js/issues/2523

plotly/figure_factory/_facet_grid.py
chriddyp: I see. So if I do 
```
fig = go.Figure(layout=go.Layout(annotations=[go.Annotation(text='...')]))
```
then `annotations` will get cast as a `tuple`.

Marking this one as a 🚨 - Breaking Change

plotly/figure_factory/_trisurf.py
chriddyp: Ah I see. I wonder if plotly.js accepted `None` before. I don't think that we should coerce the case-insensitive matches, this seems good to me.

Nice to see that this new validation is improving our own code 😉

plotly/graph_objs/graph_objs.py
chriddyp: Yeah, that's odd. It seems like the more semantic way to write this would be 
```
import plotly.graph_objs as graph_objs
```
or, the way we've been doing it all along
```
```
In those examples (https://plot.ly/python/scatterplot-matrix/) `graph_objs` isn't even used.

OK, well let's keep it for backwards compatibility but I'd be OK with deprecating it.

plotly/graph_reference.py
chriddyp: OK yeah, this seems like something we can delete. I forgot all about `graph_reference` - this is from the old days when we didn't have a plotly.s plot-schema!

@Kully - Can you chime in here? Do we still need `graph_reference.py`?

plotly/tests/test_core/test_graph_objs/test_figure.py
chriddyp: OK sounds good, let's revert back to the previous test then (cc @Kully )

plotly/tests/test_core/test_graph_objs/test_scatter.py
chriddyp: Ah fantastic! I'm OK with adopting pytest, even if that means that some of our tests are in `pytest` and others are in nosetest. This is where the tests are invoked on CI, so let's add the `pytest` commands to this file here: https://github.com/plotly/plotly.py/blob/master/tox.ini#L74

plotly/tests/test_optional/test_figure_factory/test_figure_factory.py
chriddyp: https://github.com/plotly/plotly.js/issues/2523

plotly/tests/test_optional/test_tools/test_figure_factory.py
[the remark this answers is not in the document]
chriddyp: Got it. Yeah, I think we should undo normalization (but keep validation). I think I've written code before that checks the colors that I've applied, something like:
```
colormap = {'blue': '#3D9970', 'red': '#DB2913'}

if fig.marker.color not in colors.values():
    fig.marker.color = colormap['blue']
```

which might fail in a confusing way if my colors were normalized

plotly/tests/test_optional/test_figure_factory/test_figure_factory.py
Kully: Yes

plotly/tests/test_optional/test_figure_factory/test_figure_factory.py
Kully: Good idea. @jmmease I believe you said you had plans to return an instance of the JSON when a graph_objs figure was instantiated correct? If this is the case, then I'd wait for that so then we can only remove the `.to_plotly_json`s throughout the tests

plotly/tests/test_optional/optional_utils.py
Kully: It's not a typo. This is the wrapper function that removes keys in `ignore` then calls the good old `assert_dict_assert` we had before.

What's a better name for this?

How about I put this helper function inside `assert_dict_equal`, so:

```
def _dicts_with_removed_unwanted_keys(...):
   # does stuff

def assert_dict_equal(...):
   new_d1, new_d2  _dicts_with_removed_unwanted_keys(...)
   # the rest of the code
```

plotly/tests/test_core/test_tools/test_make_subplots.py
Kully: It's a mistake. Removing it  produced the same `expected` subplot, but I changed anchor to `y1` and `xaxis1` to `xaxis` and everything works as expected

Kully:
### Current Update:
All minor tests are now fixed/working. Here are the deeper issues that need fixing. (Note: I am using the most recent commits from ipyplotly_integration.):
- `_raise` is not a keyword and some tests are failing as a result: `__init__() got an unexpected keyword argument '_raise'`
- test_grid oddly fails
```
The 'xsrc' property is a string and must be specified as:
    - A string
```
- `ValueError: Invalid property specified for object of type plotly.graph_objs.Heatmap: 'fontcolor'`
- Test2D Density breaking change `('ncontours', 'colorscale', 'reversescale', 'showscale') not valid in Scatter`
- The following fails:
```
X = np.array([[1, 2, 3, 4], [1, 1, 3, 4], [1, 2, 1, 4], [1, 2, 3, 1]])
dendro = ff.create_dendrogram(X)
-----
[Error]



cc. @chriddyp @jmmease

**Comment by Kully** [comment]
Are you guys getting this error `__init__() missing 1 required positional argument: 'class_strs_map'` as well? Maybe I overwrote the wrong file in a merge

File: plotly/tests/test_optional/optional_utils.py (hunk 21 above)
Comment (jackparmer) [reply plotly/tests/test_optional/optional_utils.py]: @Kully I think Chris means the line, "Helper function for asser_dict_equal" (note "asser")

**Comment by jmmease** [comment]
@Kully I'm not getting that error, but it could be a problem with the generated classes being out of sync.  Try rerunning codegen, and if that doesn't work could you post a full stacktrace?

File: plotly/tests/test_optional/test_figure_factory/test_figure_factory.py (hunk 23 above)
Comment (jmmease) [reply plotly/tests/test_optional/test_figure_factory/test_figure_factory.py]: @Kully The `Figure` class itself is going to stay a non-`dict` class, so we will need to use `to_plotly_json()` to compare it to dictionaries.  Moving it into `assert_fig_equal` makes sense to me.

File: plotly/tests/test_optional/test_figure_factory/test_figure_factory.py (hunk 27 above)
Comment (jmmease) [reply plotly/tests/test_optional/test_figure_factory/test_figure_factory.py]: @chriddyp @Kully I just pushed ccbd940df2cd576d5de1e7ef650bc602aceb67bc which supports the `strict` option discussed in plotly/plotlyjs#2523. Now numbers should be accepted as strings for most string properties.  So we should be able to revert the number->str testing changes now.

File: plotly/tests/test_optional/optional_utils.py (hunk 21 above)
Comment (Kully) [reply plotly/tests/test_optional/optional_utils.py]: thank you. corrected

File: plotly/tests/test_optional/test_figure_factory/test_figure_factory.py (hunk 27 above)
Comment (jmmease) [reply plotly/tests/test_optional/test_figure_factory/test_figure_factory.py]: Oh I see, this change is in the 'expected' result.  With my latest change numbers are accepted by non-strict string properties (like annotation.text), but they are converted to strings internally.

File: plotly/tests/test_optional/test_tools/test_figure_factory.py (hunk 29 above)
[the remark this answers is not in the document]
Comment (jmmease) [reply plotly/tests/test_optional/test_tools/test_figure_factory.py]: Done in 4e5780570328ffaadb7635b276151989a74ce81c
@Kully , we should be able to revert this change now

**Comment by Kully** [comment]
@jmmease I think you removed `to_dataframe` from the PlotlyData methods in one of your commits. Any plans to bring it back?

**Comment by jmmease** [comment]
@chriddyp and I chatted about that briefly and I think he was fine with removing the `to_dataframe` `to_string`, `strip_styles`, `get_data`, and `validate` methods. 

Does that sound right Chris?

**Comment by Kully** [comment]
> @chriddyp and I chatted about that briefly and I think he was fine with removing the to_dataframe to_string, strip_styles, get_data, and validate methods.

If that's the case I will swiftly remove tests that test those methods

**Comment by Kully** [comment]
@jmmease  I'm noticing that invalid `marker.symbol` values are not getting ignored but are raising ValueErrors. Eg. this gives an error:

```
fig.add_scatter(x=[1], y=[2], marker=dict(size=20, symbol='foo'))
```

```
ValueError: 
    Invalid value of type 'builtins.str' received for the 'symbol' property of scatter.marker
        Received value: 'foo'

    The 'symbol' property is an enumeration that may be specified as:
      - One of the following enumeration values:
            [0, 'circle', 100, 'circle-open', 200, 'circle-dot', 300,
            'circle-open-dot', 1, 'square', 101, 'square-open', 201,
            'square-dot', 301, 'square-open-dot', 2, 'diamond', 102,
      ...
```

Are you keeping this strict validation on the parameters (I assume it's like this for more but I have been seeing a lot of this error with symbol) or going to coerce an invalid param value to a default i.e. `circle`?

**Comment by jmmease** [comment]
@Kully Overall I've been starting out with strict validation based on the schema and only adding value change coercion in cases where we really need it for compatibility (e.g. the `x1` -> `x` coercion for anchors)

In the case of symbols, I would personally prefer to keep it strict as I think this is better feedback for the user. What do you think @chriddyp ?

**Comment by Kully** [comment]
> to_dataframe to_string, strip_styles, get_data, and validate methods.

Is `_raise` one of these methods as well?

**Comment by jmmease** [comment]
I don't think @chriddyp and I discussed `_raise` at all.  I don't think we're going to be able to support the `_raise=False` option directly because validation is pretty baked into the new object hierarchy.  Could you try to understand what the use cases are for the `_raise=False` path? If we understand that we can make an informed decision on whether we can simply remove it, or if there is anything we need to add to continue to support these use cases.

File: plotly/plotly/plotly.py [hunk 32]
```diff
@@ -474,7 +474,7 @@ def get_figure(file_owner_or_url, file_id=None, raw=False):
 
     if raw:
         return figure
-    return tools.get_valid_graph_obj(figure, obj_type='Figure')
```
Kully: @chriddyp @jmmease 
The purpose of this commit is to refactor `get_valid_graph_obj` function to _not_ remove invalid properties of a figure when `.get_figure()` is called. I think this is important because:
-it is consistent with Jon's preference to validation over coercing to default attributes as a default (which I agree with)
-if a user does want to bypass validation errors, setting `raw=True` is still an option in the get_figure function 
-one step on the road to finishing https://github.com/plotly/plotly.py/issues/283 fwiw

What do y'all think?

jmmease: I like this. The user will either get a `dict` with whatever was on the server (`raw=True`) or a fully valid `Figure`.

Do you think the stripping of invalid properties is something that regular users are going to need? I'm not clear on why we end up with figures with invalid properties on the server in the first place.

plotly/tests/test_core/test_tools/test_make_subplots.py
```diff
@@ -103,98 +103,99 @@ def test_single_plot(self):
         expected = Figure(
             data=Data(),
             layout=Layout(
-                xaxis1=XAxis(
+                xaxis=XAxis(
```
jmmease: I think these changes from `xaxis1` -> `xaxis` in the constructors are no longer neccesary after the fix in the following commit:

*Rework handling of unknown kwargs in datatype classes* (a89dcb57f5c0f966ccc990d6c1f6dc5687430fba)
@Kully

jmmease:
@Kully The  `Key error 'type'` exception raised in `figure_factory/_scatterplot.py` should be fixed by 4fb462f560cdb6323ac56576cccc26eb55e558a8

plotly/plotly/plotly.py
Kully: I wondered the same thing. In refactoring these tests I noticed that a few of the example figures I pulled from were 4 years old and contained params that are no longer part of the valid params.

jmmease: I'd be inclined to update the test data, but I'll leave it to you and @chriddyp to decide whether we need to support some kind of legacy property removal.

plotly/tests/test_core/test_tools/test_make_subplots.py
Kully: Doesn't look like it's working:
```
expected = Figure(
    data=Data(),
    layout=Layout(
        xaxis1=XAxis(
            domain=[0.0, 1.0],
            anchor='y1'
        ),
        yaxis1=YAxis(
            domain=[0.0, 1.0],
            anchor='x1'
        )
    )
)
expected.to_plotly_json()
```
Output:
```
{'data': [], 'layout': {}}
```

Kully:
> should be fixed

it is

Kully:
@jmmease `from plotly.graph_objs.graph_objs import PlotlyDict, PlotlyList
` doesn't work anymore. Is there a way to import `PlotlyDict` and `PlotlyList`?

jmmease:
@Kully No, `PlotlyDict` and `PlotlyList` don't exist anymore

Kully:
> @Kully No, PlotlyDict and PlotlyList don't exist anymore

ok, I will remove their associated test

jmmease:
@Kully @chriddyp I'm ready to start working on adding some new tests against the new functionality of graph_objs. So I'm going to go ahead and merge this branch into `ipyplotly_integration` so I can use some of the new testing utilities.

Kully:
> Yup, that sounds right to me 👍

Does the same go for `.update`?

PR #976, opened 2018-03-18
Remove trailing whitespace

jbampton:


Kully:
@jbampton Is this all the trailing white space in our modules? If not, I would prefer tackling a much bigger portion of the code, preferably all of it so the PR is useful.

Kully:
Closing in preference of a much bigger PR that deals with all/most whitespace in the code.

Issue #996, opened 2018-04-24
Investigate http:base.domain vs. https:base.domain grid creation failure. 


BRONSOLO:
I've experienced that grid creation fails with the following error when my base domain is specified to use HTTP instead of HTTPS. 

![screen shot 2018-04-24 at 11 22 30 am](https://user-images.githubusercontent.com/1909107/39200720-e81b7004-47ba-11e8-9002-897b576424e6.png)

We should investigate why this occurred and look to add better handling of HTTP vs. HTTPS config file specifications. 

Perhaps we should warn a user if they have specified HTTP?

@Kully @cldougl

Kully:
> Perhaps we should warn a user if they have specified HTTP?

How often do requests use HTTP over HTTPS anyways? If it's substantial then a warning might be a good idea. Coercing to HTTPS doesn't ever seem like a good idea to me.

BRONSOLO:
@Kully no requests should be made over HTTP / all requests should be made over HTTPS.

Kully:
@BRONSOLO in that case I can write either a warning as you suggested or an error message that pops up before the more vague Plotly Request error to give a heads up. WDYT?

BRONSOLO:
>an error message that pops up before the more vague Plotly Request error to give a heads up

I say go with the error message. @cldougl what do you think? 

I think it would also be useful if someone could investigate/reproduce what I saw in the top comment re: the "Listing all files not supported" when making a POST request to the grid V2 create endpoint when HTTP is specified for the base/api domain in the config file. It might give us more insight into how best to handle this fix.

cldougl:
+1 for error message

@Kully was able to reproduce on etest

BRONSOLO:
Thanks @Kully and @cldougl!

@Kully - were you able to identify why the "Listing all files not supported" error occurred when the POST request to create a grid was made over HTTP?

Kully:
@BRONSOLO No I didn't investigate why.

BRONSOLO:
OK, I suppose it's not so necessary. I'm mostly just curious : )

Kully:
closing as issue is fixed

PR #997, opened 2018-04-25
validate plotly_(api)_domain - HTTPS, not HTTP

Kully:
Addresses https://github.com/plotly/plotly.py/issues/996

Notes:
1. This error message does not get triggered if the original url in the request starts with `zhttp://...`. A separate error message is thrown for that.
2. The original url sent by a user is getting coerced into all lowercase, so a lowercase to lowercase comparison is not necessary.

### DIFF
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -186,7 +186,6 @@ def set_config_file(plotly_domain=None,
     :param (bool) world_readable: True = public, False = private
 
     """
-    # import ipdb; ipdb.set_trace()
     if not check_file_permissions():
         raise exceptions.PlotlyError("You don't have proper file permissions "
                                      "to run this function.")
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -276,9 +276,20 @@ def get_config_file(*args):
     """
     if check_file_permissions():
         ensure_local_plotly_files()  # make sure what's there is OK
-        return utils.load_json_dict(CONFIG_FILE, *args)
+        returned_obj = utils.load_json_dict(CONFIG_FILE, *args)
     else:
-        return FILE_CONTENT[CONFIG_FILE]
+        returned_obj = FILE_CONTENT[CONFIG_FILE]
+
+    list_of_domains = []
+    for domain in ['plotly_domain', 'plotly_api_domain']:
+        if domain in returned_obj:
+            list_of_domains.append(returned_obj[domain])
+
+    validate_domains(*list_of_domains)
+
+    return returned_obj
+
+
 
 
 def reset_config_file():
--- a/plotly/tools.py
+++ b/plotly/tools.py
@@ -49,36 +49,6 @@ DEFAULT_HISTNORM = 'probability density'
 ALTERNATIVE_HISTNORM = 'probability'
 
 
-http_msg = (
-    "The plotly_domain and plotly_api_domain of your config file must start "
-    "with 'https', 'http'.\n If you are not using On-Prem then run the "
-    "following code to ensure your plotly_domain and plotly_api_domain start "
-    "with 'https':\n\n\n"
-    "import plotly\n"
-    "plotly.tools.set_config_file(\n"
-    "    plotly_domain='https://plot.ly',\n"
-    "    plotly_api_domain='https://api.plot.ly'\n"
-    ")\n\n\n"
-    "If you are using On-Prem then you will need to use your company's "
-    "domain and api_domain urls:\n\n\n"
-    "import plotly\n"
-    "plotly.tools.set_config_file(\n"
-    "    plotly_domain='https://plotly.your-company.com',\n"
-    "    plotly_api_domain='https://plotly.your-company.com'\n"
-    ")\n\n\n"
-    "Make sure to replace `your-company.com` with the URL of your Plotly "
-    "On-Premise server.\nSee "
-    "https://plot.ly/python/getting-started/#special-instructions-for-plotly-onpremise-users "
-    "for more help with getting started with On-Prem."
-)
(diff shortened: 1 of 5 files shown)


Kully:
cc @BRONSOLO @cldougl

BRONSOLO:
Thanks @Kully! 

@Kully @cldougl what do you y'all think of approaching this fix as a validation on the `.config` file as opposed to error handling on a failed request? 

Perhaps a helper function like `validated_config_file` can be written and called in the `get/set_config_file` helper functions to make sure the structure of the domains are valid (in this case, set to use HTTPS).

Kully:
> what do y'all think of approaching this fix as a validation on the .config file as opposed to error handling on a failed request?

I like this approach. What about we have both just in case? I don't know why the helper function could be bypassed if we set it up correctly though

BRONSOLO:
>What about we have both just in case?

I think the validation will be sufficient, personally!

cldougl:
Yep I think the config validation could be more beneficial for the user.
Also we suggest that on-premise users set only their `plotly_domain` rather than both `plotly_domain` and `plotly_api_domain` (https://plot.ly/python/getting-started/#special-instructions-for-plotly-onpremise-users) so I think it would be confusing to mention the `plotly_api_domain` in a message returned to the user.

BRONSOLO:
@Kully I think this should just be issued as a warning for now and not a full error. Though SSL/TLS is now required for On-Prem, there are some clients that still may be running older versions without SSL/TLS enabled.

Kully:
@BRONSOLO I've deleted the error msg and made a warning instead.

I only placed the `validate_config_file()` in `set_config_file()` as `get_config_file()` appears in the `set` function.

There is the possibility of added a `validate` param in `get_config` that calls the `validate_config_file` within.

I personally think that we need to validate `get_config_file` without a bunch of the same error message popping up

plotly/tools.py
```diff
@@ -244,9 +276,10 @@ def get_config_file(*args):
     """
     if check_file_permissions():
         ensure_local_plotly_files()  # make sure what's there is OK
-        return utils.load_json_dict(CONFIG_FILE, *args)
+        returned_obj = utils.load_json_dict(CONFIG_FILE, *args)
     else:
-        return FILE_CONTENT[CONFIG_FILE]
+        returned_obj = FILE_CONTENT[CONFIG_FILE]
+    return returned_obj
```
BRONSOLO: Is this change necessary? Could you add it to a separate commit with a commit msg that explains why if so?

plotly/tools.py
```diff
@@ -49,6 +49,36 @@
...
+    "    plotly_api_domain='https://plotly.your-company.com'\n"
+    ")\n\n\n"
+    "Make sure to replace `your-company.com` with the URL of your Plotly "
+    "On-Premise server.\nSee "
+    "https://plot.ly/python/getting-started/#special-instructions-for-plotly-onpremise-users "
+    "for more help with getting started with On-Prem."
+)
+
+
+def validate_config_file(*domains):
```
BRONSOLO: This is really just validating domains - what do you think about renaming this as `validate_domains`?

Kully: good idea

Kully:
notes:
- moved test to plotly.utils
- new test should be passing 🎉 
- only raise UserWarning on `set_config_file`, not `get_config_file`

Waiting for tests to pass.

Kully:
Looks like only streaming errors. We are ignoring these for now @cldougl ?

cldougl:
@kully, the streaming tests should pass now

plotly/tools.py
```diff
@@ -186,12 +186,14 @@ def set_config_file(plotly_domain=None,
     :param (bool) world_readable: True = public, False = private
 
     """
+    # import ipdb; ipdb.set_trace()
```
BRONSOLO: please remove this commented code as it is unnecessary 🔪 : )

plotly/utils.py
```diff
@@ -297,6 +322,7 @@ def encode_as_decimal(obj):
         else:
             raise NotEncodable
 
+
```
BRONSOLO: Is this line needed?

BRONSOLO:
@Kully could you please provide a screenshot of what that warning message looks like when hit?

plotly/utils.py
```diff
@@ -32,6 +33,30 @@
 lock = threading.Lock()
 
 
+http_msg = (
+    "The plotly_domain and plotly_api_domain of your config file must start "
+    "with 'https', 'http'.\n If you are not using On-Prem then run the "
```
BRONSOLO: Why include `http` here? --->  "with 'https', 'http'. 

Also I would suggest using `On-Premise` instead of `On-Prem` to be consistent with the rest of the codebase.

Kully: That was a typo. I meant to type `'http', not 'https'`

Kully:
@BRONSOLO all comments addressed. Waiting for tests again.

Kully:
yay tests are passing 🎉

plotly/utils.py
```diff
@@ -35,15 +35,15 @@
 
 http_msg = (
     "The plotly_domain and plotly_api_domain of your config file must start "
-    "with 'https', 'http'.\n If you are not using On-Prem then run the "
+    "with 'https', not 'http'.\n If you are not using On-Premise then run the "
```
BRONSOLO: I believe the `\n If` ... spacing still needs to be fixed. I think it would look cleaner without a new line actually.

BRONSOLO:
One small comment and then 💃 after that.

PR #1007, opened 2018-05-09
update plotlyjs for offline for plotlyjs 1.37.1 bug fix

cldougl:
cc @Kully


Kully:
@cldougl  looks good 💃

PR #1021, opened 2018-05-30
Updated offline.py function iplot, takes height and width as argument

xhlulu:


Kully:
@xhlulu 
Playing around in offline mode, it looks like `image_width` and `image_height` don't even work at all. It's possible it was an oversight to leave them in the code as they do nothing. I am only able to change the width and height in layout. Eg.

```
fig = {
  'data': [...],
  'layout': {
      'height':50,
      'width': 100
  }
}
```

So I think this PR is not necessary as it doesn't solve the problem underneath it. If you want to dig into this issue a little further you can.

You can also create an issue at https://github.com/plotly/plotly.py/issues so others know that is a problem that needs to get solved.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2.7.1] - UNRELEASED
+### Updated
+- `plotly/offline.py` function `iplot` now takes arguments `weight` and `height`, and overwrites `image_height` and `image_width` when provided.
```
eliasdabbas: I think you mean 'width' and not 'weight'?

PR #1042, opened 2018-06-28
expose new plotly.js config option

itsb:
Expose new JS config option https://github.com/plotly/plotly.js/pull/2607


jonmmease:
Captured by #1056

PR #1056, opened 2018-07-13
Add configkeys to offline matching plotly js 1.39.2

jeffreypriebe:
Looking at plotly js, there are a number of configkeys that are valid that aren't in this offline whitelist. This adds [all configkeys from current plotly js](https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_config.js#L19).

Note, this PR encompasses: #1042

### DIFF
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -179,13 +179,19 @@ def _plot_html(figure_or_data, config, validate, default_width,
                               cls=utils.PlotlyJSONEncoder)
 
     configkeys = (
+        'staticPlot'
+        'plotlyServerURL',
         'editable',
+        'edits',
         'autosizable',
+        'queueLength'
         'fillFrame',
         'frameMargins',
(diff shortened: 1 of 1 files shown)


jonmmease:
Thanks @jeffreypriebe , I'll take a look over this when we upgrade to Plotly.js 1.39.x (which will become plotly.js 3.1.0)

jeffreypriebe:
Sounds good @jonmmease 

Right now, can't output certain valid js configs from python, so running on a fork.

jonmmease:
Thanks @jeffreypriebe!
Eventually we should also work through which of these options should be exposed to the new `FigureWidget` class.

jeffreypriebe:
Great!

Tagging @alextpg in case he wants these changes for future use (3.0 branch).

alextpg:
Yup, good. Will be useful in the 3.0. Thanks @jeffreypriebe  and @jonmmease!

PR #1089, opened 2018-08-02
Fix ColorScale presentation when specified as string

### LINKED ISSUE #1087: ColorScaleValidator should probably return a string when it's specified as a string.
Currently, color scale attributes have a kinda weird behavior:

```python
go.scatter.Marker(colorscale="Viridis").colorscale
```
```
(('V',), ('i',), ('r',), ('i',), ('d',), ('i',), ('s',))
```

This is unlike the behavior (to the best of my knowledge) for other attributes specified as strings:

```python
go.scatter.Marker(color="red").color
```
```
'red'
```

And this is pretty unintuitive behavior when doing stuff like:

```python
marker1 = go.scatter.Marker(colorscale="Viridis")
go.scatter.Marker(colorscale=marker1.colorscale)
```
```
ValueError: 
    Invalid value of type 'builtins.tuple' received for the 'colorscale' property of scatter.marker
        Received value: (('V',), ('i',), ('r',), ('i',), ('d',), ('i',), ('s',))
```

Though `.update(marker1)` does work for some reason.

My guess is all that needs to happen is to add an `elif isinstance(v, str): return v` to this function, since you won't be breaking that immutability requirement.

https://github.com/plotly/plotly.py/blob/d670bb0f535fbd0a884c699aaa835ba624721859/_plotly_utils/basevalidators.py#L1276-L1281

ivirshup:
Fixes #1087.

I kinda guessed with where I put the test, let me know if there's a better place for it.


_plotly_utils/basevalidators.py
```diff
@@ -1274,9 +1274,11 @@ def validate_coerce(self, v):
         return v
 
     def present(self, v):
-        # Return tuple of tuples so that colorscale is immutable
+        # Return-type must be immutable
         if v is None:
             return None
+        elif isinstance(v, str):
```
jonmmease: Please use `isinstance(v, string_types)` for Python 2 compatibility

plotly/tests/test_core/test_graph_objs/test_properties_validated.py
```diff
@@ -102,6 +102,16 @@ def test_present_colorscale(self):
...
+    
+    def test_present_colorscale_str(self):
+        self.assertIsNone(self.scatter.marker.colorscale)
+
+        # Assign string
+        self.scatter.marker.colorscale = "Viridis"
+
+        # Presented as a string
+        self.assertEqual(self.scatter.marker.colorscale,
+                         "Viridis")
```
jonmmease: In addition to this test, we should update the `test_acceptance_named` validator test in `_plotly_utils/tests/validators/test_colorscale_validator.py`

It should work a bit more like the `test_acceptance_array` test in that same file (the second one, it's a mistake that there are two with the same name). Where the coerced value is held onto and then run through the validators `present` method.

If it's not obvious how this should work don't spend much time on it, I need to do some cleanup in there soon anyway.

ivirshup: I've gone ahead and updated that test. Would it be worthwhile to set the `named_colorscale` fixtures possible values to be set either by `ColorScaleValidator.named_colorscales` or some schema to be sure it's up to date?

Also, should Cividis be on those lists?

jonmmease: I think I'd like to keep the explicit list of named color scales in the tests. It is duplicated with the list in `ColorScaleValidator.named_colorscales`, but if we ever try to pull the validator's named colorscales list out of the schema I'd like to have this explicit list in the tests

Yes, good call on Cividis. Could you add that to both the test fixture and the `named_colorscales` list? Thanks!

ivirshup: Added. I think that's everything, right?

Issue #1100, opened 2018-08-08
`tickmode` set to invalid value in mpltools


nikhase:
* OS: Ubuntu 16.04
* Python 3.6.6
* plotly version: 3.1.0

As you can convert matplotlib figures to plotly, it should be possible with seaborn figures as well. I tried the following code snipped:

```python
import seaborn as sns
import plotly.offline as ply
import plotly.tools as plytls

exercise = sns.load_dataset("exercise")
g = sns.catplot(x="time", y="pulse", hue="kind", data=exercise)

ply.iplot_mpl(g.fig)
```

How the figure looks like:

![image](https://user-images.githubusercontent.com/18031651/43833681-e5e9da74-9b0b-11e8-87b8-ff1a9c47f188.png)

This results in the following error:
```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-11-523d3e17715d> in <module>()
----> 1 ply.iplot_mpl(g.fig)

~/anaconda3/envs/mlapp/lib/python3.6/site-packages/plotly/offline/offline.py in iplot_mpl(mpl_fig, resize, strip_style, verbose, show_link, link_text, validate, image, image_filename, image_height, image_width)
    682     ```
    683     """
--> 684     plotly_plot = tools.mpl_to_plotly(mpl_fig, resize, strip_style, verbose)
    685     return iplot(plotly_plot, show_link, link_text, validate,
    686                  image=image, filename=image_filename,

~/anaconda3/envs/mlapp/lib/python3.6/site-packages/plotly/tools.py in mpl_to_plotly(fig, resize, strip_style, verbose)
    463     if matplotlylib:
    464         renderer = matplotlylib.PlotlyRenderer()
--> 465         matplotlylib.Exporter(renderer).run(fig)
    466         if resize:
    467             renderer.resize()

~/anaconda3/envs/mlapp/lib/python3.6/site-packages/plotly/matplotlylib/mplexporter/exporter.py in run(self, fig)
     47             import matplotlib.pyplot as plt
     48             plt.close(fig)
---> 49         self.crawl_fig(fig)
     50 
     51     @staticmethod

~/anaconda3/envs/mlapp/lib/python3.6/site-packages/plotly/matplotlylib/mplexporter/exporter.py in crawl_fig(self, fig)
    114                                        props=utils.get_figure_properties(fig)):
    115             for ax in fig.axes:
--> 116                 self.crawl_ax(ax)
    117 
[...]

PR #1101, opened 2018-08-08
Ref #1100 , tickmode set to invalid value in mpltools

nikhase:
* basevalidator throws an error on the boolean

Associated Issue: #1100

### DIFF
--- a/plotly/matplotlylib/mpltools.py
+++ b/plotly/matplotlylib/mpltools.py
@@ -453,13 +453,13 @@ def prep_ticks(ax, index, ax_type, props):
         else:
             axis_dict['tick0'] = tick0
             axis_dict['dtick'] = dtick
-            axis_dict['tickmode'] = False
+            axis_dict['tickmode'] = None
     elif scale == 'log':
         try:
             axis_dict['tick0'] = props['axes'][index]['tickvalues'][0]
             axis_dict['dtick'] = props['axes'][index]['tickvalues'][1] - \
                             props['axes'][index]['tickvalues'][0]
-            axis_dict['tickmode'] = False
+            axis_dict['tickmode'] = None
         except (IndexError, TypeError):
             axis_dict = dict(nticks=props['axes'][index]['nticks'])
         base = axis.get_transform().base


plotly/matplotlylib/mpltools.py
```diff
@@ -453,13 +453,13 @@ def prep_ticks(ax, index, ax_type, props):
         else:
             axis_dict['tick0'] = tick0
             axis_dict['dtick'] = dtick
-            axis_dict['tickmode'] = False
+            axis_dict['tickmode'] = None
```
jonmmease: I think it would be better to be explicit here and use `'auto'` rather than `None`

plotly/matplotlylib/mpltools.py
```diff
@@ -453,13 +453,13 @@ def prep_ticks(ax, index, ax_type, props):
...
     elif scale == 'log':
         try:
             axis_dict['tick0'] = props['axes'][index]['tickvalues'][0]
             axis_dict['dtick'] = props['axes'][index]['tickvalues'][1] - \
                             props['axes'][index]['tickvalues'][0]
-            axis_dict['tickmode'] = False
+            axis_dict['tickmode'] = None
```
jonmmease: Seems like this should be tick mode of `'log'`.  Would you mind trying this out on a simple matplotlib figure with a log axis?

plotly/matplotlylib/mpltools.py
nikhase: Can you maybe help me to understand what is going on? Are you trying to catch the axis scaling properties {“linear”, “log”, “symlog”, “logit”} from [matplotlib.axes.Axes.set_xscale](https://matplotlib.org/api/_as_gen/matplotlib.axes.Axes.set_xscale.html#matplotlib.axes.Axes.set_xscale) (other axis respectively)? There is no `auto` option. Maybe `None` is treated as auto?
It seems like despite the `tickmode` is set to `None`, it is inferred later on, for example the parsed JSON shows:
```...
'layout': {'autosize': False,
               'height': 360,
               'hovermode': 'closest',
               'margin': {'b': 41, 'l': 61, 'pad': 0, 'r': 79, 't': 14},
               'showlegend': False,
               'width': 434,
               'xaxis':{...
                         'type': 'linear',
                         ...},
               'yaxis': {'anchor': 'x',
                         ...
                         'type': 'linear',
                         ...}
...
```

plotly/matplotlylib/mpltools.py
nikhase: This is a [matplotlib log example](https://matplotlib.org/examples/pylab_examples/log_demo.html):
image
**With `None`, it looks like:**
image

and brings the warnings:
```
plotly.graph_objs.Annotation is deprecated. 
...
Converted non-base10 x-axis log scale to 'linear'
...
Dang! That path collection is out of this world. I totally don't know what to do with it yet! Plotly can only import path collections linked to 'data' coordinates
```
**With `log`, it looks like:**
image
Same warning as in the `None` setting here.
I cannot see a difference between `None` and setting to `log`. What should it then be?
Despite that, there are obviously troubles converting non base10 logs.

plotly/matplotlylib/mpltools.py
nikhase: btw, i chose `None` in the first place because of the code snippet `axis_dict.pop('tickmode', None)` in line 491.

jonmmease:
Thanks for looking into setting the axis type to values other than `None`.  Based on the fact that the `type` is inferred to be `linear` later on, I think your solution of just changing `False` to `None` makes sense.

Thanks for the contribution!

PR #1109, opened 2018-08-13
CircleCI 2.0 migration

jonmmease:
Port of test suite to 2.0, with improvements. Thanks to @scjody for the reminders 🙂 

General approach: There is now a workflow for each Python version / test suite combination.  This way all of the jobs can run in parallel. I also refactored the `core` tests to pull out the tests that rely on `plot.ly` interaction. These are currently the slowest tests, and pulling them out means that the core tests run really fast. It also means that if there is an intermittent failure due to plot.ly communication, we can restart only the plot.ly tests for the particular version of Python that had the failure.

Updates:
 - [x] Split plot.ly tests into separate directory
 - [x] Add core and core+optional workflows for 2.7, 3.4, 3.5, 3.6, and 3.7 (These run pretty fast now without the plot.ly tests)
 - [x] plot.ly tests for 2.7, 3.4, and 3.7?
 - [x] Add pytest validator tests to optional (these tests need the extra dependencies)
 - ~Add useful caching~ Caching tox sped builds up by a minute at most and would add a lot of extra complexity to the build file. So holding off for now.

Future testing wishlist: https://github.com/plotly/plotly.py/issues/1113

cc @cldougl @chriddyp @Kully @scjody


jonmmease:
Alright, time to merge this thing. Goodbye CircleCI 1.0!

PR #1110, opened 2018-08-14
Show colors at the top of the gantt's colorbar

omridanan:
When creating gantt with custom colors and there are a lot of tasks, the color legends are located at the bottom of the colorbar and not visible until the user scroll down.
This fix creates the dummy data with the colors before the data itself. This way, we still have a scroll on the colorbar, but the colors are located at the top and the user can see them immediately.


jonmmease:
Hi @omridanan , could you include some before and after screenshots of this change?

Looks like there are also some failing figure factory tests that would need to be updated along with this.

Thanks!

omridanan:
Sure @jonmmease 

Before:
image

When you scroll down:
image

After the fix:
image

jonmmease:
Thanks @omridanan ,

Rather than moving the dummy legend data to the top, could we avoid displaying all of the other traces in the legend to start with?  We could do this by setting the `showlegend=False`. This way there wouldn't be the scroll bar all.

What do you think?

omridanan:
@jonmmease works great, I'll do it.

jonmmease:
Hi @omridanan , could you merge master into this branch?  There are some updates to the testing infrastructure that make it easier to pinpoint test failures.

omridanan:
@jonmmease seems to be fine now

Issue #1116, opened 2018-08-15
FIgureWidget not working with nbviewer


tarzzz:
Example Notebook: http://nbviewer.jupyter.org/gist/tarzzz/7aa6eb0673d5b11b55748c1dd9f74537#

The Plot is not displayed when the chart is viewed on nbviewer (or as the HTML version).
Not sure if this is a bug, or if it is not intended to work in static views.

cc: @jonmmease

empet:
@tarzz  On my machine it works with both Python 2.7.13 and Plotly 3.1.0, respectively  Python 3.6.4 and Plotly 3.1.1.

tarzzz:
@empet It works on my machine, but not on the external/HTML viewers (like nbviewer.jupyter.org).

empet:
@tarzz  
Is the notebook version >=5.3?  This is a requirement.

empet:
@tarzz  In a comment to this issue [https://github.com/jupyter-widgets/ipywidgets/issues/2063](https://github.com/jupyter-widgets/ipywidgets/issues/2063) is suggested how to get the widgets displayed in nbviewer.
I followed the steps explained here: [https://ipywidgets.readthedocs.io/en/latest/embedding.html#embedding-widgets-in-html-web-pages](https://ipywidgets.readthedocs.io/en/latest/embedding.html#embedding-widgets-in-html-web-pages), but  the FigureWidget isn't rendered: [http://nbviewer.jupyter.org/gist/empet/f6a363e1d0140e69f5e9139bb742849f](http://nbviewer.jupyter.org/gist/empet/f6a363e1d0140e69f5e9139bb742849f)

jonmmease:
The behavior I'm seeing is that the figure axes are being displayed, but not the data. Is that what you're both seeing?

I think this is an issue related to the way that numpy arrays are encoded as binary buffers.  These buffers are stored as JavaScript TypedArrays in the JavaScript widget model. And when the buffers are saved (through the "Save notebook widget state" process), and then restored (when a page is viewed on nbviewer) they are not being restored as typed arrays, but as `DataView` objects.

To test the theory, here's a notebook with a figure that doesn't use numpy arrays. I saved widget state and pushed to GitHub and then loaded with nbviewer.

https://nbviewer.jupyter.org/github/jonmmease/plotly_ipywidget_notebooks/blob/master/notebooks/HelloWorldFigureWidget.ipynb

[...]

Issue #1041, opened 2018-06-27
autoresizing behaviour


dhirschfeld:
As mentioned in https://github.com/plotly/plotly.py/pull/942#issuecomment-400135253, in the JupyterLab display, in a lot of cases, it would be very nice if the plot would automatically resize when the window/tab resized.

Since there are situations where this wouldn't be desirable e.g. in the case of large/slow plots, it would be very good if this behaviour could be chosen by the user with an `autoresize` property/argument/config

jmmease:
Alright, I just pushed out `plotlywidget@0.1.1` and `plotly==3.0.0rc10` and I added responsive resizing in JupyterLab (the classic notebook doesn't seem to share the same PhosphorJS resize event).

The intended behavior is for plots to expand to fill their container on construction, and when the container is resized. But, if the figure has `layout.width` or `layout.height` specified explicitly then no resizing takes place.  

I think this is a nice balance. Figure's will resize unless you say what size you want. For plots that are expensive to resize, the user just needs to decide what size they want and it will stick there.

Give it a try and see what you think

jonmmease:
I think what you were talking about here was included in 3.0.0. Can you confirm @dhirschfeld?

jonmmease:
Going to close this, as I believe it was sufficiently addressed prior to the release of version 3.0.0. We can reopen or create a new issue if there are any other outstanding issues.

dhirschfeld:
Sorry, other priorities got in the way of testing this.

I'm now using plotly 3.1.0 with jupyterlab 0.33.7 and the plot is now automatically resizing which is great!

If I specify the height of the figure the auto-resizing no longer happens - as you mentioned above. I think it would be nice though if only the height is specified for the *width* to continue to automatically resize and vice-versa. i.e. auto-resizing would only be completely disabled if both height and width were specified.

Not sure if that's feasible?

jonmmease:
Thanks for the suggestion @dhirschfeld.  The width resizing with fixed height is being implemented in #1117

slishak:
I'm getting a similar issue to this in jupyterlab with Plotly Express. If I use `plotly_graph_objs.FigureWidget` setting fixed height but no width, the width scales as expected but the height does not. However, if I use `plotly_express.scatter`, the `height` argument is not respected and it auto-resizes both the width and the height. 

![image](https://user-images.githubusercontent.com/12515343/56958698-512af600-6b4b-11e9-8c88-a8ea224b0823.png)

PR #1117, opened 2018-08-17
Support for embedding plotlywidget instances that use numpy arrays

jonmmease:
See https://github.com/plotly/plotly.py/issues/1116 for background.

With these updates it is now possible to embed `FigureWidget` instances containing numpy arrays in static contexts (e.g. nbviewer, and soon Plotly Cloud).

I also took to opportunity to update the auto-resizing logic to:
  - Support the classic notebook
  - Still resize width even if the height is set explicitly (See https://github.com/plotly/plotly.py/issues/1041#issuecomment-411297737)

This change involves no changes to the Python library, so we should be able to release this as `plotlywidget@0.2.2` without making a release of plotly.py yet.

Also, to help with embedding, this version emits an event named `'plotlywidget-after-render'` when the widget rendering process is complete.

cc @tarzzz


jonmmease:
@tarzzz , if I publish this branch as `plotlywidget@0.2.2-rc.1` will you be able to test with this version?  
Being an rc, you won't get it without specifying the version explicitly, and I'm not sure if you're able to do that.

tarzzz:
@jonmmease I tested  `0.2.2-rc.2` with Plotly Cloud and everything looks good.. !!

jonmmease:
Merged in the CircleCI 2.0 integration

jonmmease:
@tarzzz Great! I'm going to merge this.  My plan is to release this along with plotly.py version 3.2.0 and `plotlywidget` 0.3 (which will include Plotly.js 1.40).  But let me know if you have particular time constraints you're working with.

Issue #1105, opened 2018-08-10
Orca integration for static image export


jonmmease:
## Overview
This is design proposal for the integration of orca into plotly.py in order to support the programmatic export of high quality static images.

## Related issues:
 - https://github.com/plotly/plotly.py/issues/880
 - https://github.com/plotly/plotly.py/issues/596
 - https://github.com/plotly/plotly.py/issues/564

## Background
The programmatic export of static raster and vector images from JavaScript-based data visualization libraries is a notoriously complicated problem.  One common solution is to combine selenium with a driver for a headless web browser like phantomjs or headless firefox/chrome. This approach is used by [Bokeh](https://bokeh.pydata.org/en/latest/docs/user_guide/export.html) and [Altair](https://altair-viz.github.io/user_guide/saving_charts.html) for example.  One challenge with this approach is that it requires the installation of dependencies that are not managed by a Python environment friendly package manager like `conda` (Although [phantomjs](https://github.com/ariya/phantomjs) is available through conda, its development has been suspended and it does not support WebGL).  This presents challenges in terms of portability and reproducibility.

The plotly.js team has taken a different approach with the [Orca project](https://github.com/plotly/orca). Orca is a standalone Electron application that can run as a command line image export tool, or it can run in a server mode and respond to image export requests interactively.  Orca is the backbone of the plot.ly image export service, and it was open sourced earlier this year.

Because Orca can be built into a standalone executable that does not depend on a system web browser, it is possible to package Orca as a conda package, and we've had [recent success](https://github.com/plotly/orca/pull/113) towards this goal.

This issue is for the discussion of how to build the best plotly.py image export experience on top of Orca.

## Goals
 - Users shouldn't need to be aware of how complicated static image export is. At most it should require a single additional conda installation command.
 - It should be as easy and reliable to use as matplotlib's image export.
 - Nothing should flash on the screen or dock or taskbar during export.
 - For raster formats, it should support png, jpg, and webp with configurable resolution.
 - For vector formats it should support svg, pdf, and eps.
 - It should be possible to save images directly to the local filesystem, or to a writable file object.
 - It should be possible to return a byte string containing the image data without specifying filenames (and ideally without actually writing anything to temp files).
 - It should be fast enough to support use as an interactive plotting backend (See #1098)
 - It should provide really helpful error messages if the orca executable isn't found.

### 1. Use command-line interface with figure as arg
The current Python instructions in the [Orca README](https://github.com/plotly/orca) suggest the following usage:

[...]

Issue #1098, opened 2018-08-07
New module proposal: plotly.io


jonmmease:
## Overview
I propose that we introduce a new module, `plotly.io`, to house a new suite of functions related to the input and output of plotly figure objects.

## Why now?
There are two immediate needs motivating this proposal.

 1. It's time to integrate [orca](https://github.com/plotly/orca) for static image export
 2. For integration with the JupyterLab [plotly-renderer](https://github.com/jupyterlab/jupyter-renderers/tree/master/packages/plotly-extension) and JupyterLab [chart editor](https://github.com/plotly/jupyterlab-chart-editor), we need a function to write out the JSON representation of a figure with the proper plotly metadata

## How do things work now?
Here's in inventory of some of our current methods that can perform I/O operations:
 - `plotly.plotly.plot`
   - Save figure to plot.ly (with optional filename) and return URL
   - Optionally open in browser
 - `plotly.offline.plot`
   - Save html to a local file
   - or return an HTML `div` string
   - optionally open in browser 
 - `plotly.image.save_as`
   - Save image to a local file with specified image parameters by making a request to plot.ly
 - `plotly.get_figure`
   - Convert plot.ly URL into a local figure object

## Why add something new?
By my taste, these APIs are a bit scattered and not very discoverable by users without consulting external documentation.  It's also unclear (to me at least) where additional I/O methods, like to two listed above, should be added.

There are good reasons that we ended up where we are (e.g. there was no offline mode when the `plotly.plotly` module was developed), but I'd like to propose we adopt a new approach moving forward.

The goal is that the options for saving/loading/converting figures should consistent and easily discoverable without the need to consult external documentation.  By triggering tab-completion on the `plotly.io` module, users should be able to learn about all of the import/export options available to them.

## Design
I propose that the initial functions in the `plotly.io` module have one of 4 prefixes, followed by the format name:

 - `read_{format}(file, ...) -> Figure`:  These methods all accept, as the first argument, either a string or a `read`-able file object. Strings are interpreted as local file paths, other objects are read from. These methods all return a `Figure` object.
 - `write_{format}(fig, file, ...) -> None`: These methods all accept the `Figure` to be written as the first argument. The second argument is either a string or a `write`-able file object. Strings are interpreted as local file paths, other objects are written to. These methods `raise` on write failures and return `None`.
[...]

PR #1120, opened 2018-08-19
Orca integration for static image export

jonmmease:
## Overview
This PR integrates [orca](https://github.com/plotly/orca) into plotly.py to support exporting figures as static images :tada:

cc @chriddyp @jackparmer @nicolaskruchten  @cldougl @Kully @etpinard @malmaud

Even if you don't have time to look at the code or test out the branch, I'd appreciate any feedback on the architecture and API notes.

Here goes...

## Background
See https://github.com/plotly/plotly.py/issues/1105 for background information and discussion of related work.

## Architecture
In this PR I went with method (3) from the issue above, "Use orca in server mode". 

The first time an image export operation is performed, an orca server process is launched in the background (as non-blocking subprocess).  Image export requests are posted to the server on a local port.

By default, the server process runs until the main process exits. But there is also a `timeout` configuration option (more on configuration options below) that allows a user to specify that the server should be automatically shut down after a certain period of inactivity.

Regardless of whether a `timeout` is set, the server may also be manually shutdown and manually started.

### Starting the server
The server subprocess is launched using [`subprocess.Popen`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) to create a long-running background process.  The server is launched in [`--graph-only` mode](https://github.com/plotly/orca/pull/114) to be as lean as possible (this avoids running processes for exporting thumbnails, dashboards, etc.)

### Communicating with the Server
Communication with the server is done using [`requests.post`](http://docs.python-requests.org/en/master/user/quickstart/). The request function is wrapped in the [`@retrying.retry` decorator](https://github.com/rholder/retrying) to handle the automatic retrying of failed requests.  The retrying logic is very convenient, as it allows an image request to be made right after the server process is launched and the request will simply block until the server responds.

### Shutting down the server
It's possible to terminate the particular process created using `subprocess.Popen` with the [`Popen.terminate`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.terminate) method.  Unfortunately, this isn't always enough to actually shut down the server. The trouble is that typical orca entry points (orca.sh, orca.js, orca.cmd) are simply wrapper scripts that call the main orca/electron executable.  In my testing on OS X, Linux, and Windows I found that `Popen.terminate` generally only terminates the shell/wrapper process, leaving the orca server running.  This is definitely not acceptable, as a user could end up with a new orca process each time they restart their kernel and export images.

I initially tried some workarounds involving process groups, and sending different signals, but the result ended up being platform dependent and still not fully reliable. I settled on introducing the [`psutil`](https://github.com/giampaolo/psutil) library as a new optional dependency.  `psutil` provides a platform agnostic API for iterating over the children of a process, and then terminating them.  In my testing, this `psutil` approach has been fully reliable in terminating the server processes across platforms.  Since our CI test suites is Linux only at this point, I'm especially glad to not need to introduce any OS X/Windows specific process management logic.

### Shutdown server after timeout
If a `timeout` is configured when the server process is launched, a [`threading.Timer`](https://docs.python.org/3/library/threading.html#timer-objects) object is created to call the shutdown function after `timeout` seconds.

Each time an image render request is made, any existing `Timer` object is [canceled](https://docs.python.org/3/library/threading.html#threading.Timer.cancel), and a new `Timer` is created.

Importantly, each timer thread has the [daemon](https://docs.python.org/3/library/threading.html#threading.Thread.daemon) property set to `True`. This prevents the main process from waiting for the timer to complete before exiting.

### Shutdown on exit
The shutdown function is annotated with the [`@atexit.register`](https://docs.python.org/3/library/atexit.html) decorator to ensure that the server is properly shutdown when the main Python process exits.

## API Design
This PR introduces the beginning of the [`plotly.io` module](https://github.com/plotly/plotly.py/issues/1098). 

### Image export
Two image export functions are introduced. These function follow the export conventions proposed in #1098.

```python
plotly.io.write_image(fig, file, format=None, scale=None, width=None, height=None)
```

This functions works very much like the matplotlib [`savefig`](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.savefig.html) function. `fig` is a `Figure` or compatible `dict`. `file` may be a string referring to a local filesystem path, or a file-like object to be written to. If `file` is a string, then the file extension is used to infer the image format if possible. The `format` may be used to explicitly specify the format, and it is required if `file` is not a string with a common extension.  Supported formats are png, jpeg (jpg extension supported as well), webp, svg, pdf, and eps (with poppler installed). `scale`, `width`, and `height` work as you would expect.

```python
plotly.io.to_image(fig, format=None, width=None, height=None, scale=None)
```
This function may be used to return the binary representation of the image directly (no temp files or messing with `io.BytesIO`!).  This can be used in conjunction with `IPython.display.Image` to display static images directly in the notebook or QtConsole.

### Orca management
If users install orca using conda or npm, they should be able to use the above methods immediately, without additional configuration. But for more technical users, and for general users if things go wrong, there is a new `plotly.io.orca` module.

#### Manual server management
The server may be manually started using `plotly.io.orca.ensure_orca_server()`, and it may be manually shut down using `plotly.io.orca.shutdown_orca_server()`

#### Orca config
`plotly.io.orca.config` is an orca configurations/settings object. Here are the properties that can be configured

```
orca configuration
------------------
    port: None
    executable: orca
    timeout: None
    default_width: None
    default_height: None
    default_scale: 1
    default_format: png
    mathjax: https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js
    topojson: None
    mapbox_access_token: None

constants
---------
    plotlyjs: /path/to/plotly/package_data/plotly.min.js 
    config_file: /Users/username/.plotly/.orca
```
If automatic port selection is not desirable, an explicit `port` value may be set here. If an executable named `orca` cannot be found on the path, then the `executable` property may be set to the absolute path to an orca executable.  This is where the `timeout` property is set. The default width, height, scale, and format, control the default values used by `to_image` when not otherwise specified.

I took the liberty of supplying a default mathjax CDN, this way latex image export just works as long as the user is online. For offline use, the `mathjax` property can be set to the path to a local mathjax installation. When `topojson` is `None` the plot.ly CDN will be used, but a local path can be supplied if working offline.  Finally, the `mapbox_access_token` property can store a mapbox token that will automatically be applied when exporting mapbox traces.

Properties can be set using property assignment
```python
plotly.io.orca.config.mapbox_access_token = 'xyz...'
```
or using the `update` method
```python
plotly.io.orca.config.update(mapbox_access_token='xyz...')
```
The constants are not settable and are listed for informational purposes.

#### Saving configuration properties
The config values may optionally be saved to the `~/.plotly` settings directory as `~/.plotly/.orca` using the `plotly.io.config.save()` method. If present, these setting are automatically loaded on import. 

#### Orca status
The current status of the orca server process can be displayed using the `plotly.io.orca.status` object.

At initial startup the `state` will be `unvalidated`
```
orca status
-----------
    executable: None
    version: None
    port: None
    pid: None
    state: unvalidated
    command: None
```
After a valid orca executable has been found, and the server is not yet running, the `state` will be `validated'

```
orca status
-----------
    executable: /anaconda3/envs/plotly_dev/bin/orca
    version: 1.1.0
    port: None
    pid: None
    state: validated
    command: None
```
Here the user can see which orca executable was found on the path, and what version it is.

When the server process is currently running, the `state` will be `running`
```
orca status
-----------
    executable: /anaconda3/envs/plotly_dev/bin/orca
    version: 1.1.0
    port: 59997
    pid: 83079
    state: running
    command: ['orca', 'serve', '-p', '59997', '--graph-only', '--plotly', '/path/to/plotly/package_data/plotly.min.js', '--mathjax', 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js', '--mapbox-access-token', 'pk...']
```
Here the user can see the details of the running process (port, pid) and the exact command line arguments that were passed to the orca server at startup.

### Error messages
There are a lot of things that can potentially go wrong here, so I've tried to make the error messages as helpful as possible.  For example here's the error that is raised if the orca executable cannot be found on the path:

```
The orca executable is required in order to export figures as static images,
but it could not be found on the system path.

Searched for executable 'orca2' on the following path:
    /anaconda3/envs/plotly_dev/bin
    /usr/local/bin
    /usr/bin
    /bin
    /usr/sbin
    /sbin
    /Applications/VMware Fusion.app/Contents/Public
    /Library/TeX/texbin

If you haven't installed orca yet, you can do so using conda as follows:

    $ conda install -c plotly plotly-orca

After installation is complete, no further configuration should be needed. 
For other approaches to installing orca, see the orca project README at
https://github.com/plotly/orca.

If you have installed orca, then for some reason plotly.py was unable to
locate it. In this case, set the `plotly.io.orca.config.executable`
property to the full path to your orca executable. For example:

    >>> plotly.io.orca.config.executable = '/path/to/orca'

After updating this executable property, try the export operation again.
If it is successful then you may want to save this configuration so that it
will be applied automatically in future sessions. You can do this as follows:

    >>> plotly.io.orca.config.save() 

If you're still having trouble, feel free to ask for help on the forums at
https://community.plot.ly/c/api/python
```
## Testing
I've added two new test suites.

 - `plotly/tests/test_orca/test_orca_server.py`. These tests cover the logic for locating and validating the orca executable. And the logic for launching and shutting it down. This testing relies on `psutil` to check that the process with the right `pid` is running and then not running. And it relies on pinging the server to make sure it's running on the right port, and that it stops responding when it should be shut down.

 - `plotly/tests/test_orca/test_to_image.py`. These tests cover the image conversion logic. I've generated a set or reference images to compare against. These ensure that valid images are produces where they should be, and that the topojson and mathjax configuration is working properly. Unfortunately, the images are not exactly reproducible between my local mac and CircleCI, so for the time being there is a separate directory of reference images for OS X and Linux (though I'm not sure `Linux` is fine grained enough).

These tests are working on CircleCI. The new tests follow a new conda environment pathway so that orca can be installed using conda. The tests are run with Python 2.7, 3.5, and 3.7.

## Performance
The whole reason for using this more complex client/server architecture is to improve image export performance. So how well does it do?

This is not an extensive performance comparison, but I did an initial comparison of matplotlib, this branch, and bokeh ([setup instructions](https://bokeh.pydata.org/en/latest/docs/user_guide/export.html)).  The test was to create a 1000 point scatter plot with varying point size and color and then save it to a png.

screen shot 2018-08-20 at 7 46 33 pm

So after the orca server is running, the export time here is right on par with matplotlib (~215ms), and much faster than bokeh (~1.7s).

Being on par with matplotlib here is really exciting, and opens up a lot of new use cases for plotly.py. I'm thinking, in particular, of the possibility of a static image backend for interactive use outside of the notebook/browser context.

**Side note:** bokeh isn't doing any wrong here. This is just how expensive it is to launch a web browser from scratch. This is also about how long it takes the orca server to start up the first time. The advantage with this orca approach is that the server only needs to start up once per session, instead of once per image.

### Produced images
And here are the images produced by matplotlib, this branch, and bokeh
screen shot 2018-08-20 at 7 47 04 pm
screen shot 2018-08-20 at 7 47 13 pm
screen shot 2018-08-20 at 7 47 21 pm

## TODO
Various things still to do/look into:
 - [x] Add `validate` option to `to_image` and `write_image`
 - [x] Look into validating poppler installation, or providing better error message on eps failure.


chriddyp:
👏 Looks great @jonmmease ! I really appreciate the extensive API and architecture notes.

> In this PR I went with method (3) from the issue above, "Use orca in server mode".
> The shutdown function is annotated with the @atexit.register decorator to ensure that the server is properly shutdown when the main Python process exits.
> Each time an image render request is made, any existing Timer object is canceled, and a new Timer is created.

👍 👍 👍 

> This PR introduces the beginning of the plotly.io module.

👍 

> If users install orca using conda or npm,

_or_ the executables, right? When the executable gets installed, will its path be found by `plotly.io`?

I'd also be interested in eventually exploring creating pip packages that embeds the built executables as package data. Something like:
```
pip install orca[windows]
pip install orca[linux]
pip install orca[mac]
```
as an alternative for our community that can't use conda or npm. 

***

```
Searched for executable 'orca2' on the following path:
    /anaconda3/envs/plotly_dev/bin
    /usr/local/bin
    /usr/bin
    /bin
    /usr/sbin
    /sbin
    /Applications/VMware Fusion.app/Contents/Public
    /Library/TeX/texbin
```

_so nice_

***

```
If you haven't installed orca yet, you can do so using conda as follows:

    $ conda install -c plotly plotly-orca

After installation is complete, no further configuration should be needed. 
For other approaches to installing orca, see the orca project README at
https://github.com/plotly/orca.
```

Could we make a slight modification to this error message to highlight the alternative methods?
```
If you haven't installed orca yet, you can do so using conda as follows:

    $ conda install -c plotly plotly-orca

Alternatively, see other installation methods in the orca project README at
https://github.com/plotly/orca.

After installation is complete, no further configuration should be needed.
```

> Being on par with matplotlib here is really exciting, and opens up a lot of new use cases for plotly.py. I'm thinking, in particular, of the possibility of a static image backend for interactive use outside of the notebook/browser context.

✨ 🌔  🌆 _a new era_ 🌆 🌔 ✨ 

***

💃 from me!

jonmmease:
Thanks for the feedback @chriddyp!

> or the executables, right? When the executable gets installed, will its path be found by plotly.io?

This isn't so clear cut at the moment. The Mac OS installer attempts to add an orca wrapper to `/usr/local/bin/`, but this approach can run into [permissions issues](https://github.com/plotly/orca/issues/111). The Windows installation instructions end with a manual step to add the installation directory to the `PATH`. And the Linux instructions are basically, put the AppImage on your `PATH`. Also, the AppImage isn't just named `orca`, because it has version info in the name, so at the moment this won't be discovered automatically even if it's on the path.

So, I'm a bit concerned that people are going to run into trouble when trying the non-conda/npm approaches. But I'm not really sure what do about this over the next couple of weeks.

> Could we make a slight modification to this error message to highlight the alternative methods?

👍 

> I'd also be interested in eventually exploring creating pip packages that embeds the built executables as package data

This would be really nice. But I'd be a little concerned about making the executable package data, since people could end up accidentally installing orca for the wrong OS (if I understand this approach correctly).  I'd be really interested to learn about any existing examples of non-python, native, command line executables being distributed as pip packages.

jonmmease:
Alright, time to merge this thing!

PR #1126, opened 2018-08-21
adding resize

timkpaine:
tmp

### DIFF
--- a/js/src/Figure.js
+++ b/js/src/Figure.js
@@ -741,7 +741,25 @@ var FigureView = widgets.DOMWidgetView.extend({
         var initialTraces = _.cloneDeep(this.model.get("_data"));
         var initialLayout = _.cloneDeep(this.model.get("_layout"));
 
-        Plotly.newPlot(that.el, initialTraces, initialLayout).then(
+        let holder = document.createElement('div');
+        holder.classList.add('js-plotly-plot-inner');
+        holder.style.height = '100%';
+        holder.style.width = '100%';
+
+        that.plot = holder;
+
+        that.el.appendChild(holder);
+        that.el.style.resize = 'both';
+        that.el.style.paddingBottom = '10px';
+
+        let observer = new MutationObserver(that.autosizeFigure.bind(that));
+        observer.observe(this.el, {attributes: true});
+
+        that.el.addEventListener('resize', function(){
+            that.autosizeFigure();
+        });
+
+        Plotly.newPlot(that.plot, initialTraces, initialLayout).then(
             function () {
 
                 // ### Send trace deltas ###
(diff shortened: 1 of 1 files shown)


timkpaine:
@jonmmease can you check the one CI, its js only so not sure why one would fail

js/src/Figure.js
```diff
@@ -741,7 +741,25 @@ var FigureView = widgets.DOMWidgetView.extend({
         var initialTraces = _.cloneDeep(this.model.get("_data"));
         var initialLayout = _.cloneDeep(this.model.get("_layout"));
 
-        Plotly.newPlot(that.el, initialTraces, initialLayout).then(
+        let holder = document.createElement('div');
+        holder.classList.add('js-plotly-plot-inner');
```
jonmmease: Is the `'js-plotly-plot-inner'` class associated with anything else? Or just a class for future use?

js/src/Figure.js
```diff
@@ -741,7 +741,25 @@ var FigureView = widgets.DOMWidgetView.extend({
...
+        that.el.appendChild(holder);
+        that.el.style.resize = 'both';
+        that.el.style.paddingBottom = '10px';
+
+        let observer = new MutationObserver(that.autosizeFigure.bind(that));
+        observer.observe(this.el, {attributes: true});
+
+        that.el.addEventListener('resize', function(){
+            that.autosizeFigure();
+        });
```
jonmmease: Could you explain why the `MutationObserver` and `addEventListener` approaches are both needed here? And add the explanations as comments 🙂

js/src/Figure.js
timkpaine: Yeah this is what the plotly object binds to, in case you want to add any additional controls or anything to the container (like bqplot and Altair have some buttons)

js/src/Figure.js
timkpaine: Mutationobserver monitors for changes in the element, while resize events are only thrown for window size changes (not for element size changes). When the window resizes, the element can just overflow (which we don't want) so we want both the element and the window to trigger resizing if they change size

jonmmease:
Thanks for the PR @timkpaine! I really like the workflow of being able to resize figures with the mouse.  I'll try to pull down the branch and play with it tomorrow, but a couple of questions in the meantime.

 1. How does this interact with the current auto-resize logic that runs when a window (or JupyterLab pane) is resized?  Does the dragged size "stick" and stay fixed during window resize? or does the window resize take over?

 2. What is the behavior when when the figure's `layout.width` and `layout.height` properties are set to fixed values?

 3. Do you know of a way to make the resize handle behave like the figure toolbar, where it's hidden until the div is hovered over? If this isn't possible, we'll need some way to disable/hide it for cases where people don't want to see the handle.

I reran the failed test and it's all green now. The `plot_ly` test jobs are a little flaky at times.

nicolaskruchten:
I'd love to see this get rolled in :)

Alexboiboi:
Hi, any chance this gets rolled in at some point ?

nicolaskruchten:
We'll try to get this in at some point, sorry for the looong delay. The way we handle figure width in JupyterLab is awkward at the moment in that we override the figure width settings :(

Issue #1129, opened 2018-08-23
Plotly stream + corporate proxy with basic auth


imscaradh:
I am trying to use Plotly streaming functionality behind a corporate proxy with basic auth. Despite the comprehensible instructions at https://plot.ly/python/proxy-configuration/ I didn't get Plotly up and running. After digging around in my system configuration I noticed that only proxy servername and port is respected while setting up the http tunnel:

https://github.com/plotly/plotly.py/blob/dbb79e43e2cc6c5762251537d24bad1dab930fff/plotly/plotly/chunked_requests/chunked_request.py#L138

Was this intentionally implemented in this way? As soon as I add the `Proxy-Authorization` header to the tunnel configuration everything works like expected.

jonmmease:
Hi @imscaradh , thanks for reaching out and I'm glad you got things working.

@scjody do you have any thoughts on this scenario, and whether something in plotly.py should be changed?

scjody:
@jonmmease It does look like we only use the hostname and port extracted from the `HTTPS_PROXY` or `HTTP_PROXY` environment variable, meaning that any other information (username, password) is ignored.

The hostname and port are extracted in https://github.com/plotly/plotly.py/blob/dbb79e43e2cc6c5762251537d24bad1dab930fff/plotly/plotly/chunked_requests/chunked_request.py#L79-L103 and used here: https://github.com/plotly/plotly.py/blob/dbb79e43e2cc6c5762251537d24bad1dab930fff/plotly/plotly/chunked_requests/chunked_request.py#L128-L138

This could be extended to support performing basic auth with the proxy.

imscaradh:
Hi @jonmmease , I forked plotly.py and committed my changes here: 4c38dc5e1e2abfa8810fc84073e81477737e2bb5

I'm not sure this is in your interest. If it is, I will be happy to create a PR.

jonmmease:
Hi @imscaradh , these changes look reasonable and generally useful so a PR would be great!

I don't have a way to test it personally, but I'll reach out to some other folks to help out there.

Thanks!

jonmmease:
Done in #1133. Thanks @imscaradh

PR #1133, opened 2018-08-24
Using basic auth for proxy tunnel

imscaradh:
This is a PR for using Plotly behind a corporate firewall with basic authentication. Although server and port were respected, the basic auth string inside environment variables like `$http_proxy` is ignored at this stage.

(My question here: https://github.com/plotly/plotly.py/issues/1129)


jonmmease:
Thanks @imscaradh!

I finally got a couple of VMs setup in a proxy configuration to try this out, and it works great!  I'm going to follow up with a commit to support the upper case versions of each of the proxy environment variables, since that tripped me up a bit.

Issue #1125, opened 2018-08-21
annotations deprecated, but not in right place


timkpaine:
Contradictory errors:
```
plotly.graph_objs.Annotations is deprecated.
Please replace it with a list or tuple of instances of the following types
  - plotly.graph_objs.layout.Annotations
```

```
ImportError: cannot import name 'Annotations' from 'plotly.graph_objs.layout' (/usr/local/lib/python3.7/site-packages/plotly/graph_objs/layout/__init__.py)
```

jonmmease:
Fixed in  #1138

PR #1138, opened 2018-08-29
Fix deprecation warning for go.Annotations.

jonmmease:
Fix for #1125.
There is no `go.layout.Annotations` (it should not have been plural)

Thanks @timkpaine


timkpaine:
thanks!

Issue #1108, opened 2018-08-13
stream.write raises error in plotly 3.1.0


Kully:
A simple streaming example that was working in Plotly 2.2 now does not work in 3.1.0+

```
import time
import plotly.plotly as py
from plotly.graph_objs import Data, Scatter, Stream, Bar

stream_id = '3luati1sxe'


py.plot(Data([Bar(x=[1], y=[1], stream=Stream(token=stream_id, maxpoints=60))]))

stream = py.Stream(stream_id) # Initialize a stream object
stream.open() # Open the stream

for i in range(1000):
    m = stream.write(dict(x=[i+2], y=[5]))
    print "i: ", i
    time.sleep(2)
    print "Sent"
```

results in the validation error

```
ValueError: The 'type' property of scatter is read-only
```

Changing `m = stream.write(dict(x=[i+2], y=[5]))` to `m = stream.write(Bar(x=[i+2], y=[5]))` bypasses the problem, allowing the script to run and print the statements. But the figure is not appending to its data, but replacing each Bar trace that is added as evidenced by the (i)plot that updates.

cc @jonmmease @tarzzz

jonmmease:
In 3.1.1 I made a change allowing the `'type'` property of a trace to be set as long as it's set to the correct value.  When I try this example on master I get a different error.

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-747009c91fc0> in <module>()
     12 
     13 for i in range(1000):
---> 14     m = stream.write(dict(x=[i+2], y=[5]))
     15     print("i: ", i)
     16     time.sleep(2)

~/Plotly/repos/plotly.py/plotly/plotly/plotly.py in write(self, trace, layout, reconnect_on)
    692                         "Here's why:\n\n{0}".format(err)
    693                     )
--> 694         del stream_object['type']
    695 
    696         if layout is not None:

AttributeError: __delitem__
```

This error can be fixed by converting the trace to a dict before removing the `'type'` key. But as you observed, even in this case the data is replaced, not appended to on each write.

The reason that the data is overwriting rather than appending seems to be related to the fact that the new `x` and `y` values are lists rather than scalar values.  If I remove the trace validation and change the write line to `m = stream.write(dict(x=i+2, y=5))` the data is appended.

I suspect the version 2 would also have the overwriting behavior if `x` and `y` are set to lists.

[...]

PR #1145, opened 2018-08-31
Remove stream write validation to avoid incorrect validation failures

jonmmease:
Fix for #1108

Removed validation since stream writes don't actually have the same validation rules as traces.

cc @gnthibault


jonmmease:
@Kully , I'm going to go ahead and merge this so it's on master for testing over the next few days.  Please still take a glance if you have a chance over the next couple of days.

PR #1148, opened 2018-09-01
Fixing docstring for manually converting from mpl to plotly

kmader:
updating documentation for the manual process of converting from matplotlib to plotly


jonmmease:
Thanks for taking the time to send this in @kmader!

PR #1178, opened 2018-09-14
Use latest plotly.js

fcollonval:
Is there a reason to lock the version of plotly.js used?

This results in a WARNING when building Jupyterlab:

```
WARNING in plotly.js
  Multiple versions of plotly.js found:
    1.40.1 ./~/plotly.js from ./~/plotly.js\src\components\colorscale\get_scale.js
    1.41.0 ./~/@jupyterlab\plotly-extension/~/plotly.js from ./~/@jupyterlab\plotly-extension\lib\index.js
```

Reference:
- Jupyterlab 0.34.8
- plotly.py 3.2.1
- plotlywidget@0.3.0 
- @jupyterlab/plotly-extension@0.17

Nota: I did not try it yet

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -31,7 +31,7 @@
     "ify-loader": "^1.1.0"
   },
   "dependencies": {
-    "plotly.js": "1.40.1",
+    "plotly.js": "^1.40.1",
     "@jupyter-widgets/base": "^1.0.0",
     "lodash": "^4.17.4"
   },


jonmmease:
Hi @fcollonval ,

We lock the plotly.js version constraint for `FigureWidget` because the Python library is code generated off of the schema for that specific version.  This is admittedly a bit conservative since the plotly.js project does work hard to keep everything backwards compatible, but there are still cases where it could cause an issue. For example, if Plotly.js fixes a bug that plotly.py had worked around then the release of a new version of plotly.js could break `FigureWidget` mid cycle (though I suppose that is just the nature of bounded dependency resolution)

In any case, there are already so many moving parts in getting the JupyterLab installation working that I'd prefer to not add another free variable that can change between releases.

I am a bit confused about why this warning is issued though. `plotlywidget` states a dependency on plotly.js version `"1.40.1"` and `@jupyterlab/plotly-extension@0.17` states a dependency of `"^1.39.2"`.  Do you have any idea why this doesn't resolve to 1.40.1?

Also, the installation still succeeds and I'm able to use both `FigureWidget` and `iplot` but I'm not sure how to check what version of plotly.js was actually installed.

fcollonval:
@jonmmease Thanks for the reply.

> Do you have any idea why this doesn't resolve to 1.40.1?

Unfortunately packaging JS code looks like black magic for me... But when looking at the build, there is something weird. In the top plotly.js library (the second one on the picture below), there are separated files in addition to the distributed `dist/plotly.js` file. May the troubles come from the content difference between the two packages?

image

jonmmease:
Interesting. @gnestor if you have a moment, do you have any insight on what might be going on here?

jonmmease:
@blink1073 or @ian-r-rose if you have a moment, do you have any thoughts on this situation?

`plotlywidget` 0.3.0 states a dependency on plotly.js version "1.40.1" and `@jupyterlab/plotly-extension@0.17.1` states a dependency on "^1.39.2". I would expect this to resolve to a single dependency on "1.40.1". But it looks like both "1.40.1" and "1.41.0" are getting installed.  It doesn't seem to cause a problem, in that both extensions are working, but it does concern me a bit.

Thanks!

blink1073:
Getting installed in the JupyterLab app directory?  If so it sounds like a bug in yarn, cf https://github.com/yarnpkg/yarn/issues/6070

jonmmease:
Thanks for the cross reference @blink1073 , that does look like what's going on.  Any thoughts on a JupyterLab workaround?  The issue suggests deleting the nodemodule directory and lockfile and reinstalling, do you know if there are any JupyterLab commands that could do this under the hood?

blink1073:
`jupyter lab clean` removes the whole `staging` directory

jonmmease:
Closing now that the plotly.js dependency has become a lower bound for both the `plotlywidget` (for `FigureWidget`) and `jupyterlab-plotly` (for `fig.show()`) extensions.

Issue #1181, opened 2018-09-17
wrong leaves determination of _dendrogram.py


444thLiao:
When I using the function for creating dendrogram, I meet some wrong presentation of the leaves.
![all zero leaves](https://user-images.githubusercontent.com/16633223/45608688-e5e25700-ba86-11e8-8e20-9ee3e28680dd.png)

In your code(**~/.local/lib/python3.6/site-packages/plotly/figure_factory/_dendrogram.py**), we can see that 
![original code which come from ~/.local/lib/python3.6/site-packages/plotly/figure_factory/_dendrogram.py](https://user-images.githubusercontent.com/16633223/45608710-0e6a5100-ba87-11e8-931f-508bd7320e4c.png)

Above code without comment is the original code for determination the y axis tick or labels. In the first image I provide, you can see that 138-140 line code will **raise silent error** which mean wrong alignment of leaves of dendrogram and labels. 

So I make a litter bit **revise** of this by 142-145 code (Of course without comment and I have use this in my computer, as far as I see, it will not raise any conflict.) .

Here is the  code
```
        if len(self.zero_vals) > len(yvals) + 1:
            l_border = int(min(self.zero_vals))
            r_border = int(max(self.zero_vals))
            self.zero_vals = [v for v in range(l_border,r_border + 1, int((r_border-l_border) / len(yvals)))]
```

#1180

jonmmease:
Hi @444thLiao , thanks for the report. I left some comments over in #1180 , so we can continue the conversation over there.

jonmmease:
Done in #1186. Thanks @444thLiao!

PR #1180, opened 2018-09-17
For wrong alignment of the leaves of dendrogram 

444thLiao:

PR #1186, opened 2018-09-18
fix wrong leaves determination bug

444thLiao:
@jonmmease Sorry for multiple pull request beacuse I am new to github cooperation.

Here is the formal request with code and example.

Before, run the code below
```
import plotly
import pandas as pd
demo_data = pd.DataFrame([[1,2,3,4],[1,2,3,4],[1,3,5,6],[1,4,2,3]])
figure = ff.create_dendrogram(demo_data)
plotly.offline.plot(figure)
```
demo_data looks like 
image
The data is some simple data and it mainly simulated the situation which has three and more rows are identical. When we using the euclidean distance, the distance between 0,1 and 3 will be identical.

we could get the dendrogram like that
wrong dendrogram

In the graph above, we can see wrong multiple x-axis ticks text raised. It will be severely when we pass the leaves name. 35 in the graph is the original x-coordinate of the leaves.

After, run the same code above.
after bug fixed

Relative issue and request: 
#1180 
#1181

### DIFF
--- a/plotly/figure_factory/_dendrogram.py
+++ b/plotly/figure_factory/_dendrogram.py
@@ -140,7 +140,10 @@ class _Dendrogram(object):
         for i in range(len(yvals_flat)):
             if yvals_flat[i] == 0.0 and xvals_flat[i] not in self.zero_vals:
                 self.zero_vals.append(xvals_flat[i])
-
+        if len(self.zero_vals) > len(yvals) + 1:
+            l_border = int(min(self.zero_vals))
+            r_border = int(max(self.zero_vals))
+            self.zero_vals = [v for v in range(l_border,r_border + 1, int((r_border-l_border) / len(yvals)))]
         self.zero_vals.sort()
 
         self.layout = self.set_figure_layout(width, height)


plotly/figure_factory/_dendrogram.py
```diff
@@ -140,7 +140,10 @@ def __init__(self, X, orientation='bottom', labels=None, colorscale=None,
         for i in range(len(yvals_flat)):
             if yvals_flat[i] == 0.0 and xvals_flat[i] not in self.zero_vals:
                 self.zero_vals.append(xvals_flat[i])
-
```
jonmmease: Let's add a nice comment here explaining why this check is here. (e.g. the situation with multiple identical values)

plotly/figure_factory/_dendrogram.py
```diff
@@ -140,7 +140,10 @@ def __init__(self, X, orientation='bottom', labels=None, colorscale=None,
...
+        if len(self.zero_vals) > len(yvals) + 1:
+            l_border = int(min(self.zero_vals))
+            r_border = int(max(self.zero_vals))
+            self.zero_vals = [v for v in range(l_border,r_border + 1, int((r_border-l_border) / len(yvals)))]
```
jonmmease: Please break this line up into two lines so that it's not so wide. Maybe create a new local variable for the `range(...)` expression.

444thLiao:
I have added comments and add a test function at suitable places. Maybe now it is suitable to merge into the master branch.

jonmmease:
Thanks @444thLiao!  I think these two test failures are due to release of matplotlib 3.0, once I confirm that I'll merge these changes in.

jonmmease:
@444thLiao Almost there. Matplotlib tests are fixed on master. Please merge master once more and push again.  Thanks!

444thLiao:
Does it have some problems? I am not familiar with the circleci. @jonmmease

Issue #1187, opened 2018-09-19
The length of the arrows are not maintained by the 2D quiver plot in figure factory.


dabana:
### Bug report

Thanks to the Plotly community for this great implementation of 2D quiver plots in the figure factory. The issue is that the length of the arrows are not maintained by the quiver plot. 

**Bug summary**

The length of the arrows in a quiver plot carries relevant information in most applications (wind speed, force field intensity, etc.). This seem to be the case with the quiver plot in figure factory if both axis have the same scale (for example if quiver plot is plotted on a map: x and y coordinates have obviously the same scale). However, as soon as the scales of both axis are different, the apparent length of the arrows is distorted. The barb angle is also affected. This should not be the case.

**Code for reproduction**

This is an instance of a quiver plot where the scales of both axis are different but where the length of each arrow is intended to be the same (using the norm and an angle to produce u an v).

```python
import plotly.plotly as py
import plotly.figure_factory as ff
import plotly
print(plotly.__version__)
import numpy as np

x,y = np.meshgrid(np.arange(0.5, 3.5, .5), np.arange(0.5, 4.5, .5))
u = x
v = y
angle = np.arctan(v / u)
norm = 0.25
u = norm * np.cos(angle)
v = norm * np.sin(angle)

fig = ff.create_quiver(x, y, u, v, scale = 1)
```

**Actual outcome**

The above code produces arrows with apparent lengths changing depending on their angle. This should not be the case.

![fig1](https://user-images.githubusercontent.com/29574757/45758771-64a8d280-bbf4-11e8-8c82-b7570765ef7f.png)

**Expected outcome**

This is an instance of the above quiver plot but where the scales of both axis are forced to be the same using specifications on the layout.

```python
x,y = np.meshgrid(np.arange(0.5, 3.5, .5), np.arange(0.5, 4.5, .5))
u = x
v = y
angle = np.arctan(v / u)
norm = 0.25
u = norm * np.cos(angle)
v = norm * np.sin(angle)

fig = ff.create_quiver(x, y, u, v, scale = 1)
fig.layout = go.Layout(
    yaxis=dict(
        scaleratio = 1,
        scaleanchor = "x"
    ))
```
![fig2](https://user-images.githubusercontent.com/29574757/45758786-6e323a80-bbf4-11e8-97f7-770ac85bab8d.png)

It would be nice to have nicely scaled arrows as the above plot but with axis of different scales.

[...]

PR #1197, opened 2018-09-24
Allow forcing a fix axis scale ratio to maintain arrow lengths

dabana:
Addresses issue #1187.

A new `scaleratio` parameter is added to the `create_quiver` function to allow forcing a fix scale ratio to the axis of the plot through the `scaleratio` option in the Layout. A distortion is introduced to the arrows length, angle and shape by having a scale ratio different than 1. In this pull request, this distortion is reversed by multiplying or dividing by `scaleratio` at key steps of the arrows construction.

### DIFF
--- a/optional-requirements.txt
+++ b/optional-requirements.txt
@@ -16,6 +16,7 @@ coverage==4.3.1
 mock==2.0.0
 nose==1.3.3
 pytest==3.5.1
+backports.tempfile==1.0
 
 ## orca ##
 psutil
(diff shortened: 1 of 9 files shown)


dabana:
There were no test class for the quiver plot, so I created it. But when trying out the tests with Nose:

```
nosetests -w plotly/tests/test_optional/test_figure_factory
```
I get the following import error:
```
ModuleNotFoundError: No module named 'retrying'
```
even though the `retrying` module is installed in my environment.

jonmmease:
Thanks @dabana, this looks great!  Thanks for the test, docstring, and example. I tried out several other examples and this makes the interpretation of the arrow lengths much more accurate.

Also, let me know if there's anything else you're interested in doing to improve `create_quiver`, it's so helpful to have contributors who also use the figure factory in their work 🙂

dabana:
Great! This PR is a good first step toward issue #1187. I would like to make the choice of `scaleratio` a little more automatic for the user.

PR #1213, opened 2018-10-03
ignore divide error in streamline

Kully:
Simple fix to warning that pops up when streamline code divides 0 by 0.

See  https://plot.ly/python/streamline-plots/#basic-streamline-plot for an example of the warning.

### DIFF
--- a/plotly/figure_factory/_streamline.py
+++ b/plotly/figure_factory/_streamline.py
@@ -354,8 +354,11 @@ class _Streamline(object):
         dif_x = arrow_end_x - arrow_start_x
         dif_y = arrow_end_y - arrow_start_y
 
+        orig_err = np.geterr()
         np.seterr(divide='ignore', invalid='ignore')
         streamline_ang = np.arctan(dif_y / dif_x)
+        np.seterr(**orig_err)
+
 
         ang1 = streamline_ang + (self.angle)
         ang2 = streamline_ang - (self.angle)
--- a/plotly/figure_factory/_streamline.py
+++ b/plotly/figure_factory/_streamline.py
@@ -354,6 +354,7 @@ class _Streamline(object):
         dif_x = arrow_end_x - arrow_start_x
         dif_y = arrow_end_y - arrow_start_y
 
+        np.seterr(divide='ignore', invalid='ignore')
         streamline_ang = np.arctan(dif_y / dif_x)
 
         ang1 = streamline_ang + (self.angle)


plotly/figure_factory/_streamline.py
```diff
@@ -354,6 +354,7 @@ def get_streamline_arrows(self):
         dif_x = arrow_end_x - arrow_start_x
         dif_y = arrow_end_y - arrow_start_y
 
+        np.seterr(divide='ignore', invalid='ignore')
```
jonmmease: Thanks @Kully!  I'd like to see if we can also restore the original numpy error state. Could we do something like:

```python
orig_err = np.geterr()
np.seterr(divide='ignore', invalid='ignore')
streamline_ang = np.arctan(dif_y / dif_x)
np.seterr(**orig_err)
```

Kully: I like that idea. Let's try it.

Kully:
it works as expected - waiting for tests to pass

jonmmease:
Great! 💃

empet:
Hey @jonmmease and @Kully,

In scientific computation  it is recommended to use `arctan2` instead of `arctan` to compute angles.
If you need that angle to get the the polar coordinates of the point(x,y), then `arctan` doesn't give the right
angle (see for example x=-3, y=np.sqrt(3), arctan(y/x)=-0.5235 radians i.e. -30 degrees ). But this point lies in the second quadrant, hence its polar angle cannot be -30 degrees.

On the other hand,  arctan2(y,x)=2.6179, i.e. 150 degrees. Hence this is  the right polar angle.

`arctan` has the range (pi/2, pi/2), while `arctan2`, (-pi, pi]. It is calculated
by the rule illustrated in this figure.

arctan2


Example:
```
x=np.array([0, 0, 3, -3, -3, 0])
y=np.array([0, 1, np.sqrt(3), np.sqrt(3), -np.sqrt(3), -1])
print(np.arctan(y/x))
print(np.arctan(y/x)*180/np.pi)

[        nan  1.57079633  0.52359878 -0.52359878  0.52359878 -1.57079633]
[ nan  90.  30. -30.  30. -90.]

radians=np.arctan2(y, x)
print(radians)
print(radians*180/np.pi)

[ 0.          1.57079633  0.52359878  2.61799388 -2.61799388 -1.57079633]
[   0.   90.   30.  150. -150.  -90.]

Issue #1227, opened 2018-10-16
Raise warning in plotly.offline.plot when output_type='file' and include_plotlyjs=False


capelastegui:
When calling plotly.offline.plot(), if we use output_type='file' and include_plotlyjs=False, the resulting html file is useless, as it cannot be rendered.

I'm using plotly 3.3.0.

```python
import plotly
import plotly.graph_objs as go

#This generates a file that includes plotly.js, and works as intended
plotly.offline.plot([go.Scatter(x=[1, 2, 3], y=[3, 2, 6])], 
                    filename='my-graph.html', output_type='file',
                   include_plotlyjs=True)

#This generates a file that doesn't include plotly.js, but fails to render
plotly.offline.plot([go.Scatter(x=[1, 2, 3], y=[3, 2, 6])], 
                    filename='my-graph-nojs.html', output_type='file', 
                    include_plotlyjs=False)
```

According to the documentation:
```
    include_plotlyjs (default=True) -- If True, include the plotly.js
        source code in the output file or string.
        Set as False if your HTML file already contains a copy of the plotly.js
        library.
```

I guess we are not supposed to use include_plotly_js=False and output_type='file' at the same time? If that is the case, the function should probably raise an error, rather than generating a useless file. However, I think the ideal solution would be to have the standalone html load plotly.js from an online CDN, like we do in Jupyter notebooks after calling `init_notebook_mode(connected=False)`. So we would insert something like the following block in the html file:

```
<script src="https://cdn.plot.ly/plotly-latest.min.js"></script>
```

jonmmease:
Hi @capelastegui , thanks for the suggestion.

Perhaps `include_plotlyjs` should be allowed to be set to `True`, `False`, or `'cdn'`.

jonmmease:
Hi @capelastegui, 

This issue was part of the motivation for the #1234 PR.  Here is the updated docstring for `include_plotlyjs`

```
    include_plotlyjs (True | False | 'cdn' | 'directory' | path - default=True)
        Specifies how the plotly.js library is included in the output html
        file or div string.

        If True, a script tag containing the plotly.js source code (~3MB)
        is included in the output.  HTML files generated with this option are
        fully self-contained and can be used offline.

        If 'cdn', a script tag that references the plotly.js CDN is included
        in the output. HTML files generated with this option are about 3MB
        smaller than those generated with include_plotlyjs=True, but they
        require an active internet connection in order to load the plotly.js
        library.

[...]

Issue #1043, opened 2018-06-28
Resizable, minimal, embeddable output from plotly.offline.plot


mchels:
In plotly.offline.plot is it intentional that resizing is only available for `output_type == 'div'` if plotly.js is included in the output: https://github.com/plotly/plotly.py/blob/master/plotly/offline/offline.py#L526-L537

This means that currently you cannot generate a minimal (no plotly.js), embeddable (`output_type == 'div'`) script that is also resizable which seems not to be the intention in the PR that made the change: https://github.com/plotly/plotly.py/pull/717

Example:
```
import plotly
import plotly.graph_objs as go

figure_or_data = {
    "data": [go.Scatter(x=[1, 2, 3, 4], y=[4, 3, 2, 1])],
    "layout": go.Layout(title="hello world"),
}
fig_div = plotly.offline.plot(
    figure_or_data,
    image_width='100%',
    image_height='100%',
    include_plotlyjs=False,
    output_type='div',
    auto_open=False,
)
```
gives for `fig_div`:
```
'<div id="0b3aff81-1632-4ea9-a11f-a6b0b1cd02c1" style="height: 100%; width: 100%;" class="plotly-graph-div"></div><script type="text/javascript">window.PLOTLYENV=window.PLOTLYENV || {};window.PLOTLYENV.BASE_URL="https://plot.ly";Plotly.newPlot("0b3aff81-1632-4ea9-a11f-a6b0b1cd02c1", [{"type": "scatter", "x": [1, 2, 3, 4], "y": [4, 3, 2, 1]}], {"title": "hello world"}, {"showLink": true, "linkText": "Export to plot.ly"})</script>'
```
Setting `include_plotlyjs=True` gives at the end
```
<script type="text/javascript">window.addEventListener("resize", function(){Plotly.Plots.resize(document.getElementById("ba33b0b5-156d-4224-b931-414fcf0f02f1"));});</script></div>'
```

Apologies if I'm missing something obvious.

ryancollingwood:
With some help from [this stack overflow post](https://stackoverflow.com/a/38033016/2805700) found a workaround for this.

Basically let it generate the html, however suppress the automatic opening with `auto_open = False`.
Then load the html generated from the offline plot and insert the script to the latest plotly javascript.

```python
import plotly
import bs4

# return as html fragment
# the include_plotlyjs argument seems to be 
# ignored as it's included regardless when outputting to div
[...]

PR #1234, opened 2018-10-22
offline.plot/iplot improvements

jonmmease:
### Overview
This PR includes a variety of small improvements to html generation in `plotly.offline.plot` and `plotly.offline.iplot`.

### Changes
 - Add support for including the auto-resize logic in the HTML div output by `offline.plot` when `output_type='div'` and `include_plotlyjs=False`. See #1043 for discussion. cc @mchels @ryancollingwood @spencercal
 - Adds resize support for iplot use in the classic notebook. This PR now superceeds #970.
 - Adds `'cdn'`, `'directory'`, and path string `include_plotlyjs` options in `offline.plot`
    - When `'cdn'`, the resulting html file/div includes a script tag reference to the plotlyjs cdn. This addresses part of #1227
    - When `'directory'`, the resulting html file/div includes a script tag reference to a plotly.min.js bundle in the same directory as the html file. If output_type is `'file'` then this plotly.min.js bundle is created in the output directory if it doesn't already exist. This is an alternative approach towards the goal of #688
    - When a string ending with `'.js'`, the resulting html file/div includes a script tag that references this exact path. This can be used to point to a plotly.js bundle from an alternative CDN.


jonmmease:
1494538 Adds resize support for `iplot` use in the classic notebook. This PR now superceeds #970.

jonmmease:
0cbcce7 adds `'cdn'` and `'directory'` `include_plotlyjs` options in `offline.plot`

When `'cdn'`, the resulting html file/div includes a script tag reference to the plotlyjs cdn.  This addresses part of #1227

When `'directory'`, the resulting html file/div includes a script tag reference to a plotly.min.js bundle in the same directory as the html file. If `output_type` is `'file'` then this plotly.min.js bundle is created in the output directory if it doesn't already exist. This is an alternative approach towards the goal of #688

jonmmease:
69dc7ba  Adds option to specify `include_plotlyjs` as a path/url to a *.js file.  This makes it possible to point the resulting html file/div to an alternative CDN location.

fu:
very nice! Thanks for picking this one up 👍

plotly/offline/offline.py
```diff
@@ -389,6 +394,17 @@ def iplot(figure_or_data, show_link=True, link_text='Export to plot.ly',
         ipython_display.display(ipython_display.HTML(script))
 
 
+def _build_resize_script(plotdivid):
+    resize_script = (
+        ''
```
jonmmease: Remove empty string here

plotly/tests/test_core/test_offline/test_offline.py
```diff
@@ -21,14 +21,29 @@
...
+
+directory_script = '<script src="plotly.min.js"></script>'
+
 
 class PlotlyOfflineBaseTestCase(TestCase):
     def tearDown(self):
         # Some offline tests produce an html file. Make sure we clean up :)
         try:
             os.remove('temp-plot.html')
+            # Some tests that produce temp-plot.html]
```
jonmmease: remove ']' character

PR #1240, opened 2018-10-23
Add support for 2D arrays in InfoArrayValidator

### LINKED ISSUE #1220: Incorrect Validation Error for layout.grid.subplots attribute
I'm getting a validation `ValueError` when I try to set the `layout.grid.subplots` attribute. This attribute should accept a 2d list of the enumerated values ([see here](https://github.com/plotly/plotly.js/blob/master/src/components/grid/index.js#L54-L60)) but it is being limited to a single dimensional list of one element.
```
ValueError: 
    Invalid value of type 'builtins.list' received for the 'subplots' property of layout.grid
        Received value: [['xy', 'x2y'], ['xy3', 'x4y4']]

    The 'subplots' property is an info array that may be specified as a
    list or tuple of up to 1 elements where:

(0) The 'subplots[0]' property is an enumeration that may be specified as:
      - One of the following enumeration values:
            ['']
      - A string that matches one of the following regular expressions:
            ['^x([2-9]|[1-9][0-9]+)?y([2-9]|[1-9][0-9]+)?$']
```
Here is an example demonstrating the problem which works if I use dicts and `validate=False`

```python
import plotly.offline as py

py.init_notebook_mode()

trace1 = go.Scatter(
    x=[1, 2, 3],
    y=[2, 3, 4]
)
trace2 = go.Scatter(
    x=[20, 30, 40],
    y=[5, 5, 5],
    xaxis='x2',
    yaxis='y'
)
trace3 = go.Scatter(
    x=[2, 3, 4],
    y=[600, 700, 800],
    xaxis='x',
    yaxis='y3'
)
trace4 = go.Scatter(
    x=[4000, 5000, 6000],
    y=[7000, 8000, 9000],
    xaxis='x4',
    yaxis='y4'
)
data = [trace1, trace2, trace3, trace4]
layout = go.Layout(
    grid = go.layout.Grid(
        rows=2, 
        columns=2,
        subplots=[['xy','x2y'],['xy3','x4y4']], 
        )
    )
fig = go.Figure(data=data, layout=layout)
py.iplot(fig, filename='shared-axes-subplots')
```

jonmmease:
Supersedes #1233, which didn't recover from the GitHub outage:

----

Rework `InfoArrayValidator` to fix #1220

### DIFF
--- a/_plotly_utils/basevalidators.py
+++ b/_plotly_utils/basevalidators.py
@@ -1611,7 +1611,8 @@ class InfoArrayValidator(BaseValidator):
             ],
             "otherOpts": [
                 "dflt",
-                "freeLength"
+                "freeLength",
+                "dimensions"
             ]
         }
     """
@@ -1621,10 +1622,14 @@ class InfoArrayValidator(BaseValidator):
                  parent_name,
                  items,
                  free_length=None,
+                 dimensions=None,
                  **kwargs):
         super(InfoArrayValidator, self).__init__(
             plotly_name=plotly_name, parent_name=parent_name, **kwargs)
+
         self.items = items
+        self.dimensions = dimensions if dimensions else 1
+        self.free_length = free_length
 
         # Instantiate validators for each info array element
         self.item_validators = []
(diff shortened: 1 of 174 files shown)


saebaxp:
This line in InfoArrayValidator.validate_coerce() doesn't look like right:
        is_v_2d = v and is_array(v[0])
for methods like 'animate' expecting their 'args' containing a first argument as a list of frame names (but it's not a really 2d array as a whole, I believe), above code gives an unexpected error.

Changing it to:
        is_v_2d = v and all([is_array(vv) for vv in v])
corrected the issue.

Could you please give a look?

jonmmease:
That's a good point @saebaxp, I've opened https://github.com/plotly/plotly.py/issues/1265 to track this

PR #1243, opened 2018-10-24
Add MathJax / Latex support for offline plot/iplot/FigureWidget

jonmmease:
## Overview
This PR adds support for displaying figures with LaTeX typesetting using:
 - `plotly.offline.plot` for standalone HTML
 - `plotly.offline.iplot` inline in the classic notebook
 - `FigureWidget`:
   - Classic notebook 
   - JupyterLab

To be effective, these changes require unreleased changes in plotly.js master. In particular (https://github.com/plotly/plotly.js/pull/2994). These will be available as plotly.js 1.42 which will be included with plotly.py 3.4.0 release.

## `plotly.offline.plot` changes
A new `include_mathjax` argument has been added to `plotly.offline.plot`.  This argument mirrors the `include_plotlyjs` argument and supports a subset of its options.  It defaults to `False`, matching the existing behavior of not adding any MathJax related code to the resulting HTML file/div.  Setting it to `'cdn'` adds a script tag referencing a predefined CDN locaiton, and setting it to a string ending with `'.js'` allows the user to specify a custom CDN path to be included in the script tag.

Full docstring

```
    include_mathjax (False | 'cdn' | path - default=False) --
        Specifies how the MathJax.js library is included in the output html
        file or div string.  MathJax is required in order to display labels
        with LaTeX typesetting.

        If False, no script tag referencing MathJax.js will be included in the
        output. HTML files generated with this option will not be able to
        display LaTeX typesetting.

        If 'cdn', a script tag that references a MathJax CDN location will be
        included in the output.  HTML files generated with this option will be
        able to display LaTeX typesetting as long as they have internet access.

        If a string that ends in '.js', a script tag is included that
        references the specified path. This approach can be used to point the
        resulting HTML file to an alternative CDN.
```

## iplot and FigureWidget
The `iplot` and `FigureWidget` changes do not involve any API updates.  They make it possible for plotly.js to take advantage of the MathJax instance already in use by the Jupyter Notebook and JupyterLab.

## Outstanding issues
The only outstanding issue is that unlike the classic notebook, JupyterLab doesn't seem to load the global MathJax instance until it is needed by the markdown renderer.  See https://github.com/jupyter-widgets/ipywidgets/issues/2253.  This may not be something we can address in the 3.4.0 timeframe.


plotly/offline/offline.py
```diff
@@ -577,31 +611,61 @@ def plot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+        include_mathjax = include_mathjax.lower()
+
+    if include_mathjax == 'cdn':
+        mathjax_script = (
+            '<script src="{url}?config=TeX-AMS-MML_SVG"></script>'.format(
+                url=('https://cdnjs.cloudflare.com'
+                     '/ajax/libs/mathjax/2.7.5/MathJax.js')))
+    elif (isinstance(include_mathjax, six.string_types) and
+          include_mathjax.endswith('.js')):
+        mathjax_script = '<script src="{url}"></script>'.format(
```
jonmmease: This needs `?config=TeX-AMS-MML_SVG` as well

plotly/offline/offline.py
```diff
@@ -577,31 +611,61 @@ def plot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
+            url=include_mathjax_orig)
+    elif not include_mathjax:
+        mathjax_script = ''
+    else:
+        raise ValueError("""\
+Invalid value of type {typ} received as the include_mathjax argument
+    Received value: {val}
+
+include_mathjax may be specified as False, 'cdn', or a string ending with '.js' 
+""".format(typ=type(include_mathjax), val=include_mathjax))
```
jonmmease: use `val=repr(include_mathjax)` to be a bit more clear

plotly/offline/offline.py
```diff
@@ -577,31 +611,61 @@ def plot(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
         plotly_js_script = ''
 
+    # Process include_mathjax and build mathjax_script
+    include_mathjax_orig = include_mathjax
+    if isinstance(include_mathjax, six.string_types):
+        include_mathjax = include_mathjax.lower()
+
+    if include_mathjax == 'cdn':
+        mathjax_script = (
+            '<script src="{url}?config=TeX-AMS-MML_SVG"></script>'.format(
```
jonmmease: Add `STIX_Web` font config from `mathjax_config`.  This way appearance of exported html will match `iplot` and `FigureWidget`

plotly/offline/offline.py
jonmmease: Also, add `STIX_Web` font config from `mathjax_config`

plotly/offline/offline.py
dealmeidavf: Is there an example on how to use `include_mathjax='cdn'` in a Jupyter notebook? For instance where would I enable LaTeX in plotly in the following block?
```python
import plotly.offline as py
py.init_notebook_mode()
from ipywidgets import interact
fig = go.FigureWidget()
```
I don't seem to get LaTeX working in the resulting interactive plots.
Thank you.

jonmmease: Hi @dealmeidavf ,

`include_mathajx` is only needed for `plotly.offline.plot`.  For `FigureWidget` and `plotly.offline.iplot` Jupyter's internal MathJax should be used automatically.

```python
go.FigureWidget(layout={'title': '$\LaTeX$'})
```
newplot 16

Could you open a new issue with your browser info, and whether you're using the classic notebook of JupyterLab? There are some issues with FireFox still, so try it out with Chrome if you haven't. Also, in JupyterLab you sometimes need to run a markdown cell containing LaTeX first so that the MathJax is initialized.

dealmeidavf: Thank you. I am using Jupyter notebook in the anaconda navigator on a Mac.  LaTeX does not work on a Mac Firefox. It works on Safari. I am new to plotly but a user of matplotlib. The interactive plots in plotly using ipywidgets are amazing!

PR #1279, opened 2018-11-20
remove plotly/colors.py in favour of utils

Kully:
Since `figure_factory` is 100% of the code that is using `colors.py`, it makes more sense to move the `colors.py` that are not in `plotly.figure_factory/utils.py` in there.

In favour of PR: https://github.com/plotly/plotly.py/pull/878

TODOs for this PR:
- [ ] move all content from colors to utils
- [ ] redirect all reference to colors over to utils
- [ ] make sure the correct plotly colorscales are being used by all figure factories

### DIFF
--- a/plotly/figure_factory/_bullet.py
+++ b/plotly/figure_factory/_bullet.py
@@ -78,7 +78,7 @@ def _bullet(df, markers, measures, ranges, subtitles, titles, orientation,
     for row in range(num_of_lanes):
         # ranges bars
         for idx in range(len(df.iloc[row]['ranges'])):
-            inter_colors = colors.n_colors(
+            inter_colors = utils.n_colors(
                 range_colors[0], range_colors[1],
                 len(df.iloc[row]['ranges']), 'rgb'
             )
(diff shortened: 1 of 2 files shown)


Kully:
ready for a review @jonmmease

Kully:
> Can you add a `plotly/colors.py` file back in that just imports all of the public functions/variables from their new locations?

Sure!

Kully:
> Can you add a `plotly/colors.py` file back in that just imports all of the public functions/variables from their new locations?

The issue I'm running into is that `figure_factory` relies on a numpy import, but none of the functions that `colors.py` is using does.

I haven't found a way to slickly import only the `figure_factory.utils` without grabbing the rest of the modules in the tree. And I don't want to change the heirarchy structure of the repo.

Any suggestions on how to proceed? Perhaps a 1-liner of python code in `colors.py` with a comment saying that colors is being depreciated in the next plotly version.

jonmmease:
@Kully, I see what you mean about the numpy dependency. Let's leave the plotly/colors.py where it is then for now. Also, I don't see any fundamental reason why we wouldn't use these functions from the core (non figure factory) code at some point.

If we leave colors.py where it is, are there other things from the PR that you want to keep?

Kully:
> Let's leave the plotly/colors.py where it is then for now.

By this I understand you mean to leave functions defined directly from plotly/colors.py rather than importing from figure_factory

> I don't see any fundamental reason why we wouldn't use these functions from the core

I agree with this too.

> If we leave colors.py where it is, are there other things from the PR that you want to keep?

There are duplicate functions in here that I removed a few commits above. I think we should for sure keep the duplicate functions removed.

PR #1301, opened 2018-12-11
No colors functions in ff utils

Kully:
in favor of https://github.com/plotly/plotly.py/pull/1279

Simply, we are making sure there are no function overlap between `plotly/colors.py` and `plotly/figure_factory/utils.py` and all references are updated so all code runs well and smoothly.

@jonmmease this ready for a review


Kully:
> can you add imports to `plotly/figure_factory/utils.py` so that these functions are still available under `plotly/figure_factory/utils.py`

Is a simple `import plotly.colors as clrs` in the file good enough? Or do you want each function explicitly imported?

```
from colors import X, Y, Z, ...
```

jonmmease:
@Kully, that latter.  The goal is that if someone has a script with

```python
from plotly.figure_factory.utils import X
```

it should still import as before.

Kully:
Review is addressed - ready to merge?

jonmmease:
💃

PR #1306, opened 2018-12-14
Emf support for save as

michaelbabyn:
Update `plotly.plotly.image.save_as` to support the `emf` file type. The tests pass on the `add_emf_backend` branch of streambed in this [PR](https://github.com/plotly/streambed/pull/12020).

cc @bpostlethwaite

PR #1307, opened 2018-12-14
retry on 429 throttled status

### LINKED ISSUE #1294: v2 API Retry on 429 status_code if users are throttled?
Right now we [Only retry on request errors with 5XX status code and use exponential backoff](https://github.com/plotly/plotly.py/pull/1146). But should we also retry for 429 status codes which users receive if they go over Plot.ly's throttling limit? Especially since calls to the image servers are now limited to 1/sec.

michaelbabyn:
resolves #1294. I also needed this change to get the image tests to pass for #1306


jonmmease:
Thanks @michaelbabyn!

PR #1375, opened 2018-12-23
Include the license in wheels

toddrme2178:
The MIT license requires all copies of the software to include the license text.


jonmmease:
Thanks @toddrme2178!

PR #1393, opened 2019-01-11
Try to coerce all objects to Numpy arrays.

malmaud:
This lets you pass a broad array of list-like objects as data to be plotted instead of the current hard-coded list of numpy arrays, lists, tuples, and Pandas objects. Objects such as PyTorch tensors (widely used in the machine-learning world), xarrays, and others know how to convert themselves into Numpy arrays and so might as well be allowed to be passed directly as plotting arguments so users don't have to explicitly convert them.

### DIFF
--- a/_plotly_utils/basevalidators.py
+++ b/_plotly_utils/basevalidators.py
@@ -156,13 +156,15 @@ def is_numpy_convertable(v):
     Return whether a value is meaningfully convertable to a numpy array
     via 'numpy.array'
     """
+    if np.isscalar(v):  # Scalar types like numpy.float64 shouldn't count as arrays
+        return False
     return hasattr(v, '__array__') or hasattr(v, '__array_interface__')
 
 
 def is_homogeneous_array(v):
     """
     Return whether a value is considered to be a homogeneous array
-    """
+    """    
     return ((np and (isinstance(v, np.ndarray) or is_numpy_convertable(v))) or
             (pd and isinstance(v, (pd.Series, pd.Index))))
 
--- a/_plotly_utils/basevalidators.py
+++ b/_plotly_utils/basevalidators.py
@@ -350,7 +350,14 @@ class DataArrayValidator(BaseValidator):
         elif is_simple_array(v):
             v = to_scalar_or_list(v)
         else:
-            self.raise_invalid_val(v)
+            # Try to coerce 'v' into an array. Useful if 'v' is a list-like object, such as a 
+            # PyTorch tensor, an xarray Dataframe, etc that knows how to turn itself into a Numpy array.
+            v_array = np.array(v)
+            # If np.array returns a scalar object, then it failed to coerce it to a list-like object.
+            if v_array.shape == ():
+                self.raise_invalid_val(v)
+            else:
+                return self.validate_coerce(v_array)
         return v
 
 
--- a/_plotly_utils/basevalidators.py
+++ b/_plotly_utils/basevalidators.py
@@ -43,12 +43,16 @@ def fullmatch(regex, string, flags=0):
 # Utility functions
 # -----------------
 def to_scalar_or_list(v):
+    if np.isscalar(v) and hasattr(v, 'item'):
+        return np.asscalar(v)
     if isinstance(v, (list, tuple)):
         return [to_scalar_or_list(e) for e in v]
     elif np and isinstance(v, np.ndarray):
         return [to_scalar_or_list(e) for e in v]
     elif pd and isinstance(v, (pd.Series, pd.Index)):
         return [to_scalar_or_list(e) for e in v]
+    elif is_numpy_convertable(v):
+        return to_scalar_or_list(np.array(v))
     else:
         return v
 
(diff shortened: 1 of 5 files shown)


malmaud:
Thanks for the feedback!  I'll implement 1) and 2). 

3) is a bit subtle - there are a few different ways that `numpy.array` tries to convert an object to an array - it could, as you said, use the array interface. It will fall back to just seeing if the object has an `__array__` method that directly returns a numpy array. Or it can also use the Python buffer interface. Or an earlier version of the array interface. I wouldn't want to re-implement that entire logic.  PyTorch tensors, for instance, have an `__array__` method but don't support the array interface.

I think just checking for `__array_interface__` or `__array__` methods on objects is probably sufficient for 99% of cases though - the coercion could always be made more general later if we see objects in the wild that this heuristic isn't sufficient for.

jonmmease:
> I think just checking for `__array_interface__` or `__array__` methods on objects is probably sufficient for 99% of cases though - the coercion could always be made more general later if we see objects in the wild that this heuristic isn't sufficient for.

👍

jonmmease:
Thanks @malmaud,

In terms of testing, it looks like something went wrong in `plotly.tests.test_optional.test_figure_factory.test_figure_factory.TestTrisurf`.  Could you have a look?

```
ERROR: test_trisurf_all_args (plotly.tests.test_optional.test_figure_factory.test_figure_factory.TestTrisurf)
  File "/home/circleci/project/plotly/tests/test_optional/test_figure_factory/test_figure_factory.py", line 889, in test_trisurf_all_args
    x, y, z, simplices
  File "/home/circleci/project/plotly/figure_factory/_trisurf.py", line 466, in create_trisurf
    edges_color=edges_color, plot_edges=plot_edges)
  File "/home/circleci/project/plotly/figure_factory/_trisurf.py", line 165, in trisurf
    showlegend=False
  File "/home/circleci/project/plotly/graph_objs/_scatter3d.py", line 1939, in __init__
    self['marker'] = marker if marker is not None else _v
  File "/home/circleci/project/plotly/basedatatypes.py", line 2820, in __setitem__
    self._set_compound_prop(prop, value)
  File "/home/circleci/project/plotly/basedatatypes.py", line 3129, in _set_compound_prop
    val = validator.validate_coerce(val, skip_invalid=self._skip_invalid)
  File "/home/circleci/project/_plotly_utils/basevalidators.py", line 2117, in validate_coerce
    v = self.data_class(v, skip_invalid=skip_invalid)
  File "/home/circleci/project/plotly/graph_objs/scatter3d/_marker.py", line 1064, in __init__
    self['color'] = color if color is not None else _v
  File "/home/circleci/project/plotly/basedatatypes.py", line 2829, in __setitem__
    self._set_prop(prop, value)
  File "/home/circleci/project/plotly/basedatatypes.py", line 3065, in _set_prop
    val = validator.validate_coerce(val)
  File "/home/circleci/project/_plotly_utils/basevalidators.py", line 1140, in validate_coerce
    invalid_els = self.find_invalid_els(v, validated_v)
  File "/home/circleci/project/_plotly_utils/basevalidators.py", line 1169, in find_invalid_els
    self.find_invalid_els(orig_el, validated_el, invalid_els)
  File "/home/circleci/project/_plotly_utils/basevalidators.py", line 1167, in find_invalid_els
    for orig_el, validated_el in zip(orig, validated):
TypeError: zip argument #1 must support iteration
```

Before merging this, we'll also need some tests that exercise the new  `__array_interface__` or `__array__` logic.  I think the easiest thing here might be to create a test file that mirrors `_plotly_utils/tests/validators/test_pandas_series_input.py` but for a library that plotly.py doesn't know about like `xarray`.  To do this, we'd need to add `xarray` to
  - `optional-requirements.txt`
  - As an `optional:` dependency in `tox.ini`
  - To the conda create command in `.circleci/create_conda_optional_env.sh`

Are you interested in adding these tests? If not, we can just leave the PR open until I have a chance to go through and add them (which would hopefully be by the end of the month).

malmaud:
Alright, I'll take a look and add the tests.

malmaud:
Tests have been added. I had to make some other changes to make the tests pass, so apparently that was a worthwhile exercise.

_plotly_utils/basevalidators.py
```diff
@@ -43,12 +43,16 @@ def fullmatch(regex, string, flags=0):
 # Utility functions
 # -----------------
 def to_scalar_or_list(v):
+    if np and np.isscalar(v) and hasattr(v, 'item'):
+        return np.asscalar(v)
```
jonmmease: I'm not sure I understand this branch.  What kinds of objects does this handle?

malmaud: I'll add a comment to the code.

tox.ini
```diff
@@ -72,7 +72,7 @@ deps=
     optional: pyshp==1.2.10
     optional: pillow==5.2.0
     optional: matplotlib==2.2.3
-    optional: xarray==0.11.2
```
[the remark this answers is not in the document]
malmaud: One of the tests reported that 0.10.9 was the highest version of `xarray` available. I guess `xarray` added some upper bound to its dependencies in the 0.11 series that rendered it incompatible with the environment of the test - I didn't really investigate too closely.

_plotly_utils/basevalidators.py
jonmmease: Thanks!

tox.ini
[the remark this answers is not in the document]
jonmmease: Thanks for the info. I figured it was a transitive dependency issue, just wanted to double check.

jonmmease:
Merged!  Thanks again for the solid PR.

PR #1432, opened 2019-02-10
create_animations improvements, progress towards removing clientresp

jonmmease:
### Background (as I understand it)
The `plotly.plotly.create_animations` function was introduced to support plotly.js animations because the legacy clientrest API that is used by `plotly.plotly.plot` does not support frames. 

 `plotly.plotly.create_animations` uses the more recent v2 API that supports frames, but it was much more limited than `plotly.plotly.plot`.  In particular is did not support uploading figures that contain inline data arrays, requiring users to manually create grids and assign grid references to their figures (See https://plot.ly/python/gapminder-example/ for example).  It also did not support overwriting existing figures and it did not support creating figures in folders.

This investigation was inspired by this discussion on the forums: https://community.plot.ly/t/trouble-with-converting-offline-animation-to-online/19129/2

### Overview
This PR addresses these limitation mentioned above

 1. When a figure contains inline data arrays, these arrays are extracted into a Grid.  The grid is them uploaded and the column references are inserted into the figure's corresponding `*src` properties. Then the figure itself is uploaded.  If the figure is named, then the associated grid is named `'{filename}_grid'`. If the figure is not named, then the grid is named `'grid_{id}'` where `id` is the url safe base64 encoding of the hash of the JSON representation of the grid.  This way repeated unnamed figure creation with the same data will result in only one grid.
 2.  Use `v2.files.lookup` to check whether the figure already exists (and retrieve the `fid` if it does).  If it exists, update it, if not create it.
 3. Use `v2.folders.create` to create folders before uploading figures.

### What's left
This implementation is nearly ready to fully replace the clientresp backend of `plotly.plotly.plot`.  The only unsupported feature is the `fileopt` option. See https://plot.ly/python/file-options/.  Right now, the only behavior available is the default `fileopt='overwrite'` behavior., and it doesn't look like the other options are directly supportable with the v2 API as it currently stands.

### Testing
There isn't currently a test suite for `create_animations` and I didn't add any additional testing in the PR.  There will be a bunch of testing work to do when this implementation is used to replace the `plotly.plotly.plot` backend.

### Questions
I've never personally used the `fileopt` argument, does anyone have opinions on whether we can deprecate this?  I think my preference would be to start raising a deprecation warning for `fileopt != 'overwrite'` in 3.7.0, and then remove it from the API in 4.0.0.  The alternative would be to do a bit of work on the v2 API in streambed.

@jackparmer @scjody @nicolaskruchten @chriddyp

### DIFF
--- a/plotly/plotly/plotly.py
+++ b/plotly/plotly/plotly.py
@@ -55,6 +55,10 @@ DEFAULT_PLOT_OPTIONS = {
     'sharing': files.FILE_CONTENT[files.CONFIG_FILE]['sharing']
 }
 
+warnings.filterwarnings(
+    'default', r'The fileopt parameter is deprecated .*', DeprecationWarning
+)
+
 SHARING_ERROR_MSG = (
     "Whoops, sharing can only be set to either 'public', 'private', or "
     "'secret'."
@@ -74,7 +78,7 @@ def sign_in(username, api_key, **kwargs):
 update_plot_options = session.update_session_plot_options
 
 
-def _plot_option_logic(plot_options_from_call_signature):
+def _plot_option_logic(plot_options_from_args):
     """
     Given some plot_options as part of a plot call, decide on final options.
     Precedence:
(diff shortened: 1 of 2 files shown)


bpostlethwaite:
Depreciation notice on `fileopt` is the way to go. Thanks for this summary and suggested path forwards!

PR #1447, opened 2019-03-04
Add new auto_play argument to offline plot and iplot

### LINKED ISSUE #697: Allow users to control autoplay feature on Python animation frames
Hey all,

Loving the new animation functionality you've introduced!

It would be great if you could allow us to decide whether or not we want the animation to autoplay. 

For example, your GapMinder plot created with the Python API [here](https://plot.ly/python/animations/) autoplays, whilst the same plot with .js [here](https://plot.ly/javascript/animations/) doesn't.

Cheers!

jonmmease:
This controls whether to automatically start the animation on page load if the figure contains frames

Closes #697

cc @nicolaskruchten @richard-muir

### DIFF
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -311,7 +311,7 @@ def init_notebook_mode(connected=False):
 
 
 def _plot_html(figure_or_data, config, validate, default_width,
-               default_height, global_requirejs):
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
Ah this is a really nice one, thanks :)

Issue #1469, opened 2019-03-19
plotly.offline.iplot - Option to save to a file and embed link in notebook


Nikolai-Hlubek:
Dear all

The following use case:
I use jupyter with plotly. In some notebooks we plot large amount of data > 1e6 points. Embedding the plot directly in the notebook causes the notebooks to get very large > 100Mb. 
This causes issues with 
- saving in jupyter [Jupyter: Large notebook fails to save](https://github.com/jupyter/notebook/issues/650)
- saving incremental progress because the files are so huge 
- with version control of the notebooks, since the diffs are huge.

The following solution seems to solve these issues for me:
```
filename = 'results/foo.html'
plotly.offline.plot(fig, filename=filename)
IPython.display.display(IPython.display.IFrame(src=filename, width=1200, height=800))
```

This saves the picture to an external file and embeds the contents of this file in Jupyter, therefore bypassing all issues mentioned above. Since googling showed that other users sometimes have similar issues maybe it is a good idea to add the above code to plotly.offline.iplot and make it available via a option?

Best regards
Nikolai

jonmmease:
Thanks for the suggestion @Nikolai-Hlubek, I can see that this could be really useful in some situations.  As it happens, I was in the middle of large rewrite of the plotly.py rendering system when I saw this, and it was actually pretty easily to integrate it as an option based on what I had so far.  See the writeup at https://github.com/plotly/plotly.py/pull/1474 for more context if you're interested!

onclick360-blog:
I am using Jupyter [notebook](https://onclick360.com/cost-function-in-machine-learning/) as a public server and accessing remotely when i am trying to run some plot with plotly it's not showing on my [remote web browser](https://onclick360.com/jenkins-installation-windows) its throw error "127.0.0.1 refused to connect." not sure in iframe how to use original server ip instead of 127.0.0.1.

Nikolai-Hlubek:
Dear Jon

I forgot to respond to your reply. Thank you that this functionality has been implemented.

Issue #1459, opened 2019-03-11
plotly.io.renderer proposal


jonmmease:
## Overview
This issue is a design proposal for a new renderer abstraction for plotly.py to replace the `iplot` functions.

## Goals
Several goals motivate this work

 1. Make it possible to support new plot rendering contexts like static images, rendering inside colab, and rendering in popup windows.
 2. Remove the "offline" language from plot rendering. As discussed in the [v4 plans](https://github.com/plotly/plotly.py/issues/1420), the Chart Studio integration (the `plotly.plotly` stuff) will be split off into a separate package for v4, and the remaining `plotly` package will be only "offline" only.  So having the "offline" language will be unnecessary and potentially confusing.
 3. Provide a single location to change the plot rendering approach across an entire notebook.  This will make it much easier for folks to copy/paste documentation examples and change the example to their preferred rendering approach.

## Background
There are currently 2 ways to display plots in plotly.py in "offline" mode:

 1. `plotly.offline.iplot`:  This will display plots inline in a Jupyter Notebook or in JupyterLab. The use in the classic notebook requires calling `plotly.offline.init_notebook_mode()` to initialize the notebook before rendering.  The use in JupyterLab requires the [`@jupyterlab/plotly-extension`](https://github.com/jupyterlab/jupyter-renderers/tree/master/packages/plotly-extension)
 2. `plotly.offline.plot`:  This will save the plot to a standalone HTML file and, by default, open it in the default system web browser.

## Design Influences
This design is somewhat inspired by the approach used by the Altair project (https://altair-viz.github.io/user_guide/renderers.html), and I expect the Altair implementation will be very helpful reference during the implementation phase of this effort.

The design also draws on elements of the `plotly.io.templates` design (See https://github.com/plotly/plotly.py/pull/1224), and is intended to have a consistent user experience.

### Default renderer
I propose we add a new configuration object named `renderers` in the `plotly.io` module.  This will sit alongside the `plotly.io.templates` and `plotly.io.orca` configuration objects.  Users will view and specify the current default renderer using property assignment. For example, to specify the equivalent of `plotly.offline.iplot`:

```python
import plotly.io as pio
pio.renderers.default = 'notebook'  # `init_notebook_mode` and output with `iplot`
```
Here are some other [`mimetype-based.`](https://jupyter.readthedocs.io/en/latest/reference/mimetype.html) renderer options

```python
pio.renderers.default = 'notebook_connected'
[...]

Issue #1420, opened 2019-02-02
Version 4 discussion


jonmmease:
## Overview
This is an issue for discussing what we'd like to work towards in a future version 4, mostly focusing on changes that would not be backwards compatible.  Nothing here is set in stone and feedback is welcome.

### Theming
Lets make the defualt theme be `plotly` or `plotly_white` (https://github.com/plotly/plotly.py/pull/1224)

### Integrate `px`
Integrate the `px` project for exploratory data analysis directly into plotly.py (https://github.com/plotly/px)

### Redo subplot support
There are a bunch of [outstanding issues](https://github.com/plotly/plotly.py/issues?q=is%3Aopen+is%3Aissue+label%3Amake_subplots) with the `make_subplots` function.  The most significant limitation being that it doesn't support non-Cartesian subplot types.

Last year, plotly.js introduced its own subplot layout system (`layout.grid`) so we could build the new subplot system on top of that (See https://github.com/plotly/plotly.js/issues/3507 ).  Hopefully we could still support the `make_subplots` API, but give most of the work to plotly.js

I would also like to add support for accessing traces by subplot.  Something like
`figure.get_traces(row=2, col=1)`.  This would allow users to construct an initial figure using `px`, and then easily grab particular traces to customize using `.update` or property assignment syntax. See https://github.com/plotly/plotly.py/issues/1484.

### Self displaying figures
I would like to allow `go.Figure` to display itself as a plotly.js figure when used in an ipython context.  We have a prototype of this in `px` with `FigurePx`

https://github.com/plotly/px/blob/610910fd5f167de2482b01753db13cd0bd01c171/plotly_express/_px.py#L22-L33

I would like to extend this a bit so that it's possible to use a global setting to configure how `go.Figure` should display itself.  `FigurePx` is using the `init_notebook_mode`/`iplot` approach, but I'd also like to support a static image display (using orca) that would be suitable for use in QtConsole/Spyder and for PDF export.

These changes would not affect `FigureWidget`.  `FigureWidget` will remain a separate class that is distinct from `go.Figure` in that is is a widget subclass and supports updating properties in place.

### Convert offline html export to `plotly.io` module
Replace the `plotly.offline.plot` function with functions in the `plotly.io` module (https://github.com/plotly/plotly.py/issues/1098)

### Split off choropleth figure factory into a separate package
[...]

PR #1224, opened 2018-10-14
Templates (themes) integration

jonmmease:
## Overview
This PR introduces figure templates (themes) to plotly.py

## Background
Earlier this year the plotly.js team discussed (https://github.com/plotly/plotly.js/issues/2469) and then implemented (https://github.com/plotly/plotly.js/pull/2761) a powerful templating approach that allows plotly.js users to specify default values for all trace and layout properties that make up a figure.  This makes it possible to create custom figure themes that can be applied repeatedly to new figures.

In addition to default property values, themes can also include default object array elements like annotations, shapes, and images.  This way, templates can be used to create figures that include default labels (e.g. "CONFIDENTIAL") and/or default images (e.g. a company/project logo).

### Plotly.js Template Representation
Templates in Plotly.js are specified as a part of the declarative structure of the figure itself.  A template is a specification object located at the `layout.template` path in the figure specification.

After a figure has been created, its template can be updated using the same `Plotly.relayout` operation that can be used to update the rest of the figure's layout properties.

## plotly.py integration
This section describes the high-level changes made in this PR to support templates in plotly.py

### Code generation / validation
The specification of the `layout.template` object can't be fully captured in the schema, so a bunch of code generation work was needed in order to provide the same level of validation that plotly.py users have come to expect from the rest of the figure property hierarchy.

In addition to the `layout.template` object itself, templates also introduce a `*defaults` property that corresponds to each object array property (e.g. `layout.annotationsdefaults` corresponds to `layout.annotations`).  These are now included in the code generation / validation process.

These changes are implemented in approximately commits 106426e through f84c8a5

### plotly.io.templates configuration object for registering named templates
Next, this PR adds a new `plotly.io.templates` configuration object for registering templates.  A user can create a template object (an instance of `go.layout.Template`), configure it to their liking, then register it by name with `plotly.io.templates` for use in the future.

If a string is assigned to the `fig.layout.template` property, plotly.py checks the `plotly.io.templates` registry to see if there is a template registered with this name. If there is, then this template is applied to the figure.

### Default template
The `plotly.io.templates` configuration object can also be used to specify a global default template. The global default template will be applied automatically when a figure is constructed (unless an explicit template was provided to the figure constructor).

### Merging templates
A string assigned to the `fig.layout.template` property, or registered as the default template, may contain multiple registered template names joined on `'+'` characters (borrowing the flaglist syntax). In this case the templates are merged together (applied to the figure from left to right).

### Extracting a template from a figure
[...]

PR #1474, opened 2019-03-21
plotly.io HTML functions, modular renderers framework, future flags system

jonmmease:
## Overview
This PR is a big one!  It tackles three [version 4 goals](https://github.com/plotly/plotly.py/issues/1420) that ended up being so intertwined that I decided to handle them in the same PR.

 1. Adds `to_html` and `write_html` functions to the `plotly.io` module and re-bases `plotly.offline.plot` on top of these new functions.  This finishes porting the existing figure export logic into the `plotly.io` module as initially described in https://github.com/plotly/plotly.py/issues/1098.
 2. Adds a new `plotly.io.renderers` configuration object to control how figures are rendered, and a new `plotly.io.show` function for displaying figures.  Reimplements `plotly.offline.init_notebook_mode` and `plotly.offline.iplot` on top of the new renderers framework.
 3. A new `_plotly_future_` module that will allow users to opt-in to future (v4) default behaviors before v4 is released.

Despite the breadth of these changes, I believe this PR is 100% backward compatible for users, and so I would like to target releasing this for plotly.py version 3.8.0

## `plotly.io` HTML functions
This PR adds two new public functions to `plotly.io`: `to_html` and `write_html`.  Per the convention discussed in https://github.com/plotly/plotly.py/issues/1098, `to_html` inputs a figure and returns an html representation as as string. And `write_html` inputs a figure and a file string/object and writes the html representation to the specified file.

### post_script
The signatures for these functions map pretty closely to `plotly.offline.plot` with one exception. `plotly.offline.plot` includes a set of image export options that can be used to add JavaScript to the exported figure that will ask plotly.js to export the figure to an image and then ask the browser to download it.

This is not a workflow that we recommend anymore now that orca is available and integrated into plotly.py (See https://github.com/plotly/plotly.py/pull/1120), so I didn't want to add this support directly to `to_html`/`write_html`.  Instead, I added a `post_script` argument that can be used to add any custom JavaScript snippet to the resulting HTML.  This custom HTML is executed after the plot is created (in a `.then` block) and it can include `'{plot_id}'` placeholders that are automatically replaced with the `id` of the `div` that the Plotly.js plot is attached to.  Here's the docstring

```
    post_script: str or None (default None)
        JavaScript snippet to be included in the resulting div just after
        plot creation.  The string may include '{plot_id}' placeholders that
        will then be replaced by the `id` of the div element that the
        plotly.js figure is associated with.  One application for this script
        is to install custom plotly.js event handlers.
```
Another use-case I had in mind where this would have come in handy was this question on the forums of how to add custom JavaScript to a plot exported by plotly.py to open hyperlinks in response to marker clicks. https://community.plot.ly/t/hyperlink-to-markers-on-map/17858/6

### responsive output
One other change I made while I was in there was to remove our custom window resize event handler logic and replace it with the (relatively) new plotly.js `'responsive': true` config parameter.  So if the input figure does not have a predefined `layout.width` and `layout.height` then`to_html` will add `'responsive': true` to the figure config.

## `plotly.io.renderers`
This PR introduces a new `plotly.io.renderers` configuration object.  This work ended up following the design I laid out in https://github.com/plotly/plotly.py/issues/1459 pretty much to the letter, and I've updated the proposal description there to match the details of this implementation.

### Design Influences
This design is somewhat inspired by the approach used by the Altair project (https://altair-viz.github.io/user_guide/renderers.html), the Altair implementation was a helpful reference in implementing this.

The design also draws on elements of the `plotly.io.templates` design (See https://github.com/plotly/plotly.py/pull/1224), and is intended to have a consistent user experience.

### Default renderer
I've added a new configuration object named `renderers` in the `plotly.io` module.  This sits alongside the `plotly.io.templates` and `plotly.io.orca` configuration objects.  Users can view and specify the current default renderer using property assignment. For example, to specify the equivalent of `init_notebook_mode(connected=False)`:

```python
import plotly.io as pio
pio.renderers.default = 'notebook+plotly_mimetype'
```
### repr and show
There are now two ways to display figures.  If one or more mime-type based renderers are defined and the `plotly.io.renderers.render_on_display` property is set to `True`, then the `go.Figure` instances will display themselves automatically as their notebook representation.  If `plotly.io.renderers.render_on_display` is set to `False`, then mimetype based rendering will not be performed when figures are displayed.

In addition, all renderer types (including external renderers that will be described below) can be invoked using a new `plotly.io.show` function.  This function will display the figure as a side effect and will return `None`.

The `show` functions will also allow the user to override the default renderer. e.g.
```python
import plotly.io as pio
pio.renderers.default = 'notebook' 

fig = go.Figure(...)

# By default show will display figure inline, override to display in browser tab
pio.show(fig, renderer='browser')
```
Then to display a figure with the equivalent of `plotly.offline.iplot`:
```python
pio.show(fig)
```

### Available Renderers
Here is a detailed description of the renderers that are included in this PR:

#### `plotly_mimetype`
The `plotly_mimetype` renderer displays figures using the `application/vnd.plotly.v1+json` mime type that is handled in JupyterLab by the [`@jupyterlab/plotly-extension`](https://github.com/jupyterlab/jupyter-renderers/tree/master/packages/plotly-extension) extension, and is handled natively by both [nteract](https://nteract.io/) and the [Visual Studio Code](https://code.visualstudio.com/docs/python/jupyter-support).

For user friendliness,  the `plotly_mimetype` renderer is also aliased as `jupyterlab`, `vscode`, and `nteract`

#### HTML-mimetype
Several renderers that produce bundles with mimetype `text/html` are included.

##### `notebook`
This renderer is equivalent to `plotly.offline.init_notebook_mode(connected=False)` and `plotly.offline.iplot`.  It works in the classic Jupyter Notebook, and works offline. This renderer is also useful for notebooks that will be converted to HTML using nbconvert/nbviewer as it will produce standalone HTML files that include interactive figures.

When the renderer is activated in loads the full plotly.js bundle (~3MB) into the notebook and registers it as `'plotly'` with `requirejs` (which is included in the classic notebook).  Then when figures are renderer, they reference plotly.js using `requirejs`.

##### `notebook_connected`
Same as `notebook` except that the initialization phase uses the plotly CDN rather than loading the full plotly.js bundle into the notebook.  This results in much smaller notebooks but requires an internet connection for use.

This is the best rendering mode for use in kaggle notebooks, so it has also been aliased as `'kaggle'`

##### `colab`
Google Colab has a different architecture than either JupyterLab or the classic notebook in that each output cell is an iframe, so there can be no global initialization for the notebook.  Also, requirejs is not available by default.  The Colab renderer borrows insight from the Altair colab renderer and produces full HTML file bundles that include plotly.js using a `<script>` tag.

##### `iframe`
This renderer was inspired by @Nikolai-Hlubek's observations in https://github.com/plotly/plotly.py/issues/1469.  It works be saving figures to standalone html files (using `plotly.io.write_html`) in a directory next to the notebook.  Then the HTML mimebundles that are inserted into the notebook are `iframe` elements with the relative path to these HTML files.

I haven't tested the performance characteristics yet, but the approach works well in the classic notebook and in JupyterLab and it will result in much smaller notebooks for cases where a notebook contains many large figures.  Export to standalone HTML still works fine as long as a copy of the `iframe_figures` directory is kept alongside the HTML file.

One limitation for this approach is that it doesn't currently support responsive resizing. And I'm not sure if this is something that's possible through an iframe.

#### HTML-external
As an alternative to the mime-type rendering approach, where renderers subclass `MimetypeRenderer`,  there is the external rendering approach, where renderers subclass `ExternalRenderer`.  These renderers are external in the sense that they render the figure outside of the notebook, and don't include any reference to the figure in the notebook itself.

##### `browser`
The `'browser'` renderer displays interactive figures in a new tab/window in the users default browser.  This is done without the use of temporary files (which would be challenging to manage) by creating a single use `HTTPServer` on a local port that responds to exactly one request with the figure's contents and then shuts down.

This approach has no Jupyter/notebook dependencies works fine in the QtConsole, ipython console, and even the vanilla python repl.

##### `firefox` and `chrome`
There are also predefined renderers that allow the user to specify that the new tabs should be opened in firefox or in chrome.

#### Static image renderers
A collection of static image renderers are included that rely on the orca integration to render figures as images.  In addition to the Jupyter Notebook/JupyterLab, these renderers work well in even static contexts like the QtConsole/Spyder.

##### 'png', 'svg', 'jpg', 'pdf'
Static rendereres are predefined for png, svg, jpg, and pdf formats.  I'm especially excited about the `pdf` static renderer because this is picked up by the LaTeX-based PDF export functionality of nbconvert.  This means that with the `pdf` renderer enabled, a notebook can be exported through LaTeX as a PDF and the figures will come along in vector form.

##### `json`
This renderer allows the figure's JSON structure to be displayed in the notebook using the JSON viewers in JupyterLab/VSCode.  This is really handy for being able to navigate the structure of a large figure with expand/collapse toggling and search-based filtering.  Here's a YouTube clip of the what this looks like https://www.youtube.com/watch?v=FRj1r7-7kiQ.

### Combining renderers
Multiple renderers can be registered as the default by separating the renderer names with `'+'` characters.  This is the same combination syntax used to combine templates in `plotly.io.templates`. When multiple mime type renderers are specified this way, a single bundle will be created with the render representation for each one.  As motivation, consider

```python
pio.renderers.default = 'notebook+plotly_mimetype+png'
```
A notebook with this renderer specification would display figures properly in the classic notebook ('notebook'), in jupyterlab/vscode/nteract ('plotly_mimetype'), in exported HTML ('notebook'), in the QtConsole/PyCharm ('png'), and in exported PDFs ('png').

Of course this would result in fairly large notebook sizes, but the user will have the flexibility to define where this code needs to render figures.

### Auto-detecting renderers
In a few cases, we're able to detect when it's appropriate to enable a specific renderer by default.  In particular, this PR includes logic to detect when plotly.py is running in Google Colab, Kaggle, or VSCode.

### Customize renderer properties
Some renderers can expose additional configuration options. In this case, a new renderer can be constructed and named. Each built-in renderer is defined by a class accessible under `plotly.io.base_renderers`.

E.g. to specify a static image renderer with custom properties:

```python
import plotly.io as pio
pio.renderers['my_png'] = pio.base_renderers.PngRenderer(width=1000, height=650, scale=1.5)
pio.renderers.default = 'my_png'
```

Or the properties of an existing renderer can be modified directly:
```python
pio.renderers['png'].width = 1000
pio.renderers['png'].height = 650
pio.renderers['png'].scale = 1.5
pio.renderers.default = 'png'
```
Or the properties can be overridden temporarily in `plotly.io.show`:
```python
pio.show(fig, renderer='png', width=1000, height=650, scale=1.5)
```

### Registering new renderers
Users or third-party-libraries may register new renderers by subclassing `plotly.io.base_renderers.MimetypeRenderer` or `plotly.io.base_renderers.ExternalRenderer`

```python
import plotly.io as pio

# Define custom mimetype renderer
class MyRenderer(pio.base_renderers.MimetypeRenderer):
    ...
    def activate():
        ...

    def to_mimebundle(self, fig_dict):
        ...

# Or a custom external renderer
class MyExternalRenderer(pio.base_renderers.ExternalRenderer):
    ...
    def activate():
        ...

    def render(self, fig_dict):
        ...

# Register for use by name
my_renderer = MyRenderer()
pio.renderers['my_renderer'] = my_renderer

# Set as default
pio.renderers.default = 'my_renderer'
```

### Backward compatibility
For backward compatibility, `plotly.offline.init_notebook_mode` and `plotly.offline.iplot` will stay around, but they will be implemented on top of the new renderer framework.

If we are able to merge this work in version 3, the default renderer will be `plotly_mimetype` and `iplot` will call `plotly.io.show`.  Calling `plotly.offline.init_notebook_mode()` will set the default renderer to `'plotly_mimetype+notebook'` and `plotly.offline.init_notebook_mode(connected=True)` will set the default renderer to `'plotly_mimetype+notebook_connected'.`

## The `_plotly_future_` system
Inspired by the Python [`__future__`](https://docs.python.org/2/library/__future__.html) mechanism for opting in to the default behavior of future version of the language, this PR introduces a `_plotly_future_` module that can be used to enable features/behaviors that will be the default in plotly.py version 4.

In every case, these `_plotly_future_` import must come before `plotly` itself is imported otherwise an error is raised.

For example, the default behavior of this PR for displaying figures matches that of plotly.py version 3. But the `renderer_defaults` flag can be used to enable the future v4 defaults, where figures will display themselves automatically.

```python
from _plotly_future_ import renderer_defaults
```

Or, to enable the `plotly` theme that we plan to make the default in version 4
```python
from _plotly_future_ import template_defaults
```

And to enable all version 4 features that are available (only the two above for now)
```python
from _plotly_future_ import v4
```

I'd like to continue adding flags here as a way to merge in [v4 updates](https://github.com/plotly/plotly.py/issues/1420) before v4 itself.  This way the community can weigh in on changes more easily (no need to download a separate version, just `from _plotly_future_ import v4` at the top of your script), and we can more easily start working on updating documentation to the v4 style before v4 is released.

cc @jackparmer @chriddyp @nicolaskruchten @cldougl @michaelbabyn 

TODO:
 - [ ] Add animation options support from https://github.com/plotly/plotly.py/pull/1503


jonmmease:
For anyone interested in trying this out...

```
$ pip install https://11125-14579099-gh.circle-artifacts.com/0/dist/plotly-3.7.0%2B40.ga3739480.tar.gz
```

GorgiAstro:
Hi,

This looks good, combining renderers is a very cool feature. Here are a few feedbacks about the new renderers:
- JPG/PNG/PDF: all good
- SVG: all good in jupyterlab, but when converting the notebook to LaTeX, nbconvert fails and gives `SVG: TypeError: Object of type bytes is not JSON serializable`. This is probably due to https://github.com/jupyter/nbconvert/issues/904, so nothing to do with plotly
- notebook: works as expected, displays the graph in Jupyter Notebooks, and displays nothing in JupyterLab
- chromium would be nice to have in addition to firefox and chrome
- Custom renderer properties: The example with a PngRenderer works as expected. However, the example might be wrong, I had to call `pio.base_renderers.PngRenderer` instead of `pio.renderers.PngRenderer`
- The combined 'notebook+pdf' renderer works as expected: an interactive plot in the Notebook, and a PDF is exported by nbconvert when exporting to LaTeX. This is exactly what I was looking for, awesome!
- What is the hierarchy when combining renderers? In JupyterLab, both 'png+pdf' and 'pdf+png' result in displaying a PDF
- I can't get Plotly to show up in JupyterLab. When setting 'jupyterlab' as renderer, I get the error message below.
  - This occurs even when using the legacy functions, (`import plotly.offline as py`, `py.init_notebook_mode(connected=False)`, `py.iplot(fig)`) 
  - I also tried using an existing conda environment I had (which I use for this project: https://nbviewer.jupyter.org/github/GorgiAstro/laser-orbit-determination/blob/master/02-orbit-determination-example.ipynb), and only installing your development version of plotly by means of the pip command you posted. This gives the same error message. After uninstalling the development version of plotly and getting back to the conda version, it works again.

At the bottom of the message are some information about my setup.

I used the following code in the notebook:
```
import numpy as np
one_hundred = np.linspace(0,100)

import plotly.io as pio
pio.renderers.default = 'jupyterlab' 

trace = go.Scatter(
    y=one_hundred,
    mode='markers'
)

data = [trace]

layout = go.Layout(
    title = 'One hundred',
    xaxis = dict(
        title = 'x'
    ),
    yaxis = dict(
        title = 'y'
    )
)

fig = dict(data=data, layout=layout)

pio.show(fig)
```

```
ValueError                                Traceback (most recent call last)
<ipython-input-8-72586bf8c561> in <module>
     20 
     21 #py.iplot(fig)
---> 22 pio.show(fig)

~/miniconda3/envs/plotlydev/lib/python3.7/site-packages/plotly/io/_renderers.py in show(fig, renderer, validate, **kwargs)
    323                 'Mime type rendering requires ipython but it is not installed')
    324 
--> 325         ipython_display.display(bundle, raw=True)
    326 
    327     # external renderers

~/miniconda3/envs/plotlydev/lib/python3.7/site-packages/IPython/core/display.py in display(include, exclude, metadata, transient, display_id, *objs, **kwargs)
    300     for obj in objs:
    301         if raw:
--> 302             publish_display_data(data=obj, metadata=metadata, **kwargs)
    303         else:
    304             format_dict, md_dict = format(obj, include=include, exclude=exclude)

~/miniconda3/envs/plotlydev/lib/python3.7/site-packages/IPython/core/display.py in publish_display_data(data, metadata, source, transient, **kwargs)
    118         data=data,
    119         metadata=metadata,
--> 120         **kwargs
    121     )
    122 

~/miniconda3/envs/plotlydev/lib/python3.7/site-packages/ipykernel/zmqshell.py in publish(self, data, metadata, source, transient, update)
    127         # hooks before potentially sending.
    128         msg = self.session.msg(
--> 129             msg_type, json_clean(content),
    130             parent=self.parent_header
    131         )

~/miniconda3/envs/plotlydev/lib/python3.7/site-packages/ipykernel/jsonutil.py in json_clean(obj)
    189         out = {}
    190         for k,v in iteritems(obj):
--> 191             out[unicode_type(k)] = json_clean(v)
    192         return out
    193     if isinstance(obj, datetime):

~/miniconda3/envs/plotlydev/lib/python3.7/site-packages/ipykernel/jsonutil.py in json_clean(obj)
    189         out = {}
    190         for k,v in iteritems(obj):
--> 191             out[unicode_type(k)] = json_clean(v)
    192         return out
    193     if isinstance(obj, datetime):

~/miniconda3/envs/plotlydev/lib/python3.7/site-packages/ipykernel/jsonutil.py in json_clean(obj)
    189         out = {}
    190         for k,v in iteritems(obj):
--> 191             out[unicode_type(k)] = json_clean(v)
    192         return out
    193     if isinstance(obj, datetime):

~/miniconda3/envs/plotlydev/lib/python3.7/site-packages/ipykernel/jsonutil.py in json_clean(obj)
    175 
    176     if isinstance(obj, list):
--> 177         return [json_clean(x) for x in obj]
    178 
    179     if isinstance(obj, dict):

~/miniconda3/envs/plotlydev/lib/python3.7/site-packages/ipykernel/jsonutil.py in <listcomp>(.0)
    175 
    176     if isinstance(obj, list):
--> 177         return [json_clean(x) for x in obj]
    178 
    179     if isinstance(obj, dict):

~/miniconda3/envs/plotlydev/lib/python3.7/site-packages/ipykernel/jsonutil.py in json_clean(obj)
    189         out = {}
    190         for k,v in iteritems(obj):
--> 191             out[unicode_type(k)] = json_clean(v)
    192         return out
    193     if isinstance(obj, datetime):

~/miniconda3/envs/plotlydev/lib/python3.7/site-packages/ipykernel/jsonutil.py in json_clean(obj)
    195 
    196     # we don't understand it, it's probably an unserializable object
--> 197     raise ValueError("Can't clean for JSON: %r" % obj)

ValueError: Can't clean for JSON: array([  0.        ,   2.04081633,   4.08163265,   6.12244898,
         8.16326531,  10.20408163,  12.24489796,  14.28571429,
        16.32653061,  18.36734694,  20.40816327,  22.44897959,
        24.48979592,  26.53061224,  28.57142857,  30.6122449 ,
        32.65306122,  34.69387755,  36.73469388,  38.7755102 ,
        40.81632653,  42.85714286,  44.89795918,  46.93877551,
        48.97959184,  51.02040816,  53.06122449,  55.10204082,
        57.14285714,  59.18367347,  61.2244898 ,  63.26530612,
        65.30612245,  67.34693878,  69.3877551 ,  71.42857143,
        73.46938776,  75.51020408,  77.55102041,  79.59183673,
        81.63265306,  83.67346939,  85.71428571,  87.75510204,
        89.79591837,  91.83673469,  93.87755102,  95.91836735,
        97.95918367, 100.        ])
```

List of packages in my conda environment:
```
# Name                    Version                   Build  Channel
attrs                     19.1.0                   py37_1  
backcall                  0.1.0                    py37_0  
blas                      1.0                         mkl  
bleach                    3.1.0                    pypi_0    pypi
ca-certificates           2019.1.23                     0  
certifi                   2019.3.9                 py37_0  
chardet                   3.0.4                    pypi_0    pypi
dbus                      1.13.6               h746ee38_0  
decorator                 4.4.0                    py37_1  
defusedxml                0.5.0                    py37_1  
entrypoints               0.3                      py37_0  
expat                     2.2.6                he6710b0_0  
fontconfig                2.13.0               h9420a91_0  
freetype                  2.9.1                h8a8886c_1  
glib                      2.56.2               hd408876_0  
gmp                       6.1.2                h6c8ec71_1  
gst-plugins-base          1.14.0               hbbd80ab_1  
gstreamer                 1.14.0               hb453b48_1  
icu                       58.2                 h9c2bf20_1  
idna                      2.8                      pypi_0    pypi
intel-openmp              2019.3                      199  
ipykernel                 5.1.0            py37h39e3cac_0  
ipython                   7.3.0            py37h39e3cac_0  
ipython_genutils          0.2.0                    py37_0  
ipywidgets                7.4.2                    py37_0  
jedi                      0.13.3                   py37_0  
jinja2                    2.10                     py37_0  
jpeg                      9b                   h024ee3a_2  
jsonschema                3.0.1                    py37_0  
jupyter                   1.0.0                    py37_7  
jupyter_client            5.2.4                    py37_0  
jupyter_console           6.0.0                    py37_0  
jupyter_core              4.4.0                    py37_0  
jupyterlab                0.35.4           py37hf63ae98_0  
jupyterlab_server         0.2.0                    py37_0  
libedit                   3.1.20181209         hc058e9b_0  
libffi                    3.2.1                hd88cf55_4  
libgcc-ng                 8.2.0                hdf63c60_1  
libgfortran-ng            7.3.0                hdf63c60_0  
libpng                    1.6.36               hbc83047_0  
libsodium                 1.0.16               h1bed415_0  
libstdcxx-ng              8.2.0                hdf63c60_1  
libuuid                   1.0.3                h1bed415_2  
libxcb                    1.13                 h1bed415_1  
libxml2                   2.9.9                he19cac6_0  
markupsafe                1.1.1            py37h7b6447c_0  
mistune                   0.8.4            py37h7b6447c_0  
mkl                       2019.3                      199  
mkl_fft                   1.0.10           py37ha843d7b_0  
mkl_random                1.0.2            py37hd81dba3_0  
nbconvert                 5.4.1                    py37_3  
nbformat                  4.4.0                    py37_0  
ncurses                   6.1                  he6710b0_1  
nodejs                    10.13.0              he6710b0_0  
notebook                  5.7.6                    py37_0  
numpy                     1.16.2           py37h7e9f1db_0  
numpy-base                1.16.2           py37hde5b4d6_0  
openssl                   1.1.1b               h7b6447c_1  
pandas                    0.24.2           py37he6710b0_0  
pandoc                    2.2.3.2                       0  
pandocfilters             1.4.2                    py37_1  
parso                     0.3.4                    py37_0  
pcre                      8.43                 he6710b0_0  
pexpect                   4.6.0                    py37_0  
pickleshare               0.7.5                    py37_0  
pip                       19.0.3                   py37_0  
plotly                    3.7.0+40.ga3739480          pypi_0    pypi
plotly-orca               1.2.1                         1    plotly
prometheus_client         0.6.0                    py37_0  
prompt_toolkit            2.0.9                    py37_0  
psutil                    5.6.1            py37h7b6447c_0  
ptyprocess                0.6.0                    py37_0  
pygments                  2.3.1                    py37_0  
pyqt                      5.9.2            py37h05f1152_2  
pyrsistent                0.14.11          py37h7b6447c_0  
python                    3.7.2                h0371630_0  
python-dateutil           2.8.0                    py37_0  
pytz                      2018.9                   py37_0  
pyzmq                     18.0.0           py37he6710b0_0  
qt                        5.9.7                h5867ecd_1  
qtconsole                 4.4.3                    py37_0  
readline                  7.0                  h7b6447c_5  
requests                  2.21.0                   pypi_0    pypi
retrying                  1.3.3                    pypi_0    pypi
send2trash                1.5.0                    py37_0  
setuptools                40.8.0                   py37_0  
sip                       4.19.8           py37hf484d3e_0  
six                       1.12.0                   py37_0  
sqlite                    3.27.2               h7b6447c_0  
terminado                 0.8.1                    py37_1  
testpath                  0.4.2                    py37_0  
tk                        8.6.8                hbc83047_0  
tornado                   6.0.1            py37h7b6447c_0  
traitlets                 4.3.2                    py37_0  
urllib3                   1.24.1                   pypi_0    pypi
wcwidth                   0.1.7                    py37_0  
webencodings              0.5.1                    py37_1  
wheel                     0.33.1                   py37_0  
widgetsnbextension        3.4.2                    py37_0  
xz                        5.2.4                h14c3975_4  
zeromq                    4.3.1                he6710b0_3  
zlib                      1.2.11               h7b6447c_3  
```

Based on the following environment YAML file:
```
name: plotlydev
channels:
  - plotly
dependencies:
  - python
  - jupyter
  - jupyterlab
  - nodejs
  - pandas
  - numpy
  - psutil
  - plotly-orca
```

List of JupyterLab plugins:
```
JupyterLab v0.35.4
Known labextensions:
   app dir: /home/yzokras/miniconda3/envs/plotlydev/share/jupyter/lab
        @jupyter-widgets/jupyterlab-manager v0.38.1  enabled  OK
        @jupyterlab/plotly-extension v0.18.2  enabled  OK
        @lckr/jupyterlab_variableinspector v0.1.0  enabled  OK
        @telamonian/theme-darcula v0.20.1  enabled  OK
        jupyterlab-chart-editor v1.1.0  enabled  OK
        plotlywidget v0.8.0  enabled  OK
```

Thank you,
Cheers
GorgiAstro

jonmmease:
@GorgiAstro, thanks so much for taking the time to try this out and for writing up your really helpful feedback!

> chromium would be nice to have in addition to firefox and chrome

Yes, good point

> Custom renderer properties: The example with a PngRenderer works as expected. However, the example might be wrong, I had to call pio.base_renderers.PngRenderer instead of pio.renderers.PngRenderer

Yes, this was a type in the overview post, `pio.base_renderers.PngRenderer` is the correct location.

> The combined 'notebook+pdf' renderer works as expected: an interactive plot in the Notebook, and a PDF is exported by nbconvert when exporting to LaTeX. This is exactly what I was looking for, awesome!

🎉 

> What is the hierarchy when combining renderers? In JupyterLab, both 'png+pdf' and 'pdf+png' result in displaying a PDF

The order when combining renderers shouldn't matter (unless multiple renderers produce the same mimetype, then the right-most takes precedence).  As I understand it, it's fully up to the jupyter front-end to choose between the available mime-types.  So based on your observations, JupyterLab has a preference of PDF over PNG, but it prefers the plotly mimetype over both of these.

> I can't get Plotly to show up in JupyterLab. When setting 'jupyterlab' as renderer, I get the error message below. 

Thanks for reporting this. Looks like I missed a JSON serialization step somewhere because the numpy array should be converted into a list before getting to the point where the error is raised.  I expect that the example would show up for you if you converted the `one_hundred` array to a list first.

```
...
y=one_hundred.tolist()
...
```

I'll fix this and add the tests that should have caught it in the first place 🙂.   And if you don't mind, I'll ping you again when I have the next update ready.

jonmmease:
Proof of out-of-the-box Colab support 🙂 

Screen Shot 2019-03-25 at 7 35 10 PM

jonmmease:
@Juanlu001,
I'm working on a big overhaul of the plotly.py rendering subsystem here and I was wondering if you'd be interesting in weighing in or testing it out in some of the documentation export contexts you've been working with. Cross reference https://github.com/plotly/plotly.py/issues/1033.

The default output behavior should match the existing `init_notebook_mode`+`iplot` behavior (i.e. using requirejs to define plotly.js). But with this PR it's possible to define custom "renderers" that present plotly figures in differernt ways.  For example, there is a `colab`-compatible renderer that outputs a complete HTML document with plotly.js defined using script tags, and there is an `iframe` renderer that saves the standalone HTML files outside of the notebook and then references them using iframes.

So with this approach it would be possible to include dedicated renderer(s) for various documentation export contexts, if that's something that would make people's lives easier.  Thanks!

Juanlu001:
Hi @jonmmease, thanks for the ping! Indeed I would love to test the new rendering machinery. As you already know from https://github.com/rtfd/readthedocs.org/issues/4367 (and a number of other issues) there is a complicated relationship between ipywidgets, Jupyter notebook, nbsphinx, Plotly... Therefore anything that potentially simplifies it, improves the current situation, or in general makes displaying Plotly in Sphinx documentation is interesting to me.

I didn't review all the changes (I skipped the collapsed files :innocent:) but I saw you implemented a bunch of renderers. Does that mean that third parties don't need to do that work?

Also, is this still WIP or do you think it's ready enough? What would be the best way to give it a shot?

jonmmease:
Hi @Juanlu001, great!

I'd say this is in a review-ready state, and you can try it out by installing from the CI build for this branch (As of 1115608) with:
```
$ pip install https://11196-14579099-gh.circle-artifacts.com/0/dist/plotly-3.7.0%2B44.g2677dafe.tar.gz
```

>  I saw you implemented a bunch of renderers. Does that mean that third parties don't need to do that work?

That's correct, the renderers listed in the top description will all included by default.  Third parties will also be able to add additional renderers as needed.

I should also be clear that all of this work is pretty much independent from the existing `FigureWidget` ipywidgets rendering path.  This is all for the rendering of plain `Figure` objects.

GorgiAstro:
Hi, here are some feedbacks about the new commits:

- Chromium support: works, thanks
- The plots are displayed correctly now in JupyterLab

Cheers
GorgiAstro

ricardofbarros:
@jonmmease tested this PR in my classic notebook and the performance is way better :+1:

I installed this:

`pip install https://11196-14579099-gh.circle-artifacts.com/0/dist/plotly-3.7.0%2B44.g2677dafe.tar.gz`

and tested like this:
```
from _plotly_future_ import renderer_defaults
import plotly.io as pio

import plotly

trace = go.Ohlc(x=df.index,
                open=df['open'],
                high=df['high'],
                low=df['low'],
                close=df['close'])

fig = dict(data=[trace], layout={'title': 'OHLC'})

pio.show(fig)
```

GorgiAstro:
> @GorgiAstro if you have some spare minutes, can you check how does this work with documentation? perhaps you already noticed [rtfd/readthedocs.org#4367 (comment)](https://github.com/rtfd/readthedocs.org/issues/4367#issuecomment-423814712) Sorry @jonmmease I haven't been able to test this!

I tested an example Notebook from poliastro, by installing the dev version of Plotly (on top of the Python environment I use for poliastro) by means of `pip install https://11196-14579099-gh.circle-artifacts.com/0/dist/plotly-3.7.0%2B44.g2677dafe.tar.gz`

After running and saving the notebook, I ran the command `make html` in the `poliastro/docs` directory.

All renderers were tested, here are the results of the HTML export to Read-The-Docs:
- plotly_mimetype/jupyterlab/vscode/nteract: exported as png
- notebook: nothing
- notebook_connected: nothing
- iframe: the carl sagan video shows up instead, wtf
- browser: nothing (expected)
- png: exported as png (expected)
- svg: exported as svg (expected)
- jpg: exported as jpg (expected)
- pdf: only the filename appears in the read-the-docs
- json: "Data type cannot be displayed: application/json"

The notebook used for the test is located here: https://github.com/GorgiAstro/poliastro/blob/test-plotly-new-renders/docs/source/examples/Going%20to%20Mars%20with%20Python%20using%20poliastro.ipynb

Nbviewer: https://nbviewer.jupyter.org/github/GorgiAstro/poliastro/blob/ec994082825fcc9496e866ec2184f29b00dfec85/docs/source/examples/Going%20to%20Mars%20with%20Python%20using%20poliastro.ipynb

jonmmease:
Thanks for giving it a try @GorgiAstro!  My expectation is that `notebook` should behave identically to `init_notebook_mode()`+`iplot` and `notebook_connected` should behave identically to `init_notebook_mode(connected=True)`+`iplot`.  Were either of these `iplot` methods working for you in poliastro docs in previous versions? From https://github.com/rtfd/readthedocs.org/issues/4367#issuecomment-423814712, I gather that this wasn't working previously, but correct me if I'm wrong.

If RTD is having a problem with the requirejs dependency, I wonder if the `'colab'` renderer would work as this loads plotly.js using a script tag.  Unfortunately this approach won't work in the classic notebook itself, but it would be interesting to know whether RTD picked it up alright.

GorgiAstro:
You're right @jonmmease, the methods were not working in poliastro docs in previous versions either (@Juanlu001, can you confirm?).

I added the colab renderer to the notebook, it does not get picked up by RTD.

jonmmease:
Thanks for confirming @GorgiAstro.

I'm planning to merge this tomorrow unless something comes up.

jonmmease:
Merging.

Now that we have this renderers interface, if someone is interested in digging in perhaps we could come up with an html renderer that works in both the classic notebook and RTD.  For example, if `requirejs` is still an issue, we could add a `'readthedocs'` renderer that automatically loads requirejs from a CDN like @Juanlu001's hack from https://rtd-js-test.readthedocs.io/en/latest/Connected%20-%20require.js%20Hack.html.

GorgiAstro:
Great to hear that Plotly 3.8.0 has been released :tada: 

Here are the new combined renderers at work: https://nbviewer.jupyter.org/github/GorgiAstro/laser-orbit-determination/blob/master/02-orbit-determination-example.ipynb

PR #1476, opened 2019-03-23
Extract chart studio functionality, optimized imports

jonmmease:
## Overview
This PR is an important step towards the [version 4 goal](https://github.com/plotly/plotly.py/issues/1420) of removing all of the chart studio (i.e. cloud-service related) functionality from plotly.py, and putting it in a separate optional package.

## chart studio extraction
For the time being, I've done this by creating a new top-level `chart_studio` package next to the top-level `plotly` package.  I've moved all of the cloud-related functionality to the `chart_studio` package, following the same structure as in the current plotly package.  For example, the `plotly.plotly` module was moved to `chart_studio.plotly`.

This PR takes advantage of the `_future_plotly_` system introduced in https://github.com/plotly/plotly.py/pull/1474 to make this refactor backward compatible.

 - By default all of the old entry points are still usable and they are aliased to the `chart_studio` package.
 - If the `extract_chart_studio` future flag is set, then deprecation warnings are raised whenever the `chart_studio` modules/functions are used from their legacy locations under the `plotly` package.
 - If the `remove_deprecations` future flag is set then the chart studio functions are fully removed from the plotly package and are accessible only under `chart_studio`.  When `remove_deprecations`  is set, `plotly` has no dependency on the `chart_studio` package.

## Usage
To remove the chart_studio functionality from the main `plotly` module, use the

```python
from _plotly_future_ import remove_deprecations
```

This will further speed up imports, and will allow for testing code to make sure it will be compatible with the package structure of plotly.py version 4.

## Import optimization
This PR also makes a relatively minor change to the code generation logic for `graph_objs` and `validator` that yields an import time reduction of ~10-20% .  Rather that creating a single file for each datatype and validator class, all of the classes in a `graph_obj` or `validator` module are specified directly in the `__init__.py` file.  This reduces the number of files significantly, which seems to yield a modest but consistent speedup while being 100% backward compatible.

## What's next
For version 4, the `chart_studio` module will be published in a separate PyPI/conda package and all of the aliases to `chart_studio` will be removed from the `plotly` PyPI package.


chriddyp:
> One benefit could be speeding up the import of the parts of plotly.py that are needed by Dash

We might also consider breaking out `plotly.utils.PlotlyJSONEncoder` into it's own module, or maybe Dash keeps its own copy. Anyway, just something to consider.

jonmmease:
> We might also consider breaking out plotly.utils.PlotlyJSONEncoder into it's own module, or maybe Dash keeps its own copy. Anyway, just something to consider.

Yeah this is definitely a possibility.  I ended up pulling out the `non_recursive_imports` flag/option from this PR because there's enough going on here already and that deserves its own dedicated discussion.

kmactavish:
@jonmmease FYI I believe this refactor broke holoviews integration (It imports plotly.offline.offline.utils, which has moved to _plotly_utils.utils).

jonmmease:
Thanks for the heads up @kmactavish.  I created an issue at https://github.com/pyviz/holoviews/issues/3643.

PR #1480, opened 2019-03-27
clean up PyPI metadata

nicolaskruchten:


setup.py
```diff
@@ -394,17 +394,20 @@ def run(self):
       maintainer='Chris P',
       maintainer_email='chris@plot.ly',
       url='https://plot.ly/python/',
-      description="Python plotting library for collaborative, "
-                  "interactive, publication-quality graphs.",
+      project_urls={"Github": "https://github.com/plotly/plotly.py"},
+      description="An open-source, interactive graphing library for Python",
       long_description=readme(),
+      long_description_content_type="text/markdown",
```
nicolaskruchten: This enables us to use Markdown

setup.py
```diff
@@ -394,17 +394,20 @@ def run(self):
...
       classifiers=[
-          'Development Status :: 4 - Beta',
+          'Development Status :: 5 - Production/Stable',
```
nicolaskruchten: I think we're there!

setup.py
```diff
@@ -394,17 +394,20 @@ def run(self):
       maintainer='Chris P',
       maintainer_email='chris@plot.ly',
       url='https://plot.ly/python/',
-      description="Python plotting library for collaborative, "
-                  "interactive, publication-quality graphs.",
+      project_urls={"Github": "https://github.com/plotly/plotly.py"},
```
nicolaskruchten: This just gets a bit more metadata on the PyPI page to show up automatically

setup.py
```diff
@@ -30,7 +30,7 @@ def plotly_js_version():
 
 
 def readme():
-    with open('README.rst') as f:
+    with open('README.md') as f:
```
nicolaskruchten: This is what's on Github by default and is much more up to date than the one on PyPI

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [metadata]
-description-file = README.rst
+description-file = README.md
```
nicolaskruchten: unsure about the impact of this...

setup.py
```diff
@@ -394,17 +394,20 @@ def run(self):
       maintainer='Chris P',
```
nicolaskruchten: We should consider making this @jonmmease

nicolaskruchten:
Might be worth a push to test.pypi.org first to check that the formatting etc looks good?

Issue #1076, opened 2018-07-27
Don't create or check permissions on .plotly config directory until it a write operation is requested


jonmmease:
Right now we always run the check to make sure read/write permissions are available for the user's `.plotly` directory.  We do it with a call to `tools.ensure_local_plotly_files()` on the import of `plotly/plotly/plotly.py`. Becuase of our recursive imports this package is always imported and so we always check for these permissions.

Because of use-cases like that in #1068, where a bunch of jobs are running on a cluster and are just doing offline plotting, we shouldn't do this check unless we're actually using the `plotly.plotly` package to communicate with the plot.ly servers.

jonmmease:
Done in #1195

pb-cdunn:
I think this is still a problem.
```
_______ ERROR collecting tests/unit/test_pbreports_plot_plotly_helper.py _______
tests/unit/test_pbreports_plot_plotly_helper.py:7: in <module>
    from pbreports.plot import plotly_helper
pbreports/plot/plotly_helper.py:12: in <module>
    from plotly import graph_objs
/scratch/cdunn/repo/pbreports/build/lib/python2.7/site-packages/plotly/__init__.py:31: in <module>
    from plotly import (plotly, dashboard_objs, graph_objs, grid_objs, tools,
/scratch/cdunn/repo/pbreports/build/lib/python2.7/site-packages/plotly/plotly/__init__.py:10: in <module>
    from . plotly import (
/scratch/cdunn/repo/pbreports/build/lib/python2.7/site-packages/plotly/plotly/plotly.py:30: in <module>
    from plotly import exceptions, files, session, tools, utils
/scratch/cdunn/repo/pbreports/build/lib/python2.7/site-packages/plotly/files.py:41: in <module>
    _file_permissions = _permissions()
/scratch/cdunn/repo/pbreports/build/lib/python2.7/site-packages/plotly/files.py:28: in _permissions
    os.mkdir(TEST_DIR)
E   OSError: [Errno 17] File exists: '/home/UNIXHOME/cdunn/.test'
```
I had to re-raise in the except-catch block of `_permissions` in order to get this. `from plotly import graph_objs` seems to trigger it, and that's in the *public* API.

pb-cdunn:
I want to make warnings into errors in my tests. (`pytest -W error`), so this is a big problem for me.

PR #1498, opened 2019-04-06
Fix race conditions

pb-cdunn:
This is a "Heisenbug", a race condition. You might not be able to reproduce it, but I'll describe it anyway.

We're running this as an automated test, as a user we cannot control, so use `export PLOTLY_DIR=whatever`

re: #1076, addressed in #1195

This is still a problem with `plotly-3.7.1`. In `plotly/files.py`:
```py
 27 def _permissions():
 28     try:
 29         if not os.path.exists(PLOTLY_DIR):
 30             os.mkdir(PLOTLY_DIR)
 31         with open(TEST_FILE, 'w') as f:
 32             f.write('testing\n')
 33         os.remove(TEST_FILE)
 34         return True
 35     except:
 36         return False
```
Lines `30` and `33` can fail, wrongly, if two processes run this at the same time. (`f.write()` is fine here, IMO.)

Note that I want to make warnings into errors in my tests. (`pytest -W error`), so this is a big problem for me even if you simply trap and `warn()`.

I'm not sure if there is still a warning in 3.7.1, but like I said, it's a Heisenbug. I can only spend so much time on this. For me, it was happening via `pytest-xdist`, in a case where I lacked permissions to create `~/.plotly`. `PLOTLY_DIR` lets me reduce the likelihood of error, but the race conditions remain.

### DIFF
--- a/plotly/files.py
+++ b/plotly/files.py
@@ -27,12 +27,20 @@ FILE_CONTENT = {CREDENTIALS_FILE: {'username': '',
 def _permissions():
     try:
         if not os.path.exists(PLOTLY_DIR):
-            os.mkdir(PLOTLY_DIR)
+            try:
+                os.mkdir(PLOTLY_DIR)
+            except Exception:
+                # in case of race
+                if not os.path.isdir(PLOTLY_DIR):
+                    raise
         with open(TEST_FILE, 'w') as f:
             f.write('testing\n')
-        os.remove(TEST_FILE)
+        try:
+            os.remove(TEST_FILE)
+        except Exception:
+            pass
         return True
-    except:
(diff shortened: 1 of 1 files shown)


pb-cdunn:
Also, bare `except:` would trap `KeyboardInterrupt`, which is usually not intended.

jonmmease:
Thanks for digging into this @pb-cdunn, I haven't run into this before but your reasoning and these changes both make sense so I'm happy to merge this.  

Regarding warnings, my intention in #1195 was to only raise a warning if a user explicitly asks to perform an operation that needs write access to the `PLOTLY_DIR` directory, and this write cannot be performed.

Issue #1402, opened 2019-01-23
Problem plotting historical time series


elysiumfield:
I have recently updated plotly to ver 3.5.0 and am having problems plotting historical time series. The test code is as follows:

```python
import numpy as np
import pandas as pd
import plotly.graph_objs as go
import plotly.offline as pyo
dt = pd.date_range(start='1969-12-31', periods=270, freq='Q')
t1 = []
t1.append(dict(x=dt, y=np.ones(270), name='cape', type='scatter'))
layout = go.Layout(title="test", width=1800, height=930, font=dict(size=27), titlefont=dict(size=42))
fig = go.Figure(data=t1, layout=layout)
pyo.plot(fig, filename='test.html', auto_open=False, show_link=False)
```
This code returns the error:
```
  File "C:\ProgramData\Anaconda3\lib\site-packages\plotly\offline\offline.py", line 645, in plot
    '100%', '100%', global_requirejs=False)

  File "C:\ProgramData\Anaconda3\lib\site-packages\plotly\offline\offline.py", line 336, in _plot_html
    jdata = _json.dumps(figure.get('data', []), cls=utils.PlotlyJSONEncoder)

  File "C:\ProgramData\Anaconda3\lib\json\__init__.py", line 238, in dumps
    **kw).encode(obj)

  File "C:\ProgramData\Anaconda3\lib\site-packages\plotly\utils.py", line 168, in encode
    encoded_o = super(PlotlyJSONEncoder, self).encode(o)

  File "C:\ProgramData\Anaconda3\lib\json\encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)

  File "C:\ProgramData\Anaconda3\lib\json\encoder.py", line 257, in iterencode
    return _iterencode(o, 0)

  File "C:\ProgramData\Anaconda3\lib\site-packages\plotly\utils.py", line 233, in default
    return encoding_method(obj)

  File "C:\ProgramData\Anaconda3\lib\site-packages\plotly\utils.py", line 297, in encode_as_datetime
    obj = obj.astimezone(pytz.utc)

OSError: [Errno 22] Invalid argument
```

However, defining dt = pd.date_range(start='1970-03-31', periods=270, freq='Q'), everything works fine. I am positive that previous versions of plotly could plot datetimeindex that was older.

jonmmease:
Hi @elysiumfield, thanks for the report.

Could you add the versions of Python and pandas that you're using?

[...]

PR #1501, opened 2019-04-07
Fix for windows os error (#1402) Changed to use py3.6 behavior also with python 3.7 and above.

mprostock:
Fix for https://github.com/plotly/plotly.py/issues/1402.
Dash/Plotly work fine with python 3.5 and 3.6., but using python 3.7 creates uncaught OS Error exceptions from plotly utils.py.

Reason:
The variable PY36 in utils.py is used to branch code for different behaviour for datetime objects depending on the python version. This was set to 3.6 by using '==', this 1-character change ('>=') fixes the problems on windows when using plotly/dash with python 3.7. Without it 3.7 gets same behaviour as 3.5, which does not work. ;-)

### DIFF
--- a/plotly/utils.py
+++ b/plotly/utils.py
@@ -36,7 +36,7 @@ sage_all = get_module('sage.all')
 lock = threading.Lock()
 
 PY36 = (
-    sys.version_info.major == 3 and sys.version_info.minor == 6
+    sys.version_info.major == 3 and sys.version_info.minor >= 6
 )
 
 


jonmmease:
Great catch @mprostock, thanks for the PR!

nicolaskruchten:
I do have some concerns about a variable called `PY36` being `True` on 3.7 ... should we rename that variable to `PY36PLUS` or something throughout?

jonmmease:
My brain didn't mind interpreting `PY36` and >= 3.6, but I'd be happy to change the name the next time one of us has a chance.

PR #1509, opened 2019-04-09
changed PY36 constant naming to be more explicit for versions >=3.6

mprostock:
as suggested by @nicolaskruchten.
I took the liberty to make it even more explicit than suggested. 
(Just to finish what I started here. ;-) )


jonmmease:
Awesome. Thanks @mprostock!

PR #1526, opened 2019-04-17
undo overly-aggressive default

nicolaskruchten:
I'm not sure how to run the templategen script... just `python templategen/__init__.py` doesn't work for me...


jonmmease:
`python templategen/__init__.py` is supposed to be enough. Are you getting an error?

nicolaskruchten:
yes it gives me:

```
  File "templategen/__init__.py", line 3, in <module>
    from templategen.definitions import builders
ModuleNotFoundError: No module named 'templategen'
```

PR #1542, opened 2019-05-01
Adding support to plotly_unselect event

denphi:
When active selections are cleared using double-click interaction over a plotly plot,  python objects are not updated properly. This code enables plotly_deselect events to update points properly

### DIFF
--- a/js/src/Figure.js
+++ b/js/src/Figure.js
@@ -750,6 +750,10 @@ var FigureView = widgets.DOMWidgetView.extend({
                     function (update) {
                         that.handle_plotly_selected(update)
                     });
+                that.el.on("plotly_deselect",
+                    function (update) {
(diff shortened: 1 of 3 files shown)


jonmmease:
Thanks for the PR @denphi!  This will be a really helpful addition.  The code looks great.  I'll test it out over the next few days and then merge it in.

jonmmease:
Great work @denphi! I made only one small adjustment, rather than setting the trace's `selectedpoints` property to `()` when `plotly_deselect` happens I set it to `None`.  The reason is that plotly.js treats `None` to mean that no selection is active (and so no points are transparent) whereas `()` means that the selection mode is active but nothing is selected (and so all of the points are transparent).

Will merge when tests are green (all but plotlyjs_dev_build, which is known to be failing right now). Thanks again!

PR #1543, opened 2019-05-01
Fix package listing in setup.py

jakevdp:
The package listing in plotly's setup.py is non-standard; rather than file paths, it should list Python package paths (see https://docs.python.org/3/distutils/setupscript.html#writing-the-setup-script). This PR fixes the issue.

I tested this by running the following
```bash
$ python setup.py sdist
$ cat plotly.egg-info/top_level.txt
_plotly_future_
_plotly_utils
chart_studio
plotly
plotlywidget
```
The top-level package listing is now correct.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -377,12 +377,12 @@ class UpdatePlotlywidgetVersionCommand(Command):
 
 
 graph_objs_packages = [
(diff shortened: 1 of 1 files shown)


jonmmease:
Looks good, thanks @jakevdp!

PR #1534, opened 2019-04-23
traces selection/update methods and "magic underscore" support

jonmmease:
Closes https://github.com/plotly/plotly.py/issues/1481 and the trace portion of https://github.com/plotly/plotly.py/issues/1484.

## Overview
This PR introduces 3 new figure methods to make it easier to make updates to traces that have already been added to a Figure (e.g. a Figure from `plotly_express` or a figure factory). It also introduces a convenient shorthand for updating nested properties that we've been calling "magic underscore" support.

### New Figure methods
All 3 methods allow the user to select certain traces from the figure using a "selector". This is an `update` style dict that can be used to specify the desired values of arbitrary properties in the trace. For example, a selector of `{'type': 'scatter'}` will select all scatter traces in the figure.  A selector of `{'type': 'scatter', 'mode': 'markers'}` will select only those scatter traces that are in `'markers'` mode. 

In addition to the selector, for figures created using `plotly.subplots.make_subplots`, the trace row and column index may also be used as selection criteria.

The lowest level method is `select_traces`. This returns a generator that iterates over all of the traces in the figure that satisfy the selection criteria.

The `for_each_trace` method applies a custom function to each trace produced by `select_traces`

The `update_traces` method accepts a "patch" to update all of the traces that satisfy the selection criteria. This "patch" is a dict that is passed to `trace.update` for all selected traces.

### "Magic underscore" support
For background on "magic underscore" support, see https://github.com/plotly/plotly.py/pull/919 and https://github.com/plotly/plotly.py/issues/1481.

The goal of magic underscore support is to make it more convenient to reference nested figure properties without the need to created deeply nested `dict` instances.

Here are the updates that have been implemented:

#### Figure and graph_objs constructors
Here's a current approach to creating a figure with a particular title:
```python
import plotly.graph_objs as go
fig = go.Figure(layout={'title': {'text': 'Hello, Figure'}})
```
With magic-underscore support:
```python
fig = go.Figure(layout_title_text='Hello, Figure')
```

#### .update operations
Here's a current approach to updating the title text for an existing figure
```python
fig = go.Figure()
fig.update(layout={'title': {'text': 'Hello, Figure'}})
```
with magic-underscore support:

```python
fig = go.Figure()
fig.update(layout_title_text='Hello, Figure')
```
In addition to using underscores to separate nested keys, when the update argument is supplied as a dict (rather than as kwargs), the path parts can be separated by periods.

```python
fig = go.Figure()
[...]

Issue #1484, opened 2019-03-28
update_*() batch methods


nicolaskruchten:
Given that `px` produces figures which often have non-trivial numbers of fairly 'regular' axes/annotations/traces, it would be very convenient to have something like `Figure.update_traces( <patch>, <selector> )`-type methods e.g. 

```python
px.scatter(..., marginal_y="violin").update_traces(
    patch=dict(marker_line_color="black"), 
    selector=dict(type="scatter")
)
```

In this example, `px.scatter()` generates some `scatter` traces but also some `violin` ones, so the patch will only apply to the `scatter` traces. Also note that this example uses some magic-underscores from #1481 

This would also be very convenient for changing axis-level attributes in the case of a facetted `px` plot with many axes, ditto the `annotation`-based facet titles etc. In fact, I could see this being useful pretty much anywhere we have a collection of things either in a list (annotations, traces, shapes, etc) or with a number suffix (xaxis, yaxis, geo, scene, etc).

A more uniform API where all the method are called `batch_update` like `Figure.data.batch_update()` or `Figure.layout.xaxis.batch_update()` might work as well? The latter case maybe less nice, but there is no "collection" of x-axes on which to attach the method there :)

jonmmease:
I like it.  Here are some places that come to mind where I can see this being immediately useful.
```
fig.update_traces()
fig.layout.update_xaxes()
fig.layout.update_yaxes()
fig.layout.update_geos()
fig.layout.update_polars()
fig.layout.update_scenes()
fig.layout.update_annotations()
fig.layout.update_shapes()

parcoords.update_dimensions()
parcats.update_dimensions()
```

Another option, would be to support JSONPath-style keys in the existing `.update` method (See https://github.com/json-path/JsonPath#path-examples).
```
# Update marker line color of all scatter traces
fig.update({'data[?(@.type == 'scatter')].marker.line.color': 'green'})

# Update colorbar title font family across all colorbars, and all nesting levels
fig.update({'..colorbar.title.font.family': 'Rockwell'})
```

This would offer pretty complete flexibility, at the cost of perhaps a slightly higher learning curve.  I'm also not sure how much we could reuse from existing libraries like https://github.com/kennknowles/python-jsonpath-rw.

nicolaskruchten:
I think the JSONPath idea is a good one, but likely not a total replacement for either what I'm proposing here or magic-underscores... 

[...]

PR #1548, opened 2019-05-04
subplot select/update methods for batch updates

jonmmease:
## Overview
This PR is part two of https://github.com/plotly/plotly.py/issues/1484 and a follow-on to https://github.com/plotly/plotly.py/pull/1534.

This adds `select_*`, `for_each_*`, and `update_*` methods to the figure class for each subplot container type.  These methods work exactly like the trace counterparts from https://github.com/plotly/plotly.py/pull/1534.

These will be especially useful when used with `plotly_express`.  Note that the `row`/`col` filtering will not work with `plotly_express` until it is updated to use the new v4 subplot logic (https://github.com/plotly/plotly.py/pull/1528) for faceting.  But the `selector` based filtering should work now.

## Examples
Here are some examples:

Set the `tickprefix` of all xaxis objects to '$';
```
fig.update_xaxes({'tickprefix': '$'})
```
Or only those in the first column
```
fig.update_xaxes({'tickprefix': '$'}, col=1)
```

In addition to row/col filters. updates can also be filtered on selectors as described in https://github.com/plotly/plotly.py/pull/1534.

So if, for some reason, you needed to update only the y axis objects that were configured in log scale to show length 5 ticks:
```
fig.update_yaxes({'ticks': 'outside', 'ticklen': 5}, selector={'type': 'log'})
```

## Other notes
With this PR, we now rely on code generation to determine the set of subplot types.  So when new ones show up, like coloraxis, these should now be handled automatically.

cc @nicolaskruchten

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -297,7 +297,7 @@ jobs:
       - checkout
       - run:
           name: Install tox
-          command: 'sudo pip install tox requests yapf pytz decorator retrying'
+          command: 'sudo pip install tox requests yapf pytz decorator retrying inflect'
       - run:
           name: Update plotlywidget version
           command: 'python setup.py updateplotlywidgetversion'
--- a/_plotly_utils/utils.py
+++ b/_plotly_utils/utils.py
@@ -2,6 +2,7 @@ import datetime
 import decimal
 import json as _json
 import sys
+import re
 
 import pytz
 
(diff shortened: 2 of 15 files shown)


nicolaskruchten:
huzzah! 🎉

jonmmease:
Merging!

Issue #1556, opened 2019-05-07
PlotlyJSONEncoder doesn't work with simplejson installed


jayvdb:
Similar to past issues (https://github.com/plotly/plotly.py/issues/415), installation of simplejson breaks plotly.py.

Example failure:
```py
======================================================================
ERROR: test_default_plot_generates_expected_html (test_offline.PlotlyOfflineTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/abuild/rpmbuild/BUILD/plotly.py-3.9.0/plotly/tests/test_core/test_offline/test_offline.py", line 92, in test_default_plot_generates_expected_html
    cls=plotly.utils.PlotlyJSONEncoder)
  File "/usr/lib64/python2.7/site-packages/simplejson/__init__.py", line 419, in dumps
    **kw).encode(obj)
TypeError: __init__() got an unexpected keyword argument 'iterable_as_array'
```

The build environment can be seen at https://build.opensuse.org/package/show/home:jayvdb:py-wheels/python-plotly

jonmmease:
Thanks for the report @jayvdb,

Looks like the problem is that we use this snippet in various places in the codebase

```python
from requests.compat import json as _json
```

Which, in turn, calls
```python
try:
    import simplejson as json
except ImportError:
    import json
```

This was done in https://github.com/plotly/plotly.py/commit/66349716fa9f3106eebcf0d0b3d61791a129bd06 in 2016.  I believe at the time this was done as a way to support Python < 2.6.  Plotly.py no longer supports Python < 2.7 so I think we should just remove this compatibility logic and always use the built-in `json` module.

jayvdb:
Bingo that solved the errors in my builds.  Submitted a PR.

PR #1561, opened 2019-05-10
Replace requests.compat.json with real json

jayvdb:
requests.compat.json may use simplejson, which is incompatible
with plotly.py usage of json.

Fixes https://github.com/plotly/plotly.py/issues/1556


jonmmease:
Thanks a lot @jayvdb, we'll pull this into the next release.

jonmmease:
Looks good. Thanks agaikn @jayvdb

PR #1574, opened 2019-05-23
Treat 'coloraxis1' as 'coloraxis'

jonmmease:
Small fix to be able to treat 'coloraxis1' as 'coloraxis' in coloraxis subplot validator


nicolaskruchten:
ref https://github.com/plotly/plotly.js/issues/3892

nicolaskruchten:
just tested it, works great.  💃 for behaviour and it didn't break anything else in px :)

PR #1576, opened 2019-05-23
Fix honouring of `validate=False` for interactive plots.

matthewwardrop:
Greetings!

We have been working with Plotly recently and attempting to plot Plotly JSON blobs generated from R in Python, and (as it happens) these blobs are not conformant to the Plotly JSON schema. Passing `validate=False` to the `plotly.offline.plot` method works fine, but not for `plotly.offline.iplot`, due to validation being enabled by default for some of the rendering backends. One approach to fixing this would be to pass `validate` down to the renderers, but it appears based on `validate=False` being passed by default to some backends that this may have been an oversight in the default passed to these mimebundle renderers. This patch makes this consistent, and causes all mimebundle generations to no longer validate. Since validation would have happened earlier if not explicitly passing `validate=False`, I believe this should be okay.

### DIFF
--- a/plotly/io/_base_renderers.py
+++ b/plotly/io/_base_renderers.py
@@ -122,7 +122,9 @@ class ImageRenderer(MimetypeRenderer):
             format=self.format,
             width=self.width,
             height=self.height,
-            scale=self.scale)
+            scale=self.scale,
+            validate=False,
+        )
 
         if self.b64_encode:
             image_str = base64.b64encode(image_bytes).decode('utf8')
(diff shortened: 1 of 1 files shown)


jonmmease:
Thanks for digging in on this @matthewwardrop, I think you're right that this was an oversight. I'll take a closer look soon and pull this in for 3.10.0. Thanks!

matthewwardrop:
Thanks for the prompt response @jonmmease :).

jonmmease:
Looks great. Thanks again @matthewwardrop

PR #1577, opened 2019-05-24
Scraper function for sphinx-gallery, in plotly.io

emmanuelle:
This PR proposes the addition of a scraper function in ``plotly.io``, to be used by the ``sphinx-gallery`` sphinx extension when a sphinx-based doc contains a gallery of examples. At the moment ``sphinx-gallery`` has a scraper for matplotlib figures, but some open-source projects have expressed their interest in having figures with plotly.py instead. 

The PR also contains a minimal sphinx doc with two gallery examples (one for 2D, one for 3D). It's not necessary to merge this but I thought that having gallery examples in the code base could be useful.

Note that this PR does not require a sphinx-gallery dependency, because it will be sphinx-gallery which calls the scraper function and not the other way around. 

One thing to decide is whether the function used in examples is ``plot`` or ``iplot``. Do most users work in notebooks and therefore use ``iplot``? The function is monkey-patched anyway so it's easy to change.


emmanuelle:
I'm struggling with installing the orca excutable on the CI ``optional`` build. Any idea of how to solve this?

jonmmease:
Hi @emmanuelle, thanks this is very interesting.

A couple of high-level thoughts.  Rather than use monkey-patching it would be nice to handle this use-case with a custom renderer (Renderers were introduced in https://github.com/plotly/plotly.py/pull/1474).

I'm also a little hesitant to introduce this logic into the plotly.py repo.  Would it make sense to add this to sphinx gallery?

jonmmease:
> I'm struggling with installing the orca excutable on the CI optional build. Any idea of how to solve this?

So far, I've had all of the orca-dependent tests run in the `orca` builds.  These builds install orca using conda

emmanuelle:
Thanks for the comments @jonmmease ! Yes, a custom renderer would be nicer, I just read through #1474 and it seems well adapted.

I initially proposed the scraper function to sphinx-gallery as a PR (https://github.com/sphinx-gallery/sphinx-gallery/pull/493) but their policy is that each plotting library should maintain its own scraper, which kinda makes sense since there is a large number of plotting libraries.  So if we want that other projects use plotly in their documentation when using sphinx-gallery (it was mentioned on the scikit-learn mailing-list that there is interest for this), I think we have to include it and maintain it...

jonmmease:
Thanks for the background @emmanuelle, having something in the plotly.py repo is fine in that case.  Looking over the code, I think we could replace the `_patched_plotly_plot` logic with a custom renderer.

Would it be possible to get away with only having a custom renderer and not the `plotly_sg_scraper` and `figure_rst` logic?  Can we make the output of plotly look just like matplotlib so that scraper built into sphinx gallery picks it up?  It doesn't *look* like there's anything in these two functions that's specific to how plotly.py works, so it would be nice to avoid them if possible. Thanks!

emmanuelle:
Thanks @jonmmease. The way sphing-gallery handles matplotlib examples is quite different, because it loops through the list of existing matplotlib figures with ``plt.get_fignums`` (https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/scrapers.py#L87), and we don't have such a list in plotly.py if I understand well. Therefore for plotly figures what I did instead was to save the figure as files and then move the files to the right location. And the ``figure_rst`` looks very much like the ``figure_rst`` of sphinx-gallery except that it's html and not a png file so another rst directive is needed. But I could ask the sphinx-gallery team if they would accept a PR for extending their ``figure_rst`` function to other formats, they didn't want to maintain the scrapers but a ``figure_rst`` function with html support might be ok for them.

emmanuelle:
@jonmmease here is a version with a custom renderer. I can add more documentation and comments but first tell me whether this was what you had in mind. I also changed the examples so that now they call ``plotly.io.show()`` since this is the syntax we'll be pushing now.

emmanuelle:
Oh and the tests failure seem to be due to something else than the tests added by this PR.

plotly/io/_renderers.py
```diff
@@ -394,6 +395,7 @@ def show(fig, renderer=None, validate=True, **kwargs):
 renderers['chrome'] = BrowserRenderer(config=config, using='chrome')
 renderers['chromium'] = BrowserRenderer(config=config, using='chromium')
 renderers['iframe'] = IFrameRenderer(config=config)
+renderers['sphinx'] = SphinxGalleryRenderer()
```
jonmmease: Will this be useful in plain sphinx (without sphinx gallery)?  If not, lets name the renderer `'sphinx_gallery'`

plotly/io/_sg_scraper.py
```diff
@@ -0,0 +1,103 @@
+# This module defines an image scraper for sphinx-gallery
+# https://sphinx-gallery.github.io/
+# which can be used by projects using plotly in their documentation.
+# This module monkey-patches plotly.offline.plot, so we do not
```
jonmmease: We can now remove mention of monkey patching right?

jonmmease:
Thanks for making the renderer update @emmanuelle, this looks nice. I left a couple of comments inline.

>  But I could ask the sphinx-gallery team if they would accept a PR for extending their figure_rst function to other formats, they didn't want to maintain the scrapers but a figure_rst function with html support might be ok for them.

Sounds like it would be worth asking, as I'd expect this to be useful for other libraries as well.  But it's not a lot of logic, so not a huge deal.

> Oh and the tests failure seem to be due to something else than the tests added by this PR.

Yeah, this has been fixed on master so if you merge/rebase it should go away.

Regarding the example in `doc/conf.py`, I think I'd rather have this moved to a separate repository that has only the sphinx gallery example.  If that sounds alright, could you move this to a new repo in the plotly org and remove it from this PR?  Thanks!

emmanuelle:
Hurray it's green! I think I addressed all your requests, I moved the sphinx files and gallery examples to plotly/plotly-sphinx-gallery.

emmanuelle:
@jonmmease no hurry but since you're always very reactive, I just wanted to make sure that this hasn't fallen off your radar :-).

jonmmease:
haha, thanks for the ping. I haven't forgotten :slightly_smiling_face: I should be able to take a look tomorrow.  BTW, are you fine leaving this for v4 or would you like to get it out into the word before that?

emmanuelle:
v4 is fine, it should be in a month or so right? I'm used to slower development cycles so no pb :-).

plotly/tests/test_optional/test_jupyter/package.json
```diff
@@ -15,6 +15,8 @@
     "ecstatic": "^2.1.0",
     "tap-parser": "^2.0.0",
     "tape": "^4.6.0",
-    "xhr": "^2.2.2"
+    "xhr": "^2.2.2",
+    "electron": "^1.8.4",
+    "orca": "^1.2.1"
```
jonmmease: I wouldn't have expected this to be needed with the scraper tests living under test_orca. Was this a holdover from earlier experiments?

plotly/io/_base_renderers.py
```diff
@@ -6,10 +6,12 @@
 import os
 
 import six
-from plotly.io import to_json, to_image
+from plotly.io import to_json, to_image, write_image
 from plotly import utils, optional_imports
 from plotly.io._orca import ensure_server
 from plotly.offline.offline import _get_jconfig, get_plotlyjs
+from plotly.offline import plot
```
jonmmease: Let's remove this import and use `plotly.io.write_html` below

plotly/io/_base_renderers.py
```diff
@@ -6,10 +6,12 @@
 import os
 
 import six
-from plotly.io import to_json, to_image
+from plotly.io import to_json, to_image, write_image
```
jonmmease: add `write_html`

plotly/io/_base_renderers.py
```diff
@@ -637,3 +639,21 @@ def render(self, fig_dict):
...
+        # Name of script from which plot function was called is retrieved
+        try:
+            filename = stack[3].filename # let's hope this is robust...
+        except: #python 2
+            filename = stack[3][1]
+        filename_root, _ = os.path.splitext(filename)
+        filename_html = filename_root + '.html'
+        filename_png = filename_root + '.png'
+        figure = return_figure_from_figure_or_data(fig_dict, True)
+        _ = plot(fig_dict, auto_open=False,
```
jonmmease: Let's change this from `plotly.offline.plot` to `plotly.io.write_html`, since that is the recommended way to write figures to HTML now.

emmanuelle:
@jonmmease sorry to have let this idle for some time. I think the tests failure are not related to this PR. Shall I try to make a new PR to the `v4_integration` branch instead?

jonmmease:
Looks, good. Thanks!

jonmmease:
@emmanuelle, could you double check that your sphinx-gallery example works with the `plotly==4.0.0a9` alpha release?  Thanks!

emmanuelle:
Yes it works! Thank you very much @jonmmease :-)

Issue #1559, opened 2019-05-09
Set default renderer with environment variable


jonmmease:
It would be helpful to be able to set the default renderer (see https://github.com/plotly/plotly.py/pull/1474) using an environment variable like `PLOTLY_RENDERER`.  This would be particularly useful when building documentation.

PR #1578, opened 2019-05-25
Support specifying default renderer using PLOTLY_RENDERER env variable

jonmmease:
Closes https://github.com/plotly/plotly.py/issues/1559

Support specifying the default renderer using the `PLOTLY_RENDERER` environment variable.  This environment variable is only checked when the `renderer_defaults` future flag is set.

cc @nicolaskruchten

### DIFF
--- a/plotly/io/_renderers.py
+++ b/plotly/io/_renderers.py
@@ -401,15 +401,27 @@ if 'renderer_defaults' in _future_flags:
     # Version 4 renderer configuration
     default_renderer = None
 
-    # Try to detect environment so that we can enable a useful
-    # default renderer
-    if ipython and ipython.get_ipython():
+    # Handle the PLOTLY_RENDERER environment variable
+    env_renderer = os.environ.get('PLOTLY_RENDERER', None)
+    if env_renderer:
         try:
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
2 questions:

1. should we temporarily have an environment variable to set things in v4 mode also?
2. do we need a new renderer for documentation or will defaulting to `notebook_connected` work for `nbconvert`ing notebooks to HTML?

jonmmease:
> 1. should we temporarily have an environment variable to set things in v4 mode also?

My first reaction is that this would be a bit too magical.  I'm comfortable with changing the renderer based on an external environment variable since rendering is inherently context dependent  But turning on v4 mode externally would change the semantics of the code with no trace of why it changed.

> 2. do we need a new renderer for documentation or will defaulting to `notebook_connected` work for nbconverting notebooks to HTML?

`notebook_connected` should work fine through nbconvert.  But adding a custom renderer for docs is certainly an option if we run into a need for it.

nicolaskruchten:
For 1 I agree in general of course I just through it might make doc writing a bit smoother until the launch :) no big deal either way

PR #1590, opened 2019-05-31
Bug fix - animation_opts docstring example

Jonathan-MW:
Bug fix: The example for the animation_opts argument in the doc strings of the plot and iplot functions doesn't work.


jonmmease:
Thanks @Jonathan-MW!

PR #1592, opened 2019-06-02
Mapbox raster

hishamkaram:
This PR include:
- [ ] add Raster Layer(EX: geoserver wms layer)

**plotly.js Pull Request [Link](https://github.com/plotly/plotly.js/pull/3928)**


jonmmease:
Hi @hishamkaram, thanks for taking a look at this.  The changes you made in this PR won't be necessary because the `plotly.validators` classes are code generated from the plotly.js schema.  So once your changes are accepted into Plotly.js and released they will automatically be available in plotly.py in the following plotly.py release.  Thanks!

jonmmease:
See `codegen/validators.py` for the validators code generation logic if you're interested in seeing how it works.

PR #1598, opened 2019-06-08
county_choropleth figure factory should return a go.Figure

emmanuelle:
Should figure factories return a `go.Figure` object? It would be nicer to be able to do `fig.show()` on the returned object.


emmanuelle:
I could easily fix the test problem, I just want to know before whether this change makes sense.

nicolaskruchten:
New PRs should probably branch from and  target the v4 integration branch starting now right @jonmmease ?

emmanuelle:
Closing because of #1599.

PR #1599, opened 2019-06-09
County choropleth

emmanuelle:
Replaces #1598, now branching from `v4_integration`. 

With this PR, the `create_choropleth` figure factory returns a `go.Figure`, so that it is possible to call `fig.show()`.


emmanuelle:
Oops forgo to target the `v4_integration` branch, closing and opening a new one.

PR #1600, opened 2019-06-09
create_choropleth figure factory returns go.Figure

emmanuelle:
With this PR, the `create_choropleth` returns a `go.Figure`, so that `fig.show()` is possible on the returned figure.


jonmmease:
Thanks @emmanuelle, this looks good and will be a great update to make things more consistent for v4.

PR #1615, opened 2019-06-15
don't override plotlyServerURL if one is set in iplot/plot

### LINKED ISSUE #1609: plotlyServerURL config option overridden by plotly_domain from ~/.plotly/.config
Setting plotlyServerURL in the config dict passed to iplot will always create a plot where the "Edit in Chart Studio" button sends the plot to the base url defined by plotly_domain. 

This code won't change the plotly url used by the "Edit in Chart Studio" button
```python
import plotly.offline as py
py.init_notebook_mode()

py.iplot([go.Scatter(y=[2,3,4])], config=dict(showSendToCloud = True, plotlyServerURL='https://myserver.com'))
```

and this code will

```python
import plotly.offline as py
import plotly.tools as tls
py.init_notebook_mode()

tls.set_config_file(plotly_domain='https://www.myserver.com')
py.iplot([go.Scatter(y=[2,3,4])], config=dict(showSendToCloud = True))
```

I think it would be good to add a check to see if plotlyServerURL is already present in the config dict before setting it to plotly_domain ([relevant code](https://github.com/plotly/plotly.py/blob/d0950b8085d26c2025333f0114c419b70e8391fc/plotly/offline/offline.py#L158-L164)).

michaelbabyn:
This resolves #1609.

I tested this out locally using so that first example mentioned in the issue which didn't set the base url for the edit in chart studio button before the change does after.

```python
import plotly.offline as py
py.init_notebook_mode()

py.iplot([go.Scatter(y=[2,3,4])], config=dict(showSendToCloud = True, plotlyServerURL='https://myserver.com'))
```

### DIFF
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -161,7 +161,7 @@ Unrecognized config options supplied: {bad_config}"""
 
     plotly_platform_url = plotly.tools.get_config_plotly_server_url()
 
-    if not clean_config.get('plotlyServerURL'):
+    if not clean_config.get('plotlyServerURL', None):
         clean_config['plotlyServerURL'] = plotly_platform_url
 
     if (plotly_platform_url != 'https://plot.ly' and
--- a/plotly/offline/offline.py
+++ b/plotly/offline/offline.py
@@ -161,7 +161,8 @@ Unrecognized config options supplied: {bad_config}"""
 
     plotly_platform_url = plotly.tools.get_config_plotly_server_url()
 
-    clean_config['plotlyServerURL'] = plotly_platform_url
+    if not clean_config.get('plotlyServerURL'):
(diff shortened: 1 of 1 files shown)


plotly/offline/offline.py
```diff
@@ -161,7 +161,8 @@ def _get_jconfig(config):
 
     plotly_platform_url = plotly.tools.get_config_plotly_server_url()
 
-    clean_config['plotlyServerURL'] = plotly_platform_url
+    if not clean_config.get('plotlyServerURL'):
```
jonmmease: I think we want `clean_config.get('plotlyServerURL', None)` in case config doesn't include `'plotlyServerURL'` at all.

jonmmease:
Great, thanks @michaelbabyn!

PR #1623, opened 2019-06-17
V4 merge plotly jupyterlab extensions

jonmmease:
## Overview
This PR merges the current `plotlywidget` JupyterLab extension and the [`@jupyterlab/plotly-extension`](https://github.com/jupyterlab/jupyter-renderers/tree/master/packages/plotly-extension) JupyterLab extension into a single JupyterLab extension simply named `jupyterlab-plotly`.

This one extension will handle displaying `FigureWidget` objects, and it will handle displaying `Figure` objects by their mimetype (the current `plotly.offline.iplot` approach).

## Other changes
I made two changes to the mimetype rendering with respect to the `@jupyterlab/plotly-extension` extension.

### Static image preview on load
The current `@jupyterlab/plotly-extension` extension automatically saves a PNG rendering of each figure back to the notebook as soon as it is displayed. This makes it possible to export the notebook to static output formats like PDF and get a static image rendering of each plot.  The `jupyterlab-plotly` extension will display this static image rendering of each figure (if it is available)  when the notebook is first loaded.  This static image will then be replaced with a Plotly.js figure when the user mouses over the image.

This workflow makes a significant improvement in notebook load time for notebooks that contain lots of plotly figures.  For the intensive notebook I've been experimenting on (20+ figures), this gives a load time improvement of 40s down to 7s.

### Remove auto-animate behavior
The `@jupyterlab/plotly-extension` extension automatically animates any figures that have frames.  Now that we've started making notebooks with plotly express with lots of figures involving frames, it's become apparent that this is not a great default.  So this PR just removes the auto-animate behavior.  In the future, we could add support for the `auto_play` argument to `plotly.offline.plot` if folks find that they need this auto play behavior back.

cc @timkpaine @gnestor


timkpaine:
"This makes it possible to export the notebook to static output formats like PDF and get a static image rendering of each plot"

Fantastic!

gnestor:
This looks great, @jonmmease! Should I unpublish @jupyterlab/plotly and add a deprecation notice to npm and the jupyter-renderers repo?

jonmmease:
Thanks @gnestor! yeah let's do those things once we ship plotly.py v4.  I'll ping you when we get to that point.

saulshanabrook:
@jonmmease Do you think that will line up with our 1.0 release? Here is our timeline: https://github.com/jupyterlab/jupyterlab/issues/6504#issue-453727845 The plotly jupyterlab extension is one of the most popular downloaded extensions, so it would be nice if there there was a 1.0 compatible one that is out by SciPy

jonmmease:
Hi @saulshanabrook, pre-SciPy is our goal too, so that should line up well. Though I'd estimate that we'll probably slip into the first week of July.  I haven't tested this against the JupyterLab 1.0 API yet, any Mimetype renderer or ipywidgets compatibility issues to watch out for? Thanks for reaching out!

PR #1643, opened 2019-06-28
Fix template error bar coloring for bar plots

Olof-Hojvall:
Previous addition of bar outline in templates also changed color of error bars, since errorbar seems to use the colorof marker line over default. This shows a minimum change to fix it back, but I suspect that it would be preferably to set correct color it a better way. Would be happy to adjust the PR, if any suggestions on that could be done. Maybe something like default_line_color should be passed as argument like initialize_template(..., default_line_color=GRAY27) to allow templates to set the color of error bars.

See this notebooks for error and fix: https://nbviewer.jupyter.org/github/Olof-Hojvall/plotly.py/blob/error_bar_notebook/Bar%20Plot%20Errorbar%20Template%20Example.ipynb

### DIFF
--- a/templategen/utils/__init__.py
+++ b/templategen/utils/__init__.py
@@ -1,4 +1,5 @@
 from plotly import graph_objs as go
+from .colors import GRAY27
 
 colorscale_parent_paths = [
     ('histogram2dcontour',),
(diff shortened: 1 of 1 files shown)


jonmmease:
Thanks a lot for the report and fix @Olof-Hojvall.  I just opened https://github.com/plotly/plotly.py/pull/1656 which has a small modification to your updates here, but built against the `v4_integration` branch.  We're very close to the version 4.0 release, so new changes are going into that branch right now.

The modification I made is to set the error bar color to the template's font color.  This way the same approach works for light and dark themes.  See examples in the PR.  Does that coloring approach work for you?

Olof-Hojvall:
Thanks, that looks good. I’ll close this one.

PR #1656, opened 2019-07-03
Error bar template colors

jonmmease:
Replacement for #1643 against the `v4_integration` branch.  Thanks to @Olof-Hojvall for the report and fix.

This sets the error bar color in templates, for bar traces, to the font color.  Before it was defaulting to the bar's line color, which is the background color, making the error bars not visible.

Before:
newplot (2)

After (`plotly` template):
newplot (3)

Afrer (`plotly_dark` template):
newplot (4)


nicolaskruchten:
works for me! 💃

PR #1657, opened 2019-07-04
Fix typo in the documentation.

jkaluzka:
Fix typo in the docstring.


jonmmease:
Hi @jkaluzka, Thanks for taking the time to look into this.  The comment that you edited here is generated automatically from the Plotly.js schema (Plotly.js is the JavaScript library that plotly.py wraps), so this will need to be fixed in that project, and then it will automatically be fixed here.

I just opened https://github.com/plotly/plotly.js/issues/4054 to track the issue, and I'm going to close this PR in favor of that issue.  Are you interested in making the change over there?

PR #1670, opened 2019-07-19
FIX: fix incorrect import in deprecation warning for Scene class

stevenjkern:
I noticed a minor issue in the text of a deprecation warning on the `Scene` class the other day, where the recommended fully-qualified class name is the same as the one being used:
```
In [1]: from plotly.graph_objs import Scene

In [2]: Scene()
/Users/skern/.edm/envs/mb_test_plotly3/lib/python3.6/site-packages/plotly/graph_objs/_deprecations.py:477: DeprecationWarning:

plotly.graph_objs.Scene is deprecated.
Please replace it with one of the following more specific types
  - plotly.graph_objs.Scene
```

The correct class that should be recommended is `plotly.graph_objs.layout.Scene`. This PR just corrects the few occurrences of that.

### DIFF
--- a/packages/python/plotly/codegen/compatibility.py
+++ b/packages/python/plotly/codegen/compatibility.py
@@ -26,7 +26,7 @@ DEPRECATED_DATATYPES = {
     "Margin": {"base_type": dict, "new": ["layout"]},
     "Marker": {"base_type": dict, "new": ["scatter", "histogram.selected", "etc."]},
     "RadialAxis": {"base_type": dict, "new": ["layout", "layout.polar"]},
-    "Scene": {"base_type": dict, "new": ["Scene"]},
+    "Scene": {"base_type": dict, "new": ["layout"]},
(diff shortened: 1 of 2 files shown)


jonmmease:
Hi @stevenjkern,
Thanks for pointing this out! The files that you modified here are automatically generated during the code generation process, so we we'll need to make these changes in the code generation logic so that they are not overwritten when plotly.js is updated.  You'll need to update this line in the code generation logic:

https://github.com/plotly/plotly.py/blob/c576adf6e89354eef02a0311702c67e0ed178474/packages/python/plotly/codegen/compatibility.py#L29

to replace `"new": ["Scene"]` with `"new": ["layout"]` and then rerun code generation with

```
$ pip install black  # If you don't have black installed yet
$ python setup.py codegen
```

Thanks!

jonmmease:
I added the codegen logic. Thanks again for pointing this out @stevenjkern!

PR #1694, opened 2019-07-29
Custom template for trace labels

weber-s:
Hi!

Thank you so much for plotly express!

Previously, I was using custom hand-made grouper loop to do what I can do now in one line.
However, I had full control of the label of the trace, which is no more the case.

This PR aims to give full access to the trace label, thanks to `template_label` kwarg.
I just made it available for `scatter` right now, and didn't run any test yet (I'm default to python 3.5, have to setup env and co first...). 

Are you interested in such feature? If so, I may continue on it.

Example:

```python
data = {"group": ["group1"]*2 + ["group2"]*2, "color": ["1","1","1","2"], "value": [1,2,3,4], "x": [10, 20, 30, 40]} 

df = pd.DataFrame(data) 
fig = px.scatter(df, x="x", y="value", color="color", symbol="group", 
             template_label="{color_key} = {color_val} ({toto_val})", 
             labels={"group": "toto", "color": "color"} 
) 
fig.show()
```

trace_labels_px

PS : I needed it because I have plots with `station` and `specie`, and all traces was displayed as `station=A, specie=SP1`, and I want to display only `A (SP1)`

### DIFF
--- a/packages/python/plotly/plotly/express/_chart_types.py
+++ b/packages/python/plotly/plotly/express/_chart_types.py
@@ -45,6 +45,7 @@ def scatter(
     template=None,
     width=None,
     height=None,
+    template_label=None,
 ):
     """
     In a scatter plot, each row of `data_frame` is represented by a symbol mark in 2D space.
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -711,6 +711,10 @@ def apply_default_cascade(args):
     if args.get("facet_row", None) and args.get("marginal_x", None):
         args["marginal_x"] = None
 
+    if "template_label" in args:
+        if args["template_label"] is None:
+            args["template_label"] = "{key}={val}"
+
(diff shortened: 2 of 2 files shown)


nicolaskruchten:
Thanks for the PR!

I'm a bit hesitant to add more kwargs to the various `px` functions at the moment... What you're describing can be done today in a less declarative way with something like:

 ```python
px.scatter(...).for_each_trace(lambda t: t.update(name= <something here> ))
```

where `<something here>` can be a regex or somesuch...

What do you think?

weber-s:
I didn't know about the `.for_each_trace`. Very useful one!

Indeed, something like `re.sub("station=(.*), specie=(.*)$", "\\1 (\\2)", t.name)` should work fine. Thanks for the trick!

I am not sure this is the most intuitive way however. Nor that the `template_label` kwarg is a more intuitive one.

We can close this PR since there is already a way to achieve what it aims to do.

PR #1704, opened 2019-08-04
Release 4.1.0

jonmmease:
Version 4.1.0 release branch

PR #1707, opened 2019-08-06
Fix row title order on PX facets

nicolaskruchten:


jonmmease:
Thanks @nicolaskruchten, I ended up doing this with a slightly upstream fix in https://github.com/plotly/plotly.py/pull/1704/commits/7d4b76bf988a820cb2d960e1d7bf46192f099249.  It amounts to the same thing but is a little more direct.

Issue #1211, opened 2018-10-03
create_gantt figure should be responsive


brylie:
The `figure_factory` `create_gantt` method outputs a figure with static dimensions. As most other Plotly charts are responsive, so should be the GANTT. Please allow the GANTT chart to be responsive to page/device dimensions, rather than expecting hard-coded width and height dimensions.

gvwilson:
Hi - we are currently trying to tidy up Plotly's public repositories to help us focus our efforts on things that will help users most. Since this issue has been sitting for several years, I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our backlog. Thanks for your help - @gvwilson

PR #1724, opened 2019-08-15
responsive Gantt-chart

csabaszan:
Hi,

#1211 mentioned that the created figure is not responsive, so I removed the default values for width and height of the figure.

Csaba


jonmmease:
Hi @csabaszan, thanks for the PR. I'm happy for the gantt to be responsive horizontally by default.  But I think we should keep the default height of 600 so that, in the notebook, gantt charts aren't displayed shorter than they are now.  Does that work for you?  If responsive height is also desired for HTML export, then the `height=None` argument could still be set.

csabaszan:
Hi @jonmmease, yeah only responsive width by default makes more sense to me as well.

jonmmease:
Thanks @csabaszan!

Issue #1717, opened 2019-08-09
overwrite argument for update methods


nicolaskruchten:
This is the only way we can do stuff like `.update_layout(annotations=None)`

PR #1726, opened 2019-08-17
Add "overwrite" kwarg to all update* figure methods.

jonmmease:
Closes https://github.com/plotly/plotly.py/issues/1717

This PR adds a new `overwrite` kwarg to all of the figure `update*` methods.  The default behavior, with `overwrite=False`, is to apply updates recursively to the existing nested property structure.  With `overwrite=True`, the prior value of existing properties is overwritten with the provided value.

### DIFF
--- a/packages/python/plotly/codegen/figure.py
+++ b/packages/python/plotly/codegen/figure.py
@@ -313,6 +313,7 @@ class {fig_classname}({base_classname}):\n"""
             self,
             patch=None,
             selector=None,
+            overwrite=False,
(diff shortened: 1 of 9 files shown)


emmanuelle:
@nicolaskruchten @jonmmease sorry I forgot about this one, I'll review it soon!

emmanuelle:
Looks good to me, thanks :dancer: We should also update the tutorial https://plot.ly/python/creating-and-updating-figures/ with an example of `overwrite=True` when the release comes out.

PR #1754, opened 2019-09-05
added minimal test for px

emmanuelle:
Of course many more tests for px, but I'm opening this PR as a placeholder to comment about how px tests should look like.

### DIFF
--- a/packages/python/plotly/plotly/tests/test_core/test_px/test_px.py
+++ b/packages/python/plotly/plotly/tests/test_core/test_px/test_px.py
@@ -1,0 +1,13 @@
+import plotly.express as px
+import numpy as np
+
+def test_scatter():
+    iris = px.data.iris()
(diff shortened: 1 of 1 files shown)


emmanuelle:
I was thinking of testing all the examples of the tutorial written by @nicolaskruchten so that we know that at least those function calls will execute.

emmanuelle:
I'm doing this since I'm about to start playing with the API of px so I'd prefer to know quickly what I'm breaking :-)

emmanuelle:
Pandas is an optional dependency for plotly.py, but is it ok to have it as a required dependency for testing?

nicolaskruchten:
I'm fine with this approach as a basic smoke test, but I wouldn't go super-far with individual attribute spot-checks, given how many there are!

nicolaskruchten:
💃 once rebased onto `master` with a passing build

PR #1763, opened 2019-09-09
new regressions Huber and Theil-Sen for trendline

martins0n:
Hi! I've added  two robust regressions for trendlines. There's an import from scipy, maybe it's not great. But I think that robust regressions will be helpful for analyzing data.

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -46,7 +46,7 @@ def set_mapbox_access_token(token):
 def get_trendline_results(fig):
     """
     Extracts fit statistics for trendlines (when applied to figures generated with
-    the `trendline` argument set to `"ols"`).
+    the `trendline` argument set to `"ols"` and `"huber"`).
 
     Arguments:
         fig: the output of a `plotly_express` charting call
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
Thanks for this PR! We've been thinking about an extension mechanism for this "trendline" feature for a while, as an alternative to baking in an increasing number of types of regressions directly into the library. I was thinking of having this argument accept a function with a signature like `(x_in, y_in) -> (y_out, results, hover_string)` or something like that, so users like yourself would be able to define/tweak/pass-in different regression types.

The advantage to this function approach is that you could do something like the following:

```python
def huber(t):
    def huber_impl(x_in, y_in):
        # math that depends on `t`
        return y_out, results, hover_string
    return huber_impl

px.scatter(..., trendline=huber(t=2.222))
```

In addition to this, I'd want to include some kind of `px.trendline_functions` dictionary where users could register "named" functions, and then refer to them by name like `"huber"` or something like that.

As our first new-regression-function contributor, how does this sound as an API?

martins0n:
Use of decorators is a great and flexible option. But it's maybe not very obvious for users.
How to make it more obvious keeping it simple I have no ideas unfortunately. Maybe good examples would help.

nicolaskruchten:
Thanks again for taking the time to submit this PR! I'm going to close it now in favour of https://github.com/plotly/plotly.py/issues/2418 but I will re-use this code when documenting the new pluggable interface, as examples of the kinds of things that can be done with it :)

PR #1768, opened 2019-09-13
more flexible type of input arguments for px functions

### LINKED ISSUE #1767: Plotly Express column input formats
Right now for all `px` methods, `data_frame` is a required argument and the values of many kwargs are column names as strings.

We should be much more flexible in what we accept, for example the following should work:

```python
px.scatter(x=[1,2,3], y=[1,2,3])
```

In which case we would internally build a data frame with column names "x" and "y".

In addition, the following should work:

```python
px.scatter(df, x=df.index, y=df.col)
```

In which case we can read the column names (and optionally index name?) from the data frame.

Finally, we should allow for mixing and matching:

```python
px.scatter(df, x=df.index, y=df.col, color=[1,2,3])
```

emmanuelle:
This is a first a draft for accepting more flexible types of input arguments for `px` functions, eg numpy arrays, dataframe columns, mixed type etc. At the moment the transformation done by `build_or_augment_dataframe` is done argument by argument, so having different arguments with different types is ok. 

Todo:
- write more tests
- decide what to do with arguments which are lists, such as `hover_data`, `custom_data`, `dimensions`. At the moment they are just skipped so they should be column names. I did this because I wanted to discuss the names they should have, should it be `customdata_0`, `customdata_1` etc. ?

For the cases which are taken into account, please see the test file.

Closes #1767 , follow-up on https://github.com/plotly/plotly_express/pull/87 (thank you @malmaud !)

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -754,12 +754,12 @@ def apply_default_cascade(args):
         args["marginal_x"] = None
 
 
-def _name_heuristic(argument, field_name, reserved_names):
+def _name_heuristic(argument, field_name, used_col_names):
     if isinstance(argument, int):
         argument = str(argument)
-    elif field_name not in reserved_names:
+    elif field_name not in used_col_names:
         return field_name
-    elif field_name + argument not in reserved_names:
+    elif field_name + argument not in used_col_names:
         return field_name + "_" + argument
     else:
         raise NameError(
@@ -769,14 +769,14 @@ def _name_heuristic(argument, field_name, reserved_names):
         )
 
 
-def _get_reserved_names(args, attrables, array_attrables):
+def _initialize_argument_col_names(args, attrables, array_attrables):
     """
     This function builds a list of columns of the data_frame argument used
     as arguments, either as str/int arguments or given as columns
     (pandas series type).
     """
     df = args["data_frame"]
-    reserved_names = set()
+    used_col_names = set()
     for field in args:
         if field not in attrables:
             continue
@@ -787,16 +787,16 @@ def _get_reserved_names(args, attrables, array_attrables):
             if arg is None:
                 continue
             if isinstance(arg, str):
-                reserved_names.add(arg)
+                used_col_names.add(arg)
             if isinstance(arg, int):
-                reserved_names.add(str(arg))
+                used_col_names.add(str(arg))
             if isinstance(arg, pd.DataFrame) or isinstance(arg, pd.core.series.Series):
                 arg_name = arg.name
                 if arg_name:
                     in_df = arg is df[arg_name]
-                    reserved_names.add(arg_name)
+                    used_col_names.add(arg_name)
 
-    return reserved_names
+    return used_col_names
 
 
 def build_dataframe(args, attrables, array_attrables):
@@ -818,24 +818,27 @@ def build_dataframe(args, attrables, array_attrables):
         argument names corresponding to iterables, such as `hover_data`, ...
     """
     # Cast data_frame argument to DataFrame (it could be a numpy array, dict etc.)
-    if args["data_frame"] is not None and not isinstance(
-        args["data_frame"], pd.DataFrame
-    ):
+    df_provided = args["data_frame"] is not None
+    if df_provided and not isinstance(args["data_frame"], pd.DataFrame):
         args["data_frame"] = pd.DataFrame(args["data_frame"])
 
-    if args["data_frame"] is not None:
-        reserved_names = _get_reserved_names(args, attrables, array_attrables)
-    else:
-        reserved_names = set()
-    canbechanged_names = {}
-    forbidden_names = set(reserved_names)  # copy method compatible with Py2
     # We start from an empty DataFrame except for the case of functions which
     # implicitely need all dimensions: Splom, Parcats, Parcoords
     # This could be refined when dimensions is given
     df = pd.DataFrame()
 
+    # Initialize sets of column names
+    if df_provided:
+        used_col_names = _initialize_argument_col_names(
+            args, attrables, array_attrables
+        )
+    else:
+        used_col_names = set()
+    canbechanged_names = {}
+    forbidden_names = set(used_col_names)  # copy method compatible with Py2
+
     if "dimensions" in args and args["dimensions"] is None:
-        if args["data_frame"] is None:
+        if not df_provided:
             raise ValueError(
                 "No data were provided. Please provide data either with the `data_frame` or with the `dimensions` argument."
             )
@@ -849,6 +852,7 @@ def build_dataframe(args, attrables, array_attrables):
         if isinstance(args["data_frame"], pd.DataFrame)
         else None
     )
+
     # Loop over possible arguments
     for field_name in attrables:
         argument_list = (
@@ -871,10 +875,11 @@ def build_dataframe(args, attrables, array_attrables):
             length = len(df)
             if argument is None:
                 continue
+            ## ----------------- argument is a col name ----------------------
             elif isinstance(argument, str) or isinstance(
                 argument, int
-            ):  # just a column name
-                if not isinstance(args.get("data_frame"), pd.DataFrame):
+            ):  # just a column name given as str or int
+                if not df_provided:
                     raise ValueError(
                         "String or int arguments are only possible when a"
                         "DataFrame or an array is provided in the `data_frame`"
@@ -908,7 +913,7 @@ def build_dataframe(args, attrables, array_attrables):
                     "Argument '%s' is a pandas MultiIndex."
                     "pandas MultiIndex is not supported by plotly express" % field
                 )
-            # Case of numpy array or df column
+            # ----------------- argument is a column / array / list.... -------
             else:
                 is_index = isinstance(argument, pd.core.indexes.range.RangeIndex)
                 # pandas series have a name but it's None
@@ -920,16 +925,14 @@ def build_dataframe(args, attrables, array_attrables):
                         col_name = "index"
                     # revert previous argument
                     if col_name in canbechanged_names:
-                        if argument is not df[col_name]:
+                        if not argument.equals(df[col_name]):
+                            print ("will revert", col_name)
                             old_field, old_i = canbechanged_names[col_name]
-                            # old_field_name = old_field + str(i) if else old_field
                             df.rename(columns={col_name: old_field}, inplace=True)
                             args[old_field] = old_field
                             del canbechanged_names[col_name]
-                            reserved_names.remove(col_name)
+                            used_col_names.remove(col_name)
                     if col_name in forbidden_names:
-                        # If the name exists but the values have changed
-                        # we do not want to keep the name, revert to field
                         name_in_dataframe = (
                             args["data_frame"] is not None
                             and col_name in args["data_frame"].columns
@@ -942,11 +945,10 @@ def build_dataframe(args, attrables, array_attrables):
                         col_name = (
                             col_name
                             if keep_name
-                            else _name_heuristic(col_name, field, reserved_names)
+                            else _name_heuristic(col_name, field, used_col_names)
                         )
                 else:  # numpy array, list...
-                    # col_name = field
-                    col_name = _name_heuristic(field, field, reserved_names)
+                    col_name = _name_heuristic(field, field, used_col_names)
                 if length and len(argument) != length:
                     raise ValueError(
                         "All arguments should have the same length."
@@ -955,14 +957,16 @@ def build_dataframe(args, attrables, array_attrables):
                         % (field, len(argument), length)
                     )
                 df[str(col_name)] = argument
-                reserved_names.add(str(col_name))
+                used_col_names.add(str(col_name))
                 forbidden_names.add(str(col_name))
                 canbechanged_names[str(col_name)] = (field_name, i)
-            # Update argument with column name now that column exists
+
+            # Finally, update argument with column name now that column exists
             if field_name not in array_attrables:
                 args[field_name] = str(col_name)
             else:
                 args[field_name][i] = str(col_name)
+
     args["data_frame"] = df
     return args
 
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -822,12 +822,11 @@ def build_dataframe(args, attrables, array_attrables):
     if df_provided and not isinstance(args["data_frame"], pd.DataFrame):
         args["data_frame"] = pd.DataFrame(args["data_frame"])
 
-    # We start from an empty DataFrame except for the case of functions which
-    # implicitely need all dimensions: Splom, Parcats, Parcoords
-    # This could be refined when dimensions is given
+    # We start from an empty DataFrame
     df = pd.DataFrame()
 
-    # Initialize sets of column names
+    # Initialize set of column names
+    # These are reserved names
     if df_provided:
         used_col_names = _initialize_argument_col_names(
             args, attrables, array_attrables
@@ -835,6 +834,7 @@ def build_dataframe(args, attrables, array_attrables):
     else:
         used_col_names = set()
 
+    # Case of functions with a "dimensions" kw: scatter_matrix, parcats, parcoords
     if "dimensions" in args and args["dimensions"] is None:
         if not df_provided:
             raise ValueError(
@@ -843,13 +843,6 @@ def build_dataframe(args, attrables, array_attrables):
         else:
             df_args = args["data_frame"]
             df[df_args.columns] = df_args[df_args.columns]
-
-    # Valid column names
-    df_columns = (
-        args["data_frame"].columns
-        if isinstance(args["data_frame"], pd.DataFrame)
-        else None
-    )
 
     # Loop over possible arguments
     for field_name in attrables:
@@ -862,7 +855,7 @@ def build_dataframe(args, attrables, array_attrables):
         # argument not specified, continue
         if argument_list is None or argument_list is [None]:
             continue
-        # Argument name: field_name if the argument is a list
+        # Argument name: field_name if the argument is not a list
         # Else we give names like ["hover_data_0, hover_data_1"] etc.
         field_list = (
             [field_name]
@@ -887,11 +880,11 @@ def build_dataframe(args, attrables, array_attrables):
                         "is of type str or int." % field
                     )
                 # Check validity of column name
-                if df_columns is not None and argument not in df_columns:
+                if argument not in args['data_frame'].columns:
                     raise ValueError(
                         "Value of '%s' is not the name of a column in 'data_frame'. "
                         "Expected one of %s but received: %s"
-                        % (field, str(list(df_columns)), argument)
+                        % (field, str(list(args['data_frame'].columns)), argument)
                     )
                 if length and len(args["data_frame"][argument]) != length:
                     raise ValueError(
@@ -907,7 +900,7 @@ def build_dataframe(args, attrables, array_attrables):
                     else:
                         args[field_name][i] = str(argument)
                 continue
-            # Case of index
+            # Case of multiindex
             elif isinstance(argument, pd.core.indexes.multi.MultiIndex):
                 raise TypeError(
                     "Argument '%s' is a pandas MultiIndex."
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -778,7 +778,6 @@ def build_or_augment_dataframe(args, attrables, array_attrables):
     ):
         args["data_frame"] = pd.DataFrame(args["data_frame"])
 
-
     # We start from an empty DataFrame except for the case of functions which
     # implicitely need all dimensions: Splom, Parcats, Parcoords
     # This could be refined when dimensions is given
@@ -851,16 +850,14 @@ def build_or_augment_dataframe(args, attrables, array_attrables):
             # Case of index
             elif isinstance(argument, pd.core.indexes.multi.MultiIndex):
                 raise TypeError("pandas MultiIndex is not supported by plotly express")
-            elif isinstance(argument, pd.core.indexes.range.RangeIndex):
-                col_name = argument.name if argument.name else "index"
-                try:
-                    df.insert(0, col_name, argument)
-                except ValueError:  # if col named index already exists, replace
-                    df[col_name] = argument
             # Case of numpy array or df column
             else:
                 try:
                     col_name = argument.name  # pandas df
+                    if col_name is None and isinstance(
+                        argument, pd.core.indexes.range.RangeIndex
+                    ):
+                        col_name = "index"
                     if (
                         args.get("data_frame") is not None
                         and col_name in args["data_frame"]
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -754,16 +754,13 @@ def apply_default_cascade(args):
         args["marginal_x"] = None
 
 
-def build_or_augment_dataframe(args, attrables, array_attrables, constructor):
+def build_or_augment_dataframe(args, attrables, array_attrables):
     """
     Constructs an implicit dataframe and modifies `args` in-place.
 
     Parameters
     ----------
     args : OrderedDict
-        
-    constructor : go Trace object
-        trace function. It is used to fine-tune some options.
     `attrables` is a list of keys into `args`, all of whose corresponding
     values are converted into columns of a dataframe.
     Used to be support calls to plotting function that elide a dataframe
@@ -773,13 +770,14 @@ def build_or_augment_dataframe(args, attrables, array_attrables, constructor):
     # We start from an empty DataFrame except for the case of functions which
     # implicitely need all dimensions: Splom, Parcats, Parcoords
     # This could be refined when dimensions is given
-    if constructor in [go.Splom, go.Parcats, go.Parcoords]:  # we take all dimensions
-        df = args["data_frame"]
-    else:
-        df = pd.DataFrame()
+    df = pd.DataFrame()
 
-    # Retrieve labels (to change column names)
-    # labels = args.get("labels")  # labels or None
+    if "dimensions" in args and args["dimensions"] is None:
+        if args.get("data_frame") is None or args["data_frame"] is None:
+            raise ValueError("No data were provided")
+        else:
+            df_args = args["data_frame"]
+            df[df_args.columns] = df_args[df_args.columns]
 
     # Valid column names
     df_columns = (
@@ -857,7 +855,7 @@ def infer_config(args, constructor, trace_patch):
     group_attrables = ["animation_frame", "facet_row", "facet_col", "line_group"]
 
     all_attrables = attrables + group_attrables + ["color"]
-    build_or_augment_dataframe(args, all_attrables, array_attrables, constructor)
+    build_or_augment_dataframe(args, all_attrables, array_attrables)
 
     attrs = [k for k in attrables if k in args]
     grouped_attrs = []
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -928,8 +928,8 @@ def build_dataframe(args, attrables, array_attrables):
                         if is_index:
                             keep_name = df_provided and argument is df_input.index
                         else:
-                            keep_name = col_name in df_input and argument.equals(
-                                df_input[col_name]
+                            keep_name = (
+                                col_name in df_input and argument is df_input[col_name]
                             )
                         col_name = (
                             col_name
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -803,7 +803,8 @@ def build_or_augment_dataframe(args, attrables, array_attrables):
             if field_name not in array_attrables
             else args.get(field_name)
         )
-        if argument_list is None:  # argument not specified, continue
+        # argument not specified, continue
+        if argument_list is None or argument_list is [None]:
             continue
         # Argument name: field_name if the argument is a list
         # Else we give names like ["hover_data_0, hover_data_1"] etc.
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -671,7 +671,7 @@ def one_group(x):
 
 
 def apply_default_cascade(args):
-    # https://github.com/plotly/dash-table/issues/597first we apply px.defaults to unspecified args
+    # first we apply px.defaults to unspecified args
     for param in (
         ["color_discrete_sequence", "color_continuous_scale"]
         + ["symbol_sequence", "line_dash_sequence", "template"]
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -762,11 +762,19 @@ def build_or_augment_dataframe(args, attrables, array_attrables, constructor):
     Used to be support calls to plotting function that elide a dataframe
     argument; for example `scatter(x=[1,2], y=[3,4])`.
     """
+
+    # We start from an empty DataFrame except for the case of functions which
+    # implicitely need all dimensions: Splom, Parcats, Parcoords
+    # This could be refined when dimensions is given
     if constructor in [go.Splom, go.Parcats, go.Parcoords]:  # we take all dimensions
         df = args["data_frame"]
     else:
         df = pd.DataFrame()
+
+    # Retrieve labels (to change column names)
     labels = args.get("labels")  # labels or None
+
+    # Valid column names
     df_columns = (
         args["data_frame"].columns if args.get("data_frame") is not None else None
     )
@@ -774,23 +782,29 @@ def build_or_augment_dataframe(args, attrables, array_attrables, constructor):
     for group_attr in group_attrs:
         if group_attr in args:
             attrables += [group_attr]
+
+    # Loop over possible arguments
     for field_name in attrables:
         argument_list = (
             [args.get(field_name)]
             if field_name not in array_attrables
             else args.get(field_name)
         )
-        if argument_list is None:
+        if argument_list is None:  # argument not specified, continue
             continue
+        # Argument name: field_name if the argument is a list
+        # Else we give names like ["hover_data_0, hover_data_1"] etc.
         field_list = (
             [field_name]
             if field_name not in array_attrables
             else [field_name + "_" + str(i) for i in range(len(argument_list))]
         )
+        # argument_list and field_list ready, iterate over them
         for i, (argument, field) in enumerate(zip(argument_list, field_list)):
             if argument is None:
                 continue
-            elif isinstance(argument, str):  # needs to change
+            elif isinstance(argument, str):  # just a column name
+                # Check validity of column name
                 try:
                     df[argument] = args["data_frame"][argument]
                     continue
@@ -807,14 +821,15 @@ def build_or_augment_dataframe(args, attrables, array_attrables, constructor):
                 try:
                     df.insert(0, col_name, argument)
                 except ValueError:  # if col named index already exists, replace
-                    df["col_name"] = argument
-            else:  # args[field] should be an array or df column
+                    df[col_name] = argument
+            # Case of numpy array or df column
+            else:
                 try:
                     col_name = argument.name  # pandas df
                 except AttributeError:
                     col_name = labels[field] if labels and labels.get(field) else field
                 df[col_name] = argument
-            # This sets the label of an attribute to be
+            # Update argument with column name now that column exists
             if field_name not in array_attrables:
                 args[field_name] = col_name
             else:
@@ -837,28 +852,6 @@ def infer_config(args, constructor, trace_patch):
 
     all_attrables = attrables + group_attrables + ["color"]
     build_or_augment_dataframe(args, all_attrables, array_attrables, constructor)
-
-    # Validate that the strings provided as attribute values reference columns
-    # in the provided data_frame
-    df_columns = args["data_frame"].columns
-
-    for attr in attrables + group_attrables + ["color"]:
-        if attr in args and args[attr] is not None:
-            maybe_col_list = [args[attr]] if attr not in array_attrables else args[attr]
-            for maybe_col in maybe_col_list:
-                try:
-                    in_cols = maybe_col in df_columns
-                except TypeError:
-                    in_cols = False
-                if not in_cols:
-                    value_str = (
-                        "Element of value" if attr in array_attrables else "Value"
-                    )
-                    raise ValueError(
-                        "%s of '%s' is not the name of a column in 'data_frame'. "
-                        "Expected one of %s but received: %s"
-                        % (value_str, attr, str(list(df_columns)), str(maybe_col))
-                    )
 
     attrs = [k for k in attrables if k in args]
     grouped_attrs = []
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -762,12 +762,13 @@ def build_or_augment_dataframe(args, attrables, array_attrables, constructor):
     Used to be support calls to plotting function that elide a dataframe
     argument; for example `scatter(x=[1,2], y=[3,4])`.
     """
-    # This will be changed so that we start from an empty dataframe
-    if args.get("data_frame") is None:
-        df = pd.DataFrame()
-    else:
-        df = args["data_frame"]
+    df = pd.DataFrame()
     labels = args.get("labels")  # labels or None
+    df_columns = (
+        args["data_frame"].columns if args.get("data_frame") is not None else None
+    )
+    if "symbol" in args:
+        attrables += ["symbol"]
     for field_name in attrables:
         argument_list = (
             [args.get(field_name)]
@@ -785,7 +786,15 @@ def build_or_augment_dataframe(args, attrables, array_attrables, constructor):
             if argument is None:
                 continue
             elif isinstance(argument, str):  # needs to change
-                continue
+                try:
+                    df[argument] = args["data_frame"][argument]
+                    continue
+                except KeyError:
+                    raise ValueError(
+                        "Value of '%s' is not the name of a column in 'data_frame'. "
+                        "Expected one of %s but received: %s"
+                        % (field, str(list(df_columns)), argument)
+                    )
             # Case of index
             elif isinstance(argument, pd.core.indexes.range.RangeIndex):
                 col_name = argument.name if argument.name else "index"
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -829,6 +829,8 @@ def build_or_augment_dataframe(args, attrables, array_attrables):
                 df[argument] = args["data_frame"][argument]
                 continue
             # Case of index
+            elif isinstance(argument, pd.core.indexes.multi.MultiIndex):
+                raise TypeError("pandas MultiIndex is not supported by plotly express")
             elif isinstance(argument, pd.core.indexes.range.RangeIndex):
                 col_name = argument.name if argument.name else "index"
                 try:
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -771,6 +771,11 @@ def _name_heuristic(argument, field_name, df):
         )
 
 
+def _get_reserved_names(args, attrables, array_attrables):
+    df = args['data_frame']
+    for field in args:
+
+
 def build_dataframe(args, attrables, array_attrables):
     """
     Constructs a dataframe and modifies `args` in-place.
@@ -795,6 +800,8 @@ def build_dataframe(args, attrables, array_attrables):
     ):
         args["data_frame"] = pd.DataFrame(args["data_frame"])
 
+    if args["data_frame"] is not None:
+        reserved_names = _get_reserved_names(args, attrables, array_attrables)
     # We start from an empty DataFrame except for the case of functions which
     # implicitely need all dimensions: Splom, Parcats, Parcoords
     # This could be refined when dimensions is given
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -817,15 +817,15 @@ def build_or_augment_dataframe(args, attrables, array_attrables):
                         "is provided in the `data_frame` argument."
                     )
                 # Check validity of column name
-                try:
-                    df[argument] = args["data_frame"][argument]
-                    continue
-                except KeyError:
+                if argument not in df_columns:
                     raise ValueError(
                         "Value of '%s' is not the name of a column in 'data_frame'. "
                         "Expected one of %s but received: %s"
                         % (field, str(list(df_columns)), argument)
                     )
+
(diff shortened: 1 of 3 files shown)


emmanuelle:
One other thing to fix : `px.scatter(tips, x="bla", y='wrong')` does not produce a meaningful error message any more, have to fix this.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,42 @@ def apply_default_cascade(args):
...
+def build_or_augment_dataframe(args, attrables, array_attrables):
+    """
+    Constructs an implicit dataframe and modifies `args` in-place.
+    `attrables` is a list of keys into `args`, all of whose corresponding
+    values are converted into columns of a dataframe.
+    Used to be support calls to plotting function that elide a dataframe
+    argument; for example `scatter(x=[1,2], y=[3,4])`.
+    """
+    if args.get("data_frame") is None:
+        df = pd.DataFrame()
```
nicolaskruchten: arguably we could just always start with an empty DF and progressively add columns to it.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,42 @@ def apply_default_cascade(args):
...
+    else:
+        df = args["data_frame"]
+        df = df.reset_index()
```
nicolaskruchten: I don't think we should automatically do this, as it will flow through to the `dimension`-supporting traces :)

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
```diff
@@ -0,0 +1,29 @@
+import plotly.express as px
+import numpy as np
+import pandas as pd
+
+
```
nicolaskruchten: here I think we can also unit-test the `build_or_augment_dataframe` function directly... probably easier to read the tests that way IMO

emmanuelle: we can do both, it's also nice to have regression tests calling the px functions (which are here mostly to be sure it does not crash :-))

emmanuelle:
TODO
Additional cases to take into account: check that error messages make sense when:
- having different lengths for different kw arguments (or dimensions not compatible with the given DataFrame)
- what happens with 2D arrays passed as parameters

At the moment the initial DataFrame is modified in-place, probably we should not do that (unless the DataFrame is very big??).

emmanuelle:
next step: start from empty dataframe and add columns

nicolaskruchten:
We should also move the "non-existent column" check and error message into this new function, which I think will help with some of this logic.

packages/python/plotly/plotly/express/_core.py
emmanuelle: done

packages/python/plotly/plotly/express/_core.py
emmanuelle: yes, this has been corrected

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,75 @@ def apply_default_cascade(args):
         args["marginal_x"] = None
 
 
+def build_or_augment_dataframe(args, attrables, array_attrables, constructor):
+    """
+    Constructs an implicit dataframe and modifies `args` in-place.
+    `attrables` is a list of keys into `args`, all of whose corresponding
+    values are converted into columns of a dataframe.
+    Used to be support calls to plotting function that elide a dataframe
+    argument; for example `scatter(x=[1,2], y=[3,4])`.
```
emmanuelle: need to write docstring here

emmanuelle:
@nicolaskruchten this is ready for a round of review. In the meantime I'll write a better docstring for the helper function.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,90 @@ def apply_default_cascade(args):
...
+    """
+
+    # We start from an empty DataFrame except for the case of functions which
+    # implicitely need all dimensions: Splom, Parcats, Parcoords
+    # This could be refined when dimensions is given
+    if constructor in [go.Splom, go.Parcats, go.Parcoords]:  # we take all dimensions
+        df = args["data_frame"]
```
nicolaskruchten: even in this case I would not directly use the provided data frame: rather I would start with an empty `df` always, and move the default `dimensions` logic here (basically if `args["dimensions"]` is present but empty then add all the columns from `args["data_frame"]` to `df`)... using this logic would also enable you to not pass `constructor` in here. (In general, the less we can condition things on `constructor` the better IMO :)

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,90 @@ def apply_default_cascade(args):
...
+    else:
+        df = pd.DataFrame()
+
+    # Retrieve labels (to change column names)
+    labels = args.get("labels")  # labels or None
```
nicolaskruchten: is it important to move the (re-)labelling logic into this function? Right now it all mostly happens later with various helper functions. If we're moving this logic into this function then should we refactor it out of where it is right now...?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,90 @@ def apply_default_cascade(args):
...
+                col_name = argument.name if argument.name else "index"
+                col_name = labels[field] if labels and labels.get(field) else col_name
+                try:
+                    df.insert(0, col_name, argument)
+                except ValueError:  # if col named index already exists, replace
+                    df[col_name] = argument
+            # Case of numpy array or df column
+            else:
+                try:
+                    col_name = argument.name  # pandas df
```
nicolaskruchten: what case does this cover?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,90 @@ def apply_default_cascade(args):
...
+                    )
+            # Case of index
+            elif isinstance(argument, pd.core.indexes.range.RangeIndex):
+                col_name = argument.name if argument.name else "index"
+                col_name = labels[field] if labels and labels.get(field) else col_name
+                try:
+                    df.insert(0, col_name, argument)
+                except ValueError:  # if col named index already exists, replace
+                    df[col_name] = argument
+            # Case of numpy array or df column
```
nicolaskruchten: I'm not sure how/when this covers the case of `df.col` ? We don't seem to actually look to see `if argument in args["data_frame"]` the way I would expect... ?

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
```diff
@@ -0,0 +1,144 @@
...
+def test_with_index():
+    tips = px.data.tips()
+    fig = px.scatter(tips, x=tips.index, y="total_bill")
+    fig = px.scatter(tips, x=tips.index, y=tips.total_bill)
+    fig = px.scatter(tips, x=tips.index, y=tips.total_bill)
+    assert fig.data[0]["hovertemplate"] == "index=%{x}<br>total_bill=%{y}"
+    # I was not expecting this to work but it does...
+    fig = px.scatter(tips, x=tips.index, y=10 * tips.total_bill)
+    assert fig.data[0]["hovertemplate"] == "index=%{x}<br>total_bill=%{y}"
```
nicolaskruchten: wow, I'm surprised that this works ... how does it work with the `total_bill` name??

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: oh I see now... This is potentially very dangerous.

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: here we should loop on columns by name and check for `is` equality on `argument`.

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: We should only do this iff `argument` is not a column (as checked by `is`) *and* if `argument.name` doesn't appear in `df.columns`. That way we can accept `Series` from other dataframes or xarrays, say, but we don't grab `df.x` by name if the user passes in `df.x*10` (which, annoyingly, passes `df.x.name` through).

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
nicolaskruchten: Oh I see. scary :)

packages/python/plotly/plotly/express/_core.py
emmanuelle: yes, it's better to leave the labelling logic later as it was before, this should be resolved now.

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
emmanuelle: For the docstrings of `px` functions I chose to use the name `array_like` which is commonly used in numpy to refer to both arrays and lists, and in the broad sense dataframes as well. This is hopefully ready for another round of review.

emmanuelle: once black is happy of course...

emmanuelle:
For the docstrings of px functions I chose to use the name array_like which is commonly used in numpy to refer to both arrays and lists, and in the broad sense dataframes as well. This is hopefully ready for another round of review.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,107 @@ def apply_default_cascade(args):
...
+
+    # Loop over possible arguments
+    for field_name in attrables:
+        argument_list = (
+            [args.get(field_name)]
+            if field_name not in array_attrables
+            else args.get(field_name)
+        )
+        if argument_list is None:  # argument not specified, continue
+            continue
```
nicolaskruchten: what if it's `[None]` ?

nicolaskruchten: I would maybe break out of the loop even earlier if `args.get(field_name) is None`

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,107 @@ def apply_default_cascade(args):
...
+            if field_name not in array_attrables
+            else [field_name + "_" + str(i) for i in range(len(argument_list))]
+        )
+        # argument_list and field_list ready, iterate over them
+        for i, (argument, field) in enumerate(zip(argument_list, field_list)):
+            if argument is None:
+                continue
+            elif isinstance(argument, str):  # just a column name
+                # Check validity of column name
+                try:
```
nicolaskruchten: what if `args["data_frame"]` isn't provided? we need a special error message for "you can't just give us strings without a dataframe!" or something

emmanuelle: done!

packages/python/plotly/plotly/express/_core.py
emmanuelle: ok, taken into account

packages/python/plotly/plotly/express/_core.py
```diff
@@ -821,6 +827,11 @@ def build_or_augment_dataframe(args, attrables, array_attrables):
                         "Expected one of %s but received: %s"
                         % (field, str(list(df_columns)), argument)
                     )
+                except TypeError:
```
nicolaskruchten: This seems overly-broad to me, compared to just directly checking if the `data_frame` argument has been passed in... I worry about try-catch control flow as you never know what new crazy thing will raise any given exception later :)

emmanuelle: ok, I hesitated indeed between the two solutions. I chose the try/except mostly for the sake of compactness since this function is getting crazily long but this is not such a good argument compared to robustness.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,119 @@ def apply_default_cascade(args):
...
+            df_args = args["data_frame"]
+            df[df_args.columns] = df_args[df_args.columns]
+
+    # Valid column names
+    df_columns = (
+        args["data_frame"].columns if args.get("data_frame") is not None else None
+    )
+    group_attrs = ["symbol", "line_dash"]
+    for group_attr in group_attrs:
+        if group_attr in args:
```
nicolaskruchten: this case is already handled below, so you may as well just add `symbol` and `line_dash` in attrables always at call-time?

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: I think I would move this up to the top of the `for` loop as `if args.get(field_name) is None: continue` that way we don't need to check for that `[None]` case at all.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,119 @@ def apply_default_cascade(args):
...
+                    col_name = argument.name  # pandas df
+                    if (
+                        args.get("data_frame") is not None
+                        and col_name in args["data_frame"]
+                    ):
+                        # If the name exists but the values have changed
+                        # we do not want to keep the name, revert to field
+                        col_name = (
+                            col_name
+                            if args["data_frame"][col_name].equals(argument)
```
nicolaskruchten: what's the tradeoff between using `.equals()` vs `is` here? the former is slower but a bit more permissive in that if it's a copy with the same name we accept it?

emmanuelle: yes this was my idea. But maybe the flexibility gain is not worth the performance cost, I'm also happy to change to `is`.

nicolaskruchten: Maybe we can switch to `is` for maximum strictness/performance, to start, and see if people complain/notice?

packages/python/plotly/plotly/express/_core.py
emmanuelle: ok done

packages/python/plotly/plotly/express/_core.py
emmanuelle: done

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,112 @@ def apply_default_cascade(args):
...
+            else [field_name + "_" + str(i) for i in range(len(argument_list))]
+        )
+        # argument_list and field_list ready, iterate over them
+        for i, (argument, field) in enumerate(zip(argument_list, field_list)):
+            if argument is None:
+                continue
+            elif isinstance(argument, str):  # just a column name
+                if not isinstance(args.get("data_frame"), pd.DataFrame):
+                    raise ValueError(
+                        "String arguments are only possible when a DataFrame"
```
nicolaskruchten: missing a trailing space here: `DataFrameis`

nicolaskruchten:
So this gives an error, as it should, due to misaligned lengths: `px.scatter(px.data.tips(), x=[1,2,3,4], y=[1,2,3])`

But this does not: `px.scatter(px.data.tips(), x=[1,2,3,4], y="tip")`

emmanuelle:
Oh wow good catch. `pandas` accepts very weird things like 
```
In [10]: df = pd.DataFrame({'a':[1, 3], 'b':[5, 6], 'c':[1, 2]})        
In [11]: df['bla'] = tips['tip']    
```
does not raise an error

nicolaskruchten:
So this does something fun: `px.bar(x=[['a','b'],['a','c'],['d','b'],['d','c']], y=[1,2,3,4])` ... it triggers our "multicategory" behaviour, but not a in good way:

image

nicolaskruchten:
so maybe we need to check on "shape" as well as just length?

emmanuelle:
I'm not sure it's possible to support multicategory axis if all values have to be in a dataframe at some point. For your example to work (like it does with `go`) we should have I think `x.bar(x=[['a','a', 'b', 'b'],[1, 2, 1, 2], y=[1,2,3,4]) ` and then because of dimension mismatch (we would need a transpose) it's not possible to put x and y in a common dataframe.

emmanuelle:
but the correct behavior is not right, that's for sure

emmanuelle:
Well the current behavior is not right because of what `go` does:
```
x=[['a','b'],['a','c'],['d','b'],['d','c']]
y=[1,2,3,4]
fig = go.Figure(data=go.Bar(x=x, y=y)
)
fig.show()
```
does the same.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,138 @@ def apply_default_cascade(args):
...
+        if args["data_frame"] is None:
+            raise ValueError(
+                "No data were provided. Please provide data either with the `data_frame` or with the `dimensions` argument."
+            )
+        else:
+            df_args = args["data_frame"]
+            df[df_args.columns] = df_args[df_args.columns]
+
+    # Cast data_frame argument to DataFrame (it could be a numpy array, dict etc.)
+    if args["data_frame"] is not None and not isinstance(
```
nicolaskruchten: should this move up above the `dimensions` block, which relies on data frames a bit?

emmanuelle:
@jonmmease we've made some progress here and your thoughts would be very welcome, since it's a quite important change (more flexibility means empowering users, but also potentially having more bugs :-)). The test file should give you an idea of cases which are supported with this PR. What do you think of this new API ? Any changes you would suggest?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,135 @@ def apply_default_cascade(args):
...
+                        and col_name in args["data_frame"]
+                    ):
+                        # If the name exists but the values have changed
+                        # we do not want to keep the name, revert to field
+                        col_name = (
+                            col_name
+                            if argument is args["data_frame"][col_name]
+                            else field
+                        )
+                except AttributeError:  # numpy array, list...
```
nicolaskruchten: this is intended to catch errors from `argument.name` primarily, right? can we just check for the presence of this directly and avoid the try-catch pattern?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,135 @@ def apply_default_cascade(args):
...
+                df[str(argument)] = args["data_frame"][argument]
+                if isinstance(argument, int):
+                    if field_name not in array_attrables:
+                        args[field_name] = str(argument)
+                    else:
+                        args[field_name][i] = str(argument)
+                continue
+            # Case of index
+            elif isinstance(argument, pd.core.indexes.multi.MultiIndex):
+                raise TypeError("pandas MultiIndex is not supported by plotly express")
```
nicolaskruchten: for all these errors we should clearly indicate which attribute is causing the issue.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,135 @@ def apply_default_cascade(args):
...
+                        # we do not want to keep the name, revert to field
+                        col_name = (
+                            col_name
+                            if argument is args["data_frame"][col_name]
+                            else field
+                        )
+                except AttributeError:  # numpy array, list...
+                    col_name = field
+                if length and len(argument) != length:
+                    raise ValueError("All arguments should have the same length.")
```
nicolaskruchten: we should indicate which attribute is causing the issue here (I know in case of length it's ambiguous, but any hint helps!)

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,135 @@ def apply_default_cascade(args):
...
+            if argument is None:
+                continue
+            elif isinstance(argument, str) or isinstance(
+                argument, int
+            ):  # just a column name
+                if not isinstance(
+                    args.get("data_frame"), pd.DataFrame
+                ) and not isinstance(args.get("data_frame"), np.ndarray):
+                    raise ValueError(
+                        "String or int arguments are only possible when a"
```
nicolaskruchten: we should indicate which arg this is

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: maybe report the previous argument whose length we compared to?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,135 @@ def apply_default_cascade(args):
...
+        for i, (argument, field) in enumerate(zip(argument_list, field_list)):
+            length = len(df)
+            if argument is None:
+                continue
+            elif isinstance(argument, str) or isinstance(
+                argument, int
+            ):  # just a column name
+                if not isinstance(
+                    args.get("data_frame"), pd.DataFrame
+                ) and not isinstance(args.get("data_frame"), np.ndarray):
```
nicolaskruchten: we no longer need to check against `ndarray` here right?

packages/python/plotly/plotly/express/_core.py
emmanuelle: ok

packages/python/plotly/plotly/express/_core.py
emmanuelle: good catch!

packages/python/plotly/plotly/express/_doc.py
```diff
@@ -15,98 +17,110 @@
     data_frame=["A 'tidy' `pandas.DataFrame`"],
```
nicolaskruchten: this line will likely need to change also :)

packages/python/plotly/plotly/express/_doc.py
```diff
@@ -1,7 +1,9 @@
 import inspect
 
-colref = "(string: name of column in `data_frame`)"
-colref_list = "(list of string: names of columns in `data_frame`)"
+colref = "(string: name of column in `data_frame`, or array_like object)"
```
nicolaskruchten: should we add something about the fact that array_like objects with `.name` defined get special treatment? maybe not worth the bloat.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,169 @@ def apply_default_cascade(args):
...
+    if argument not in df.columns:
+        print ("no pb", argument)
+        return argument
+    elif field_name not in df.columns:
+        print ("fallback field", field_name)
+        return field_name
+    elif field_name + argument not in df.columns:
+        return field_name + "_" + argument
+    else:
+        raise NameError("A name conflict was encountered.")
```
emmanuelle: I will improve the error message here, I pushed just to show the logics.

emmanuelle:
Our name setting heuristic is interferring with `hover_data` and `custom_data`, will look into this

emmanuelle:
OK, with this name-changing heuristic it will be very complicated if we accept too many things. What I suggest is that
- if a dataframe is provided as `data_frame` then we get all string arguments as forbidden names, and we will change the name of iterables if they have a name in this list of forbidden names.
- if several iterables have the same name we can either have a "first come - first serve" policy but it's not great, or we could change all names like "df1_x" and "df2_x".

emmanuelle:
Now we have a better naming heuristic but the code is very ugly, please don't review now as I need to improve this

emmanuelle:
The `build_dataframe` function is now very long, mostly because of the necessity to handle a number of different cases for column names. This part of the code does not smell good, to the extent that I'm wondering whether we should not support less fancy things, like
- if the `data_frame` argument is provided, then arguments with str column names or column names from `df` (as tested with `is` and not `equals`) keep their names in df
- if `data_frame` is not provided, all arguments have kw names (`x`, `y`, `size`)
- if you want more control, you use the `labels` argument. 

This would be a basic but robust behaviour. Conversely, with the current behaviour we do some magics in order to use column names as much as possible. This is useful when several dataframes are used in the arguments, but it should be a marginal case I guess. 

With both solutions all `px` calls in the tests will work ie display the same data, up to hover data and axis labels.

To be discussed! @nicolaskruchten @jonmmease

nicolaskruchten:
Yeah, this is closer to the original goal here... It's been worthwhile exploring the idea of using `.name` from arbitrary vectors but in the details it does seem like the main source of issues IMO.

emmanuelle:
@nicolaskruchten ready for a round of review when you find the time. One thing left to discuss if the name that should be used when the argument is not a column from the `data_frame` argument. In the example below
```
import plotly.express as px
import pandas as pd
df = pd.DataFrame(dict(a=[0, 1], b=[1, 10], c=[0.1, 0.8]))
df2 = pd.DataFrame(dict(time=[23, 26], money=[100, 200]))
fig = px.scatter(df, x=df.a, y=df2.money, size=df.b)
```
what should be the name used for the `y` argument? At the moment it's just `y`, but we could also use `y_money`, ie concatenate the name of the keyword and the name of the column. This would maybe be more informative. At the moment `y_money` is only used if there is already a column `y` of `data_frame` which is used in the plot.

nicolaskruchten:
> At the moment y_money is only used if there is already a column y of data_frame which is used in the plot

This is fine. What happens if `data_frame` also contains `y_money`? :)

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,202 @@ def apply_default_cascade(args):
...
+            continue
+        names = args[field] if field in array_attrables else [args[field]]
+        if names is None:
+            continue
+        for arg in names:
+            if arg is None:
+                continue
+            if isinstance(arg, str):
+                used_col_names.add(arg)
+            if isinstance(arg, pd.DataFrame) or isinstance(arg, pd.core.series.Series):
```
nicolaskruchten: we accept full `DataFrame`s as well as `Series`?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,202 @@ def apply_default_cascade(args):
...
+    for field in args:
+        if field not in attrables:
+            continue
+        names = args[field] if field in array_attrables else [args[field]]
+        if names is None:
+            continue
+        for arg in names:
+            if arg is None:
+                continue
+            if isinstance(arg, str):
```
nicolaskruchten: what if `arg` isn't in `df` ?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,202 @@ def apply_default_cascade(args):
...
+                used_col_names.add(arg)
+            if isinstance(arg, pd.DataFrame) or isinstance(arg, pd.core.series.Series):
+                arg_name = arg.name
+                if arg_name and hasattr(df, arg_name):
+                    in_df = arg is df[arg_name]
+                    if in_df:
```
nicolaskruchten: can probably inline this check

packages/python/plotly/plotly/express/_core.py
emmanuelle: it will be caught later on, L882 
```
 if argument not in args["data_frame"].columns:
                    raise ValueError(
                        "Value of '%s' is not the name of a column in 'data_frame'. "
                        "Expected one of %s but received: %s"
                        % (field, str(list(args["data_frame"].columns)), argument)
                    )
```
is it ok for you?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,202 @@ def apply_default_cascade(args):
...
+                    raise ValueError(
+                        "Value of '%s' is not the name of a column in 'data_frame'. "
+                        "Expected one of %s but received: %s"
+                        % (field, str(list(args["data_frame"].columns)), argument)
+                    )
+                if length and len(args["data_frame"][argument]) != length:
+                    raise ValueError(
+                        "All arguments should have the same length."
+                        "The length of column argument `df[%s]` is %d, whereas the"
+                        "length of previous arguments is %d"
```
nicolaskruchten: is it possible to name the previous argument here? "previous" won't mean anything to the user as they don't know the iteration order

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: yes, I was just curious about why it ends up in the reserved names

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: oh I just meant `if arg is df[arg_name]` ... doesn't matter much :)

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,207 @@ def apply_default_cascade(args):
...
+                        args[field_name] = col_name
+                    else:
+                        args[field_name][i] = col_name
+                df[col_name] = args["data_frame"][argument]
+                continue
+            # Case of multiindex
+            elif isinstance(argument, pd.core.indexes.multi.MultiIndex):
+                raise TypeError(
+                    "Argument '%s' is a pandas MultiIndex."
+                    "pandas MultiIndex is not supported by plotly express" % field
```
nicolaskruchten: maybe add "at the moment" ? :)

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,207 @@ def apply_default_cascade(args):
...
+                if (
+                    hasattr(argument, "name") and argument.name is not None
+                ) or is_index:
+                    col_name = argument.name  # pandas df
+                    if col_name is None and is_index:
+                        col_name = "index"
+                    if not df_provided:
+                        col_name = field
+                    else:
+                        keep_name = hasattr(
```
nicolaskruchten: instead of `hasattr`/`getattr` we could do `col_name in args["data_frame"].columns`/`args["data_frame"][col_name]` or something right?

packages/python/plotly/plotly/express/_doc.py
nicolaskruchten: worth adding that `int` column names are accepted also?

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
```diff
@@ -0,0 +1,267 @@
...
+    + ["lat", "lon", "locations", "animation_group"]
+)
+array_attrables = ["dimensions", "custom_data", "hover_data"]
+group_attrables = ["animation_frame", "facet_row", "facet_col", "line_group"]
+
+all_attrables = attrables + group_attrables + ["color"]
+
+
+def test_numpy():
+    fig = px.scatter(x=[1, 2, 3], y=[2, 3, 4], color=[1, 3, 9])
```
nicolaskruchten: can we assert something here? :)

emmanuelle: done

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
```diff
@@ -0,0 +1,267 @@
...
+    before_tip = tips.total_bill - tips.tip
+    fig = px.bar(tips, x="day", y=before_tip, labels={"y": "bill"})
+    assert fig.data[0].hovertemplate == "day=%{x}<br>bill=%{y}"
+
+
+def test_several_dataframes():
+    df = pd.DataFrame(dict(x=[0, 1], y=[1, 10], z=[0.1, 0.8]))
+    df2 = pd.DataFrame(dict(time=[23, 26], money=[100, 200]))
+    fig = px.scatter(df, x="z", y=df2.money, size="y")
+    assert fig.data[0].hovertemplate == "z=%{x}<br>y_money=%{y}<br>y=%{marker.size}"
```
nicolaskruchten: this result surprises me a bit...

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
```diff
@@ -0,0 +1,267 @@
...
+    fig = px.scatter(df2, x=df.z, y=df2.money, size=df.y)
+    assert fig.data[0].hovertemplate == "x=%{x}<br>money=%{y}<br>size=%{marker.size}"
+
+
+def test_name_conflict():
+    df = pd.DataFrame(dict(x=[0, 1], y=[3, 4]))
+    fig = px.scatter(df, x=[10, 1], y="y", color="x")
+    assert np.all(fig.data[0].x == np.array([10, 1]))
+    assert fig.data[0].hovertemplate == "x_x=%{x}<br>y=%{y}<br>x=%{marker.color}"
```
nicolaskruchten: `x_x` seems very surprising here also :)

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
emmanuelle: why? It's a bit weird but it corresponds to the logics
- use the field name if possible
- if not use field + _ + name
Hopefully users will use sensible column names most of the time...

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
```diff
@@ -0,0 +1,267 @@
...
+    # If we tinker with the column then the name is the one of the kw argument
+    fig = px.scatter(tips, x=tips.index, y=10 * tips.total_bill)
+    assert fig.data[0]["hovertemplate"] == "index=%{x}<br>y=%{y}"
+    fig = px.scatter(tips, x=tips.index, y=tips.total_bill, labels={"index": "number"})
+    assert fig.data[0]["hovertemplate"] == "number=%{x}<br>total_bill=%{y}"
+    # We do not allow "x=index"
+    with pytest.raises(ValueError) as err_msg:
+        fig = px.scatter(tips, x="index", y="total_bill")
+        assert (
+            "ValueError: Value of 'x' is not the name of a column in 'data_frame'"
```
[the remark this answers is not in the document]
emmanuelle: we could, but this is quite a specific case. Would you like to print this only when the str is 'index'?

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
```diff
@@ -0,0 +1,267 @@
...
+            in str(err_msg.value)
+        )
+
+
+def test_pandas_series():
+    tips = px.data.tips()
+    before_tip = tips.total_bill - tips.tip
```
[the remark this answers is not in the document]
emmanuelle: yes I will remove one of those

packages/python/plotly/plotly/express/_core.py
emmanuelle: done

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
```diff
@@ -0,0 +1,267 @@
...
+        fig.data[0]["hovertemplate"]
+        == "total_bill=%{x}<br>tip=%{y}<br>hover_data_0=%{customdata[0]}"
+    )
+    tips = px.data.tips()
+    fig = px.scatter(
+        tips,
+        x="total_bill",
+        y="tip",
+        hover_data=[np.random.random(tips.shape[0])],
+        labels={"hover_data_0": "suppl"},
```
nicolaskruchten: this is nice

emmanuelle: and it should also go in the tutorial

packages/python/plotly/plotly/express/_core.py
emmanuelle: hum now, it won't work with a full `DataFrame` actually (an error will be raised because it has no `name`). I will correct the code and the doc.

packages/python/plotly/plotly/express/_doc.py
emmanuelle: done

packages/python/plotly/plotly/express/_core.py
emmanuelle: ok, corrected

packages/python/plotly/plotly/express/_core.py
emmanuelle: we could also raise the error in the function building the list of reserved names, which is always called

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
emmanuelle: this has been chnged, it would raise an error

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
emmanuelle: would raise an error

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,207 @@ def apply_default_cascade(args):
...
+                if isinstance(argument, int):
+                    col_name = _name_heuristic(argument, field, reserved_names)
+                    if field_name not in array_attrables:
+                        args[field_name] = col_name
+                    else:
+                        args[field_name][i] = col_name
+                df[col_name] = args["data_frame"][argument]
+                continue
+            # Case of multiindex
+            elif isinstance(argument, pd.core.indexes.multi.MultiIndex):
```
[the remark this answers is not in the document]
nicolaskruchten: yep.

packages/python/plotly/plotly/tests/test_core/test_px/test_px_input.py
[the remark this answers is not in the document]
nicolaskruchten: Yeah, that's a nice touch I think.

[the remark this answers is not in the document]
emmanuelle: done!

emmanuelle:
@nicolaskruchten I think I addressed all your comments. If you have a quick idea about the chart-studio related CI problem I'm all ears, otherwise I'll investigate elsewhere

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,215 @@ def apply_default_cascade(args):
...
+    if df_provided and not isinstance(args["data_frame"], pd.DataFrame):
+        args["data_frame"] = pd.DataFrame(args["data_frame"])
+
+    # We start from an empty DataFrame
+    df = pd.DataFrame()
+
+    # Initialize set of column names
+    # These are reserved names
+    if df_provided:
+        reserved_names = _initialize_argument_col_names(
```
nicolaskruchten: consistency-wise, maybe we should rename this function `_get_reserved_col_names` or something?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,215 @@ def apply_default_cascade(args):
...
+
+
+def _initialize_argument_col_names(args, attrables, array_attrables):
+    """
+    This function builds a list of columns of the data_frame argument used
+    as arguments, either as str/int arguments or given as columns
+    (pandas series type).
+    """
+    df = args["data_frame"]
+    used_col_names = set()
```
nicolaskruchten: here we call it `used_col_names` but elsewhere it's `reserved_names`

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,215 @@ def apply_default_cascade(args):
         args["marginal_x"] = None
 
 
+def _name_heuristic(argument, field_name, used_col_names):
```
nicolaskruchten: here we call it `used_col_names` but elsewhere it's `reserved_names`

emmanuelle: right, there were a few iterations, hence the different names. I will change this.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,215 @@ def apply_default_cascade(args):
...
+                            field,
+                            len(args["data_frame"][argument]),
+                            str(list(df.columns)),
+                            length,
+                        )
+                    )
+                col_name = argument
+                if isinstance(argument, int):
+                    col_name = _name_heuristic(argument, field, reserved_names)
+                    if field_name not in array_attrables:
```
nicolaskruchten: could we remove this `if/else` block as well as the `continue` and let it "fall through" to the bottom of the loop where we have the same logic?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -813,6 +813,12 @@ def build_dataframe(args, attrables, array_attrables):
     array_attrables : list
         argument names corresponding to iterables, such as `hover_data`, ...
     """
+    args = dict(input_args)
```
nicolaskruchten: I don't think this makes a deep copy...

```
>>> x= [1,2,3]
>>> y = dict(a=x)
>>> z = dict(y)
>>> z["a"] is x
True
>>> z["a"][0]=100
>>> z
{'a': [100, 2, 3]}
>>> x
[100, 2, 3]
```

But also think this is only an issue for `array_attrables` where we write into `args[something][something]` so if we just copy those that's enough:

```
>>> x = [1,2,3]
>>> y = list(x)
>>> y is x
False
>>> y[0] = 100
>>> x
[1, 2, 3]
```

packages/python/plotly/plotly/express/_core.py
```diff
@@ -813,6 +813,12 @@ def build_dataframe(args, attrables, array_attrables):
...
+    for field in args:
+        if field in array_attrables and isinstance(
+            args[field], pd.core.indexes.base.Index
+        ):
+            args[field] = list(args[field])
```
nicolaskruchten: following from comment above... if we just always do this, instead of in an `if` then I think we're safe (for now!)

packages/python/plotly/plotly/express/_core.py
emmanuelle: oh no! I knew it does a deep copy for a list so I had wrongly assumed I could transpose this pattern to a `dict`.

packages/python/plotly/plotly/express/_core.py
emmanuelle: good idea! Then we don't have to do a deepcopy of the whole `args`, it's better not to copy the `dataframe` in particular.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,214 @@ def apply_default_cascade(args):
...
+            continue
+        names = args[field] if field in array_attrables else [args[field]]
+        if names is None:
+            continue
+        for arg in names:
+            if arg is None:
+                continue
+            if isinstance(arg, str):
+                reserved_names.add(arg)
+            if isinstance(arg, pd.core.series.Series):
```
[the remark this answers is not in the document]
emmanuelle: ok thanks, I got the names by asking `type(df)` but more readable names are better!

nicolaskruchten:
> What is the simplest way to explain when the .name property of an input series is kept vs. when the arg name is used?

I believe that with what we landed on, the simplest explanation is that `.name` is used only to check if the input is the same object as the correspondingly-named column in `data_frame` (if `data_frame` is provided) otherwise it's ignored. Doing more than that caused all sorts of weird(er) conflicts!

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,214 @@ def apply_default_cascade(args):
...
+    for field in args:
+        if field not in attrables:
+            continue
+        names = args[field] if field in array_attrables else [args[field]]
+        if names is None:
+            continue
+        for arg in names:
+            if arg is None:
+                continue
+            if isinstance(arg, str):
```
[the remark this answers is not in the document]
nicolaskruchten: In general I agree except for cases where the first `if` results in a `continue` or errors out ;)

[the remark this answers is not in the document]
emmanuelle: not really... but this list of reserved names is tested against the names of keyword arguments which will never be ints (I think) so it should be ok. On the other hand it makes no harm to add int names to the list...

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,214 @@ def apply_default_cascade(args):
...
+                        % (
+                            field,
+                            len(df_input[argument]),
+                            str(list(df_output.columns)),
+                            length,
+                        )
+                    )
+                col_name = argument
+                if isinstance(argument, int):
+                    col_name = _name_heuristic(argument, field, reserved_names)
```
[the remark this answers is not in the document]
emmanuelle: turns out we could write simpler code here. Thanks !

emmanuelle:
Thanks a lot for your review @jonmmease, we have iterated a lot on this piece of code and having a fresh eye is wonderful. I think I addressed all your comments, which helped to improve the readability of the code.
Regarding when to use `name`, as explained by @nicolaskruchten we only used names of series equal to columns of the `dataframe` argument. With the possibility to pass arguments coming from different dataframes, handling conflicts was very tricky and hard to explain. Another possibility would have been to take the name whenever it exists and raise an error when a conflict arises, but we preferred to have a strict approach in the beginning and it is still possible in future versions to see if we want to change this behaviour.

nicolaskruchten:
Let's do a pass where we ensure all error messages have correct spacing between words: 

`ValueError: All arguments should have the same length.The length of argument `color` is 3, whereas thelength of previous arguments ['x', 'y', 'facet_row', 'facet_col'] is 4` ... "length.The length" and "thelength".

packages/python/plotly/plotly/express/_core.py
```diff
@@ -754,6 +754,210 @@ def apply_default_cascade(args):
...
+                    if col_name is None and is_index:
+                        col_name = "index"
+                    if not df_provided:
+                        col_name = field
+                    else:
+                        if is_index:
+                            keep_name = df_provided and argument is df_input.index
+                        else:
+                            # we use getattr/hasattr because of index
+                            keep_name = hasattr(
```
nicolaskruchten:
```python
df = px.data.tips()
px.scatter(df, x=df["size"], y=df.tip)
```

This has a weird behaviour: it doesn't pick up the name "size" even though it is a valid column... This is almost certainly because `df.size` doesn't return `df["size"]`. Thanks Pandas, again :(

nicolaskruchten:
OK, other than the spaces in the error messages and the `df.size` thing I think we're good! and we finally have the start of a decent PX test suite: yay!

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: I think the solution is to look up columns via `df_input["name"]` and not `getattr` and just special-case the `index` thing.

emmanuelle: actually it was already a special case, so the modification was easy

packages/python/plotly/plotly/express/_core.py
```diff
@@ -929,9 +929,9 @@ def build_dataframe(args, attrables, array_attrables):
                             keep_name = df_provided and argument is df_input.index
                         else:
                             # we use getattr/hasattr because of index
-                            keep_name = hasattr(
-                                df_input, col_name
-                            ) and argument is getattr(df_input, col_name)
+                            keep_name = col_name in df_input and argument.equals(
```
nicolaskruchten: Is it intentional to move from `is` to `.equals()` here?

nicolaskruchten: (also we should remove the comment above now ;)

emmanuelle: yes it was intentional, I think it's enough that the values are all equal. What do you think?

emmanuelle: ok I reverted to `is`

nicolaskruchten:
💃 💃 💃 thanks so much for getting this thing over the finish line!

emmanuelle:
yeehaw ! Merging :-)

thatneat:
Would love to consume this from PyPI! Is there a release planned?

nicolaskruchten:
Version 4.2 should come out this week, yes!

PR #1782, opened 2019-09-20
Update README.md for using install plotly jupyter widget

yoonghm:
The build system needs to have git and node.js.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,11 @@ or conda.
 conda install "notebook>=5.3" "ipywidgets=7.5"
 ```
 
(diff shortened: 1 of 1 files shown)


jonmmease:
Hi @yoonghm, thanks for the suggestions. You are right that nodejs is required for installing JupyterLab extensions, and we should add more info to the README on how to handle that dependency.  But it's not clear to me why git is required. Could you clarify?

nicolaskruchten:
I've updated the README to say we need node :)

PR #1785, opened 2019-09-22
Update _chart_types.py

yoonghm:
barnorm is default to None instead of ''

### DIFF
--- a/packages/python/plotly/plotly/express/_chart_types.py
+++ b/packages/python/plotly/plotly/express/_chart_types.py
@@ -779,7 +779,7 @@ def bar_polar(
     labels={},
     color_discrete_sequence=None,
     color_discrete_map={},
-    barnorm="",
+    barnorm=None,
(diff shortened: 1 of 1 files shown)


yoonghm:
The default parameter barnorm in bar_polar() should be None, as stated in the document rather than the code. Please review.

jonmmease:
Hi @yoonghm , thanks for your interest.  You're right that plotly.py does not explicitly include the default values of every property, and this is by design.  When a property is `None` on the Python side, this means that that the default will be chosen by the JavaScript library.

Is there a particular problem that you were trying to solve?

nicolaskruchten:
This seems like a good modification, I can't think of a valid reason for it to be set to `""` in `bar_polar()` and `None` in `histogram()`.

nicolaskruchten:
We should think about adding `barnorm` to `bar()` :)

PR #1786, opened 2019-09-22
Update _chart_types.py

yoonghm:
`points` parameter in `plotly.express.box()` is default to `"outliers"` instead of `None`


nicolaskruchten:
In this case, `None` is the correct default in Python, and the Javascript layer applies its own default behaviour, which generally works out to `"outliers"`.

PR #1791, opened 2019-09-27
use generators and comprehensions instead of lists

tweakimp:
I tried to improve the speed by preventing the creation of lists that are not needed and the use of generators and comprehensions. In some cases this could mean significant speedups.

### DIFF
--- a/packages/python/chart-studio/chart_studio/grid_objs/grid_objs.py
+++ b/packages/python/chart-studio/chart_studio/grid_objs/grid_objs.py
@@ -292,9 +292,7 @@ class Grid(MutableSequence):
                 break
 
         if column_id is None:
-            col_names = []
-            for column in self._columns:
-                col_names.append(column.name)
+            col_names = [column.name for column in self._columns]
             raise _plotly_utils.exceptions.PlotlyError(
                 "Whoops, that column name doesn't match any of the column "
                 "names in your grid. You must pick from {cols}".format(cols=col_names)
--- a/packages/python/chart-studio/chart_studio/grid_objs/grid_objs.py
+++ b/packages/python/chart-studio/chart_studio/grid_objs/grid_objs.py
@@ -200,22 +200,17 @@ class Grid(MutableSequence):
                         )
             # collect and sort all orders in case orders do not start
             # at zero or there are jump discontinuities between them
-            all_orders = [
-                columns_or_json["cols"][column_name]["order"]
-                for column_name in columns_or_json["cols"].keys()
-            ]
+            all_orders = [value["order"] for value in columns_or_json["cols"].values()]
             all_orders.sort()
 
             # put columns in order in a list
             ordered_columns = []
             for order in all_orders:
-                for column_name in columns_or_json["cols"].keys():
-                    if columns_or_json["cols"][column_name]["order"] == order:
+                for value in columns_or_json["cols"].values():
+                    if value["order"] == order:
                         break
 
-                ordered_columns.append(
-                    Column(columns_or_json["cols"][column_name]["data"], column_name)
-                )
+                ordered_columns.append(Column(value["data"], column_name))
             self._columns = ordered_columns
 
             # fill in column_ids
--- a/packages/python/chart-studio/chart_studio/grid_objs/grid_objs.py
+++ b/packages/python/chart-studio/chart_studio/grid_objs/grid_objs.py
@@ -164,9 +164,7 @@ class Grid(MutableSequence):
                 raise exceptions.InputError(err)
 
             # create columns from dataframe
-            all_columns = []
-            for name in columns_or_json.columns:
-                all_columns.append(Column(columns_or_json[name].tolist(), name))
+            all_columns = [Column(columns_or_json[name].tolist(), name) for name in columns_or_json.columns]
             self._columns = all_columns
             self.id = ""
 
@@ -199,9 +197,7 @@ class Grid(MutableSequence):
                         )
             # collect and sort all orders in case orders do not start
             # at zero or there are jump discontinuities between them
-            all_orders = []
-            for column_name in columns_or_json["cols"].keys():
-                all_orders.append(columns_or_json["cols"][column_name]["order"])
+            all_orders = [columns_or_json["cols"][column_name]["order"] for column_name in columns_or_json["cols"].keys()]
             all_orders.sort()
 
             # put columns in order in a list
--- a/packages/python/chart-studio/chart_studio/grid_objs/grid_objs.py
+++ b/packages/python/chart-studio/chart_studio/grid_objs/grid_objs.py
@@ -164,7 +164,10 @@ class Grid(MutableSequence):
                 raise exceptions.InputError(err)
 
             # create columns from dataframe
-            all_columns = [Column(columns_or_json[name].tolist(), name) for name in columns_or_json.columns]
+            all_columns = [
+                Column(columns_or_json[name].tolist(), name)
+                for name in columns_or_json.columns
+            ]
             self._columns = all_columns
             self.id = ""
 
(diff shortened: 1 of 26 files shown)


tweakimp:
I dont know what goes wrong with the 2.7 tests... :(

nicolaskruchten:
Thanks for the PR! I think there is something generally broken with that 2.7 test which I will try to fix.

In general are you able to quantify the performance gains here? Are there any cases where performance would degrade as a result of these changes?

tweakimp:
I can give you some simple examples:
```python
import time
from contextlib import contextmanager

n = 10000000


@contextmanager
def time_this(label):
    start = time.perf_counter_ns()
    try:
        yield
    finally:
        end = time.perf_counter_ns()
        print(f"TIME {label}: {(end-start)/1000000000} sec")


with time_this("append"):
    lst = []
    for x in range(n):
        if x % 2 == 0:
            lst.append(x)

with time_this("comprehension"):
    lst = [x for x in range(n) if x % 2 == 0]

with time_this("list call"):
    lst = list(x for x in range(n) if x % 2 == 0)
```
Returns:
```
TIME append: 1.3849639 sec
TIME comprehension: 0.8442628 sec
TIME list call: 1.0024824 sec
```

nicolaskruchten:
Yep, the principle makes sense, but I'm interested in seeing what real-world applications of `plotly` are sped by this PR... Do you have any examples of figures that are slow to generate today and faster as a result of this PR?

tweakimp:
> Yep, the principle makes sense, but I'm interested in seeing what real-world applications of `plotly` are sped by this PR... Do you have any examples of figures that are slow to generate today and faster as a result of this PR?

No, what would be the way to test this?

nicolaskruchten:
Similar to your code snippet above, just using `plotly` instead of synthetic examples :)

tweakimp:
I dont know what part of plotly I improved. I dont see what is called where, what is called how many times and where the speedups should be.

If tell me what kind of plot to create, I can test this.

nicolaskruchten:
OK. While I thank you for your work and interest in maintaining our codebase, I am reticent to merge this change, because some of the parts of the code this PR touches are not all that well-tested, and most of the changes you are proposing are refactoring away lists which in basically every case have 10-100 elements, so the absolute speedups in real-world use-cases are likely to be very small. As a result, the potential benefits of this PR don't really outweigh the risks of merging it from my perspective today. Re testing, I should note that I've resolved the previous issue with the 2.7 tests and the current failure is likely due to changes from this PR.

If you're able to point to specific figures (say some of the sample figures in our documentation at https://plot.ly/python) and find significant speedups this might change the conversation, however.

tweakimp:
Your fix of the 2.7 tests made it possible for me to fix an actual bug, it all works now.

sursu:
@nicolaskruchten I think it's easier to look at functions being changed, than at plots.
For instance, take the [endpoints_to_interval](https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/figure_factory/_scatterplot.py#L17) function being which is changed  [here](
https://github.com/plotly/plotly.py/pull/1791/commits/6d2435bf891a2d813fd55352bd853919a8e1632a) (last).

It is easy to see that that function is now cleaner. And it happens to be faster (but, of course, we are talking about microseconds here).

I would suggest modifying it to:

```python
        # add -inf to intervals
        intervals = [[float("-inf"), endpts[0]]]
        for k in range(length - 1):
            intervals.append([endpts[k],endpts[k + 1]])
```

packages/python/plotly/plotly/figure_factory/_scatterplot.py
```diff
@@ -57,10 +57,7 @@ def endpts_to_intervals(endpts):
         # add -inf to intervals
         intervals = [[float("-inf"), endpts[0]]]
         for k in range(length - 1):
-            interval = []
-            interval.append(endpts[k])
-            interval.append(endpts[k + 1])
-            intervals.append(interval)
+            intervals.append([endpts[k],endpts[k + 1]])
```
sursu: Proposed change:
```python
        # add -inf to intervals
        intervals = [[float("-inf"), endpts[0]]]
        for k in range(length - 1):
            intervals.append([endpts[k], endpts[k + 1]])
```

packages/python/chart-studio/chart_studio/presentation_objs/presentation_objs.py
```diff
@@ -973,6 +973,7 @@ def _markdown_to_presentation(self, markdown_string, style, imgStretch):
             if slide.count("```") % 2 != 0:
                 raise _plotly_utils.exceptions.PlotlyError(CODE_ENV_ERROR)
 
+            wdw_size = len("```")
```
sursu: Why not simply set `wdw_size = 3`? or why even create this variable, it's being used only in one place?

packages/python/chart-studio/chart_studio/grid_objs/grid_objs.py
```diff
@@ -199,9 +197,7 @@ def __init__(self, columns_or_json, fid=None):
                         )
             # collect and sort all orders in case orders do not start
             # at zero or there are jump discontinuities between them
-            all_orders = []
-            for column_name in columns_or_json["cols"].keys():
-                all_orders.append(columns_or_json["cols"][column_name]["order"])
+            all_orders = [columns_or_json["cols"][column_name]["order"] for column_name in columns_or_json["cols"].keys()]
```
sursu: I think this can be simplified to:
```python
all_orders = [D['order'] for D in columns_or_json["cols"].values()]
```
but, I'm not sure what `columns_or_json` looks like. I guess it's a dictionary.

nicolaskruchten:
> but, of course, we are talking about microseconds here

So this is kind of my concern... It doesn't seem to be worth the risk of introducing a new bug in reasonably-working, marginally-tested code for this kind of improvement, either in performance or in style.

packages/python/plotly/plotly/figure_factory/_scatterplot.py
tweakimp: You looked at outdated code, it is already changed in the current version of the pull request.

packages/python/chart-studio/chart_studio/presentation_objs/presentation_objs.py
tweakimp: I tried to change as little as possible. There might be a reason (readability, further use in the future) why this variable was there before.

packages/python/chart-studio/chart_studio/grid_objs/grid_objs.py
tweakimp: You are right. Please dont mix single and double quotes :)

tweakimp: You are right, I changed this there and below.

nckswt:
@nicolaskruchten I stumbled across this PR when trying to solve my own problem. It's quite possible that Plot.ly isn't the right tool for this, but I wanted to animate a scatterplot of 200 points for 1000+ frames. I used [this page](https://plot.ly/python/animations/#moving-frenet-frame-along-a-planar-curve) as a starting point. 

Having to pre-generate the frames means I have to wait for nearly a minute before my graph shows up. Is there some other way to generate frames during an animation, instead of pre-generating all frames before the animation? Passing a generator to `frames` would be the ideal solution, but I [didn't see any reference to it here](https://plot.ly/python/reference/).

SankarGaneshb:
@tweakimp I like this pull request for the right clean up of existing code. Idea for performance with very big data.

@nicolaskruchten True, the impact it is going to make is very minimal and more over the Chart Studio package is moved to different source this pull request might not need. 

@nckswt Even As per @tweakimp the impact for millisecond improvement is for the count of 10000000, but as such you were talking about 200000+ alone so this change might not be of great help.

This pull request can be marked as not requried further. 

Thanks.

marthacryan:
Hey @tweakimp! After reading through this discussion, I think I'll close this. It seems like it would be a good idea to use these more, but it might not be worth the risks associated with changing this much logic. Thank you for this PR though!

PR #1796, opened 2019-10-03
[WIP] document API using sphinx

emmanuelle:
This is very much WIP! Just a proof of concept showing how to use sphinx in order to document the API. I have only included a few modules, but if we decide to pursue this path it's not a big effort to add the other ones.


emmanuelle:
Closing in favor of #1802, #1803 and https://github.com/plotly/plotly.py-docs/pull/134

PR #1803, opened 2019-10-06
__all__ for figure_factory and io (for sphinx)

emmanuelle:
I modified the `__all__` variable of `figure_factory` and `io` so that sphinx finds the functions when autodocumenting modules.


nicolaskruchten:
💃

PR #1817, opened 2019-10-14
Add convenience methods for annotations, shapes, and images

jonmmease:
## Overview
This PR adds a set of new convenience methods for annotations, shapes, and images
 - `fig.add_*`
 - `fig.select_*` 
 - `fig.for_each_*`
 - `fig.update_*`

Where `*` is annotation(s), shape(s), or image(s).

cc @nicolaskruchten


nicolaskruchten:
Possible issue: we’re going to have an ‘image’ trace type before long so there would be a conflict around add_image... should we use add_layout_image?

jonmmease:
Oh, good point. Thanks for catching that! I'm fine with `add_layout_image`/`select_layout_images`/etc.  Should we do the same for shapes and annotations or treat image as a special case?

nicolaskruchten:
Tough call but I think special case for image makes the most sense to me. There isn’t and likely won’t be a non-layout annotation or shape...

jonmmease:
Renamed in fd0ae43

nicolaskruchten:
Looks good, let's merge it and I'll QA the whole thing this afternoon? 💃

Issue #1829, opened 2019-10-18
Plotly 4.2 breaks figure_factory (when scikit-image isn't installed)


ethanopp:
`scikit-image` needs to be added to dependencies

```
C:\Users\OPPENE\PyCharm\Projects\fitness\venv\Scripts\python.exe C:/Users/OPPENE/PyCharm/Projects/fitness/index.py
Traceback (most recent call last):
  File "C:/Users/OPPENE/PyCharm/Projects/fitness/index.py", line 6, in <module>
    from pages import performance, home, settings, power
  File "C:\Users\OPPENE\PyCharm\Projects\fitness\pages\home.py", line 5, in <module>
    import plotly.figure_factory as ff
  File "C:\Users\OPPENE\PyCharm\Projects\fitness\venv\lib\site-packages\plotly\figure_factory\__init__.py", line 27, in <module>
    from plotly.figure_factory._ternary_contour import create_ternary_contour
  File "C:\Users\OPPENE\PyCharm\Projects\fitness\venv\lib\site-packages\plotly\figure_factory\_ternary_contour.py", line 11, in <module>
    from skimage import measure
ModuleNotFoundError: No module named 'skimage'

Process finished with exit code 1
```

nicolaskruchten:
Thanks for bringing this up, it's a regression and we'll fix it!

nicolaskruchten:
Our intention was not to have *all* figure factories depend on scikit-image, just the ternary contours one I believe.

The workaround for 4.2 until the fix is out is just to install the needed dependencies *if you are using a figure factory*.

ethanopp:
Yep that's what I did :)

nicolaskruchten:
Fixed in 4.2.1

PR #1832, opened 2019-10-18
fixed skimage import

### LINKED ISSUE #1826: Does Plotly 4.2.0 depend on scikit-image?
I failed to `import plotly.figure_factory` because plotly seems not to install `scikit-image` when running `pip install -U plotly`. After I manually installed `scikit-image`, `import plotly.figure_factory` worked.

This was not a problem in version 4.1.1.

But the source code shows it depends on it.
https://github.com/plotly/plotly.py/blob/b7ad5433c4e0882715781fa6c4816fc7fff62965/packages/python/plotly/plotly/figure_factory/_ternary_contour.py#L11

emmanuelle:
Closes #1826 and #1829

### DIFF
--- a/packages/python/plotly/plotly/express/__init__.py
+++ b/packages/python/plotly/plotly/express/__init__.py
@@ -4,6 +4,7 @@ data exploration and figure generation. See the gallery at https://plotly.github
 """
 from __future__ import absolute_import
 from plotly import optional_imports
+
 pd = optional_imports.get_module("pandas")
 if pd is None:
     raise ImportError(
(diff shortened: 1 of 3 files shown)


nicolaskruchten:
So with this we just can’t see certain factories when the deps aren’t installed? Shouldn’t we just throw a call-time error instead? Otherwise someone will try to run the docs and be confused about import failures, no?

nicolaskruchten:
Or does “optional imports” warn or something?

emmanuelle:
We can probably import the dep inside the `ff` function instead, yes. I just copied the behaviour which was used for `pandas` and `county_choropleth`, I did not give it much thought

nicolaskruchten:
💃

PR #1833, opened 2019-10-19
remove pywavelets from CI

emmanuelle:
The pywavelets team has been pretty quick at fixing the py2 issue, so let's remove this tox dep before we forget why it was introduced....


nicolaskruchten:
💃

PR #1835, opened 2019-10-20
Reformat docstrings

joelostblom:
@nicolaskruchten As per our discussion in https://github.com/plotly/plotly_express/issues/154

I tried not making more changes than necessary so I didn't go after things like changing "numbers" to "int" or "float". I used `make_subplots.py` as a guide and tried to be consistent with the formatting there (with the one exception being dict that I changed there instead according to what seemed to be the most frequently used elsewhere in plotly).

Please have a thorough look through since I might have missed some parenthesis somewhere. 

Close https://github.com/plotly/plotly_express/issues/154

### DIFF
--- a/packages/python/plotly/plotly/express/_doc.py
+++ b/packages/python/plotly/plotly/express/_doc.py
@@ -2,12 +2,6 @@ import inspect
 from textwrap import fill, indent
 
 
-colref = ("str or int or Series or array-like",
-          "Either a name of a column in `data_frame`, or a pandas Series or array_like object.")
-
-colref_list = ("list of str or int, or Series or array-like",
-               "Either names of columns in `data_frame`, or pandas Series, or array_like objects")
-
 # TODO contents of columns
 # TODO explain categorical
 # TODO handle color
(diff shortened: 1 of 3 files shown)


joelostblom:
I'm note sure how to fix the Python2 related test errors, I believe unpacking operators have been available in Python for quote some time, so not sure why `*colref` is marked as invalid syntax.

emmanuelle:
Thank you for your pull request, it is clearly an improvement, also in plain `ipython` :-). Yeah, it looks like `*colref` is not valide py2 syntax. You have several options to fix this I think, either use list comprehension, or `colref[0]` and `colref[1]`. I'll merge once the CI is fixed. Thanks again!

joelostblom:
@emmanuelle I had to reformat one more thing for Python2 comparability and now all tests are passing!

packages/python/plotly/plotly/express/_doc.py
```diff
@@ -11,377 +10,444 @@
...
-        colref,
-        "Values from this column or array_like are be interpreted according to `locationmode` and mapped to longitude/latitude.",
+        colref_type,
+        colref_desc,
+        "Values from this column or array_like are to be interpreted according to `locationmode` and mapped to longitude/latitude.",
     ],
     dimensions=[
-        "(list of strings, names of columns in `data_frame`)",
-        "Columns to be used in multidimensional visualization.",
+        "list of str",
```
nicolaskruchten: I think this is a `colref_list_type` no? @emmanuelle thoughts?

nicolaskruchten:
This is pretty great, thank you! Modulo my one comment we can merge this :)

packages/python/plotly/plotly/express/_doc.py
emmanuelle: yes, correct, we forgot to update this one. @joelostblom would you mind making this change? Otherwise I can push a commit to your branch if you don't have the time. And then we'll merge :-)

joelostblom: @emmanuelle Updated!

joelostblom:
The failed test seems unrelated.

nicolaskruchten:
This is really nice work, thanks very much! 💃

PR #1837, opened 2019-10-20
docstring refactoring in figure factory

emmanuelle:
Goals
- update docstring `iplot` --> `fig.show()`
- format code blocks so that they render nicely (syntax highlighting) in API ref

image

Note that the `>>>` do not prevent pasting in a notebook cell, jupyter know how to handle them.

I'll work through the list of figure factories but I wanted to open this first as WIP for the sake of discussing this update.


emmanuelle:
@nicolaskruchten @jonmmease ideally we would include this in the release, are you fine with this change (adding the `>>>`)

nicolaskruchten:
I’m fine with this whole class of change so feel free to merge these types of doc string PRs at will unless you’re concerned about something and want input!

emmanuelle:
ok thanks!

jonmmease:
looks great to me! :dancer:

PR #1840, opened 2019-10-22
Update _orca.py

standbyme:
Currently, this link guides user to [orca.](https://github.com/plotly/orca.) instead of [orca](https://github.com/plotly/orca)


emmanuelle:
Thank you very much! Merging.

standbyme:
: )

Issue #1570, opened 2019-05-20
Allow configure orca to send requests to remote server


cooperok:
According to [code](https://github.com/plotly/plotly.py/blob/master/plotly/io/_orca.py#L1228) static image export will work only if orca is installed on same server as plotly.py

`server_url = 'http://{hostname}:{port}'.format(hostname='localhost', port=orca_state['port'])`

But, there is an [option](https://github.com/plotly/orca/tree/master/deployment) to run orca as standalone server. Would be greate if param hostname can be specified in options same as port.

My case is that I have Google App Engine application, and it's problematic to install orca. So I want to set up standalone image server and still use plotly to get images like plotly.io.to_image(...)

jonmmease:
Hi @cooperok, yes this is a great idea.
Happy to help someone through a PR on this if anyone is interested in tackling this.

I think we could probably add a `plotly.io.orca.config.server_url` parameter. Setting this would then unset (or override) the `executable`, `port`, and `timeout`, `mathjax`, `topojson`, and `mapbox_access_token` properties.

maresb:
For a quick-and-dirty workaround to specify a custom hostname, I found the following two possibilities. 

1. If you are able to edit the source code, then edit `_orca.py` and manually set the desired hostname and port in the `server_url = ` command.  Since we won't be using the `orca` executable, at the end of `_orca.py` add the line `ensure_server = lambda: None` to disable the corresponding checks.  (This sets `ensure_server` to the do-nothing function.)

2. If for some reason you are unable or unwilling to edit the `plotly.py` source code, then on the machine running `plotly.py`, run the command 

        socat tcp-listen:9091,reuseaddr,fork,bind=127.0.0.1 tcp:orca-host:9091

    where `orca-host` is the hostname of the orca server.  This creates a tunnel to access `orca-host` from `localhost` of the machine running plotly.py, using the default port 9091.  Finally, import plotly.io with the following extra commands:

        import plotly.io as pio
        pio._orca.ensure_server = lambda: None
        pio._orca.orca_state["port"] = 9091


With either of these techniques, I can set up docker-compose to run [Jupyter](https://github.com/jupyter/docker-stacks) + plotly.py, and [orca](https://github.com/plotly/orca/tree/master/deployment).

miriad:
@jonmmease  , Any chance you could review the PR on this?

jonmmease:
done in #1850

PR #1850, opened 2019-10-28
Add server_url to plotly.io.orca to allow for external orca server

miriad:
Add server_url to plotly.io.orca.config to allow an external orca server.

Implementation of #1570


jonmmease:
Thanks @miriad! We'll try to take close look soon, and we should be able to get this in for version 4.3.0 (probably a week or two out).

emmanuelle:
@miriad @jonmmease any idea of how to add tests for this? Open a local orca server and communicate with it?

miriad:
@emmanuelle 

That's something I struggled with, because of the need to run an external service.

Seemingly, running a local orca server and communicating with it via localhost would be sufficient.

jonmmease:
Looks great @miriad. I added a test, made a couple small cleanup changes, and merged master.   I'm very glad to have this option available now. Thanks!

ferran-candela:
Hello,

I have tried to configure the external server by means of setting up the new available property 'server_url' but after doing it like this:

```
    import plotly.io as pio
    pio.orca.config.server_url = "localhost"
    pio.orca.config.port = 9091
```

The property 'executable' does not change to 'None' so the executable continues being required.

After debugging I have seen that the setter does not change the property since self._props is empty.
```
    @executable.setter
    def executable(self, val):

        if val is None:
            self._props.pop("executable", None)
```

Maybe I am not configuring it correctly.
Thanks.

dev-dsp:
I can confirm that it just doesn't work. 
If I set executable to None by editing few lines in io/_orca.py, validate_executable() fails with exception, so it's not possible to run pio.to_image()

miriad:
Please try including the port directly in server_url, such as "localhost:9091"

The port field is intended for use with a locally managed server, and setting will override the server_url field

dev-dsp:
Please find my comment in above mentioned issue.

dev-dsp:
Opened PR for fixing that: https://github.com/plotly/plotly.py/pull/1915

PR #1855, opened 2019-10-30
Imshow

emmanuelle:
This is still WIP, I need to track some corner cases, but this is to give an idea of what the `imshow` function could look like.

### DIFF
--- a/packages/python/plotly/plotly/express/_imshow.py
+++ b/packages/python/plotly/plotly/express/_imshow.py
@@ -1,0 +1,127 @@
+import plotly.graph_objs as go
+import numpy as np  # is it fine to depend on np here?
+
+_float_types = []
+
+# Adapted from skimage.util.dtype
+_integer_types = (
+    np.byte,
+    np.ubyte,  # 8 bits
+    np.short,
+    np.ushort,  # 16 bits
+    np.intc,
+    np.uintc,  # 16 or 32 or 64 bits
+    np.int_,
+    np.uint,  # 32 or 64 bits
+    np.longlong,
+    np.ulonglong,
+)  # 64 bits
+_integer_ranges = {t: (np.iinfo(t).min, np.iinfo(t).max) for t in _integer_types}
+
+
+def _vectorize_zvalue(z):
+    if z is None:
+        return z
+    elif np.isscalar(z):
+        return [z] * 3 + [1]
+    elif len(z) == 1:
+        return list(z) * 3 + [1]
+    elif len(z) == 3:
+        return list(z) + [1]
+    elif len(z) == 4:
+        return z
+    else:
+        raise ValueError(
+            "zmax can be a scalar, or an iterable of length 1, 3 or 4. "
+            "A value of %s was passed for zmax." % str(z)
+        )
+
+
+def _infer_zmax_from_type(img):
+    dt = img.dtype.type
+    if dt in _integer_types:
+        return _integer_ranges[dt][1]
+    else:
+        return img[np.isfinite(img)].max()
+
+
+def imshow(
+    img, zmin=None, zmax=None, origin=None, colorscale=None, showticks=True, **kwargs
+):
+    """
+    Display an image, i.e. data on a 2D regular raster.
+
+    Parameters
+    ----------
+
+    img: array-like image
+        The image data. Supported array shapes are
+
+        - (M, N): an image with scalar data. The data is visualized
+          using a colormap.
+        - (M, N, 3): an image with RGB values.
+        - (M, N, 4): an image with RGBA values, i.e. including transparency.
+
+    zmin, zmax : scalar or iterable, optional
+        zmin and zmax define the scalar range that the colormap covers. By default,
+        zmin and zmax correspond to the min and max values of the datatype for integer
+        datatypes (ie [0-255] for uint8 images, [0, 65535] for uint16 images, etc.), and
+        to the min and max values of the image for an image of floats.
+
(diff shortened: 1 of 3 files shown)


packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -0,0 +1,128 @@
...
+
+    origin : str, 'upper' or 'lower' (default 'upper')
+        position of the [0, 0] pixel of the image array, in the upper left or lower left
+        corner. The convention 'upper' is typically used for matrices and images.
+
+    colorscale : str
+        colormap used to map scalar data to colors (for a 2D image). This parameter is not used for
+        RGB or RGBA images.
+
+    showticks : bool, default True
```
nicolaskruchten: Hmm this one isn't in the matplotlib version and isn't in any other PX method... I would be in favour of removing it and documenting using `.update_xaxes(ticks=False).update_yaxes(ticks=False)` instead. Thoughts?

emmanuelle: ok ! It was a matter of convenience but we don't want to trade too much consistency for convenience so I'll remove this.

emmanuelle: I have also updated the doc tutorial accordingly

nicolaskruchten: thanks!

packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -0,0 +1,122 @@
+import plotly.graph_objs as go
+import numpy as np  # is it fine to depend on np here?
```
nicolaskruchten: it is fine, but we should have a little friendly error message if it's not installed, similar to the pandas one for px, no?

in fact, px depends on numpy but doesn't have a friendly error message... maybe we can do both in a separate PR.

How about: let's open a separate issue to discuss the `numpy` dependency and move forward with this as is :)

packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -0,0 +1,122 @@
...
+            "A value of %s was passed for zmax." % str(z)
+        )
+
+
+def _infer_zmax_from_type(img):
+    dt = img.dtype.type
+    if dt in _integer_types:
+        return _integer_ranges[dt][1]
+    else:
+        return img[np.isfinite(img)].max()
```
nicolaskruchten: Per our conversation, I think this should return "the smallest power of 255 which is greater than the max".

My rationale here is that if you have a pipeline that's intended to produce output between 0-X, and you use this to display the output of multiple inputs, it would be nice for them to have the same bounds, instead of having the bounds vary. And it feels like the most likely values of X are 1 and 255, and then possibly thereafter some powers thereof, if the data is 16-bit or 32-bit or whatnot :)

packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -0,0 +1,122 @@
...
+        img = 255 * img.astype(np.uint8)
+
+    # For 2d data, use Heatmap trace
+    if img.ndim == 2:
+        if colorscale is None:
+            colorscale = "gray"
+        trace = go.Heatmap(z=img, zmin=zmin, zmax=zmax, colorscale=colorscale)
+        autorange = True if origin == "lower" else "reversed"
+        layout = dict(
+            xaxis=dict(scaleanchor="y", constrain="domain"),
```
nicolaskruchten: did the domain constraints get added in by default in the JS layer finally? If so, we could skip them here

nicolaskruchten: ah never mind, this is the heatmap path.

emmanuelle: that's for the heatmap part here, for single-channel images. The JS Heatmap does not use a scaleanchor (as opposed to the Image trace)

packages/python/plotly/plotly/express/_imshow.py
emmanuelle: one thing which bothers me though is the following case: suppose an image is in the [0-1] range but some numerical computation (filter etc.) changes the max to 1 + some small value. Then the zmax will be 255 for a max of say 1.05, and it will look really bad.

emmanuelle: I can add a multiplicative tolerance factor like 10%

nicolaskruchten: I'd be OK with this

emmanuelle: how about having a keyword argument `colorrange` (or a better name consistent maybe with other parts of the library) which could be `data` (take min/max) or `dtype` (inference based on what we've discussed)? With this I think most people would be happy... Or does this complicate too much the API ?

nicolaskruchten: My feeling is that that complicates too much

emmanuelle:
OK, so here is a summary of the zmax behaviour
- for single-channel images, we pass directly to Heatmap (which takes the max of the image)
- for RGB images
      * for integer types we take the max of the dtype (for uint8 images, the most common type, it corresponds to the 255 default of go.Image)
      * for floats the max of the image is computed and zmax is the smallest power of 256 (1, 255, 65535) greater than this max value, with a 5% tolerance (to account for overflows resulting from numerical computations, eg filters).

In all cases, the user can override the zmax value.

nicolaskruchten:
Awesome! It's nice to have someone who knows this world on the team :)  💃

PR #1863, opened 2019-11-04
Remove marginal plot grids as hardcoded default

joelostblom:
I suggest to remove that hardcoded grids for marignal plots and instead inherit the grid settings from the template. This is particularly useful for templates that don't use grids.

The second commit sets two defaults that I think are almost always wanted: No y-axis line or ticks for the marginal plots. I am happy to drop this change with the same logic as above if you think these should be configurable from templates as well.

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -375,11 +375,11 @@ def configure_cartesian_marginal_axes(args, fig, orders):
 
     # Configure axis ticks on marginal subplots
     if args["marginal_x"]:
-        fig.update_yaxes(showticklabels=False, row=nrows)
+        fig.update_yaxes(showticklabels=False, showline=False, ticks="", row=nrows)
         fig.update_xaxes(row=nrows)
 
     if args["marginal_y"]:
-        fig.update_xaxes(showticklabels=False, col=ncols)
+        fig.update_xaxes(showticklabels=False, showline=False, ticks="", col=ncols)
         fig.update_yaxes(col=ncols)
 
     # Add axis titles to non-marginal subplots
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -375,16 +375,12 @@ def configure_cartesian_marginal_axes(args, fig, orders):
 
     # Configure axis ticks on marginal subplots
     if args["marginal_x"]:
-        fig.update_yaxes(
-            showticklabels=False, showgrid=args["marginal_x"] == "histogram", row=nrows
-        )
-        fig.update_xaxes(showgrid=True, row=nrows)
+        fig.update_yaxes(showticklabels=False, row=nrows)
+        fig.update_xaxes(row=nrows)
 
     if args["marginal_y"]:
-        fig.update_xaxes(
-            showticklabels=False, showgrid=args["marginal_y"] == "histogram", col=ncols
-        )
-        fig.update_yaxes(showgrid=True, col=ncols)
+        fig.update_xaxes(showticklabels=False, col=ncols)
+        fig.update_yaxes(col=ncols)
 
     # Add axis titles to non-marginal subplots
     y_title = get_decorated_label(args, args["y"], "y")
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -375,12 +375,11 @@ def configure_cartesian_marginal_axes(args, fig, orders):
 
     # Configure axis ticks on marginal subplots
     # Don't change the marginal grid settings if the template already specifies `showgrid`
-    if isinstance(
-        pio.templates[args["template"]].layout.xaxis.showgrid, bool
-    ) or isinstance(pio.templates[args["template"]].layout.yaxis.showgrid, bool):
-        grid_set = True
-    else:
+    if (pio.templates[args["template"]].layout.xaxis.showgrid is None
+            or pio.templates[args["template"]].layout.yaxis.showgrid is None):
         grid_set = False
+    else:
+        grid_set = True
 
     if args["marginal_x"]:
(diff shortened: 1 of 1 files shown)


joelostblom:
The failing test seems unrelated to the changes I made.

nicolaskruchten:
Can you post a screenshot of some figures with marginals and the default template, given these changes please?

I’m not sure about “always wanted” btw... every explicit setting in PX is there for a reason ;) in this case I found the grid-less marginals with background color pretty ugly, and the bar heights harder to compare, as Weber’s Law would predict.

joelostblom:
Sorry, I didn't mean to imply that the current settings were there for no good reason and I agree that grid lines often facilitates comparisons. I just think the grid lines should be configurable and inherited from the template. So for the default template, there will still be grids in the marginal plots:

image

image

image

But with templates that don't include grids, the marginal plots will not add extra grids:
image

nicolaskruchten:
Hmm I don’t love the loss of the continuity of grid lines from the main plot... 

We could leave the current PX settings alone if you set your new template’s grid lines to white tho!

nicolaskruchten:
PS the new template looks very slick, looking forward to seeing it in 4.3!

joelostblom:
> We could leave the current PX settings alone if you set your new template’s grid lines to white tho!

I considered that, but think it would be confusing if someone tried to add grids by setting `showgrid=True` and the lines would still not be visible because they are white. I can turn off the marginal grids in the template via the secondary and tertiary xy-axes, I am just not sure if that would remove grids from other plots in a unexpected manner. But before I go that route, just one more comment on this PR.

> Hmm I don’t love the loss of the continuity of grid lines from the main plot...

Except for histograms, plotly currently only adds gridlines on the y-axis for marginal plots, so although there is continuity from the main plot grid lines, there is still some visual dissonance when looking at a plot like the below since there are square grids only on 2 of 3 plots:

image

I see the value in the xy-grids for the histogram as both axes represent a value and not a grouping as in the rug plots. However, the default for histograms when created as standalone plots is to have only y-grid lines.

image

Although I they are much closer to the x-axis in the standalone plot, it is still easy to see the x-values for the histogram in the marginal plot since the grid lines from the main plot ends just below the histogram and there is little utility in continuing them.

Just some aspects to consider before deciding.

> PS the new template looks very slick, looking forward to seeing it in 4.3!

Thank you! =D

nicolaskruchten:
I think maybe a better fix would probably be to have PX check to see if the active template sets `showgrid` and not touch it if it does? I don't think the current default template sets this attr. 

I doubt that trying to set this in the second/third plots is going to go well template-wise :)

nicolaskruchten:
PX in general does need to tread more lightly to let templates show through, and I think this sort of "if the template doesn't set it, it's fair game, otherwise no" approach is a good approach... thoughts?

joelostblom:
I like it! I will try that tomorrow, first I am going to get a PR up for the template suggestions so you can have a look.

joelostblom:
@nicolaskruchten Added a commit with your suggested approach.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -374,17 +374,27 @@ def configure_cartesian_marginal_axes(args, fig, orders):
...
+    else:
+        grid_set = False
+
     if args["marginal_x"]:
-        fig.update_yaxes(
-            showticklabels=False, showgrid=args["marginal_x"] == "histogram", row=nrows
-        )
-        fig.update_xaxes(showgrid=True, row=nrows)
+        fig.update_yaxes(showticklabels=False, showline=False, ticks="", row=nrows)
+        fig.update_xaxes(row=nrows)
```
nicolaskruchten: this line won't do anything

packages/python/plotly/plotly/express/_core.py
```diff
@@ -374,17 +374,27 @@ def configure_cartesian_marginal_axes(args, fig, orders):
...
+            fig.update_yaxes(showgrid=args["marginal_x"] == "histogram", row=nrows)
+            fig.update_xaxes(showgrid=True, row=nrows)
 
     if args["marginal_y"]:
-        fig.update_xaxes(
-            showticklabels=False, showgrid=args["marginal_y"] == "histogram", col=ncols
-        )
-        fig.update_yaxes(showgrid=True, col=ncols)
+        fig.update_xaxes(showticklabels=False, showline=False, ticks="", col=ncols)
+        fig.update_yaxes(col=ncols)
```
nicolaskruchten: this line won't do anything

packages/python/plotly/plotly/express/_core.py
```diff
@@ -374,17 +374,27 @@ def configure_cartesian_marginal_axes(args, fig, orders):
         set_cartesian_axis_opts(args, xaxis, "x", orders)
 
     # Configure axis ticks on marginal subplots
+    # Don't change the marginal grid settings if the template already specifies `showgrid`
+    if isinstance(
+        pio.templates[args["template"]].layout.xaxis.showgrid, bool
```
nicolaskruchten: here you should check `fig.layout.template` actually

nicolaskruchten: Sorry, that's not right... you can just check `args["template"]` directly, at this point it should already be a template object.

nicolaskruchten: Sorry, that's not right either! but it should be... I'm going to PR that in.

packages/python/plotly/plotly/express/_core.py
joelostblom: woopsie =p

packages/python/plotly/plotly/express/_core.py
joelostblom: I removed the `pio.templates` part to be compatible with your upcoming PR. I also changed the comparison to be against `None` for clarity.

nicolaskruchten:
Thanks for pushing on this! I've rolled these changes into #1875 so I'll close this PR :)

PR #1864, opened 2019-11-04
Add mpl-like template

### LINKED ISSUE #1834: Add white templates with axes lines, tickmarks, and different color rotation
Would you consider adding two templates with axis lines, tickmarks, and white background (similar to "xticks" in seaborn)? I believe this is a rather common style in various plotting libraries that also looks good in print and it would be nice to have it easily accessible, especially when teaching so that custom templates need not to be explained early on.

To illustrate what I mean

image

And with gridlines

image

I also changed the color rotations to D3, viridis, and RdBu, which are common defaults elsewhere. Below is the code I used to create the first figure, it is not complete with the figure specific template properties that are in the other default templates but gives an idea of what I am referring to.

```python
go.layout.Template(
    layout=go.Layout(
        hovermode='closest',
        hoverlabel_align='left',
        plot_bgcolor='white',
        paper_bgcolor='white',
        font_size=13,
        xaxis=dict(showline=True, ticks='outside', showgrid=False,
                   linewidth=1.5, zeroline=False),
        yaxis=dict(showline=True, ticks='outside', showgrid=False,
                   linewidth=1.5, zeroline=False),
        colorway=px.colors.qualitative.D3,
        colorscale=dict(sequential=px.colors.sequential.Viridis,
                        diverging=px.colors.diverging.RdBu), 
    )
)
```

joelostblom:
@nicolaskruchten The PR is finally here! There are still a few things to test, but please have a look nd let me know if this makes sense so far. Notably, I still have to tend to these items

- [x] ~test table colors~ #1869 
- [x] test annotations
- [x] ~no axes lines in scatter matrix~ see comment posted below
- [x] ~no ticks on colorbars~ #1867
- [x] consider adding weak transparency on plots that are likely to overlap such as scatters
- [x] take final decision on the sequential colormap

I suggest naming this template mpl as a reference to the default style in matplotlib. The most notable difference would be that the top and rightmost axis lines are omitted here whereas matplotlib plots full boxes around the plots by default. The ticks, bg color, and color rotations are all the same as the default matplotlib style (altuogh I kept gray14 as the black color for now). I thought it made sense with this consistency so I included Viridis as the default color map instead of Inferno or Blues which I would be more likely to choose myself. For complete adherence, I could add the full box and then make another template called "despine" or similar to remove the topmost and rightmost axis lines.

I would also be happy to call this template xticks or axes lines and be more liberal in some of the choices, let me know what you think makes the most sense.

Some screens:
image

image

image

image

close #1834


emmanuelle:
How about naming the template `simple_white` or `basic_white` ?

joelostblom:
Sounds good to me, I think I would prefer "simple" over "basic", and open to other suggestions also.

Any idea why the axes lines are hidden in scatter matrices? I have not been able to find how to make it read the layout from the template instead of defaulting to hiding axes, but no luck so far.

image

nicolaskruchten:
Let's go with `simple_white` then.

The reason you don't get axis lines in the inner plots in a SPLOM is that there aren't any axes there. Unlike the output of `make_subplots` wherein every subplot has its own X and Y axes that lock using `matches`, in a SPLOM each column shares an X axis (physically located at the bottom) and each row shares a Y axis (on the left) so those are the only places you'll ever see a line.

joelostblom:
Done! I wont add any new things now, just fix what you think need fixing. This is what the scatter matrix looks like with the added transparency.

image

Ideally I would remove the upper half and show histograms on the diagonal, but I suspect that would be too much alteration in an official template so I will make those changes locally instead.

nicolaskruchten:
I don't think I'm a fan of the opacity changes to scatter... none of our other templates do this and I'd like for there to be at least some measure of consistency about this :)

nicolaskruchten:
I also think the lines around the bars are too strong... can we use the same thickness as in the `plotly` template family please?

joelostblom:
> I don't think I'm a fan of the opacity changes to scatter... none of our other templates do this and I'd like for there to be at least some measure of consistency about this :)

I think it is a useful way to indicate saturated plots, but I also get your points about consistency, removed.

> I also think the lines around the bars are too strong... can we use the same thickness as in the `plotly` template family please?

Hmm, doesn't the plotly templates use 0 linewidth? I changed it to 0.6 in my latest commit, looks like this:

image

nicolaskruchten:
This is looking pretty solid, thanks very much! Once #1875 lands we can rebase it on top of that one and then I can kick the tires a bit more and merge it in :)

Thanks for all the patience and enthusiasm on these PRs and issues over the past few weeks!

nicolaskruchten:
OK, if can rebase onto `master` and force-push, I'll do more QA today and we can merge

joelostblom:
@nicolaskruchten Rebased

nicolaskruchten:
Looking really good!

Some observations from rendering the [Plotly Express gallery notebook](https://github.com/plotly/plotly.py-docs/blob/master/python/plotly-express.md) with `px.defaults.template="simple_white"`:
* Marginal plots in PX now have both axis lines set, which is not quite as nice as your screenshot above
* Ternary plot axis lines are grey rather than black, which doesn't seem strictly required

These don't necessarily have to change but they're worth a think

Other than those two items, this is ready for merging!

joelostblom:
> Looking really good!

Thanks!

>* Marginal plots in PX now have both axis lines set, which is not quite as nice as your screenshot above

Yes, this was a change I suggested in #1863 but it didn't make it into #1875. I created a new PR  #1882.
 
>* Ternary plot axis lines are grey rather than black, which doesn't seem strictly required

Done

nicolaskruchten:
OK, I'm very happy to merge this now and we can figure out https://github.com/plotly/plotly.py/pull/1882 separately (hopefully still for 4.3!)

A big thank you for working through this whole thing with us :)

joelostblom:
Thanks for helping me contribute!

PR #1866, opened 2019-11-04
Docstring touchups

joelostblom:
I noticed these minor issues while working on the other PRs.

### DIFF
--- a/packages/python/plotly/plotly/express/_chart_types.py
+++ b/packages/python/plotly/plotly/express/_chart_types.py
@@ -461,6 +461,7 @@ def box(
     """
     In a box plot, rows of `data_frame` are grouped together into a
     box-and-whisker mark to visualize their distribution.
+
     Each box spans from quartile 1 (Q1) to quartile 3 (Q3). The
     second quartile (Q2) is marked by a line inside the box. By
     default, the whiskers correspond to the box' edges +/- 1.5
--- a/packages/python/plotly/plotly/express/_chart_types.py
+++ b/packages/python/plotly/plotly/express/_chart_types.py
@@ -461,6 +461,11 @@ def box(
     """
     In a box plot, rows of `data_frame` are grouped together into a
     box-and-whisker mark to visualize their distribution.
+    Each box spans from quartile 1 (Q1) to quartile 3 (Q3). The
+    second quartile (Q2) is marked by a line inside the box. By
+    default, the whiskers correspond to the box' edges +/- 1.5
+    times the interquartile range (IQR: Q3-Q1), see "points" for
+    other options.
     """
     return make_figure(
         args=locals(),
--- a/packages/python/plotly/plotly/express/_doc.py
+++ b/packages/python/plotly/plotly/express/_doc.py
@@ -302,17 +302,17 @@ docs = dict(
     ],
     marginal=[
         "str",
-        "One of `'rug'`, `'box'`, `'violin'`, `'histogram'`.",
+        "One of `'rug'`, `'box'`, `'violin'`, or `'histogram'`.",
         "If set, a subplot is drawn alongside the main plot, visualizing the distribution.",
     ],
     marginal_x=[
         "str",
-        "One of `'rug'`, `'box'`, `'violin'`, `'histogram'`.",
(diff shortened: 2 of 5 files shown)


packages/python/plotly/plotly/express/_doc.py
```diff
@@ -444,7 +446,7 @@
 
 
 def make_docstring(fn):
-    tw = TextWrapper(width=79, initial_indent="    ", subsequent_indent="    ")
+    tw = TextWrapper(width=77, initial_indent="    ", subsequent_indent="    ")
```
emmanuelle: not sure I understand this change...

packages/python/plotly/plotly/express/_chart_types.py
```diff
@@ -461,6 +461,11 @@ def box(
     """
     In a box plot, rows of `data_frame` are grouped together into a
     box-and-whisker mark to visualize their distribution.
+    Each box spans from quartile 1 (Q1) to quartile 3 (Q3). The
```
emmanuelle: could you please add a blank line before this sentence? The convention is to have a 1-line sentence to start the docstring, then if needed explanations in another paragraph.

packages/python/plotly/plotly/express/_doc.py
```diff
@@ -302,22 +302,22 @@
...
     marginal_y=[
         "str",
-        "One of `'rug'`, `'box'`, `'violin'`, `'histogram'`.",
+        "One of `'rug'`, `'box'`, `'violin'`, or `'histogram'`.",
         "If set, a vertical subplot is drawn to the right of the main plot, visualizing the y-distribution.",
     ],
     trendline=[
         "str",
-        "One of `'rug'`, `'box'`, `'violin'`, `'histogram'`.",
+        "One of `'ols'` or `'lowess'`.",
```
emmanuelle: oops good catch!

packages/python/plotly/plotly/express/_doc.py
joelostblom: It's to prevent spillover in the help message that pops up when clicking shift + tab in Jupyter notebooks. I thought the width for that was 79 when I reformatted the docstrings previously, but noticed today that it actually is 77.

Before:
image

After:
image

emmanuelle: OK thank you for the explanation! It's weird however because I don't have the problem 
image

emmanuelle: Which version of jupyter are you using? Mine is 
```
$ jupyter --version
jupyter core     : 4.5.0
jupyter-notebook : 6.0.1
qtconsole        : 4.5.5
ipython          : 7.8.0
ipykernel        : 5.1.2
jupyter client   : 5.3.1
jupyter lab      : 1.1.4
nbconvert        : 5.6.0
ipywidgets       : 7.5.1
nbformat         : 4.4.0
traitlets        : 4.3.2

```

emmanuelle: But I'm using a dark theme for jupyter which might tune the configuration. I also tested on binder and there is even more spillover I think, it looks like it does not use wrapping at all? 
image

emmanuelle: This issue seems more complicated than I orginally thought, would you mind reverting here to 79 and opening an issue so that we can take the time to understand what's going on, and merge the rest of this PR?

joelostblom: I think it is the custom dark theme that changes the width of the help box in the first screen shot (it looks like you have even more room on the right). I tried the default Jupyter Light and Dark theme and they both perform the wrapping as I indicated above:

image

On your binder screenshot, it appears as if you are running without PR #1835. I'm on jupyter 4.4.0 and jupyterlab 1.0.5.

joelostblom: Do you prefer if I make a new commit that to revert the change, or can I drop the old one in a rebase and then force push?

emmanuelle: Thanks a lot for helping to document this. OK, so no need to revert the change, I'll merge as it is.

PR #1871, opened 2019-11-05
PX templates should override, not merge into default

nicolaskruchten:
Right now, passing `template={}` to any `px` method gives strange output: it uses the `pio.templates.default` for most things but then uses `px`'s internal defaults for colors. This is pretty weird and annoying, and it means that if a user wants to inline any template information, they *first* have to set `pio.templates.default` to `"none"`, or they have to override the entire default template. Also they can't use `template={}`, they have to use `template="none"`.

I think the behaviour in this PR is better, but I welcome feedback.

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1295,7 +1295,7 @@ def make_figure(args, constructor, trace_patch={}, layout_patch={}):
     # Add traces, layout and frames to figure
     fig.add_traces(frame_list[0]["data"] if len(frame_list) > 0 else [])
     fig.layout.update(layout_patch)
-     if "template" in args and args["template"] is not None:
+    if "template" in args and args["template"] is not None:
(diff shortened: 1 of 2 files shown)


emmanuelle:
With this modification, what would be the recommended way to create a template copying an existing one but changing a few parameters? Defining the new template in `pio.templates` and then passing `template="plotly_dark+my_patch"` to the `px` function? Or another way?
We could document this in https://plot.ly/python/templates/

nicolaskruchten:
Re unit-tests: this is basically what the Percy test does, although you don't get to see the "before" :)

nicolaskruchten:
> Defining the new template in pio.templates and then passing template="plotly_dark+my_patch" to the px function?

Yes, basically. Right now you have to do this to overwrite, and this PR basically flips this around.

nicolaskruchten:
bundled into #1875 with tests

Issue #1869, opened 2019-11-04
go.Table appearance ignores the deafult PX template


joelostblom:
`go.Table` seems to be fixed at using the plotly template, and using PX to change to other templates has no effect. Updating the plotly template even after switching to another PX default template changes the appearance of the table.

I couldn't figure out if the template `go.Table` is using is set at a location that is not overwritten when changing the PX default template, or if the plotly template is hardcoded to always be used within the table function.

I appreciate that `go.Table` is not a PX function, but since it seemingly is reading from a default template somewhere and the PX templates all specify Table colors, it would be useful if the Table appearance would change when the default PX template changes.

Here is an example of what I mean, note the comments.

```python
import plotly.express as px
import plotly.io as pio
import plotly.graph_objects as go

# Plotly template
fig1 = go.Figure(data=[go.Table(header=dict(values=['A Scores', 'B Scores']),
                cells=dict(values=[[100, 90, 80, 90], [95, 85, 75, 95]]))]) 
fig1.show()

# Changing to ggplot2 doesn't anything
px.defaults.template = 'ggplot2'
fig2 = go.Figure(data=[go.Table(header=dict(values=['A Scores', 'B Scores']),
                cells=dict(values=[[100, 90, 80, 90], [95, 85, 75, 95]]))]) 
fig2.show()

# Updating the plotly template shows that go.Table is reading from the 
# plotly template rather than ignoring templates altogether.
pio.templates['plotly'].data.table[0].cells.fill.color = 'orange'
fig3 = go.Figure(data=[go.Table(header=dict(values=['A Scores', 'B Scores']),
                cells=dict(values=[[100, 90, 80, 90], [95, 85, 75, 95]]))]) 
fig3.show()
```

Fig1
![image](https://user-images.githubusercontent.com/4560057/68167009-6c484980-ff64-11e9-9586-b9e989d60178.png)

Fig2
![image](https://user-images.githubusercontent.com/4560057/68167033-7407ee00-ff64-11e9-8b3b-14b4381f7acf.png)

Fig3
![image](https://user-images.githubusercontent.com/4560057/68167044-7b2efc00-ff64-11e9-9655-3f494a066631.png)

nicolaskruchten:
Right, the PX templates apply only to the PX functions.

nicolaskruchten:
[...]

PR #1872, opened 2019-11-05
Use the default template's background color for swatches

joelostblom:
This is a suggestion to use the default template's background color for swatches Instead of always using the plotly template. The potential benefit of this change would be that it is easier to tell which colors sequence go well with the currently used template.

It would be neat if this also checked if there was a PX template set, and I asked a related question about how PX and PIO templates synergize in #1869.

Screenshots:

image
---

image
---

image
---

image
---

image
---

image

### DIFF
--- a/packages/python/plotly/_plotly_utils/colors/_swatches.py
+++ b/packages/python/plotly/_plotly_utils/colors/_swatches.py
@@ -6,6 +6,7 @@ def _swatches(module_names, module_contents):
         template's background.
     """
     import plotly.graph_objs as go
+    import plotly.express as px
 
     sequences = [
         (k, v)
@@ -29,6 +30,7 @@ def _swatches(module_names, module_contents):
             title=module_names,
             barmode="stack",
             barnorm="fraction",
+            template=px.defaults.template,
             bargap=0.5,
             showlegend=False,
             xaxis=dict(range=[-0.02, 1.02], showticklabels=False, showgrid=False),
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
Looks good to me if you can fix the formatting problem in CI?

joelostblom:
I believe this is the same CI error as on the other PRs and unrelated to the changes I have made. The error is

```
error: cannot format /home/circleci/project/packages/python/plotly/plotly/basedatatypes.py: cannot use --safe with this file; failed to parse source file.  AST error message: misplaced type annotation (<unknown>, line 256)
```
I think that [line 256](https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/basedatatypes.py#L256) need to have `collections.` added in front of `OrderedDict`

nicolaskruchten:
If you want to have this check the `px.defaults.template` and use that I'd be fine with it.

joelostblom:
Thanks, added!

I also added the default width, so that the autoadjusting behavior can be overridden if desired.

width 600 samples:
image


image

nicolaskruchten:
re the synergization of px and pio templates... Really the way to do this would be to rewrite this little swatch function on top of `px.bar` somehow :)

nicolaskruchten:
the plot titles are also pretty ugly at the moment... they used to say `px.colors.whatever` but after to move into plotly.py, things got shuffled around

joelostblom:
Rebased and cleaned up the titles 

image

emmanuelle:
Closing since #1886 was merged, thanks @joelostblom !

PR #1874, opened 2019-11-05
Add additional templates to control grid lines

joelostblom:
Together with the xgridoff template these allow full control over which gridlines are visible.

### DIFF
--- a/packages/python/plotly/plotly/io/_templates.py
+++ b/packages/python/plotly/plotly/io/_templates.py
@@ -42,6 +42,7 @@ class TemplatesConfig(object):
             "plotly_dark",
             "presentation",
             "xgridoff",
+            "ygridoff",
             "none",
         ]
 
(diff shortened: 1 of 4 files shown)


emmanuelle:
Adding `ygridoff` is a good idea I think, for symmetry with `xgridoff`. On the other hand I'm not sure we need the `gridon` one since most default templates already display grid lines?

nicolaskruchten:
> most default templates already display grid lines

I think the idea here is that @joelostblom's new template doesn't, so I like this idea personally.

joelostblom:
True, I believe it's only the template I proposed that doesn't have grid lines by default. I think it would be nice with `gridon` since the three together then covers all possible scenarios (adding or removing one or two grid lines) which makes it feel like a complete modular set, but I can remove it if you think there are too many templates.

emmanuelle:
Sounds good, I'm convinced now!

emmanuelle:
@joelostblom if you can find the time, could you please document these templates (at least the new one `simple_white`) in https://plot.ly/python/templates/? (the source is https://github.com/plotly/plotly.py-docs/blob/master/python/templates.md). You could just add the new template to the figure showing how a chart looks like with the different templates.

joelostblom:
@emmanuelle Done! https://github.com/plotly/plotly.py-docs/pull/169

emmanuelle:
great, merging

PR #1882, opened 2019-11-07
Hide ticks and axis line for the marginal value axes

joelostblom:
**Current behavior**
image

**Proposed behavior**
image


nicolaskruchten:
@emmanuelle thoughts?

I like this better, personally

nicolaskruchten:
Looks like https://seaborn.pydata.org/examples/hexbin_marginals.html

nicolaskruchten:
OK let's do it :)

PR #1884, opened 2019-11-08
Update contribution guidelines

joelostblom:
I was setting up a plotly dev env again and documented a few things I did differently from the current guidelines. The only thing that I didn't include was that I had issues installing the optional dependencies via `pip`. This might be a local problem, but fwiw I installed these packages via conda instead:

```
conda create -n plotly-dev pandas seaborn jupyterlab scipy simplegeneric
conda activate plotly-dev
conda install -c plotly plotly-orca psutil requests
```

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -28,9 +28,16 @@ cd plotly.py
 ```
 
 ### Create a virtual environment for plotly development
-TODO: Use virtualenv or conda, activate it
 
-http://docs.python-guide.org/en/latest/dev/virtualenvs/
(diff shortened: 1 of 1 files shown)


emmanuelle:
Thanks @joelostblom. I think most people take a look at this info on github, so is it OK if we review/merge after 4.3?

jonmmease:
Thanks @joelostblom!  I'm going to merge this and then add a follow-up commit removing the submodules section now that these are all inlined (as of https://github.com/plotly/plotly.py/pull/1942)

PR #1885, opened 2019-11-08
Move imshow closer to px pattern

nicolaskruchten:
@emmanuelle here's the first pass on the PX-ification of the `imshow` function. I'm realizing that to align better with the other `px` methods we should probably add `height`, `width` and `template` and the rest of the `px` defaulting machinery...

### DIFF
--- a/packages/python/plotly/plotly/express/_imshow.py
+++ b/packages/python/plotly/plotly/express/_imshow.py
@@ -1,5 +1,6 @@
 import plotly.graph_objs as go
 from _plotly_utils.basevalidators import ColorscaleValidator
+from ._core import apply_default_cascade
 import numpy as np  # is it fine to depend on np here?
 
 _float_types = []
@@ -63,6 +64,10 @@ def imshow(
     color_continuous_scale=None,
     color_continuous_midpoint=None,
     range_color=None,
+    title=None,
+    template=None,
+    width=None,
+    height=None,
 ):
     """
     Display an image, i.e. data on a 2D regular raster.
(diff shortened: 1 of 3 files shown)


nicolaskruchten:
Here I've bolted on the applicable parts of the "defaults cascade" logic for which I added a test suite the other day :)

nicolaskruchten:
This lets us do stuff like this:

image

packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -74,16 +88,38 @@ def imshow(img, zmin=None, zmax=None, origin=None, colorscale=None):
...
         with a 5% tolerance. For a single-channel image, the max of the image is used.
 
     origin : str, 'upper' or 'lower' (default 'upper')
         position of the [0, 0] pixel of the image array, in the upper left or lower left
         corner. The convention 'upper' is typically used for matrices and images.
 
-    colorscale : str
-        colormap used to map scalar data to colors (for a 2D image). This parameter is not used for
-        RGB or RGBA images.
+    color_continuous_scale : str or list of str
```
emmanuelle: `str (name of colorscale) or list of str (list of css colors)` would be clearer here.

emmanuelle: ok it's explained below, it's fine

emmanuelle:
:dancer: when the CI is green. Which reminds me that we should add imshow examples to the percy test file, one with single-channel and one with RGB. If you don't want to add it here (I should have done this in the initial PR) I'll make another PR for this.

nicolaskruchten:
Ah right yes... If you could do it that would be great :)

nicolaskruchten:
Ah drat also we need to tweak the PR in py-docs

PR #1886, opened 2019-11-09
apply defaults cascade to swatches()

nicolaskruchten:


nicolaskruchten:
Supersedes #1872 except for the title ... @joelostblom what do you think of this? It now respects `pio.templates.default` and `px.default.templates` and lets you override inline with a `template` kwarg. These swatches basically can't be made as `px.bar` figures so this is probably the next best thing

nicolaskruchten:
And I've fixed the title.

packages/python/plotly/_plotly_utils/colors/_swatches.py
```diff
@@ -1,10 +1,23 @@
-def _swatches(module_names, module_contents):
+def _swatches(module_names, module_contents, template=None):
     """
-    Returns:
+    Parameters
+    ----------
+
```
joelostblom:
```suggestion
```

packages/python/plotly/_plotly_utils/colors/_swatches.py
```diff
@@ -1,10 +1,23 @@
...
+    template : str or dict or plotly.graph_objects.layout.Template instance
+        The figure template name or definition.
+
+    Returns
+    -------
+
```
joelostblom:
```suggestion
```

packages/python/plotly/_plotly_utils/colors/_swatches.py
```diff
@@ -25,13 +38,13 @@ def _swatches(module_names, module_contents):
...
-            title=module_names,
+            title="plotly.colors." + module_names.split(".")[-1],
             barmode="stack",
             barnorm="fraction",
-            template="plotly",
             bargap=0.5,
             showlegend=False,
             xaxis=dict(range=[-0.02, 1.02], showticklabels=False, showgrid=False),
             height=max(600, 40 * len(sequences)),
+            template=args["template"],
```
joelostblom: Could we somehow add control of the width and maybe also height parameter? I had this in the previous PR:

```
height=max(px.defaults.height, 40 * len(sequences)),
width=px.defaults.width,
```

packages/python/plotly/_plotly_utils/colors/_swatches.py
joelostblom: To make it consistent with how the other px docstrings are written

packages/python/plotly/_plotly_utils/colors/_swatches.py
nicolaskruchten: We could, but I'm not sure there's such a strong use-case for this... The template thing was quite important to be able to see things against the proper background etc :)

nicolaskruchten: Let's do it in another PR?

emmanuelle:
Look good to me, thanks! :dancer:

PR #1888, opened 2019-11-11
Remove "paper" as default reference when adding annotations/shapes/images

### LINKED ISSUE #1881: xref/yref in annotation/shape/image without row/col
When row/col are not provided, we shouldn't write to those keys, to let the JS defaults shine through.

jonmmease:
This PR updates `add_annotations`/`add_shapes`/`add_images` so that if no `row`/`col` are specified, the `xref`/`yref` properties are left unset rather than being set to `"paper"`.  Now, to use `add_annotations` to add an annotation in "paper" coordinates you leave `row`/`col` unset and simply pass `xref="paper" and yref="paper` kwargs.

Similarly, `select_annotations` no longer accepts `row="paper"`/`col="paper`. Instead, to select only the annotations in paper coordinates, you would use `xref="paper"`, `yref="paper"` in the `selector`.

@nicolaskruchten 

Closes #1881

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/ambv/black
-    rev: stable
+    rev: 19.10b0
     hooks:
     - id: black
       language_version: python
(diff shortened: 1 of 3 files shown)


nicolaskruchten:
@emmanuelle this resolves this issues you raised right? 💃 from me if so!

emmanuelle:
Yes, looks good ! :dancer: I might have to update the doc tutorial.

Issue #1873, opened 2019-11-05
Effect on JupyterLab bundle size


vidartf:
Hi!

I've been looking at the effect of the plotly jupyterlab extensions on the final bundle size, and I've made these observations:
- The plotly.js part for a prod build is roughly 17 MB.
- Both `plotly.js/dist/plotly.js` and `plotly.js/dist/plotly.min.js` gets included in the final lab bundle (one from plotlywidget and one from jupyterlab-plotly). Both end up taking roughly equal amount of space (6-7 MB) when the lab build is run in prod mode.
- Also some parts of the `plotly.js/src` dir gets included (from plotlywidget).

Would it be possible to consolidate the imports in the jupyterlab packages such that only one of the dists gets imported from (and if possible also the `src` import)?

nicolaskruchten:
Yikes, we should be using the `min` version everywhere for sure. The duplication is a bummer. 

@jonmmease could we imagine merging these extensions someday?

jonmmease:
> could we imagine merging these extensions someday?

Yes, and we almost did for v4.  But the distinct plotlywidget package is needed in order for nbviewer to find the extension, and I didn't spend enough time on it to figure out how to combine things in such a way that wouldn't break something :slightly_smiling_face:

nicolaskruchten:
Ah right. Does JLab have a way for extensions share dependencies yet that we know of?

vidartf:
The packages doesn't need to be merged to resolve this, they just need to import from the same dist. If the dependency versions are compatible, jupyterlab will de-duplicate them.

vidartf:
Note: even now the dependency *package* is already duplicated, but the *code* is duplicated in the final bundle since it is imported from two different dists.

nicolaskruchten:
Ah nice. So if we changed https://github.com/plotly/plotly.py/blob/master/packages/javascript/plotlywidget/src/Figure.js#L5 to load the unminified bundle and not from `src` then in principle webpack should dedupe and minify all of Plotly. We'll try! Probably for 4.4 :)

vidartf:
Sounds like a plan. Please ping this issue with any progress, and I'll be happy to verify if it works :)

nicolaskruchten:
[...]

PR #1891, opened 2019-11-11
load plotly.js from unminified dist bundle in plotlywidget

jonmmease:
With this PR, both `jupyterlab-plotly` and `plotlywidget` now load plotly.js from the unminified dist bundle, so jupyterlab should be able to de-duplicate plotly.js when both extensions are installed.

Closes https://github.com/plotly/plotly.py/issues/1873

cc @vidartf

### DIFF
--- a/packages/javascript/plotlywidget/package-lock.json
+++ b/packages/javascript/plotlywidget/package-lock.json
@@ -9,9 +9,9 @@
       "resolved": "https://registry.npmjs.org/3d-view/-/3d-view-2.0.0.tgz",
       "integrity": "sha1-gxrpQtdQjFCAHj4G+v4ejFdOF74=",
       "requires": {
-        "matrix-camera-controller": "2.1.3",
-        "orbit-camera-controller": "4.0.0",
-        "turntable-camera-controller": "3.0.1"
+        "matrix-camera-controller": "^2.1.1",
+        "orbit-camera-controller": "^4.0.0",
+        "turntable-camera-controller": "^3.0.0"
       }
     },
     "@jupyter-widgets/base": {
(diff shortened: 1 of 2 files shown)


emmanuelle:
So the `index.js` file is now larger (7 Mo instead of 4). Is this a problem (for other environments, eg Dash applications etc.). Probably a naive question :-)

jonmmease:
@nicolaskruchten any thoughts on this. We could use the `min` bundle in both extensions instead of the regular one. My thinking was that this should be getting compressed everywhere it matters, but I don't have a strong opinion here.

nrbgt:
Perhaps when making this change, please also consider registering the exports in the lab plugin with a [lazy load](https://webpack.js.org/guides/code-splitting/#dynamic-imports): this will keep it from 
- hitting the browser until it's actually requested
- getting compiled in with the already-titanic `vendor~main`

Here's how it interacts with the crucial `loadClass`: https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/jupyterlab-manager/src/manager.ts#L383

Since it appears to be commonjs, you may have to use the vendor-specific `webpack.ensure` (or `require.ensure`?) rather than the standards-based `await import`. But doing it at the widget load level is far more convenient than trying to do it within the backbone-related code, which doesn't async much.

nrbgt:
Here's an example with `jupyter-threejs`:

https://github.com/jupyter-widgets/pythreejs/pull/297

vidartf:
@nrbgt While I agree with your point, I believe it should be made as a separate issue.

nrbgt:
@vidartf made issue #1913

nicolaskruchten:
OK so as I understand it right now, this will bloat the nbextension size because we're not minifying in webpack.config.js.

Issue #1870, opened 2019-11-05
Problems opting out of the default template


nicolaskruchten:
Original observation: neither call here suppresses the default template:

```python
from plotly.offline import iplot
iplot(dict(data=[dict(type="scatter")], layout=dict(template="none")))
iplot(dict(data=[dict(type="scatter")], layout=dict(template={})))
```

Popping the hood on that method, it looks like it's doing the equivalent of:

```python
import plotly.graph_objects as go
fig = go.Figure(dict(data=[dict(type="scatter")], layout=dict(template={})))

import plotly.io as pio
pio.show(fig.to_dict())
```

The problem being that `fig.to_dict()` here returns `{'data': [{'type': 'scatter'}], 'layout': {}}` and when `pio.show()` gets it, it seems to applies the default template. This doesn't happen with `pio.show(fig.to_dict(), validate=False)` or with `pio.show(fig.update_layout(template={}, overwrite=True).to_dict(), validate=False)`.

So should `pio.show()` be applying the default template at all here? (Hard to say "no" here unless we do validation somehow without just creating a `Figure` ?)

And/or should `fig.to_dict()` include the `template: {}` to signal to functions like `pio.show()` that no defaulting should happen?

nicolaskruchten:
@jonmmease would love your thoughts on this!

nicolaskruchten:
> And/or should fig.to_dict() include the template: {} to signal to functions like pio.show() that no defaulting should happen?

let's do this, but for 4.4

PR #1892, opened 2019-11-11
Propagate empty templates

jonmmease:
When a template is explicitly set to `{}`, it is now replaced by `{"data": {"scatter": [{}]}}` so that it will not get overridden by a default template when a figure is passed along to other functions/libraries.

cc @nicolaskruchten

### DIFF
--- a/packages/python/plotly/_plotly_utils/basevalidators.py
+++ b/packages/python/plotly/_plotly_utils/basevalidators.py
@@ -2712,6 +2712,13 @@ class BaseTemplateValidator(CompoundValidator):
             # v is un-hashable
             pass
 
+        # Check for empty template
+        if v == {} or isinstance(v, self.data_class) and v.to_plotly_json() == {}:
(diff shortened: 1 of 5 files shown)


nicolaskruchten:
makes sense to me I think!

emmanuelle:
Would it be possible to add a test which would have failed before this change? Like checking for a potential override of the template? Otherwise, :dancer: (we can also add the test later -- if it makes sense to you)

jonmmease:
> Would it be possible to add a test which would have failed before this change? Like checking for a potential override of the template?

@nicolaskruchten I don't remember exactly which use-cases you were having trouble with, are there any particular tests that you'd like to have in here?

nicolaskruchten:
The original issue was https://github.com/plotly/plotly.py/issues/1870 so it was with `plotly.io.show()` ... hard to test.

jonmmease:
Ok, I'll go ahead and merge then.

PR #1893, opened 2019-11-11
Template bis

emmanuelle:


jonmmease:
Looks good to me.  The changes in the percy tests all look like improvements. And the chart studio test failure is unrelated so :dancer: from me.

emmanuelle:
@jonmmease please bear with me I just pushed another commit with also pie automargin. Waiting for percy to run... If eveything is well I should be done

jonmmease:
great. looks like the chart studio tests are passing now. Just a black formatting issue left

nicolaskruchten:
This change moves the titles a bit closer to the tick labels by default, which I’m ok with. Maybe in some near future we can have the built in templates reduce the margins a bit too :)

emmanuelle:
thanks Jon!

PR #1896, opened 2019-11-12
version update in README

emmanuelle:
@jonmmease I think this one needed to be updated as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
 
 ## Quickstart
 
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
This is odd, I think I did this already... confused about why this PR is in this state.

emmanuelle:
Maybe because I pushed the PR just before you pushed your change (directly to master). But I would have expected a conflict indeed... Anyway, closing since it's fixed.

PR #1905, opened 2019-11-15
Use main readme as long_description for PyPi

andriyor:
https://pypi.org/project/plotly/

### DIFF
--- a/packages/python/plotly/setup.py
+++ b/packages/python/plotly/setup.py
@@ -37,7 +37,7 @@ def plotly_js_version():
 
 
 def readme():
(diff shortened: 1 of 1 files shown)


emmanuelle:
Thank you @andriyor . It seems indeed that we do not use the right `README` on pypi (@nicolaskruchten I think you were wondering about this). However, it seems that there is a problem with paths, since our CI is failing with a `FileNotFoundError`. I don't know if you have the time to investigate this @andriyor otherwise we'll take a look.

nicolaskruchten:
Done differently in #1949 ... thanks for the push though!

PR #1909, opened 2019-11-18
adding pie, sunburst, funnel and funnelarea to px

emmanuelle:


packages/python/plotly/plotly/express/_core.py
```diff
@@ -305,11 +305,38 @@ def make_trace_kwargs(args, trace_spec, g, mapping_labels, sizeref):
             elif k == "locations":
                 result[k] = g[v]
                 mapping_labels[v_label] = "%{location}"
+            elif k == "values":
+                result[k] = g[v]
+                _label = "value" if v_label == "values" else v_label
+                mapping_labels[_label] = "%{value}"
+            elif k == "parents":
```
emmanuelle: if we want to remove parents and ids from the hover template we can do it here but we could also keep them

packages/python/plotly/plotly/express/_doc.py
```diff
@@ -442,21 +462,41 @@
...
+        "int",
+        "Positive integer",
+        "Sets the number of rendered sectors from any given `level`. Set `maxdepth` to -1 to render all the"
+        "levels in the hierarchy.",
+    ],
 )
 
 
-def make_docstring(fn):
+def make_docstring(fn, override_dict={}):
```
nicolaskruchten: 👍

packages/python/plotly/plotly/express/_chart_types.py
```diff
@@ -1115,3 +1115,172 @@ def parallel_categories(
...
+    labels={},
+    title=None,
+    template=None,
+    width=None,
+    height=None,
+):
+    """
+    In a funnel area plot, each row of `data_frame` is represented as a trapezoidal sector of a funnel.
+    """
+    return make_figure(args=locals(), constructor=go.Funnelarea,)
```
nicolaskruchten: here we should force `showlegend=True` just like `pie`

packages/python/plotly/plotly/express/_chart_types.py
```diff
@@ -1115,3 +1115,172 @@ def parallel_categories(
...
+        trace_patch=dict(branchvalues=branchvalues, maxdepth=maxdepth),
+    )
+
+
+treemap.__doc__ = make_docstring(treemap)
+
+
+def funnel(
+    data_frame=None,
+    x=None,
```
nicolaskruchten: can we add all the other applicable kwargs from `bar` like `facet_*` and `color_continuous_sequence` etc?

packages/python/plotly/plotly/express/_chart_types.py
```diff
@@ -1115,3 +1115,172 @@ def parallel_categories(
...
+        args=locals(),
+        constructor=go.Funnel,
+        trace_patch=dict(opacity=opacity, orientation=orientation, textinfo=textinfo),
+    )
+
+
+funnel.__doc__ = make_docstring(
+    funnel,
+    override_dict=dict(
+        textinfo=[
```
nicolaskruchten: I hesitate about exposing `textinfo` in PX... `texttemplate` is basically the new and improved way to do this, and I think if we're going to expose one or the other, it should be `texttemplate` and it should be probably be added everywhere that text is supported. My feeling is that we should leave out `textinfo` for now and do the broader change in a separate PR.

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: This makes sense to me ATM.

nicolaskruchten:
This is all pretty good, modulo inline comments.

I'd really like to have support for `color` in pie/funnel/treemap/sunburst the day we release this though... I feel like adding continuous-color support for treemap/sunburst should be pretty trivial (i.e. it should "just work") and then discrete-color support will be harder because for these 4 traces it doesn't work at all the way it does in the other traces: it's not "groupable". But we should try to do it anyway :)

nicolaskruchten:
> I'd really like to have support for color in pie/funnel/treemap/sunburst the day we release this

This would basically be the "value added" of PX for these traces IMO, as compared to using the graph objects directly, so that's why I think it's important :)

nicolaskruchten:
The way I would implement discrete color is basically in  `make_trace_kwargs` with another gross type-specific exception inside the `elif k == "color":` block, which would just hardcode the marker colors.

nicolaskruchten:
Extra requirement: if `color_discrete_sequence` is passed into pie/funnelarea/treemap/sunburst, we should probably assign it to `layout.<whatever>colorway`, so that something like `px.pie(values=[1,2,3], color_discrete_sequence=["red", "green", "blue"])` will do the expected thing.

packages/python/plotly/plotly/express/_chart_types.py
emmanuelle: addressed

packages/python/plotly/plotly/express/_chart_types.py
emmanuelle: done

packages/python/plotly/plotly/express/_chart_types.py
emmanuelle: ok, so I did not put anything and we'll add texttemplate whereever it's relevant in another PR

emmanuelle:
Hum, interesting to see that the test fail because the color of the funnel is not the same.

emmanuelle:
Do we want to accept `color_discrete_sequence='Viridis'`? At the moment it only works with a `px` object.

nicolaskruchten:
it's normal that `color_discrete_sequence` only works with an array of colors, yes. It's not clear what we would do if we got `Viridis`, as that scale doesn't have a fixed number of categorical colors...?

emmanuelle:
ok good, so nothing more to do for this :-).

nicolaskruchten:
@emmanuelle some tweaks in that last commit, LMK what you think.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -304,26 +291,28 @@ def make_trace_kwargs(args, trace_spec, g, mapping_labels, sizeref):
...
-                        g[v], args["color_discrete_sequence"]
-                    )
+                        result["marker"]["colors"] = []
+                        mapping = {}
+                        for cat in g[v]:
+                            if mapping.get(cat) is None:
+                                mapping[cat] = args["color_discrete_sequence"][
+                                    len(mapping) % len(args["color_discrete_sequence"])
+                                ]
+                            result["marker"]["colors"].append(mapping[cat])
```
emmanuelle: Oh interesting, I did not know that `marker_colors` could be a dictionary, I thought it had to be a list/array.

emmanuelle:
Thanks, look good to me. Should we give each other a :dancer: ? :-)

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: I think it does have to be a list, and in this case it is...?

nicolaskruchten: this is the same code you had, just inlined :)

emmanuelle: oh yes I had read too fast!

PR #1915, opened 2019-11-21
Fix validate_executable behavior

### LINKED ISSUE #1914: Clarify work with external server
I started orca as shown here: https://github.com/plotly/orca#method-3-docker
So I got it on 9091 port on localhost accessible

I've installed plotly==4.3.0 with latest "feature" from #1850 

Now, when running following code, I get two terminal screens of errors and cool stories about how to install that orca executable, which I actually should not have to keep in my host system when using orca *server*:
```
import plotly.io as pio
pio.orca.config.server_url = "localhost"
pio.orca.config.port = 9091
pio.to_image(go.Figure())
```

Could you _please_ clarify, how to use that new feature? since for now it is just useless without binary in PATH.

dev-dsp:
Validate orca executable only if server_url is not provided

Fixes #1914

### DIFF
--- a/packages/python/plotly/plotly/io/_orca.py
+++ b/packages/python/plotly/plotly/io/_orca.py
@@ -1384,11 +1384,10 @@ Install using conda:
 """
         )
 
-    # Validate orca executable
-    if status.state == "unvalidated":
-        validate_executable()
-
     if not config.server_url:
+        # Validate orca executable only if server_url is not provided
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
@jonmmease this seems fine to me, do you agree?

jonmmease:
@dev-dsp, thanks!

Issue #1920, opened 2019-11-22
pio.to_templated not behaving as expected


emmanuelle:
```
>>> fig = go.Figure(layout={'title': 'Figure Title', 'font': {'size': 20, 'family': 'Courier'}})                                             
>>> fig                                                                                                                                      
Figure({
    'data': [],
    'layout': {'font': {'family': 'Courier', 'size': 20}, 'template': '...', 'title': {'text': 'Figure Title'}}
})
>>> pio.to_templated(fig)                                                                                                                    
Figure({
    'data': [], 'layout': {'template': '...', 'title': {'text': 'Figure Title'}}
})
```

The style properties should go to the template I think?

nicolaskruchten:
This is working as intended, it's just that `template` is suppressed in the output AND your figure also has the default template on board... Here's another way of looking at it:

![image](https://user-images.githubusercontent.com/203523/69460150-6498eb00-0d41-11ea-94f5-8e88cc3c03aa.png)

nicolaskruchten:
Also we should document `pio.to_templated()` in https://plot.ly/python/templates/ :)

emmanuelle:
Got it, thanks. I got confused by the fact that the template is suppressed in the output (which is a good thing!). I'll let this one open until we document the feature.

gvwilson:
Hi - we are currently trying to tidy up Plotly's public repositories to help us focus our efforts on things that will help users most. Since this issue has been sitting for several years, I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our backlog. Thanks for your help - @gvwilson

Issue #1919, opened 2019-11-22
[BUG] create_facet figure factory when color_names are numbers


emmanuelle:
Was caught by this docstring example
```
   >>> import plotly.figure_factory as ff
    >>> import pandas as pd
    >>> tips = pd.read_csv('https://raw.githubusercontent.com/plotly/datasets/master/tips.csv')

    >>> fig = ff.create_facet_grid(
    ...     tips,
    ...     x='total_bill',
    ...     y='tip',
    ...     facet_row='sex',
    ...     facet_col='smoker',
    ...     color_name='size',
    ...     colormap='Viridis',
    ... )
    >>> fig.show()

```

The error is 
```
Traceback (most recent call last):
  File "<ipython-input-45-2e217ba09405>", line 10, in <module>
    color_is_cat=True,
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/figure_factory/_facet_grid.py", line 952, in create_facet_grid
    kwargs_marker,
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/figure_factory/_facet_grid.py", line 257, in _facet_grid_color_categorical
    trace, j + 1 if facet_row else 1, 1 if facet_row else j + 1
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/basedatatypes.py", line 1789, in append_trace
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/basedatatypes.py", line 1615, in add_trace
    cols=[col] if col is not None else None,
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/basedatatypes.py", line 1676, in add_traces
    # Validate traces
  File "/home/emma/code/plotly.py/packages/python/plotly/_plotly_utils/basevalidators.py", line 2645, in validate_coerce
    skip_invalid=skip_invalid, **v_copy
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/graph_objs/__init__.py", line 37591, in __init__
    self["name"] = name if name is not None else _v
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/basedatatypes.py", line 3479, in __setitem__
    else:
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/basedatatypes.py", line 3766, in _set_prop
    else:
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/basedatatypes.py", line 3761, in _set_prop
    try:
[...]

PR #1921, opened 2019-11-22
first round of changes towards doctesting

emmanuelle:
Not a full doctest workflow yet, but some required changes.

### DIFF
--- a/packages/python/plotly/plotly/basedatatypes.py
+++ b/packages/python/plotly/plotly/basedatatypes.py
@@ -458,7 +458,7 @@ class BaseFigure(object):
         --------
         >>> import plotly.graph_objs as go
         >>> fig = go.Figure(data=[{'y': [1, 2, 3]}])
-        >>> fig.update(data=[{'y': [4, 5, 6]}]) #doctest: +ELLIPSIS
+        >>> fig.update(data=[{'y': [4, 5, 6]}]) # doctest: +ELLIPSIS
         Figure(...)
         >>> fig.to_plotly_json() # doctest: +SKIP
             {'data': [{'type': 'scatter',
@@ -469,7 +469,7 @@ class BaseFigure(object):
         >>> fig = go.Figure(layout={'xaxis':
         ...                         {'color': 'green',
         ...                          'range': [0, 1]}})
-        >>> fig.update({'layout': {'xaxis': {'color': 'pink'}}}) #doctest: +ELLIPSIS
+        >>> fig.update({'layout': {'xaxis': {'color': 'pink'}}}) # doctest: +ELLIPSIS
         Figure(...)
         >>> fig.to_plotly_json() # doctest: +SKIP
             {'data': [],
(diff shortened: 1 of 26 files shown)


emmanuelle:
Also see #1919 and #1920.

packages/python/plotly/plotly/figure_factory/_ohlc.py
```diff
@@ -161,91 +161,9 @@ def create_ohlc(open, high, low, close, dates=None, direction="both", **kwargs):
     >>> from plotly.figure_factory import create_ohlc
     >>> from datetime import datetime
 
-    >>> import pandas.io.data as web
```
emmanuelle: here I removed all of them because the doc was broken and we have the trace anyway

packages/python/plotly/plotly/basedatatypes.py
```diff
@@ -468,8 +469,9 @@ def update(self, dict1=None, overwrite=False, **kwargs):
         >>> fig = go.Figure(layout={'xaxis':
         ...                         {'color': 'green',
         ...                          'range': [0, 1]}})
-        >>> fig.update({'layout': {'xaxis': {'color': 'pink'}}})
-        >>> fig.to_plotly_json()
+        >>> fig.update({'layout': {'xaxis': {'color': 'pink'}}}) #doctest: +ELLIPSIS
```
nicolaskruchten: another approach would be `fig = fig.update( ... )` ?

emmanuelle:
Note that sphinx removes the doctest flags
image

nicolaskruchten:
Cool :)

nicolaskruchten:
All seems pretty good to me! I notice that Sphinx grabs the first sentence or line of docstrings to make these "TOC" type pages: https://plot.ly/python-api-reference/plotly.figure_factory.html so we'll want to take a look at this page with these new docstrings to make sure we're happy with it.

Maybe the deprecated ones all start with `**deprecated**` and show up at the end or something?

emmanuelle:
So here is how the table of contents looks like
image
For figure factories replaced by a trace I just wrote **deprecated** at the start. For some other functions it's not a 1-1 correspondence and there is an example in the docstring, so I did not want to put off people from clicking on the link by having **deprecated** at the very start.

nicolaskruchten:
Looks good! Can we order them such that the non-deprecated ones are up top?

emmanuelle:
Sure!

emmanuelle:
Yehaah, I think this one is ready for review @nicolaskruchten !

nicolaskruchten:
Nice!

So (how) does this test the docstrings for e.g. `go.Scatter` ? this runs after codegen?

nicolaskruchten:
Ah, haha, there are no code examples in `go.Scatter` :)

emmanuelle:
Exactly :-). Actually we don't have a lot of docstrings with code examples. You can see a list of the files with docstrings doctested in the corresponding CI job (python3.7-orca)

packages/python/plotly/plotly/figure_factory/_table.py
```diff
@@ -36,7 +36,8 @@ def create_table(
     **kwargs
 ):
     """
-    BETA function that creates data tables
+    **deprecated**, use instead the plotly.graph_objects trace
```
nicolaskruchten: I actually realize now that this isn't an older way of making tables, but it's a convenience method for making tables, because `go.Table` is pretty hard to use. So maybe this one shouldn't be `deprecated` after all.

nicolaskruchten:
💃 modulo table comment. Thanks for this cleanup and lock-down!

PR #1932, opened 2019-11-28
dataframe parameter in px docstrings

### LINKED ISSUE #1926: PX docstrings and `data_frame`
Right now all the PX doctrings talk about "rows of `data_frame`" but this is a bit out of date with respect to input formats. We should update it somehow, perhaps with a note that `data_frame`, while technically optional, gets constructed under the hood if you pass in array-like objects.

emmanuelle:
Closes #1926 . Is this enough?

### DIFF
--- a/packages/python/plotly/plotly/express/_doc.py
+++ b/packages/python/plotly/plotly/express/_doc.py
@@ -22,6 +22,7 @@ docs = dict(
         "DataFrame or array-like or dict",
         "This argument needs to be passed for column names (and not keyword names) to be used.",
         "Array-like and dict are tranformed internally to a pandas DataFrame.",
+        "Optional, since a DataFrame gets constructed under the hood using the other arguments."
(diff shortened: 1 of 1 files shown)


packages/python/plotly/plotly/express/_doc.py
```diff
@@ -22,6 +22,7 @@
         "DataFrame or array-like or dict",
         "This argument needs to be passed for column names (and not keyword names) to be used.",
         "Array-like and dict are tranformed internally to a pandas DataFrame.",
+        "Optional, since a DataFrame gets constructed under the hood using the other arguments."
```
nicolaskruchten:
```suggestion
        "Optional: if missing, a DataFrame gets constructed under the hood using the other arguments."
```

nicolaskruchten:
Yup, modulo small fix! 💃

PR #1933, opened 2019-11-29
Reversed colorscale

### LINKED ISSUE #1681: add support for the _r convention to reverse colorscales
...the way Matplotlib does it.

emmanuelle:
Closes #1681


nicolaskruchten:
To record my verbal comments from earlier: this looks right from the point of view of `...colorscale='Viridis_r'` but breaks the current symmetry we have between  `...colorscale='Viridis'...` and `...colorscale=plotly.colors.sequential.Viridis...`. It would be great if all the colorscales in `plotly.colors` also had a `_r` variant. 

The tricky bit is that this would include all the colorscales in `plotly.colors.sequential/diverging/cyclical` *and* all those same scales as they appear in other collections like `plotly.colors.carto` etc (if possible!)

emmanuelle:
I need to add a couple of tests as well, but let's see first what percy thinks about this (I did not include the reversed colorscales in the swatches, so there should be no change).

packages/python/plotly/_plotly_utils/colors/_swatches.py
```diff
@@ -33,7 +33,7 @@ def _swatches(module_names, module_contents, template=None):
                 marker=dict(color=colors),
                 hovertemplate="%{y}[%{customdata}] = %{marker.color}<extra></extra>",
             )
-            for name, colors in reversed(sequences)
+            for name, colors in (sequences)
```
nicolaskruchten: i actually kind of liked the old order, otherwise they read bottom up...

nicolaskruchten:
looks great to me! could you revert the codegen'ed piece of this plz though leaving just the changes to the core lib in color an util? 💃

packages/python/plotly/_plotly_utils/colors/_swatches.py
emmanuelle: oh good catch, I had modified this while debugging and then forgot to change it back.

emmanuelle:
Not sure I understand your comments about the codegened piece: a lot of files are modified because docstrings are modified. Shouldn't we keep these modifications? Or are they going to be rolled out by something else? (the release etc.). Or should I just revert the `package.json-lock` changes? Please bear with me :-).

packages/python/plotly/plotly/graph_objs/_figurewidget.py
```diff
@@ -15147,7 +15147,7 @@ def add_violin(
             or greater than 4*Q3-3*Q1 are highlighted (see
             `outliercolor`) If "all", all sample points are shown
             If False, only the violins are shown with no sample
-            points and the whiskers extend to the range of the sample.
+            points
```
jonmmease: @emmanuelle, I don't think this docstring change should be coming in this PR.  Maybe try rerunning codegen in case you had a different schema around when you generated this.

packages/javascript/plotlywidget/package-lock.json
```diff
@@ -9,26 +9,26 @@
       "resolved": "https://registry.npmjs.org/3d-view/-/3d-view-2.0.0.tgz",
```
jonmmease: It shouldn't be necessary for this file to have changed unless you're changing the bundled version of plotly.js.  Could you revert it?

packages/python/plotly/plotlywidget/static/index.js
```diff
@@ -10495,27 +10495,27 @@ module.exports = function(opts) {
 /* 10 */
```
jonmmease: Revert these changes as well

packages/javascript/plotlywidget/package-lock.json
emmanuelle: This one will go away when rebasing on master. I don't think I changed the bundled version of plotly.js but I ran the codegen to update the docstrings.

packages/python/plotly/plotlywidget/static/index.js
emmanuelle: ok

packages/python/plotly/plotly/graph_objs/_figurewidget.py
emmanuelle: Actually #1988 already removed those. I think they were not added at the right place (I think it was a community PR, I must have merged it without thinking that this should have been in the codegen instead.)

emmanuelle: What do you suggest to do here @jonmmease ?

jonmmease: It looks like this code was somehow generated from the schema for plotly.js 1.51.2 (which this branch shouldn't have yet).  In any case, why don't you try merging in master and then running codegen again.  Hopefully then the diff will only show the colorscale docstring changes.

packages/javascript/plotlywidget/package-lock.json
jonmmease: When you run codegen, are you using `python setup.py codegen`?  This shouldn't trigger any JavaScript/npm logic that would alter this file.  When you merge in master, go ahead and accept masters version of this file as-is.

emmanuelle:
OK, I ended up with a big git mess. I'll work on it later this evening, hope it won't be too late

jonmmease:
No worries, I'll do the release tomorrow morning. Thanks for working through it!

emmanuelle:
Note there yet but making progress :-)

emmanuelle:
Hey @jonmmease I think this one is ready now. Thank you for your patience!

emmanuelle:
Hum, not quite yet, CI not happy.

jonmmease:
Thanks @emmanuelle! I made one small testing change in https://github.com/plotly/plotly.py/pull/1933/commits/0c555f6c88f57b874562cbad6c2b250d140a0e55 to allow all of the `*_r` colorscales to be included in the parametrized tests.

PR #1934, opened 2019-11-30
fix a PX input bug when using data frame indices

nicolaskruchten:


nicolaskruchten:
@jonmmease this is an annoying-enough bug that I'd like to do a 4.3.1 release this coming week if we can :)

emmanuelle:
Thanks for noticing this bug and fixing it right away! :dancer:

PR #1936, opened 2019-12-01
Doc migration

emmanuelle:


.circleci/config.yml
```diff
@@ -385,6 +385,116 @@ jobs:
...
+      # use `-browsers` prefix for selenium tests, e.g. `3.6.1-browsers`
+      - image: circleci/python:3.6.8-node
+
+
+        working_directory: ~/project
+
+    steps:
+      - add_ssh_keys:
+          fingerprints:
+            - "a1:22:8b:da:c9:9f:88:db:be:cb:1d:c5:12:de:70:4f"
```
emmanuelle: I was not sure how this line is used but I copied it none the less, to be discussed.

emmanuelle:
This one should be ready for review.

.circleci/config.yml
nicolaskruchten: In order to be able to push to the `plotly.py-docs` repo, yu'll need to manually build a keypair on your machine and add the public key to Github as a read/write deploy key, and you'll need to give the private key to Circle. The fingerprint for that key appears here for Circle to be able to use it.

Unfortunately you'll have to do this whole dance because I'm pretty sure I deleted the keypair from my machine once I was done setting this up ;)

@jdamiba just set this up in the `plotly.r-docs` repo so he can help you!

.circleci/config.yml
```diff
@@ -385,6 +385,116 @@ jobs:
...
+              python3 setup.py install
+              cd ../../../doc      
+            fi 
+            echo 'export PATH="/home/circleci/project/doc/node_modules/.bin:$PATH"' >> $BASH_ENV
+            sudo apt-get install -y poppler-utils libxtst6 xvfb libgtk2.0-0 libgconf-2-4 libnss3 libasound2
+            cd ..
+
+      - save_cache:
+          paths:
+            - ./venv
```
nicolaskruchten: I think these need to be prefixed with `./doc/`

nicolaskruchten:
Thanks very much for taking this one on!

.circleci/config.yml
emmanuelle: OK, thanks! Yeah, maybe it will be faster if @jdamiba can give me a hand :-).

emmanuelle: OK, so I generated the new pair of keys (thank you @jdamiba!) and did the part on circle CI (fingerprint updated on the last commit). Now to add the public part I think I need to be made admin of this repo since I cannot access the settings. Otherwise I can just send you @nicolaskruchten the key.

emmanuelle: ok done!

emmanuelle:
I think this one is ready now!

.circleci/config.yml
```diff
@@ -385,6 +385,116 @@ jobs:
...
+
+            cd ../..
+
+
+      - store_artifacts:
+          path: doc/build
+          destination: doc/build
+     
+ 
+      - store_artifacts:
```
nicolaskruchten: is this step really necessary? It adds like 45s to the build, which is already at 5min...

.circleci/config.yml
```diff
@@ -385,6 +385,116 @@ jobs:
...
+            python front-matter-ci.py build/html
+            python check-or-enforce-order.py build/html
+            if [ "${CIRCLE_BRANCH}" == "doc-prod" ]; then
+              cd build/html
+              git init
+              git config user.name nicolaskruchten
+              git config user.email nicolas@plot.ly
+              git add *
+              git commit -m build
+              git push --force git@github.com:plotly/plotly.py.git doc-prod:built
```
nicolaskruchten: I don't think that `doc-prod:built` at the end will work here, I think this refers to the branch name of the repo you `init` a few lines higher, so the default branch is `master`

.circleci/config.yml
```diff
@@ -385,6 +385,116 @@ jobs:
...
+            # fallback to using the latest cache if no exact match is found
+            - v1-dependencies-
+
+      - run:
+          name: install dependencies
+          command: |
+            cd doc
+            python3 -m venv venv
+            . venv/bin/activate
+            npm install electron@1.8.4
```
nicolaskruchten: I'm curious about why you needed a different set of things here than I did in the other repo...?

Was it just `npm install orca` that was missing?

emmanuelle: It's the same instructions I think, I just put the installation of electron and orca on two separate lines for the sake of debugging, I had trouble with finding the path where orca had been installed.

.circleci/config.yml
emmanuelle: oh yes thanks

.circleci/config.yml
emmanuelle: We can remove it I think. It can be nice when reviewing a PR because then you don't have to do the whole process of getting the branch and building the doc to look at how changes are rendered, but it's more useful for the build of tutorials than the apidoc (for which changes won't be that frequent).

nicolaskruchten:
💃 ! Thanks again! Plz fork the doc-prod branch once merged and confirm it deploys :)

PR #1937, opened 2019-12-02
New mapbox functions in PX

nicolaskruchten:
I'll add a couple of tests here and then this should be good to go :)


nicolaskruchten:
And then we'll need to update some docs to show off the new functions!

packages/python/plotly/plotly/express/_chart_types.py
```diff
@@ -998,6 +998,85 @@ def scatter_mapbox(
...
+    """
+    In a Mapbox choropleth map, each row of `data_frame` is represented by a
+    colored region on a Mapbox map.
+    """
+    return make_figure(
+        args=locals(),
+        constructor=go.Choroplethmapbox,
+        trace_patch=dict(
+            geojson=geojson
+            if not hasattr(geojson, "__geo_interface__")
```
nicolaskruchten: this is geopandas-specific code so that you can just say `...geojson=geo_df.geometry...` :)

emmanuelle:
Exactly is the doc migration PR is merged then you can add directly the updates to the doc!

emmanuelle:
So the percy script fails because it's missing the `center` argument which has just been introduced here.

nicolaskruchten:
Hah, yes, hurray tests :)

emmanuelle:
:dancer:, thanks!

PR #1939, opened 2019-12-02
Set docstring linewidth to 75 to avoid wrapping in jupyterlab help popup

joelostblom:
I noticed that I didn't cover the one line chart descriptions in #1866, so adding them in this PR.

Current:
image

Suggested:
image

When I did this I noticed that the exact line width that jupyterlab breaks at is 76, not 78, so I updated the autoformatting of the docs as well.

### DIFF
--- a/packages/python/plotly/plotly/express/_chart_types.py
+++ b/packages/python/plotly/plotly/express/_chart_types.py
@@ -49,8 +49,8 @@ def scatter(
     height=None,
 ):
     """
-    In a scatter plot, each row of `data_frame` is represented by a symbol mark
-    in 2D space.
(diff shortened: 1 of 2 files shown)


joelostblom:
I should note that the [pep8 recommendation for doscstrings is a line width of 72](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) so I can change to that if desired. I looked around briefly in the rest of the plotly code and it seems like a both 72 and 75 is used, so either should be fine for consistency within the project.

nicolaskruchten:
Thanks for this! Let's stick with 75 for now unless we can easily force everything to 72 :)

PR #1941, opened 2019-12-02
bypass confirmation when uninstalling stable plotly

emmanuelle:
The doc build is currently failing on master since it waits for a confirmation which never comes, this should fix this failure.


nicolaskruchten:
💃

PR #1949, opened 2019-12-03
symlink root readme

### LINKED ISSUE #1898: Readme gone from PyPI
Back in 3.10 days this is what our PyPI page looked like https://pypi.org/project/plotly/3.10.0/ but now it looks like https://pypi.org/project/plotly/ ... Something got lost in the shuffle with 4.0 I think.

nicolaskruchten:
Closes #1898 and replaces #1905

### DIFF
--- a/packages/python/plotly/README.md
+++ b/packages/python/plotly/README.md
@@ -1,1 +1,1 @@
-The main plotly distribution package
+../../../README.md


emmanuelle:
:dancer: We'll see if pypi likes it too!

PR #1951, opened 2019-12-03
bingroup

Mahdis-z:


doc/python/2D-Histogram.md
```diff
@@ -70,6 +70,33 @@ fig = go.Figure(go.Histogram2d(x=x, y=y, histnorm='probability',
...
+This example shows how to use [bingroup](https://plot.ly/python/reference/#histogram-bingroup) attribute to have a compatible bin settings for both histograms. To define `start`, `end` and `size` value of x-axis and y-axis seperatly, set [ybins](https://plot.ly/python/reference/#histogram2dcontour-ybins) and `xbins`.
+
+```python
+import plotly.graph_objects as go
+from plotly.subplots import make_subplots
+fig = make_subplots(2,2)
+fig.add_trace(go.Histogram2d(
+    x = [ 1, 2, 2, 3, 4 ],
+    y = [ 1, 2, 2, 3, 4 ],
+    xbins = {'start':1, 'size':1}), 1,1)
```
emmanuelle: sorry one last comment: these subplots should share the same coloraxis, at the moment the colorbar has several sets of labels superimposed

doc/python/histograms.md
```diff
@@ -350,6 +350,28 @@ fig1.show()
...
+fig = go.Figure(go.Histogram(
+    x=np.random.randint(7, size=100),
+    bingroup=1))
+
+fig.add_trace(go.Histogram(
+    x=np.random.randint(7, size=20),
+    bingroup=1))
+
+fig.update_layout(
+    barmode="overlay",
```
emmanuelle: From a quick glance at the figure one could think that the `barmode` is `stack` (that is the blue bars on top of the red ones). How about decreasing opacity and changing the sample size so that we can see both traces? Like
```
import numpy as np
fig = go.Figure(go.Histogram(
    x=np.random.randint(7, size=100),
    bingroup=1))
fig.add_trace(go.Histogram(
    x=np.random.randint(7, size=100),
    bingroup=1, 
    opacity=0.5))
fig.update_layout(
    barmode="overlay",
    bargap=0.1)
fig.show()
```

emmanuelle:
So I had new ideas :-). If it's clear for you how to make these changes, :dancer: once you've done the update (don't wait for me, merge it right away after doing the changes). Thanks @Mahdis-z !

PR #1979, opened 2019-12-05
multiple tooltips

Mahdis-z:


nicolaskruchten:
💃

emmanuelle:
Should this be a PR for `doc-prod` instead? (ie is this a feature which has already been released?)

PR #1981, opened 2019-12-05
correct zmin/zmax/range_color bug

### LINKED ISSUE #1931: [BUG] zmin and zmax not working in imshow for 2d data
See for example
```
import plotly.express as px
import numpy as np
img = np.random.random((10, 10))
px.imshow(img, zmin=0.3, zmax=0.7)
```
I think this is because the Heatmap trace uses instead the `range_color` argument which was quite introduced quite late in the process. It's easy to fix, we just have to decide first if the API is ok as it is.

emmanuelle:
Closes #1931


nicolaskruchten:
💃

PR #1987, opened 2019-12-09
'stack' barmode also works for px.bar

Ddedalus:
Hi, this is my first pull request, if I choose the wrong branch, please let me know :)

Just a small extension to the available options of `px.bar` to begin with.

### DIFF
--- a/packages/python/plotly/plotly/express/_doc.py
+++ b/packages/python/plotly/plotly/express/_doc.py
@@ -398,9 +398,10 @@ docs = dict(
     ],
     barmode=[
         "str (default `'relative'`)",
-        "One of `'group'`, `'overlay'` or `'relative'`",
+        "One of `'group'`, `'overlay'`, `'relative'` or `'stack'`",
+        "In `'stack'` mode, bars are placed on top of each other.",
         "In `'relative'` mode, bars are stacked above zero for positive values and below zero for negative values.",
-        "In `'overlay'` mode, bars are drawn on top of one another.",
+        "In `'overlay'` mode, bars are drawn over each other.",
         "In `'group'` mode, bars are placed beside each other.",
     ],
     boxmode=[


nicolaskruchten:
Thanks for this PR! I had initially left out "stack" because it's so easy to misunderstand what it does with respect to negative values... We probably should indeed add it but with a clear warning about how unintuitive the output can be.

Ddedalus:
I've tried it out on a small example and added a brief note. If you think more explanation is needed there (or my understanding is wrong), please go ahead. It may be a good idea to put a longer description in `go.Bar` docs as well...

Out of curiosity, how would you create a stack plot where positive and negative values are stacked separately on the opposite sides of the x-axis?

emmanuelle:
Thanks for updating the PR @Ddedalus ! @nicolaskruchten I think we can merge this one?

packages/python/plotly/plotly/express/_doc.py
```diff
@@ -418,9 +418,11 @@
     ],
     barmode=[
         "str (default `'relative'`)",
-        "One of `'group'`, `'overlay'` or `'relative'`",
+        "One of `'group'`, `'overlay'`, `'relative'` or `'stack'`",
+        "In `'stack'` mode, bars are placed on top of each other."
+        "Positive and negative values will cancel out, drawn in order of appearance in `data_frame`.",
```
emmanuelle:
```suggestion
        "Positive and negative values will cancel out, drawn in order of appearance of categories in `data_frame`.",
```

nicolaskruchten:
Right to stack above and below zero the option is “relative”. The “stack” option is supported but needs a clear and scary warning like “this is almost never what you want to use and will often mislead viewers when negative values are involved” or something like that. We have to support it for backwards compatibility reasons but IMO it’s quite a dangerous option.

PR #1988, opened 2019-12-09
update to plotly.js 1.51.2

jonmmease:


emmanuelle:
hum a doctesting problem... let me know if you need help to fix this, I did some modifications when adding the doctesting and I might have broken stuff.

jonmmease:
Thank @emmanuelle, yeah if you have any ideas on this failure I'd appreciate it! Were these doctests passing previously as far as you know?

packages/python/plotly/plotly/graph_objs/__init__.py
```diff
@@ -95064,31 +95067,27 @@ def template(self):
...
           - The name of a registered template where current registered templates
             are stored in the plotly.io.templates configuration object. The names
             of all registered templates can be retrieved with:
-
                 >>> import plotly.io as pio
-                >>> list(pio.templates) # doctest: +ELLIPSIS
-                ['ggplot2', 'seaborn', 'simple_white', 'plotly', 'plotly_white', 'plotly_dark', ...]
-
-
+                >>> list(pio.templates)
```
emmanuelle: This is the part which makes the doctest fails, the expected result is the line which was removed (probably I should have added it in the codegen and not here??).

jonmmease: Ohh, I missed that in the diff. Yeah, I'll add it in the codegen. Thanks!

PR #1995, opened 2019-12-10
px examples for pie, funnel/funnelarea, sunburst/treemap

emmanuelle:
~Sunburst and treemap examples will go to another PR because I did not want to slow down the review process.~ 

For sunburst and treemap I included only a minimal example since we will probably have a nicer API in the future, with `path`.

### DIFF
--- a/doc/python/pie-charts.md
+++ b/doc/python/pie-charts.md
@@ -5,8 +5,8 @@ jupyter:
     text_representation:
       extension: .md
       format_name: markdown
-      format_version: '1.1'
-      jupytext_version: 1.1.1
+      format_version: '1.2'
+      jupytext_version: 1.3.0
   kernelspec:
     display_name: Python 3
     language: python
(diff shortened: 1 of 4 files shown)


emmanuelle:
I deleted the problematic file on chart studio, the next commit should be ok (saving on electrons and not relaunching the build :-p).

emmanuelle:
ready for review !

doc/python/pie-charts.md
```diff
@@ -33,9 +33,63 @@ jupyter:
...
-### Basic Pie Chart ###
+A pie chart is a circular statistical chart, which is divided into sectors to illustrate numerical proportion. 
 
-A pie chart ``go.Pie`` object is a circular statistical chart, which is divided into sectors to illustrate numerical proportion. Data visualized by the sectors of the pie is set in `values`. The sector labels are set in `labels`. The sector colors are set in `marker.colors`. 
+If you're looking instead for a multilevel hierarchical pie-like chart, go to the
+[Sunburst tutorial](/python/sunburst-charts/).
+
+### Pie chart with plotly express
+
+`plotly.express` is a functional API for `plotly`. Its functions take as input pandas dataframe or objects that can be converted to dataframes such as arrays or dicts. For more details, see the [plotly.express tutorial](/plotly-express/).
```
nicolaskruchten: It would be nice to consistently use the same little sentence I added into all the `px` blocks at the top of pages like https://plot.ly/python/line-and-scatter/ and the corresponding one in the first non-`px` block 

> Plotly Express is the easy-to-use, high-level interface to Plotly, which operates on "tidy" data.

and 

> If Plotly Express does not provide a good starting point, it is possible to use the more generic go.Scatter function from plotly.graph_objects.

etc (with links)

If we could use this little formula for all the doc pages in this PR that would be 🎉

nicolaskruchten:
💃

PR #1996, opened 2019-12-10
Add support for MultiIndex levels in px

Ddedalus:
This is one feature I'm missing in my standard workflow where I often end up with a nice MultiIndex but need to drop its levels (or use horribly verbose `index.get_level_values()`) every time I try to plot something.

Here I provide an implementation which in a situation like `px(df, x='x')` searches for 'x' in the levels of `df.index` if its not found in `df.columns`.

I also provided a test function and updated suitable docstring. I hope to see this extension in future versions of Plotly :)

### DIFF
--- a/doc/python/px-arguments.md
+++ b/doc/python/px-arguments.md
@@ -87,6 +87,17 @@ fig = px.scatter(iris, x='sepal_length', y='sepal_width', color='species', size=
 fig.show()
 ```
 
+If the `data_frame` has a MultiIndex with named levels, it is possible to pass the name of the level as well.
+In such case its values are extracted with `get_level_values()`.
+If a column and an index level have the same name, the column is preferred.
+
+```python
+iris = px.data.iris()
+mi_iris = iris.set_index('species', append=True, drop=False)
+# Use level name for color. This is the same chart as above.
+fig = px.scatter(mi_iris, x='sepal_length', y='sepal_width', color='species', size='petal_length')
+fig.show()
+```
 #### Using the index of a DataFrame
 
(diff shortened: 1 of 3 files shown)


emmanuelle:
Hi @Ddedalus thanks for your pull request. One item on our plotly.py roadmap is to decide if/how we want to support multiindex, and your contribution helps us to think about this. However we're just at the beginning of this reflection so we cannot tell you as of know if we can merge this. We have to decide how much "magic" we want to do for users, with the upside of convenience and the downside of potential confusion.

Also, at the moment the tests are failing, if you have the time to investigate why it would be great (run `pytest` in the `/home/emma/code/plotly.py/packages/python/plotly/plotly/tests/test_core/test_px` directory), otherwise we'll take a look.

Ddedalus:
@emmanuelle thanks for the clarification on that. I appreciate that handling MultiIndex in full generality will be a complex effort of the team. Here is my nudge in what I think is the right direction.

I'm sorry for the failed tests, I misunderstood what the expectations are in terms of indexing `df_output`. Now it should almost always come with a RangeIndex. I've also silenced an error about int32/64 selection inside Pandas (occurred in `test_build_df_from_lists`). I get this error on master as well, so I believe it is not related to my change.

gvwilson:
Hi - we are currently trying to tidy up Plotly's public repositories to help us focus our efforts on things that will help users most. Since this issue has been sitting for several years, I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our backlog. Thanks for your help - @gvwilson

PR #1998, opened 2019-12-11
mapbox examples with px

emmanuelle:


doc/python/mapbox-density-heatmaps.md
```diff
@@ -42,14 +41,34 @@ To plot on Mapbox maps with Plotly you *may* need a Mapbox account and a public
...
+import plotly.express as px
+fig = px.density_mapbox(df, lat='Latitude', lon='Longitude', z='Magnitude', radius=10,
+                        center=dict(lat=0, lon=180), zoom=0,
+                        mapbox_style="stamen-terrain")
+fig.show()
+```
+
+### Stamen Terrain base map (no token needed): density mapbox with `plotly.graph_objects`
+
+f Plotly Express does not provide a good starting point, it is also possible to use the more generic `go.Densitymapbox` function from `plotly.graph_objects`.
```
nicolaskruchten:
```suggestion
If Plotly Express does not provide a good starting point, it is also possible to use the more generic `go.Densitymapbox` function from `plotly.graph_objects`.
```

nicolaskruchten:
💃 pending my "suggestion" comment

thanks very much :)

PR #2006, opened 2019-12-16
Sunburst/treemap path

emmanuelle:
Still WIP, TODO
- [x] write doc
- [x] discuss the different cases we want to support
- [x] discuss case of hover columns


packages/python/plotly/plotly/express/_doc.py
```diff
@@ -82,6 +82,7 @@
         colref_desc,
         "Values from this column or array_like are used to set ids of sectors",
     ],
+    path=[colref_type, colref_desc],
```
emmanuelle: needs work

mwouts:
Thank you @emmanuelle for working on this! This will make is way easier for everyone to produce these beautiful plots, and I will certainly be a user of this.

I have a few remarks/questions:
- Would it make sense to could extend this to non-rectangular trees by using 'None' in the path?
- Sometime want to compute aggregates on all nodes, not only for the 'value', but also for the hovertext, like the life expectancy in the plot below. Do you think we could cover this use case with a dataframe? Maybe with intermediate nodes? (I was using a function)
[image](https://mwouts.github.io/easyplotly/#Treemaps)
- (technical) do you want to add epsilon when you compute the sum, until https://github.com/plotly/plotly.js/issues/4405 gets fixed? See https://github.com/mwouts/EasyPlotly/issues/3 for an example that produces an empty plot because of float comparisons issues.

nicolaskruchten:
Thanks for taking a look @mwouts !

Re intermediate nodes, I think we should try to accept null/None/NaN for trailing path elements somehow, yes.

Re values for intermediate nodes, let's see how far we can get by having users provide them with the approach above. We could accept an aggregating function as a parameter eventually if this is too limiting :)

Also, I don't think we should add an epsilon, as this will be released as part of 4.5.0, which will include the fix for https://github.com/plotly/plotly.js/issues/4405

emmanuelle:
Thanks for taking a look @mwouts ! @nicolaskruchten you were mentioning that we could accept a dictionary of lambda functions for doing the aggregation, is this what you had in mind?

emmanuelle:
In order to address the case of lines with None, maybe we should have `branchvalues='remainder'` as default and when doing the groupby, put the value of the parent to 0 is it's not already specified? (instead of computing the sum as it is implemented now?).

emmanuelle:
@mwouts we were also looking for a way to address the case of https://plot.ly/python/sunburst-charts/#sunburst-chart-with-a-continuous-colorscale but it seems hard to use a generic API for this...

emmanuelle:
After all I included an EPS fix because otherwise my doc example (with `tips`) did not work! On second thoughts, I also kept the groupby mechanism (including computing the sum of children) since if we want to generalize to more complex aggregations we will have a similar syntax, whereas switching to a remainder mode implied a somewhat different code.

emmanuelle:
I wrote the doc for sunburst but not for treemap since it will be mostly a copy so I'll update after a round of review.

mwouts:
Thanks! Glad that helps!

> @mwouts we were also looking for a way to address the case of https://plot.ly/python/sunburst-charts/#sunburst-chart-with-a-continuous-colorscale but it seems hard to use a generic API for this...

Well don't you think the same approach, generalized to magic underscore notation, could work? In my [example](https://mwouts.github.io/easyplotly/#Treemaps) (scroll to the second treemap plot) I have defined `marker_color` as a function, but I believe a column of the dataframe (maybe extended to include the parent nodes) could work as well.

packages/python/plotly/plotly/express/_doc.py
nicolaskruchten: hehe yes :)

emmanuelle: oops yes. The PR is not completely ready but I'll still appreciate your thoughts about the API and how it's implemented, then I will polish if this is the right direction, unless there is some refactoring to do first.

nicolaskruchten:
Two bits of feedback to start with:

1. The path should go from parent to child rather than the other way around
2. For the continuous-color case, it makes sense to me that for "generated" parent nodes (i.e. ones that don't have a dedicated row whose `path` ends in `None`!) for the `color` of the parent to be the `value`-weighted average of the children's `color`. (NB color here is a scalar!)
3. For the discrete-color case, it's trickier: there's no sensible 'implicit' behaviour that I can think of other than imputing some dummy value to color by, or to error out

nicolaskruchten:
Overall this is really, really awesome though, I love the symmetry with `parcoords`/`parcats`/`splom` !

nicolaskruchten:
PS: I think we can probably use a similar API for `px.sankey`

nicolaskruchten:
As a motivating example for the implicit behaviour in continuous color, think about a case like `px.sunburst(gapminder2007, path=["continent", "country"], values="pop", color="lifeExp")`

With my scheme, the `continent` sectors would be colored by the population-weighted average life expectancy, which is pretty much what you'd want. Ditto for `values="gdpPercap"` etc.

emmanuelle:
Question about color: if `color_continuous_scale` is passed but `color=None`, then a discrete color sequence will be used. Is it ok or should we take the `values` argument instead? I'd rather keep the current behaviour (not too much magics) but just checking.

nicolaskruchten:
Yep the current behaviour is fine when color is not passed IMO

emmanuelle:
Still need to update the doc tutorial

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1015,9 +1108,9 @@ def infer_config(args, constructor, trace_patch):
         + ["names", "values", "parents", "ids"]
         + ["error_x", "error_x_minus"]
         + ["error_y", "error_y_minus", "error_z", "error_z_minus"]
-        + ["lat", "lon", "locations", "animation_group"]
```
nicolaskruchten: I'm wondering if it's necessary to add it here and to array_attrables, as we're basically rewriting the dataframe here and not passing `path` down any further into the machinery

emmanuelle: it's because `build_dataframe` is called before `process_dataframe_hierarchy`. This way, it works when a user passes arrays or columns in the path instead of just column names.

nicolaskruchten: ah perfect. so... much... magic!

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1007,6 +1007,99 @@ def build_dataframe(args, attrables, array_attrables):
...
+
+def process_dataframe_hierarchy(args):
+    """
+    Build dataframe for sunburst or treemap when the path argument is provided.
+    """
+    df = args["data_frame"]
+    path = args["path"][::-1]
+
+    # ------------ Define aggregation functions --------------------------------
+    lambda_discrete = _discrete_agg
```
nicolaskruchten: I would just `def lambda_discrete(x):` here instead of above with `_` and then assigning here

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1007,6 +1007,99 @@ def build_dataframe(args, attrables, array_attrables):
...
+            # trick to be sure the col name is unused: take the sum of existing names
+            count_colname = "".join([str(el) for el in list(df.columns)])
+            # we can modify df because it's a copy of the px argument
+            df[count_colname] = 1
+
+    if args["color"]:
+        if df[args["color"]].dtype.kind not in "bifc":
+            aggfunc_color = lambda_discrete
+        elif not aggfunc_color:
+            aggfunc_color = lambda x: np.average(
```
nicolaskruchten: flake8 doesn't like this lambda-assignment, so I would `def` it above as `lambda_continuous` or something

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1007,6 +1007,99 @@ def build_dataframe(args, attrables, array_attrables):
...
+            df[args["values"]] = pd.to_numeric(df[args["values"]])
+        except ValueError:
+            raise ValueError(
+                "Column `%s` of `df` could not be converted to a numerical data type."
+                % args["values"]
+            )
+
+        if args["color"]:
+            if args["color"] == args["values"]:
+                aggfunc_color = "sum"
```
nicolaskruchten: this confuses me a little bit... can you explain?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1007,6 +1007,99 @@ def build_dataframe(args, attrables, array_attrables):
...
+    # ----------------------------------------------------------------------------
+
+    df_all_trees = pd.DataFrame(columns=["labels", "parent", "id"] + cols)
+    #  Set column type here (useful for continuous vs discrete colorscale)
+    for col in cols:
+        df_all_trees[col] = df_all_trees[col].astype(df[col].dtype)
+    for i, level in enumerate(path):
+        df_tree = pd.DataFrame(columns=df_all_trees.columns)
+        if not agg_f:
+            dfg = df.groupby(path[i:]).sum(numerical_only=True)
```
nicolaskruchten: under what circumstances do we do this, and what's the reasoning?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1007,6 +1007,99 @@ def build_dataframe(args, attrables, array_attrables):
...
+        df_tree["parent"] = ""
+        df_tree["id"] = dfg[level].copy().astype(str)
+        if i < len(path) - 1:
+            j = i + 1
+            while j < len(path):
+                df_tree["parent"] += dfg[path[j]].copy().astype(str)
+                df_tree["id"] += dfg[path[j]].copy().astype(str)
+                j += 1
+        else:
+            df_tree["parent"] = ""
```
nicolaskruchten: redundant `else`?

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: I think doing the weighted-average still makes sense to me here... if I have two children of value=color=[2 and 3], then my value is 5 but my color should be (3*3+2*2)/(3+2) = 2.6

nicolaskruchten: hmmm now that I write it out, this makes no sense. `sum` is the better approach here indeed.

nicolaskruchten:
This causes an error: `px.sunburst(df, path=["continent", "country"], values="pop", color="continent")`

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1007,6 +1007,99 @@ def build_dataframe(args, attrables, array_attrables):
...
+            dfg = df.groupby(path[i:]).agg(agg_f)
+        dfg = dfg.reset_index()
+        df_tree["labels"] = dfg[level].copy().astype(str)
+        df_tree["parent"] = ""
+        df_tree["id"] = dfg[level].copy().astype(str)
+        if i < len(path) - 1:
+            j = i + 1
+            while j < len(path):
+                df_tree["parent"] += dfg[path[j]].copy().astype(str)
+                df_tree["id"] += dfg[path[j]].copy().astype(str)
```
nicolaskruchten: Here the IDs for `["Path","To","Node"]` end up being `"NodeToPath"` and are visible in hoverlabels... it would be nice for them to be `"Path/To/Node"` IMO :)

emmanuelle: I would also prefer this but then I had a hard time implementing it :-). I'll give it another go.

packages/python/plotly/plotly/express/_core.py
emmanuelle: ok

packages/python/plotly/plotly/express/_core.py
emmanuelle: it's if neither `values` nor `color` was passed. I think I need to use some sort of aggregation function because after I call `reset_index` to get the groups, but maybe it's possible to do it in a more elegant/efficient way.

packages/python/plotly/plotly/express/_core.py
emmanuelle: indeed

packages/python/plotly/plotly/express/_core.py
emmanuelle: I could get rid of this part with the `count_colname` logic.

packages/python/plotly/plotly/express/_core.py
emmanuelle: ok wasn't that hard :-)

packages/python/plotly/plotly/express/_core.py
emmanuelle: ok

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1007,6 +1007,104 @@ def build_dataframe(args, attrables, array_attrables):
...
+        df_tree["labels"] = dfg[level].copy().astype(str)
+        df_tree["parent"] = ""
+        df_tree["id"] = dfg[level].copy().astype(str)
+        if i < len(path) - 1:
+            j = i + 1
+            while j < len(path):
+                df_tree["parent"] += dfg[path[j]].copy().astype(str)
+                if j < len(path) - 1:
+                    df_tree["parent"] += "/"
+                df_tree["id"] += "/" + dfg[path[j]].copy().astype(str)
```
nicolaskruchten: these are still "backwards" though right?

Would `df_tree["id"] = dfg[path[j]].copy().astype(str) + "/" + df_tree["id"]` not work?

PR #2007, opened 2019-12-17
standardize on df variable name

nicolaskruchten:


doc/python/px-arguments.md
```diff
@@ -189,9 +190,9 @@ import plotly.express as px
 import numpy as np
 import pandas as pd
 
-gapminder = px.data.gapminder()
-gdp = np.log(gapminder['pop'] * gapminder['gdpPercap'])  # NumPy array
-fig = px.bar(gapminder, x='year', y=gdp, color='continent', labels={'y':'log gdp'},
+df = px.data.gapminder()
+gdp = np.log(df['pop'] * gapminder['gdpPercap'])  # NumPy array
```
emmanuelle: gapminder -> df

emmanuelle: (thanks to the CI !)

nicolaskruchten: yep, just saw that one in the tests. I think it's the only one I missed though!

nicolaskruchten:
FYI my search'n'replace approach only targets the datasets in `px.data` and not others we may have named while loading from files etc... should be a pretty good start though!

Do you recall any other names we may have used repeatedly?

emmanuelle:
I did a `grep "pd.read_csv('https" *` and only found a couple of places where it's not `df` or `df_something`, like `milk_data` in `peak_finding.md`

PR #2016, opened 2019-12-25
first and last color of cyclicals should match

nicolaskruchten:


nicolaskruchten:
Without this fix:

image

With this fix:

image

nicolaskruchten:
Note that the last two look markedly better, and the other ones that changed don't look all that different, because the last color didn't change all that much.

emmanuelle:
Looks good ! Ideally we would document the new cyclical swatches function but maybe it should go instead to a new "colors" page of the documentation (I think you mentioned it previously). Maybe open an issue to remember that we need to document this? But this should not block the PR, so :dancer:

nicolaskruchten:
I've just added this page in doc-prod https://plot.ly/python/builtin-colorscales/ so I'll update this PR to expand that page a bit :)

PR #2025, opened 2020-01-01
Add an example with `constrain='domain'`

jingw222:
Add a following example demonstrating Fixed Ratio Axes with Compressed domain to [axes tutorial](https://plot.ly/python/axes/).


emmanuelle:
Great example, thank you very much for your contribution to `plotly.py`! Merging :dancer:.

PR #2030, opened 2020-01-04
Color docs

nicolaskruchten:
Once this is merged, I'll update https://github.com/plotly/plotly.py/pull/2016 and https://github.com/plotly/plotly.py/pull/2017

### DIFF
--- a/doc/python/categorical-colors.md
+++ b/doc/python/categorical-colors.md
@@ -1,0 +1,231 @@
+---
+jupyter:
+  jupytext:
+    notebook_metadata_filter: all
+    text_representation:
+      extension: .md
+      format_name: markdown
+      format_version: "1.2"
+      jupytext_version: 1.3.1
+  kernelspec:
+    display_name: Python 3
+    language: python
+    name: python3
+  language_info:
+    codemirror_mode:
+      name: ipython
+      version: 3
+    file_extension: .py
+    mimetype: text/x-python
+    name: python
+    nbconvert_exporter: python
+    pygments_lexer: ipython3
+    version: 3.6.8
+  plotly:
+    description:
+      How to use and configure categorical (also known as qualitative or
+      discrete) color sequences.
+    display_as: file_settings
+    has_thumbnail: true
+    ipynb: ~notebook_demo/187
+    language: python
+    layout: base
+    name: Categorical Colors
+    order: 27
+    permalink: python/categorical-color/
+    thumbnail: thumbnail/heatmap_colorscale.jpg
+    v4upgrade: true
+---
+
+### Categorical vs Continuous Color
+
+In the same way as the X or Y position of a mark in cartesian coordinates can be used to represent continuous values (i.e. amounts or moments in time) or categories (i.e. labels), color can be used to represent continuous or categorical data. This page is about using color to represent **categorical** data, but Plotly can also [represent continuous values with color](/python/colorscales/).
+
+### Categorical Color Concepts
+
+This document explains the following categorical-color-related concepts:
+
+- **color sequences** are lists of colors to be mapped onto discrete data values. No interpolation occurs when using color sequences, unlike with [continuous color scales](/python/colorscales/), and each color is used as-is. Color sequence defaults depend on the `layout.colorway` attribute of the active [template](/python/templates/), and can be explicitly specified using the `color_discrete_sequence` argument for many [Plotly Express](/python/plotly-express/) functions.
+- **legends** are visible representations of the mapping between colors and data values. Legend markers also change shape when used with various kinds of traces, such as symbols or lines for scatter-like traces. [Legends are configurable](/python/legend/) under the `layout.legend` attribute. Legends are the categorical equivalent of [continous color bars](/python/colorscales/)
+
+### Categorical Color with Plotly Express
+
+Most Plotly Express functions accept a `color` argument which automatically assigns data values to categorical colors **if the data is non-numeric**. If the data is numeric, the color will automatically be considered [continuous](/python/colorscales/). This means that numeric strings must be parsed to be used for continuous color, and conversely, numbers used as category codes must be converted to strings.
+
+For example, in the `tips` dataset, the `smoker` column contains strings:
+
+```python
+import plotly.express as px
+df = px.data.tips()
+fig = px.scatter(df, x="total_bill", y="tip", color="smoker", title="String 'smoker' values mean categorical colors")
+
+fig.show()
+```
+
+The `size` column, however, contains numbers:
+
+```python
+import plotly.express as px
+df = px.data.tips()
+fig = px.scatter(df, x="total_bill", y="tip", color="size", title="Numeric 'size' values mean continous color")
+
+fig.show()
+```
+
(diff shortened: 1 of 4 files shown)


doc/python/categorical-colors.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Categorical vs Continuous Color
+
+In the same way as the X or Y position of a mark in cartesian coordinates can be used to represent continuous values (i.e. amounts or moments in time) or categories (i.e. labels), color can be used to represent continuous or categorical data. This page is about using color to represent **categorical** data, but Plotly can also [represent continuous values with color](/python/colorscales/).
+
+### Categorical Color Concepts
+
+This document explains the following categorical-color-related concepts:
+
+- **color sequences** are lists of colors to be mapped onto discrete data values. No interpolation occurs when using color sequences, unlike with [continuous color scales](/python/colorscales/), and each color is used as-is. Color sequence defaults depend on the `layout.colorway` attribute of the active [template](/python/templates/), and can be explicitly specified using the `color_discrete_sequence` argument for many [Plotly Express](/python/plotly-express/) functions.
```
emmanuelle: `layout.colorway`: except for pie, treemap, sunburst etc. but maybe we don't want to go into such details here.

doc/python/categorical-colors.md
```diff
@@ -0,0 +1,231 @@
...
+### Categorical Color with Plotly Express
+
+Most Plotly Express functions accept a `color` argument which automatically assigns data values to categorical colors **if the data is non-numeric**. If the data is numeric, the color will automatically be considered [continuous](/python/colorscales/). This means that numeric strings must be parsed to be used for continuous color, and conversely, numbers used as category codes must be converted to strings.
+
+For example, in the `tips` dataset, the `smoker` column contains strings:
+
+```python
+import plotly.express as px
+df = px.data.tips()
+fig = px.scatter(df, x="total_bill", y="tip", color="smoker", title="String 'smoker' values mean categorical colors")
```
emmanuelle: break line before `title` to avoid scrolling

doc/python/categorical-colors.md
```diff
@@ -0,0 +1,231 @@
...
+print(px.colors.qualitative.Plotly)
+```
+
+Here is an example that creates a scatter plot using Plotly Express, with points colored using the built-in qualitative `G10` color sequence.
+
+```python
+import plotly.express as px
+df = px.data.gapminder()
+fig = px.line(df, y="lifeExp", x="year", color="continent", line_group="country",
+              line_shape="spline", render_mode="svg",
```
emmanuelle: Can we keep the example short by removing this line? Is it a bug that `spline` gives an error when not using `render_mode='svg'`?

doc/python/categorical-colors.md
```diff
@@ -0,0 +1,231 @@
...
+             color_discrete_sequence=px.colors.qualitative.G10,
+             title="Built-in G10 color sequence")
+
+fig.show()
+```
+
+### Explicity Constructing a Color Sequence
```
emmanuelle: Explicitly

doc/python/categorical-colors.md
nicolaskruchten: yeah, I think that's too much detail, and also `layout.*colorway` defaults to `layout.colorway` if it's not set IIRC.

doc/python/categorical-colors.md
nicolaskruchten: I think webgl doesn't support spline, so maybe yes, spline should force svg? maybe too magical... To be honest I just didn't like the look of the figure in webgl :)

doc/python/categorical-colors.md
```diff
@@ -0,0 +1,231 @@
...
+fig.show()
+```
+
+```python
+import plotly.express as px
+df = px.data.gapminder().query("year == 2007")
+fig = px.bar(df, y="continent", x="pop", color="continent", orientation="h", hover_name="country",
+             color_discrete_sequence=["red", "green", "blue", "goldenrod", "magenta"],
+             category_orders={"continent": ["Oceania", "Europe", "Asia", "Africa", "Americas"]},
+             title="Explicit color mapping with explicit ordering"
```
emmanuelle: I'm a bit confused about the titles of these two examples. Shouldn't the previous one be "Explicit color mapping with explicit ordering" and this one "Explicit color sequence with explicit ordering"?

doc/python/colorscales.md
```diff
@@ -22,27 +22,64 @@ jupyter:
...
+- **color bars** are legend-like visible representations of the color range and color scale with optional tick labels and tick marks. Color bars can be configured with attributes inside `layout.coloraxis.colorbar` or in places like `marker.colorbar` in `go.Scatter` traces.
+- **color axes** connect color scales, color ranges and color bars to a trace's data. By default, any colorable attribute in a trace is attached to its own local color axis, but color axes may also be shared across attributes and traces by setting e.g. `marker.coloraxis` in `go.Scatter` traces. Local color axis attributes are configured within traces e.g. `marker.showscale` whereas shared color axis attributes are configured within the Layout e.g. `layout.coloraxis.showscale`.
```
emmanuelle: Suggestion: 
e.g. `marker.coloraxis` in `go.Scatter` or `coloraxis` in `go.Heatmap` traces.

doc/python/colorscales.md
```diff
@@ -22,27 +22,64 @@ jupyter:
...
+
+fig.show()
+```
+
+Converting this column to strings is very straightforward:
+
+```python
+import plotly.express as px
+df = px.data.tips()
+df["size"] = df["size"].astype(str)
```
emmanuelle: show how to go back to float as in the categorical page? It's a common pattern to open a csv file with string columns which are in fact numerical values.

doc/python/colorscales.md
```diff
@@ -84,11 +121,154 @@ import plotly.express as px
...
+df = px.data.tips()
+fig = px.density_heatmap(df, x="total_bill", y="tip", title="Customized color bar on this density plot")
+
+fig.update_layout(coloraxis_colorbar=dict(
+    title="Number of Bills per Cell",
+    thicknessmode="pixels", thickness=50,
+    lenmode="pixels", len=200,
+    yanchor="top", y=1,
+    ticks="outside", ticksuffix=" bills",
+    dtick=5
```
emmanuelle: Suggestion: add
`tickvals=[0, 10, 20, 30]`
here, this is something that users often want to configure.
Maybe also add a link to the axes tutorial which has a number of examples about ticks.

emmanuelle: oops I had not seen that it was done in the next example, please ignore this suggestion

emmanuelle:
This is amazing work, thank you! 
* I left a couple of minor comments, none of them is very important.
* In the existing colorscale page which you modified, I wonder if it would be possible to intertwine more the px and graph_objects examples. At the moment it's one big block of px and then another block of graph_objects, but the concepts are actually the same so maybe it would make more sense. It can also be done later also.
* the CI seems to fail because of the order of the files?

nicolaskruchten:
I addressed most comments but left PX/GO separate for now... We can do another pass later I think.

emmanuelle:
:dancer: when everything's green!

PR #2032, opened 2020-01-05
Add some comments about the configuration of the legend

WenjieZ:
~~Remove "Legend Entries" in the title since it is already in an individual page.~~

Add some comments about the configuration of the legend.

### DIFF
--- a/doc/python/figure-labels.md
+++ b/doc/python/figure-labels.md
@@ -22,12 +22,12 @@ jupyter:
     pygments_lexer: ipython3
     version: 3.7.3
   plotly:
-    description: How to set the global font, title, and axis-titles
+    description: How to set the global font, title, legend-entries, and axis-titles
       in python.
     display_as: file_settings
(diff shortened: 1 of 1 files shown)


emmanuelle:
Hi @WenjieZ , I would prefer to keep "legend entries" here because setting the `name` parameter of a trace sets its legend entry. But you're right that this example could be improved by
- adding an inline comment in the first comment to say that `name` is used as legend entry
- and to add sentence mentioning that another page is dedicated to legends, with a link to this page (the link would be `/python/legend` since we use relative links in the doc).

If you agree with these suggestions, would you have the time to implement it? Thanks anyway, it's very helpful to get feedback and improvements about the doc.

WenjieZ:
Hi @emmanuelle , let me know whether that modification is good enough.

emmanuelle:
Great, thanks a lot @WenjieZ ! Merging :dancer:

PR #2033, opened 2020-01-06
switch ternary scatter and ternary contour doc order

jdamiba:
closes https://github.com/plotly/documentation/issues/1553

The purpose of this PR is to swap the order of the ternary contour and scatter plot examples so that the countour example shows up on the index page.

### DIFF
--- a/doc/python/ternary-contour.md
+++ b/doc/python/ternary-contour.md
@@ -28,7 +28,7 @@ jupyter:
     layout: base
     name: Ternary contours
     order: 4
-    page_type: u-guide
+    page_type: example_index
     permalink: python/ternary-contour/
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
I actually think that plotly/documentation#1553 needs to be solved by having the `ternary-plots` thumbnail show a ternary scatter and not a contour. 

You can use some crop/resize of this: https://plot.ly/python/plotly-express/#ternary-coordinates

nicolaskruchten:
Actually we already have that one: https://images.plot.ly/plotly-documentation/thumbnail/v4-migration.png

jdamiba:
superseded by https://github.com/plotly/plotly.py/pull/2112

PR #2034, opened 2020-01-06
document usage of an externally hosted Orca instance

antoinerg:
This PR adds documentation for the new feature introduced in https://github.com/plotly/plotly.py/pull/1850 and https://github.com/plotly/plotly.py/pull/1915 that enables using an externally running Orca to produce static images.

Please feel free to rephrase/modify the text!

cc @nicolaskruchten

### DIFF
--- a/doc/python/static-image-export.md
+++ b/doc/python/static-image-export.md
@@ -64,6 +64,19 @@ If you are unable to install conda or npm, you can install orca as a precompiled
 ```
 $ pip install psutil requests
 ```
+
+##### Externally Hosted Orca + pip
+If you already run an instance of [orca](https://github.com/plotly/orca) on your network, there's no need to install it.
+You can simply point to it in your Python code.
+
+```python
+import plotly
+# specify URL to Orca's endpoint
+plotly.io.orca.config.server_url = "http://localhost:9091"
+```
+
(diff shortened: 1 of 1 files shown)


doc/python/static-image-export.md
```diff
@@ -64,6 +64,19 @@ If you are unable to install conda or npm, you can install orca as a precompiled
 ```
 $ pip install psutil requests
 ```
+
+##### Externally Hosted Orca + pip
```
nicolaskruchten: why "+ pip" ?

antoinerg: > why "+ pip" ?

I copied the title style of the other subsections. Would you prefer me to remove it?

P.S. One still needs to install dependencies via `pip` namely `requests` and `psutil` (the requirement for the latter could/should be removed but atm it's mandatory) for this feature to work.

nicolaskruchten:
OK so I actually think this should go in https://plot.ly/python/orca-management/ instead... also, if you could branch off of the `doc-prod` branch and target that branch with your PR that would get things into the docs faster :) The easiest way to do this is to browse to https://plot.ly/python/orca-management/ and click on the "edit this page in Github" link :)

PR #2038, opened 2020-01-07
bumping JS to 1.52pre

nicolaskruchten:


packages/javascript/plotlywidget/package.json
```diff
@@ -35,7 +35,7 @@
     "typescript": "~3.1.1"
   },
   "dependencies": {
-    "plotly.js": "^1.51.2",
+    "plotly.js": "https://64237-45646037-gh.circle-artifacts.com/0/plotly.js.tgz",
```
emmanuelle: was this changed when running the `setup.py` or did you change it manually? I'm asking because I had first tried to change this line manually by giving a URL as here, but then when I ran `npm install` npm was throwing an error.

nicolaskruchten: `setup.py` did this.

emmanuelle:
So the CI failure here is related to the name of the plotly.js version, I expect the test did not expect that we use something else than a release of plotly.js. We might want to fox this in the future if it becomes a pattern to work on pre-releases in the interest of time.

emmanuelle:
Percy changes correspond to the new zoom in/out modebar buttons for mapbox, so it's ok. The doc build CI job passed, so it will still be possible to deploy the doc even with the failing tests on the other jobs.

emmanuelle:
However, since failing CI builds are a pain, you might want to fix manually one of `package.json` or `_plotlyjs_version.py` so that they agree. The failing test is
```
@attr("nodev")
    def test_plotlyjs_version(self):
        path = os.path.join(packages_root, "javascript", "plotlywidget", "package.json")
        with open(path, "rt") as f:
            package_json = json.load(f)
            expected_version = package_json["dependencies"]["plotly.js"]
            if expected_version[0] == "^":
                expected_version = expected_version[1:]

        self.assertEqual(expected_version, plotly.offline.get_plotlyjs_version())
```
and indeed there are not the same.

nicolaskruchten:
OK well it's pretty easy to loosen the test so that it ignores this check when we're building from a circleci artifact.

nicolaskruchten:
Maybe not worth doing this time around, as 1.52 will come out very soon.

emmanuelle:
closed by #2047.

PR #2043, opened 2020-01-07
fix CI by disabling test

nicolaskruchten:
Same PR as the last one, but in doc-prod... I should have done it here first and then merged doc-prod to master but oh well.


emmanuelle:
:dancer:

PR #2053, opened 2020-01-09
config list

emmanuelle:
workaround for now, we would need to do this through the codegen instead


nicolaskruchten:
💃

emmanuelle:
actually I need to remove the `notifyOnLogging` since it's not a `config` option.

PR #2057, opened 2020-01-10
categorical choropleths

nicolaskruchten:


nicolaskruchten:
missing docs, of course

packages/python/plotly/plotly/express/_chart_types.py
```diff
@@ -866,7 +870,13 @@ def choropleth(
     return make_figure(
         args=locals(),
         constructor=go.Choropleth,
-        trace_patch=dict(locationmode=locationmode),
+        trace_patch=dict(
+            locationmode=locationmode,
+            featureidkey=featureidkey,
+            geojson=geojson
+            if not hasattr(geojson, "__geo_interface__")
```
emmanuelle: add a comment here to explain why this is needed?

nicolaskruchten: sure, this is a special hook for `geopandas`... if you have a geopandas frame, it has a special property `.geometry` which has a special `.__geo_interface__` property which returns GeoJSON, so doing this here allows the user to say `px.choropleth(df, geojson=df.geometry,...)`

emmanuelle: thanks, I was wondering whether this explanation could be added to the source code as a comment, for future maintainers

emmanuelle:
Looks good to me, and the example with the `gapminder` and `continent` color will be straightforward to add. It would be good practice to add it here but it can also be another PR, so :dancer: for this part. The only issue I experienced is that when you hide a legend item you get the "double-click on the legend to isolate a trace" message, which I do not really understand (or it does not work), but this is not related to this PR.

nicolaskruchten:
> the "double-click on the legend to isolate a trace" message, which I do not really understand (or it does not work)

it does work for me... although it is odd that it's just that one hint we give to users

emmanuelle:
:dancer:

emmanuelle:
:dancer:

PR #2058, opened 2020-01-11
Update plotly-express.md

jdamiba:
I think it would be a good idea to include a link that shows all the built in data sets. I had to go looking into the package code to find this file.

### DIFF
--- a/doc/python/plotly-express.md
+++ b/doc/python/plotly-express.md
@@ -53,6 +53,8 @@ print(px.data.iris.__doc__)
 px.data.iris().head()
 ```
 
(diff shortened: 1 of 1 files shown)


doc/python/plotly-express.md
```diff
@@ -53,6 +53,8 @@ print(px.data.iris.__doc__)
 px.data.iris().head()
 ```
 
+To see the full list of built-in datasets, click [here](https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/data/__init__.py).
```
emmanuelle: The best thing would probably be to add an `__all__` to `data/__init__.py` and to `print(px.data.__all__)`.

Issue #2049, opened 2020-01-09
Document quartilemethod and  "pre-computed" attributes for Box traces


jdamiba:
a part of https://github.com/plotly/plotly.py/issues/2044

references work done in https://github.com/plotly/plotly.js/pull/4432

- [x] create example with q1/median/q3/lowerfence/upperfence/mean/sd/notchspan - https://codepen.io/etpinard/pen/gObgzdL?editors=1010

- [x] create example with `inclusive`, `exclusive`, and `linear` quartile-computing algorithm - https://codepen.io/etpinard/pen/GRgrdyZ

Issue #10, opened 2014-03-16
Wrong facebook link?


jiayiliu:
Hi,
Nice library!

I am going to 'like' you on facebook, however the facebook link is wrong and I can't access the plot.ly page on facebook neither.

PR #2063, opened 2020-01-13
Update box plots doc with precomputed quartiles and quartilemethod

jdamiba:
closes https://github.com/plotly/plotly.py/issues/2049

The purpose of this PR is to update the box plots doc to include information about using precomputed quartile values and using the `quartilemethod` attribute to select algorithms for quartile computation.

### DIFF
--- a/doc/python/box-plots.md
+++ b/doc/python/box-plots.md
@@ -73,7 +73,9 @@ fig.show()
 
 ### Choosing The Algorithm For Computing Quartiles
 
-By default, quartiles for box plots are computed using a linear algorithm method (see #10 listed on [http://www.amstat.org/publications/jse/v14n3/langford.html](http://www.amstat.org/publications/jse/v14n3/langford.html) for more details). However, you can also choose to use an `exclusive` or an `inclusive` algorithm to compute quartiles. 
+By default, quartiles for box plots are computed using a linear algorithm method (see #10 listed on [http://www.amstat.org/publications/jse/v14n3/langford.html](http://www.amstat.org/publications/jse/v14n3/langford.html) and [https://en.wikipedia.org/wiki/Quartile](https://en.wikipedia.org/wiki/Quartile) for more details). 
+
+However, you can also choose to use an `exclusive` or an `inclusive` algorithm to compute quartiles. 
 
 The *exclusive* algorithm uses the median to divide the ordered dataset into two halves. If the sample is odd, it does not include the median in either half. Q1 is then the median of the lower half and Q3 is the median of the upper half.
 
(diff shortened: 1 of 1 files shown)


doc/python/box-plots.md
```diff
@@ -71,6 +71,49 @@ fig = px.box(df, x="time", y="total_bill", points="all")
 fig.show()
 ```
 
+### Choosing The Algorithm For Computing Quartiles
+
+By default, quartiles for box plots are computed using a linear algorithm method (see #10 listed on [http://www.amstat.org/publications/jse/v14n3/langford.html](http://www.amstat.org/publications/jse/v14n3/langford.html) for more details). However, you can also choose to use an `exclusive` or an `inclusive` algorithm to compute quartiles.
```
emmanuelle: By `linear algorithm method`, did you mean a method using linear interpolation between consecutive values? I'm not sure I understand.

doc/python/box-plots.md
```diff
@@ -138,6 +181,45 @@ fig = go.Figure(data=[go.Box(y=[0, 1, 1, 2, 3, 5, 8, 13, 21],
 fig.show()
 ```
 
+### Choosing The Algorithm For Computing Quartiles
+
```
emmanuelle: Can you link to the explanations given above, using the internal anchor link? (it would probably be "/python/box-plots/#difference-between-quartile-algorithms#)

doc/python/box-plots.md
emmanuelle: For a reference, you could also link to the wikipedia page https://en.wikipedia.org/wiki/Quartile which describes all three algorithms and is more likely to stay in the future. But the amstat page is useful for other algorithms so you could have both references.

emmanuelle:
I just left a couple of comments, it looks good otherwise, thanks!

doc/python/box-plots.md
jdamiba: I myself am not quite sure what makes the default algorithm `linear` in this context. Perhaps your suggestion is correct?

However, as `linear` is the attribute name used to refer to default algorithm, I felt that it was safe to refer to it as a `linear` algorithm in these docs.

emmanuelle: ok, so how about "using the `linear` method (for linear interpolation, see #10...)"?

doc/python/box-plots.md
jdamiba: Sure, added a link to the explanations above.

doc/python/box-plots.md
jdamiba: Sure, added a link to the wiki page.

jdamiba: How about this?

```
By default, quartiles for box plots are computed using the `linear` method (for more about linear interpolation, see #10 listed on [http://www.amstat.org/publications/jse/v14n3/langford.html](http://www.amstat.org/publications/jse/v14n3/langford.html) and [https://en.wikipedia.org/wiki/Quartile](https://en.wikipedia.org/wiki/Quartile) for more details). 
```

emmanuelle:
:dancer" thank you!

PR #2065, opened 2020-01-13
correct confusing indentation level

emmanuelle:
Cosmetic change, but the identation in https://plot.ly/python/3d-axes/#range-of-axes was confusing so I corrected this.


nicolaskruchten:
💃

PR #2069, opened 2020-01-14
Added missing dependency to make cell standalone

emmanuelle:


emmanuelle:
I think that when we click on an "edit in Github" link in the doc it branches off master and not doc-prod (although the PR is correctly against doc-prod). This, or I don't understand the origin of the percy changes :-).

nicolaskruchten:
I think the git-level stuff is good, but Percy does indeed compare to the latest master snapshots I think.

💃

emmanuelle:
ok, got it :-)

PR #2076, opened 2020-01-15
Fix for Python 3.10: don't compare to sys.version string

hugovk:
When dealing with version numbers, it's generally safer to use the `sys.version_info` tuple that the `sys.version` string.

```pycon
>>> sys.version
'3.7.5 (default, Nov  1 2019, 02:16:32) \n[Clang 11.0.0 (clang-1100.0.33.8)]'
>>> sys.version[:3]
'3.7'
>>> fake_version310_info = '3.10.5 (default, Nov  1 2019, 02:16:32) \n[Clang 11.0.0 (clang-1100.0.33.8)]'
>>> fake_version310_info[:3]
'3.1'
```

When Python 3.10 comes around (probably in [May 2020 when Python 3.9 reaches beta](https://www.python.org/dev/peps/pep-0596/#schedule)), it will claim to be on Python 3.1!

In this case, Python 3.2 is no longer supported, so this check can be removed. (And in fact, it wouldn't have failed until Python 3.20, but it's good to clean up.)

---

Found with https://github.com/asottile/flake8-2020

### DIFF
--- a/packages/python/plotly/_plotly_utils/utils.py
+++ b/packages/python/plotly/_plotly_utils/utils.py
@@ -227,7 +227,7 @@ def iso_to_plotly_time_string(iso_string):
 
 def template_doc(**names):
     def _decorator(func):
-        if sys.version[:3] != "3.2":
+        if not sys.version_info[:2] == (3, 2):
             if func.__doc__ is not None:
                 func.__doc__ = func.__doc__.format(**names)
         return func


emmanuelle:
Thank you @hugovk. I have no idea why we excluded python 3.2 here... any idea @nicolaskruchten @jonmmease ?

hugovk:
If it helps, it was added in PR https://github.com/plotly/plotly.py/pull/1476.

nicolaskruchten:
No idea, but probably for a good reason, in which case I'd rather upgrade the check to use the right thing rather than removing it :)

hugovk:
Updated:
```diff
-        if sys.version[:3] != "3.2":
+        if not sys.version_info[:2] == (3, 2):
```

nicolaskruchten:
Thanks! I know my feedback is coming across as very-conservative, and I appreciate the followup :)

packages/python/plotly/_plotly_utils/utils.py
```diff
@@ -227,7 +227,7 @@ def iso_to_plotly_time_string(iso_string):
 
 def template_doc(**names):
     def _decorator(func):
-        if sys.version[:3] != "3.2":
+        if not sys.version_info[:2] == (3, 2):
```
nicolaskruchten: should there be a `[:2]` in there? None of the changes in https://github.com/plotly/plotly.py/pull/2078/files include this index...

In fact, I think it might be nice to add this commit to https://github.com/plotly/plotly.py/pull/2078 and close this PR so we can merge it all at once, if you don't mind :)

hugovk: The `[:2]` is needed for `==` comparisons. It can be used but isn't required for `>`, `>=`, `<`, `<=` etc.

```pycon
>>> sys.version_info
sys.version_info(major=3, minor=8, micro=1, releaselevel='final', serial=0)
>>> sys.version_info[:2]
(3, 8)
>>> sys.version_info[:2] == (3, 8)
True
>>> sys.version_info == (3, 8)
False
>>> sys.version_info >= (3, 8)
True
>>> sys.version_info[:2] >= (3, 8)
True
>>> sys.version_info >= (3, 9)
False
>>> sys.version_info[:2] >= (3, 9)
False
```

---

Yep, will combine.

hugovk:
> Thanks! I know my feedback is coming across as very-conservative, and I appreciate the followup :)

No problem at all, it's good to be careful with widely used code!

By the way, how about (in another PR) adding `python_requires` to the `setup.py` files? That way modern pip (>=9) installs the right version of the package for user's running Python. For example, pip itself has: 

```python
    python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*',
```

Which allows install on Python 2.7 and 3.5+.

https://github.com/pypa/pip/blob/fed360a64e907f3ce8f5d497cc806df2fa9a0799/setup.py#L86

packages/python/plotly/_plotly_utils/utils.py
hugovk: > In fact, I think it might be nice to add this commit to #2078 and close this PR so we can merge it all at once, if you don't mind :)

Added, closing.

PR #2078, opened 2020-01-15
Fix for Python 4

hugovk:
Python 4 is some way off, but it may be a good idea to clean some of these up, similar to #2076.

---

This comparison is `True` for Python 3.3 - 3.99, but is `False` for Python 4 and would run the Python 2 - 3.2 code:

```python
if sys.version_info.major == 3 and sys.version_info.minor >= 3:	
    from unittest.mock import patch
else:
    from mock import patch	
```

Because Python 3.3 is the minimum Python 3.x supported, it can be simplified:

```diff
-if sys.version_info.major == 3 and sys.version_info.minor >= 3:	
+if sys.version_info.major >= 3:	
```

---

Another couple of cases which fail on Python 4.0, here we need to compare the `version_info` tuple and avoid comparing the `sys.version_info.minor` integer:

```diff
-if sys.version_info.major == 3 and sys.version_info.minor >= 4:
+if sys.version_info >= (3, 4):
```

```diff
-if sys.version_info.major != 3 or sys.version_info.minor < 6:
+if sys.version_info < (3, 6):
```

---

Finally, there's some code which uses `six.PY3`, a bit like:

```python
if six.PY3:
    print("Python 3+ code")
else:
    print "Python 2 code"
```

Where in six.py:

```python
PY3 = sys.version_info[0] == 3
```

When run on Python 4, this will run the Python 2 code!

Instead, use `six.PY2`.

---

Found with https://github.com/asottile/flake8-2020.

### DIFF
--- a/packages/python/plotly/_plotly_utils/utils.py
+++ b/packages/python/plotly/_plotly_utils/utils.py
@@ -7,7 +7,7 @@ from _plotly_utils.optional_imports import get_module
 from _plotly_utils.basevalidators import ImageUriValidator
 
 
-PY36_OR_LATER = sys.version_info.major == 3 and sys.version_info.minor >= 6
+PY36_OR_LATER = sys.version_info >= (3, 6)
 
 
 class PlotlyJSONEncoder(_json.JSONEncoder):
--- a/packages/python/plotly/_plotly_utils/utils.py
+++ b/packages/python/plotly/_plotly_utils/utils.py
@@ -227,7 +227,7 @@ def iso_to_plotly_time_string(iso_string):
 
 def template_doc(**names):
     def _decorator(func):
-        if sys.version[:3] != "3.2":
+        if not sys.version_info[:2] == (3, 2):
             if func.__doc__ is not None:
                 func.__doc__ = func.__doc__.format(**names)
         return func
--- a/packages/python/chart-studio/chart_studio/tests/test_plot_ly/test_api/__init__.py
+++ b/packages/python/chart-studio/chart_studio/tests/test_plot_ly/test_api/__init__.py
@@ -8,7 +8,7 @@ from chart_studio.tests.utils import PlotlyTestCase
 import sys
 
 # import from mock
-if sys.version_info.major == 3 and sys.version_info.minor >= 3:
+if sys.version_info >= (3, 3):
     from unittest.mock import patch
 else:
     from mock import patch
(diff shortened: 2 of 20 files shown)


nicolaskruchten:
Thanks for this! I don't understand the following reasoning though:

> Because Python 3.3 is the minimum Python 3.x supported, it can be simplified:

-if sys.version_info.major == 3 and sys.version_info.minor >= 3:	
+if sys.version_info.major >= 3:	

wouldn't this be a change in behaviour for 3.2?

hugovk:
Yes, it would be a change in behaviour for 3.2.

Does plotly.py still support 3.2? 

The classifiers at https://pypi.org/project/plotly/ suggest 2.7 and 3.3 - 3.7.

nicolaskruchten:
No, but I'd rather these changes be pure refactorings that don't change behaviour if possible, to avoid unintended side-effects :)

hugovk:
Sure thing, will update tomorrow!

nicolaskruchten:
Thanks very much for bringing this up and fixing it for us :)

hugovk:
You're welcome, and updated:

```diff
-if sys.version_info.major == 3 and sys.version_info.minor >= 3:
+if sys.version_info >= (3, 3):
```

packages/python/plotly/_plotly_utils/utils.py
```diff
@@ -7,9 +7,6 @@
 from _plotly_utils.basevalidators import ImageUriValidator
 
 
-PY36_OR_LATER = sys.version_info.major == 3 and sys.version_info.minor >= 6
```
nicolaskruchten: we should probably leave this and change it to the new pattern, in case someone had the bad idea of referring to it

nicolaskruchten:
LGTM... @jonmmease @emmanuelle thoughts?

packages/python/plotly/_plotly_utils/utils.py
hugovk: Good idea, updated in https://github.com/plotly/plotly.py/pull/2078/commits/0fd758010549b2f1cfbb51743bca3b7c2a2e0b78.

hugovk:
As suggested in https://github.com/plotly/plotly.py/pull/2076#discussion_r367441023, I've combined PR https://github.com/plotly/plotly.py/pull/2076 with this one.

nicolaskruchten:
OK nice so this basically mostly impacts tests, and all the tests are passing! We'll merge this in before the 4.5 release next week unless anyone brings up any objections by then. Thanks again!

emmanuelle:
Thank you very much!

hugovk:
You're welcome!

PR #2085, opened 2020-01-16
removing dash-example section

jdamiba:
The purpose of this PR is to remove the `dash example` section from these docs per https://github.com/plotly/documentation/issues/1654


emmanuelle:
Thanks @jdamiba ! :dancer:

PR #2087, opened 2020-01-16
modified trendline formatting

### LINKED ISSUE #1984: [bug] Suboptimal number formatting for trendline coefficients in plotly.express
In https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/express/_core.py#L239 we should use `%g` instead of `%f` so that enough significant digits are displayed even for very small numbers (for the slope in particular). Raised by https://community.plot.ly/t/how-to-extend-ols-trendline/32270

emmanuelle:
Closes #1984. 
For example 
```
import plotly.express as px

df = px.data.tips()
fig = px.scatter(df, x="total_bill", y=1.e-6 * df.tip, trendline="ols")
fig.show()
```
Master:
image

Now:
image

Existing examples should not be modified, I checked with the doc example of `trendline`.


nicolaskruchten:
if existing examples are not impacted, then this is a clear win :)

💃

PR #2088, opened 2020-01-16
fix deprecation warning in inspect

### LINKED ISSUE #2045: inspect.getargspec() is deprecated
[This line](https://github.com/plotly/plotly.py/blob/81f3922722672fbd6ce7ff50acfac8771c7c7ac2/packages/python/plotly/plotly/express/_doc.py#L503) triggers a `DeprecationWarning` on Python 3:
```
[...]/lib/python3.6/site-packages/plotly/express/_doc.py:503: DeprecationWarning:
inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
```
`getfullargspec` from the same `inspect` package is usually a [drop-in replacement on Python 3](https://docs.python.org/3/library/inspect.html#inspect.getargspec).

emmanuelle:
Closes #2045

### DIFF
--- a/packages/python/plotly/plotly/express/_doc.py
+++ b/packages/python/plotly/plotly/express/_doc.py
@@ -2,7 +2,7 @@ import inspect
 from textwrap import TextWrapper
 try:
     getfullargspec = inspect.getfullargspec
-except ImportError:
+except ImportError: # python 2
     getfullargspec = inspect.getargspec
 
 # TODO contents of columns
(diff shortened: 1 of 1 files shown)


emmanuelle:
The other `getargspec` in `io/_base_renderers.py` was already in a py2-only code block, so nothing else to change I think.

nicolaskruchten:
💃 thanks!

Issue #2080, opened 2020-01-15
Feature Request: Filled Polygon


peterdsharpe:
Hey all,

First off - thanks for all your hard work on this beautiful visualization library. I have a feature request that would be very handy in my scientific computing work, and I wanted to submit it to you for your consideration:

Matplotlib has a very useful command to draw a 2D filled polygon (matplotlib.pyplot.fill), exhibited here:
https://matplotlib.org/3.1.1/gallery/lines_bars_and_markers/fill.html

It would be very helpful to have an equivalent command available through Plotly. As far as I'm aware, this feature does not yet exist. The closest equivalent that I could find is an SVG Path, exhibited here:
https://plot.ly/python/shapes/#svg-paths

However, the input syntax is pretty clunky, especially if one wants to plot a filled polygon with many vertices defined from, say, a NumPy ndarray:
```python
        go.layout.Shape(
            type="path",
            path=" M 3,7 L2,8 L2,9 L3,10, L4,10 L5,9 L5,8 L4,7 Z",
            fillcolor="PaleTurquoise",
            line_color="LightSeaGreen",
        ),
```

A more streamlined interface, with syntax that lends itself to scientific computing (like matplotlib.pyplot.fill), would be greatly appreciated. Is this a feature the Plotly team would consider adding?

Thank you so much!
-Peter

empet:
@peterdsharpe It's not so awful to define a fiiled polygon with many points. You can use f-string concatenation:

```
n_points =50
polygon = f'M {x[0]}, {y[0]}'
for k in range(1, n_points):
    polygon += f' L{x[k]}, {y[k]}'
polygon += ' Z'

fig = go.Figure(go.Scatter())
fig.update_layout(width=600, height=600,
                  shapes=[dict(path =polygon,
                          fillcolor='RoyalBlue',
                          line_width=2,
                          line_color='RoyalBlue')])
```

nicolaskruchten:
You can also use a `scatter` trace with the `fill="toself"` option to create closed, filled shapes out of arrays. This approach gives you hover-labels which you can fill with useful information.

![image](https://user-images.githubusercontent.com/203523/72529605-a32bbd00-383b-11ea-9da5-7b82d10087b0.png)

nicolaskruchten:
A single trace can also encode many shapes if you interrupt the series with `None`:

![image](https://user-images.githubusercontent.com/203523/72529799-fe5daf80-383b-11ea-9fed-85465071c60e.png)

[...]

PR #2089, opened 2020-01-17
filled area

Mahdis-z:
closes https://github.com/plotly/plotly.py/issues/2080

### DIFF
--- a/doc/python/shapes.md
+++ b/doc/python/shapes.md
@@ -6,7 +6,7 @@ jupyter:
       extension: .md
       format_name: markdown
       format_version: '1.1'
-      jupytext_version: 1.1.1
+      jupytext_version: 1.2.1
   kernelspec:
     display_name: Python 3
     language: python
@@ -679,6 +679,26 @@ fig.update_layout(
     ]
 )
 
+fig.show()
+```
+
+### Filled Area Chart
+
+So far we saw [layout.shapes](https://plot.ly/python/reference/#layout-shapes-items-shape-type) defines the shape type to be drawn, which can be rectangle, circle, line, or path (a custom SVG path). You also can set an area to be filled with a solid color by defining [Scatter.fill="toself"](https://plot.ly/python/reference/#scatter-fill) that connects the endpoints of the trace into a closed shape. 
+
+```python
+import plotly.graph_objects as go
+
+fig = go.Figure(go.Scatter(x=[0,1,2], y=[0,2,0], fill="toself"))
+fig.show()
+```
+
+You can have more shapes either by adding [more traces](https://plot.ly/python/filled-area-plots/) or interupting the series with `None`.
+
+```python
+import plotly.graph_objects as go
+
+fig = go.Figure(go.Scatter(x=[0,1,2,None,3,3,5,5], y=[0,2,0,None,0.5,1.5,1.5,0.5], fill="toself"))
 fig.show()
 ```
 
--- a/doc/python/shapes.md
+++ b/doc/python/shapes.md
@@ -35,12 +35,12 @@ jupyter:
 
 ### Filled Area Chart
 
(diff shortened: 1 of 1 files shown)


doc/python/shapes.md
```diff
@@ -682,5 +682,25 @@ fig.update_layout(
 fig.show()
 ```
 
+### Filled Area Chart
```
nicolaskruchten: I would move this to the very top, with a little intro saying "there are two ways to draw filled shapes: scatter traces and layout.shapes" with a note that layout.shapes is mostly useful for cartesian 2d subplots, but you can use e.g. scatterpolar and scattergeo and scattermapbox to draw filled shapes on any kind of subplots.

nicolaskruchten: You can add a link to the scatterpolar page where we have filled shapes, and to the "filled shapes on mapbox" page that you wrote last year?

doc/python/shapes.md
```diff
@@ -33,6 +33,26 @@ jupyter:
...
 ---
 
+### Filled Area Chart
+
+There are two ways to draw filled shapes: scatter traces and [layout.shapes](https://plot.ly/python/reference/#layout-shapes-items-shape-type) which is mostly useful for the 2d subplots, and defines the shape type to be drawn, and can be rectangle, circle, line, or path (a custom SVG path). You also can use [scatterpolar](https://plot.ly/python/polar-chart/#categorical-polar-chart), scattergeo, [scattermapbox](https://plot.ly/python/filled-area-on-mapbox/#filled-scattermapbox-trace) to draw filled shapes on any kind of subplots. To set an area to be filled with a solid color, you need to define [Scatter.fill="toself"](https://plot.ly/python/reference/#scatter-fill) that connects the endpoints of the trace into a closed shape. 
+
+```python
+import plotly.graph_objects as go
+
+fig = go.Figure(go.Scatter(x=[0,1,2], y=[0,2,0], fill="toself"))
```
nicolaskruchten: can you add a final point that's the same as the initial point, so the line goes all the way around?

doc/python/shapes.md
```diff
@@ -33,6 +33,26 @@ jupyter:
...
+fig.show()
+```
+
+You can have more shapes either by adding [more traces](https://plot.ly/python/filled-area-plots/) or interrupting the series with `None`.
+
+```python
+import plotly.graph_objects as go
+
+fig = go.Figure(go.Scatter(x=[0,1,2,None,3,3,5,5], y=[0,2,0,None,0.5,1.5,1.5,0.5], fill="toself"))
```
nicolaskruchten: can you add a final point to each shape that's the same as the initial point, so the line goes all the way around?

doc/python/shapes.md
```diff
@@ -33,6 +33,26 @@ jupyter:
     thumbnail: thumbnail/shape.jpg
 ---
 
+### Filled Area Chart
+
+There are two ways to draw filled shapes: scatter traces and [layout.shapes](https://plot.ly/python/reference/#layout-shapes-items-shape-type) which is mostly useful for the 2d subplots, and defines the shape type to be drawn, and can be rectangle, circle, line, or path (a custom SVG path). You also can use [scatterpolar](https://plot.ly/python/polar-chart/#categorical-polar-chart), scattergeo, [scattermapbox](https://plot.ly/python/filled-area-on-mapbox/#filled-scattermapbox-trace) to draw filled shapes on any kind of subplots. To set an area to be filled with a solid color, you need to define [Scatter.fill="toself"](https://plot.ly/python/reference/#scatter-fill) that connects the endpoints of the trace into a closed shape.
```
nicolaskruchten: Maybe add a sentence saying that if `mode` contains `"lines"` (which is the default) then you'll likely want to repeat the first point of a shape at the end of the sequence, so the line goes all the way around.

Mahdis-z: Yeah to have a closed shape, all right!

nicolaskruchten:
💃 when the build is green

PR #2091, opened 2020-01-17
add a geojson dataset and link from election dataset

nicolaskruchten:
This PR adds a compressed GeoJSON file for choropleth demos. It has an `id` and something in `properties` to demo `featureidkey`. In order for this demo to work I had to add a column to the tabular `elections` dataset.


nicolaskruchten:
ah hang on, I committed a doc file I didn't mean to

nicolaskruchten:
ready to go!

emmanuelle:
:dancer: thanks!

PR #2103, opened 2020-01-20
Use plotly.min.js in both JS bundles

### LINKED ISSUE #1873: Effect on JupyterLab bundle size

nicolaskruchten:
Closes #1873 the opposite way from #1891 so as not to bloat the classic nbextension size. If we wanted to do things the #1891 way we would need to add minification in `plotlywidget/webpack.config.js` so that the nbextension stays small. I'm indifferent but this way seemed easy :) 

This PR also inlines the `randstr` function the same way, and removes the `prepublish: webpack` line that fails if webpack isn't globally installed, as well as the nonfunctional `build` target that calls `tsc`, replacing them with a `build: webpack` target that gets called explicitly in the CI step.

The bundle size savings of this change are substantial on a percentage basis: both bundles together are now roughly the size of either bundle, as expected, instead of twice. Also PS I resurrected the jupyterlab-chart-editor and it uses the `min` bundle right now so the savings should be compounded if all three are installed.

(closes #1891 if merged!)

### DIFF
--- a/packages/javascript/jupyterlab-plotly/package.json
+++ b/packages/javascript/jupyterlab-plotly/package.json
@@ -31,7 +31,7 @@
     "typescript": "~3.1.1"
   },
   "dependencies": {
-    "plotly.js": "^1.51.2",
+    "plotly.js": "^1.52.1",
     "@types/plotly.js": "^1.44.9",
     "@jupyterlab/rendermime-interfaces": "^1.3.0",
     "@phosphor/messaging": "^1.2.3",
--- a/packages/javascript/plotlywidget/package.json
+++ b/packages/javascript/plotlywidget/package.json
@@ -25,7 +25,6 @@
     "build": "npm run build:src",
     "build:src": "rimraf dist && tsc",
     "clean": "rimraf dist/ && rimraf  ../../python/plotly/plotlywidget/static'",
-    "prepublish": "webpack",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "devDependencies": {
--- a/packages/javascript/plotlywidget/package.json
+++ b/packages/javascript/plotlywidget/package.json
@@ -22,8 +22,7 @@
     "style/*.*"
   ],
   "scripts": {
-    "build": "npm run build:src",
-    "build:src": "rimraf dist && tsc",
+    "build": "webpack",
     "clean": "rimraf dist/ && rimraf  ../../python/plotly/plotlywidget/static'",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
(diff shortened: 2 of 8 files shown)


packages/javascript/plotlywidget/package.json
```diff
@@ -22,10 +22,8 @@
     "style/*.*"
   ],
   "scripts": {
-    "build": "npm run build:src",
-    "build:src": "rimraf dist && tsc",
+    "build": "webpack",
     "clean": "rimraf dist/ && rimraf  ../../python/plotly/plotlywidget/static'",
-    "prepublish": "webpack",
```
nicolaskruchten: this `prepublish` thing is deprecated. It seems like NPM tries to run it when doing `npm install` and fails unless you have `webpack` installed globally for some reason.

packages/javascript/plotlywidget/package.json
```diff
@@ -22,10 +22,8 @@
     "style/*.*"
   ],
   "scripts": {
-    "build": "npm run build:src",
-    "build:src": "rimraf dist && tsc",
```
nicolaskruchten: this isn't a TypeScript module so this doesn't seem to do much

packages/javascript/plotlywidget/package.json
```diff
@@ -22,10 +22,8 @@
...
+    "build": "webpack",
```
nicolaskruchten: The build for this package is, in fact webpack, and running `npm run build` does use the installed webpack in node_modules. So now we have to run `npm install && npm run build` and there are no deprecation warnings :)

packages/javascript/jupyterlab-plotly/package.json
```diff
@@ -31,7 +31,7 @@
     "typescript": "~3.1.1"
   },
   "dependencies": {
-    "plotly.js": "^1.51.2",
+    "plotly.js": "^1.52.1",
```
jonmmease: Did you intend to downgrade plotly.js here?

jonmmease:
One question above, but this looks good to me, assuming you've smoke tested the extensions in JupyterLab.

Thanks @nicolaskruchten!

packages/javascript/jupyterlab-plotly/package.json
nicolaskruchten: That’s an upgrade actually ;)

nicolaskruchten:
I’ve smoke tested by u installing, building and reinstalling both extensions... is that about right?

PR #2105, opened 2020-01-21
preload val_map from orders

### LINKED ISSUE #2084: Facet category order with missing data
This seems like a bug: https://community.plot.ly/t/python-plotly-express-facet-col-facet-row-order-categories/33693

nicolaskruchten:
Closes #2084 by populating `val_map` in a group-independent way. @emmanuelle  thoughts on the approach?

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1141,6 +1141,7 @@ def get_orderings(args, grouper, grouped):
     """
     orders = {} if "category_orders" not in args else args["category_orders"].copy()
     group_names = []
+    group_values = {}
     for group_name in grouped.groups:
         if len(grouper) == 1:
             group_name = (group_name,)
@@ -1154,6 +1155,7 @@ def get_orderings(args, grouper, grouped):
                     for val in uniques:
                         if val not in orders[col]:
                             orders[col].append(val)
+                group_values[col] = sorted(uniques, key=orders[col].index)
 
     for i, col in reversed(list(enumerate(grouper))):
         if col != one_group:
@@ -1162,7 +1164,7 @@ def get_orderings(args, grouper, grouped):
                 key=lambda g: orders[col].index(g[i]) if g[i] in orders[col] else -1,
             )
 
-    return orders, group_names
+    return orders, group_names, group_values
 
 
 def make_figure(args, constructor, trace_patch={}, layout_patch={}):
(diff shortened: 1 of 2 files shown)


packages/python/plotly/plotly/express/_core.py
```diff
@@ -1162,7 +1164,7 @@ def get_orderings(args, grouper, grouped):
                 key=lambda g: orders[col].index(g[i]) if g[i] in orders[col] else -1,
             )
 
-    return orders, group_names
+    return orders, group_names, group_values
```
emmanuelle: what is the difference between `orders` and `group_values`? In the examples I tried (including with facets and user-specified category orders) they were the same (the keys were not in the same order, but keys of dicts are not ordered I think?)

nicolaskruchten: `orders` includes all values given by the user even if they don't show up in the data

nicolaskruchten: whereas `group_values` has keys only per-group and has only values that show up in the data, so it's more specific. I'm pretty sure we need both, or at least I'm not comfortable trimming down `orders` at the moment :)

emmanuelle: Thanks. Would you mind adding a comment? The number of variables exceeds by working memory :-)

nicolaskruchten: pushed!

emmanuelle: thanks

packages/python/plotly/plotly/tests/test_core/test_px/test_px.py
```diff
@@ -182,3 +182,43 @@ def test_px_templates():
...
+                facet_col="day",
+                color="time",
+                symbol="day",
+                symbol_sequence=symbol_sequence,
+                color_discrete_sequence=color_sequence,
+                category_orders=dict(day=days, time=times),
+            )
+
+            for col in range(len(days)):
+                for trace in fig.select_traces(col=col + 1):
```
emmanuelle: oh great I didn't know `fig.select_traces` :-). Not sure it's documented anywhere.

packages/python/plotly/plotly/tests/test_core/test_px/test_px.py
```diff
@@ -182,3 +182,43 @@ def test_px_templates():
     assert fig.layout.xaxis3.showgrid is None
     assert fig.layout.yaxis2.showgrid
     assert fig.layout.yaxis3.showgrid
+
+
+def test_orthogonal_orderings():
```
emmanuelle: great test!

emmanuelle: Is it worth testing that adding in `category_orders` a value not present in the data does not break things?

emmanuelle:
The approach (defining col and row numbers in advance) looks sound to me, and the test is very useful. :dancer:

packages/python/plotly/plotly/tests/test_core/test_px/test_px.py
nicolaskruchten: yeah I tried but it's a harder test to write and much slower :)

nicolaskruchten: OK I think I have a way to spot-check a few options

PR #2132, opened 2020-01-27
changed order of imports for better error messages

### LINKED ISSUE #2125: plotly.express import raises ModuleNotFound in environment without pandas.
Importing plotly.express when pandas is not available raises `ModuleNotFoundError: No module named 'pandas'`, instead of the intended `ImportError: Plotly express requires pandas to be installed.`
This happens on `from ._imshow import imshow`.
Perhaps this import should be moved below the code that will output a more helpful message?

emmanuelle:
Closes #2125 (check if pandas if installed first for a meaningful error message)


nicolaskruchten:
💃

PR #2127, opened 2020-01-24
Px bool

emmanuelle:
Closes #2121

PR #2133, opened 2020-01-27
Sunburst improvements

emmanuelle:


nicolaskruchten:
:dancer:

maybe let's merge this and then rebase https://github.com/plotly/plotly.py/pull/2127 onto it and see if there's a fix?

PR #2136, opened 2020-01-29
Actually display the tidy table in example

havrikov:


emmanuelle:
Good idea, thanks @havrikov. :dancer:, merging.

PR #2155, opened 2020-02-03
add the word background to bgcolor description so it will get indexed by algolia

jdamiba:
The purpose of this PR is to make sure that `plot_bgcolor` and `paper_bgcolor` show up in algolia search results for `background`. 

Related to https://github.com/plotly/plotly.js/pull/4536

### DIFF
--- a/packages/python/plotly/codegen/resources/plot-schema.json
+++ b/packages/python/plotly/codegen/resources/plot-schema.json
@@ -57324,14 +57324,14 @@
                 "role": "style",
                 "dflt": "#fff",
                 "editType": "plot",
-                "description": "Sets the color of paper where the graph is drawn."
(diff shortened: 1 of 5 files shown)


nicolaskruchten:
Thanks for this PR! It's not actually necessary, though, as this code is automatically-generated by the release process, and only deployed to the docs when `master` is merged to `doc-prod` thereafter.

emmanuelle:
@nicolaskruchten you mean, because the JS doc has already been updated?

nicolaskruchten:
Oh, no, I meant this change will come in automatically when we codegen for 1.52.2, and then we'll deploy to prod. We don't need to make these changes piecemeal here in Python when we do them in JS.

PR #2156, opened 2020-02-03
finally: how to style px

nicolaskruchten:


doc/python/styling-plotly-express.md
```diff
@@ -0,0 +1,183 @@
...
+    nbconvert_exporter: python
+    pygments_lexer: ipython3
+    version: 3.6.8
+  plotly:
+    description: Plotly Express is a terse, consistent, high-level API for rapid data
+      exploration and figure generation.
+    display_as: file_settings
+    language: python
+    layout: base
+    name: Plotly Express
```
emmanuelle: Styling Plotly Express?

emmanuelle:
It looks great! The only thing which is missing maybe is the case where you cannot call one of the `update_x` methods, for example when you want to change the duration of frames in an animation etc. It might be a good idea to link here to the structure of a figure (https://plot.ly/python/creating-and-updating-figures/), say that doing `print(fig)` can be a good idea, and that it's also possible to modify manually the contents of the figure. Or does it make the message too complicated?

emmanuelle:
but it can also be merged as it is, so :dancer: anyway

nicolaskruchten:
Yeah I should definitely link to that page... I toyed with printing a figure and saying "here are all the things you can change" but it did dilute the message a fair bit. The animation thing is a bit of a special case in how deep the edit goes ;)

doc/python/styling-plotly-express.md
nicolaskruchten: ah yeah I forgot to actually edit the front matter here, it's a straight copy.

PR #2161, opened 2020-02-04
shapes on subplots

Mahdis-z:


doc/python/shapes.md
```diff
@@ -638,6 +638,34 @@ fig.update_layout(
...
 ```
 
+#### Adding Shapes to Subplots
+
+```python
+from plotly.subplots import make_subplots
+
+# Create Subplots
+fig = make_subplots(rows=2, cols=2, start_cell='bottom-left')
```
emmanuelle: So why do we need to change the `start_cell` here? I would need an explanation, so probably the tutorial too ;-)

doc/python/shapes.md
```diff
@@ -638,6 +638,34 @@ fig.update_layout(
...
+fig.add_trace(go.Scatter(x=[2, 6], y=[1,1]), row=1, col=1)
+fig.add_trace(go.Bar(x=[1,2,3], y=[4,5,6]), row=1, col=2)
+fig.add_trace(go.Scatter(x=[10,20], y=[40,50]), row=2, col=1)
+fig.add_trace(go.Bar(x=[11,13,15], y=[8,11,20]), row=2, col=2)
+
+# Add shapes
+fig.update_layout(
+    shapes=[
+        dict(type="line", xref="x1", yref="y1",
+            x0=3, y0=0.5, x1=5, y1=0.8),
```
emmanuelle:
```suggestion
            x0=3, y0=0.5, x1=5, y1=0.8, line_width=3),
```

emmanuelle: the line width was too small, I did not see the line and did not understand why this one was not there :-)

doc/python/shapes.md
emmanuelle: It looked good to me also without changing the `start_cell`, so I'd suggest to limit the complexity here unless of course I am missing something.

doc/python/shapes.md
```diff
@@ -638,6 +638,34 @@ fig.update_layout(
 fig.show()
 ```
 
+#### Adding Shapes to Subplots
+
```
emmanuelle:
```suggestion
 
 Here we use the different axes (`x1`, `x2`) created by `make_subplots` as reference in order to draw shapes in figure subplots.
```

emmanuelle:
Thank you @Mahdis-z ! Just a couple of suggestions and then we can merge :-).

Mahdis-z:
> Thank you @Mahdis-z ! Just a couple of suggestions and then we can merge :-).

Thanks for the thorough review @emmanuelle

doc/python/shapes.md
emmanuelle: why didn't you include this suggestion? Was something wrong with it?

doc/python/shapes.md
```diff
@@ -638,6 +638,34 @@ fig.update_layout(
 fig.show()
 ```
 
+#### Adding Shapes to Subplots
+Here we use the different axes (`x1`, `x2`) created by `make_subplots` as reference in order to draw shapes in figure subplots.
+```python
```
emmanuelle: maybe you need a blank line before the python cell here, not sure (please check)

emmanuelle:
thank you :dancer: !

doc/python/shapes.md
Mahdis-z: would try, not sure why CI fails

nicolaskruchten:
The orca failure here should resolve itself soon when pytest upgrades, or if we downgrade pytest..

emmanuelle:
This one is good to go @Mahdis-z you can merge it (the orca failure has already been fixed in another merged commit).

PR #2166, opened 2020-02-06
xarray support in imshow

emmanuelle:
I'm planning to complete the docstring and add examples in the tutorials, but I would like first to discuss whether this xarray support should rather go here in imshow or/and in a `px.matrix_heatmap` / `px.heatmap` function which is yet to be written (but could be adapted from the imshow code). @nicolaskruchten interested in your thoughts here :-).


emmanuelle:
Oh and I will fix the CI too of course (add xarray to CI deps). My hesitation here is in particular for arrays generated by datashader: since their dimensions are sometimes heterogeneous, it does not always make sense to plot them with square pixels.

nicolaskruchten:
This looks great, but per semver rules belongs in 4.6 :)

emmanuelle:
no worries :-). Anything to change ?

emmanuelle:
Now that the patch release is out, I think this one is ready for review :-).

doc/python/imshow.md
```diff
@@ -110,6 +110,34 @@ fig.update_layout(coloraxis_showscale=False)
...
+### Display an xarray image with px.imshow
+
+[xarrays](http://xarray.pydata.org/en/stable/) are labeled arrays (with labeled axes and coordinates). If you pass an xarray image to `px.imshow`, its axes labels and coordinates will be used for ticks. (If you don't want this behavior, just pass `img.values` which is a NumPy array if `img` is an xarray).
+
+```python
+import plotly.express as px
+import xarray as xr
+# Load xarray from dataset included in the xarray tutorial
+# We remove 273.5 to display Celsius degrees instead of Kelvin degrees
+airtemps = xr.tutorial.open_dataset('air_temperature').air.isel(lon=20) - 273.5
```
nicolaskruchten: would be cleaner to do `xr.tutorial.open_dataset('air_temperature').air.sel(lon=250.0)` no? that way we're doing a data-space selection (lon == 250.0 rather than 20th lon value) and we're not subtracting anything, so the underlying `long_name` is preserved... If we don't subtract, the attrs stick around and look like this:

```
<xarray.DataArray 'air' (time: 2920, lat: 25)>
array([[246.5    , 242.     , 240.39   , ..., 295.5    , 297.     , 298.29   ],
       [244.     , 241.59999, 240.59999, ..., 296.     , 297.9    , 299.19998],
       [242.2    , 239.59999, 239.89   , ..., 296.79   , 297.9    , 298.69998],
       ...,
       [246.68999, 246.89   , 248.89   , ..., 296.29   , 297.99   , 298.88998],
       [244.39   , 244.59   , 245.59   , ..., 296.99   , 298.19   , 299.49   ],
       [242.79   , 244.39   , 246.09   , ..., 296.09   , 297.99   , 299.19   ]],
      dtype=float32)
Coordinates:
  * lat      (lat) float32 75.0 72.5 70.0 67.5 65.0 ... 25.0 22.5 20.0 17.5 15.0
    lon      float32 250.0
  * time     (time) datetime64[ns] 2013-01-01 ... 2014-12-31T18:00:00
Attributes:
    long_name:     4xDaily Air temperature at sigma level 995
    units:         degK
    precision:     2
    GRIB_id:       11
    GRIB_name:     TMP
    var_desc:      Air temperature
    dataset:       NMC Reanalysis
    level_desc:    Surface
    statistic:     Individual Obs
    parent_stat:   Other
    actual_range:  [185.16 322.1 ]
```

nicolaskruchten: not sure which of these is "standard" for us to pick up on for additional labelling... the `actual_range` looks juicy for color_range ?

packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -137,11 +153,33 @@ def imshow(
...
+            # np.datetime64 is not handled correctly by go.Heatmap
+            for ax in [x_label, y_label]:
+                if np.issubdtype(img.coords[ax].dtype, np.datetime64):
+                    img.coords[ax] = img.coords[ax].astype(str)
+            x = img.coords[x_label]
+            y = img.coords[y_label]
+            img_is_xarray = True
+            if aspect is None:
+                aspect = "auto"
+            z_name = img.attrs["long_name"] if "long_name" in img.attrs else "z"
```
nicolaskruchten: someday maybe we allow this to be overridden somehow? maybe someone likes their `long_name` the way it is but wants to have something else? Like in the `air` example above, the built-in `long_name` is `4xDaily Air temperature at sigma level 995` but the `var_desc` is `Air Temperature` ...

nicolaskruchten: OK I see, this is what they use in http://xarray.pydata.org/en/stable/plotting.html ... works for me for now I think! maybe we can eventually use `labels` to override this

doc/python/imshow.md
nicolaskruchten: ah in the xarray docs they do this:

```
# Convert to celsius
In [7]: air = airtemps.air - 273.15

# copy attributes to get nice figure labels and change Kelvin to Celsius
In [8]: air.attrs = airtemps.air.attrs

In [9]: air.attrs['units'] = 'deg C'
```

packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -137,11 +153,33 @@ def imshow(
...
+
+    If an xarray is passed, dimensions names and coordinates are used for
+    axes labels and ticks.
     """
     args = locals()
     apply_default_cascade(args)
+    img_is_xarray = False
+    if xarray_imported:
+        if isinstance(img, xarray.DataArray):
+            y_label, x_label = img.dims[0], img.dims[1]
```
nicolaskruchten: should we include units like they do in xarray?

emmanuelle: I decided not to because the logics start to be complicated, especially with `labels` (should we add the units also when the label was overriden with `labels`?).

nicolaskruchten: I think we should have the default label include the units, but then let the user override the whole thing. Otherwise we'll get users saying "but the built-in matplotlib version does all this stuff"

packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -95,6 +104,14 @@ def imshow(
         position of the [0, 0] pixel of the image array, in the upper left or lower left
         corner. The convention 'upper' is typically used for matrices and images.
 
+    labels : dict with str keys and str values (default `{}`)
+        Overrides names used in the figure for axis titles (keys ``x`` and ``y``),
+        colorbar title (key ``colorbar``) and hover (key ``color``). The values
```
nicolaskruchten: to match other PX functions there should be no `colorbar` and `color` should be used there....

emmanuelle: I liked the idea of having the possibility of having different labels for the hover and the colorbar title...

emmanuelle: it can still be overriden of course but not in px land.

packages/python/plotly/plotly/express/_imshow.py
emmanuelle: ok, I was reluctant to code too much xarray-specific stuff but then I figured out that we could just use their function for this :-).

packages/python/plotly/plotly/express/_imshow.py
emmanuelle: ok, I compromised here :-)

nicolaskruchten: thank you :)

doc/python/datashader.md
```diff
@@ -98,32 +98,18 @@ fig.show()
...
-
-fig = go.Figure(go.Heatmap(
-    z=np.log10(agg), x=x, y=y,
-    hoverongaps=False,
-    hovertemplate='Scheduled departure: %{x:.1f}h <br>Depature delay: %{y} <br>Log10(Count): %{z}',
-    colorbar=dict(title='Count (Log)', tickprefix='1.e')))
-fig.update_xaxes(title_text='Scheduled departure')
-fig.update_yaxes(title_text='Departure delay')
+agg.values = np.log10(agg.values)
+agg.attrs['long_name'] = 'Log10(count)'
```
nicolaskruchten: both this line and the `coloraxis_colorbar` line two below could be removed and replaced with a `labels` arg now I think and in fact we could add `x` and `y` as well to get axis titles and hover labels :)

packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -137,11 +161,42 @@ def imshow(
...
+            x = img.coords[x_label]
+            y = img.coords[y_label]
+            img_is_xarray = True
+            if aspect is None:
+                aspect = "auto"
+            z_name = xarray.plot.utils.label_from_attrs(img).replace("\n", "<br>")
+
+    if labels is not None:
+        if "x" in labels:
+            y_label = labels["x"]
```
nicolaskruchten:
```suggestion
            x_label = labels["x"]
```

packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -185,12 +241,25 @@ def imshow(
...
-        if args[v]:
-            layout_patch[v] = args[v]
+    for attr_name in ["title", "height", "width"]:
+        if args[attr_name]:
+            layout_patch[attr_name] = args[attr_name]
     if "title" not in layout_patch and args["template"].layout.margin.t is None:
         layout_patch["margin"] = {"t": 60}
     fig = go.Figure(data=trace, layout=layout)
     fig.update_layout(layout_patch)
+    if img_is_xarray:
```
nicolaskruchten: I actually think we can do this for `np` inputs also, no, given the `labels` array? We can just default `x_label` to `x` etc

nicolaskruchten:
@emmanuelle LMK what you think of my changes... missing additional tests, docstring and docs but I will add them if you agree.

Basically what I've done is generalized `x` and `y` a little bit so that a user can pass them in even with a normal `numpy` array for `img`. In essence, an alternative way to look at this PR with my changes is that it first adds user-controllable `x`, `y` and `labels` to `imshow()`, and then adds special handling for `xarray`s to prepopulate those values if unspecified ;)

emmanuelle:
Thank you this looks great ! Looks like you added the tests already, :dancer: !

Issue #2168, opened 2020-02-06
px.defaults too permissive


nicolaskruchten:
`px.defaults.whatever = "whatever"` should raise an exception: only the specific keys in the object should be settable.

PR #2183, opened 2020-02-11
prevent the creation of new px.default properties at run time

jdamiba:
closes https://github.com/plotly/plotly.py/issues/2168

The purpose of this PR is to prevent the creation of new px.default properties at run time.

Testing Plan:

Prior to this PR, the following code does not throw an exception:

```python
import plotly.express as px

px.defaults.template = "ggplot2"
px.defaults.color_continuous_scale = px.colors.sequential.Blackbody
px.defaults.width = 600
px.defaults.height = 400
px.defaults.should_not_work = "test"

df = px.data.iris()
fig = px.scatter(df, x="sepal_width", y="sepal_length", color="sepal_length", width=400)
fig.show()
```

When running this PR, the above code throws:

```
  File "app.py", line 6, in <module>
    px.defaults.should_not_work = "test"
AttributeError: 'PxDefaults' object has no attribute 'should_not_work'
```


emmanuelle:
Very elegant solution! I didn't know `__slots__` before :tada:. Could you also please add a test in https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/tests/test_core/test_px/test_px.py which would check that an AttributeError is raised? See for example https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/tests/test_core/test_px/test_px_functions.py#L159 for a typical pattern.

jdamiba:
@emmanuelle 

I have added a test- thanks for the reminder. I will try to proactively add them where relevant going forward :)

packages/python/plotly/plotly/tests/test_core/test_px/test_px.py
```diff
@@ -232,3 +233,9 @@ def assert_orderings(days_order, days_check, times_order, times_check):
     for days in permutations(df["day"].unique()):
         for times in permutations(df["time"].unique()):
             assert_orderings(days, days, times, times)
+
+
+def check_permissive_defaults():
```
emmanuelle: I think your function should be called `test_something` so that pytest catches it in the test suite (otherwise it will consider it a helper function). You can verify this by executing locally pytest in your machine (for example in the same directory, `pytest -v test_px.py` and check whether the new function appears). Maybe I'm wrong but please check :-).

jdamiba: You're right about tests needing to start with `test` in order to be picked up by `pytest`.

After modifying to above code, the output of running `pytest -v test_px.py` is:

```
test_px.py::test_scatter PASSED
test_px.py::test_custom_data_scatter PASSED
test_px.py::test_labels PASSED
test_px.py::test_px_templates PASSED
test_px.py::test_orthogonal_orderings PASSED
test_px.py::test_permissive_defaults PASSED

emmanuelle:
Thanks @jdamiba ! Just wait for the CI to finish (everything should be green except for one of the orca builds which failure is not related to this PR), and then :dancer: !

packages/python/plotly/plotly/tests/test_core/test_px/test_px.py
emmanuelle: awesome!

Issue #2177, opened 2020-02-10
Better error message if passed in a column name instead of a list of column names into `custom_data`


chriddyp:
`custom_data` is unlike most other `px` properties in that it accepts a list of columns instead of just a single column.

It would be helpful if we said something like "Looks like you supplied a single column name, you should supply a list of columns."

Here's the current error: 
> ValueError: Value of 'custom_data_0' is not the name of a column in 'data_frame'. Expected one of ['uid', 'from_uid', 'dropout', 'embedding_size', 'epoch', 'lr', 'train f1', 'valid ppl'] but received: 19751736_1

emmanuelle:
Good idea. Same for `hover_data` and maybe a couple of others. We'll have to find a clever workaround though to see at which level to catch the error.

chriddyp:
It would also be helpful if we noted _which_ attribute had an error. So, for the error message above, saying something like "You supplied a single column name _to the property custom_data_. This property only accepts a list of column names. Try supplying custom_data=["your-column"] instead of custom_data="your-column""

wornbb:
Do you think it might be better to automatically convert single column data "your-column" to ["your-column"] instead of throwing an error?

wornbb:
If we let it convert "your-column" to ["your-column"], will there be any consequence?

nicolaskruchten:
I'd rather we provide a helpful error message for all the `array_attrables` if what's provided for them is not list-list :)

PR #2185, opened 2020-02-12
a bandage solution to issue #2177

wornbb:
A naive attempt to fix issue #2177.
When the field is "custom_data", it ensures the input is a list.

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -928,6 +928,10 @@ def build_dataframe(args, attrables, array_attrables):
             length = len(df_output)
             if argument is None:
                 continue
+            # Ensure the field of "custom_data" is always list
+            if argument == "custom_data":
+                if not isinstance(field, list):
+                    field = [field]
             # Case of multiindex
             if isinstance(argument, pd.MultiIndex):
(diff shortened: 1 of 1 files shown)


emmanuelle:
@wornbb thank you for your contribution. There are pros and cons about this kind of "under-the-hood magic" fixes, so I'm a bit hesitant here. Let's wait maybe for other devs to comment? But we already do this in some parts of `plotly.py`, like for the `data` argument of a `go.Figure`. 

At the moment we accept several kinds of iterables for `custom_data` (and `hover_data`, so if decide to go for your solution for `hover_data`, we would need the same for `hover_data`), like tuple, numpy arrays etc. For a tuple we also need to keep the argument as it is.

One tricky corner case is what to do with `custom_data=np.arange([1, 2, 3])`, since it could correspond either to column names (for a dataframe created from a numpy array) or to the `custom_data` themselves... Passing column names inside a numpy array is a really weird syntax but it works as of now, basically we can take everything which can be indexed. 

Another solution can be to convert only when the type is str of pandas series. 

To be discussed together! Thank you again @wornbb

emmanuelle:
@nicolaskruchten interested in your opinion here when you have the time :-).

PR #2195, opened 2020-02-14
Minor correction

ashwin-nayak:
Name of the window around "ideal" line plot should be "ideal" but is instead marked as "fair".


emmanuelle:
Indeed! Thank you very much @ashwin-nayak , :dancer:!

PR #2201, opened 2020-02-14
Updates to examples in fundamentals section

jdamiba:
The purpose of this PR is to update 3 of our examples.

### DIFF
--- a/doc/python/creating-and-updating-figures.md
+++ b/doc/python/creating-and-updating-figures.md
@@ -39,11 +39,11 @@ jupyter:
 
 ### Representing Figures
 
-The goal of the `plotly.py` package is to provide a pleasant Python interface for creating figure specifications which are displayed by the [`plotly.js`](https://plot.ly/javascript) JavaScript graphing library. 
+The goal of the plotly.py package is to provide a pleasant Python interface for creating figure specifications which are displayed by the [plotly.js](https://plot.ly/javascript) JavaScript graphing library. 
 
-In the context of the `plotly.js` library, a figure is specified by a declarative [JSON](https://www.json.org/json-en.html) data structure. 
+In the context of the plotly.js library, a figure is specified by a declarative [JSON](https://www.json.org/json-en.html) data structure. 
 
-Therefore, you should always keep in mind as you are creating and updating figures using the `plotly.py` package that its ultimate goal is to help users produce Python [dictionaries](https://docs.python.org/3/tutorial/datastructures.html#dictionaries) that can be automatically [serialized](https://en.wikipedia.org/wiki/Serialization) into the JSON data structure that the `plotly.js` graphing library understands.
+Therefore, you should always keep in mind as you are creating and updating figures using the plotly.py package that its ultimate goal is to help users produce Python [dictionaries](https://docs.python.org/3/tutorial/datastructures.html#dictionaries) that can be automatically [serialized](https://en.wikipedia.org/wiki/Serialization) into the JSON data structure that the plotly.js graphing library understands.
 
 #### Figures As Dictionaries
 
@@ -612,7 +612,7 @@ There are also `for_each_xaxis()` and `for_each_yaxis()` methods that are analog
 
 ### Other Update Methods
 
-Figures created with the `plotly.graph_objects` module also support:
+Figures created with the plotly.py graphing library also support:
     - the `update_layout_images()` method in order to [update background layout images](/python/images/), 
     - `update_annotations()` in order to [update annotations](/python/text-and-annotations/#multiple-annotations), 
     - and `update-shapes()` in order to [update shapes](/python/shapes/).
--- a/doc/python/creating-and-updating-figures.md
+++ b/doc/python/creating-and-updating-figures.md
@@ -5,8 +5,8 @@ jupyter:
     text_representation:
       extension: .md
       format_name: markdown
-      format_version: "1.1"
-      jupytext_version: 1.1.1
+      format_version: '1.2'
+      jupytext_version: 1.3.2
   kernelspec:
     display_name: Python 3
     language: python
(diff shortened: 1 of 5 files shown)


doc/python/renderers.md
```diff
@@ -223,12 +226,28 @@ fig = go.Figure(
...
-Dash is a Python framework for building web applications, and it provides built-in support for displaying Plotly figures. See the [Dash User Guide](https://dash.plot.ly/) for more information.
+### Displaying Figures Using Dash
+[Dash](https://dash.plot.ly) is a Python framework for building web applications, and it provides built-in support for displaying figures created with Plotly's graphing libraries. See the [Dash User Guide](https://dash.plot.ly/) for more information.
+
+It is important to note that Dash does not use the renderers framework discussed above, so you should not use the `show()` figure method or the `plotly.io.show` function inside Dash applications.
+
+Instead, pass your figure as the `figure` parameter to the `dcc.Graph` component. 
+
+```python
+import dash_core_components as dcc
```
emmanuelle: You need to add `dash_core_components` to `doc/requirements.txt` and `binder/requirements.txt` (the CI fails at the moment).

doc/python/renderers.md
```diff
@@ -223,12 +226,28 @@ fig = go.Figure(
...
-### Displaying figures using Dash
-Dash is a Python framework for building web applications, and it provides built-in support for displaying Plotly figures. See the [Dash User Guide](https://dash.plot.ly/) for more information.
+### Displaying Figures Using Dash
+[Dash](https://dash.plot.ly) is a Python framework for building web applications, and it provides built-in support for displaying figures created with Plotly's graphing libraries. See the [Dash User Guide](https://dash.plot.ly/) for more information.
+
+It is important to note that Dash does not use the renderers framework discussed above, so you should not use the `show()` figure method or the `plotly.io.show` function inside Dash applications.
+
+Instead, pass your figure as the `figure` parameter to the `dcc.Graph` component. 
+
+```python
```
nicolaskruchten: does this actually work here? I would think you'd need to escape this block somehow so it doesn't actually execute :)

doc/python/subplots.md
```diff
@@ -37,9 +37,11 @@ jupyter:
 
 #### Simple Subplot
 
-Figures with subplots are created using the `make_subplots` function from the `plotly.subplots` module.
+Figures with subplots are created using the `make_subplots()` function from the `plotly.subplots` module. 
 
-Here is an example of creating a figure with two scatter traces in side-by-side subplots.
+This function accepts two integer parameters (`rows` and `cols`), which specifiy how the subplots should be laid out.
```
nicolaskruchten: this is probably misleading, as this function accepts many other parameters as well...

doc/python/subplots.md
```diff
@@ -203,9 +205,9 @@ fig.show()
 ```
 
 #### Customizing Subplot Axes
-After a figure with subplots is created using the `make_subplots` function, its axis properties (title, font, range, grid style, etc.) can be customized using the `update_xaxes` and `update_yaxes` graph object figure methods.  By default, these methods apply to all of the x axes or y axes in the figure. The `row` and `col` arguments can be used to control which axes are targeted by the update.
+After a figure with subplots is created using the `make_subplots()` function, its axis properties (title, font, range, grid style, etc.) can be customized using the `update_xaxes` and `update_yaxes` graph object figure methods.  By default, these methods apply to all of the `x` axes or `y` axes in the figure. The `row` and `col` arguments can be used to control which axes are targeted by the update.
```
nicolaskruchten: why put `x` and `y` in backticks here?

doc/python/subplots.md
```diff
@@ -241,17 +243,17 @@ fig.show()
...
+Here is an example that creates a figure with 3 vertically stacked subplots with linked `x` axes. A small vertical spacing value is used to reduce the spacing between subplot rows.
 
 ```python
 
-fig = make_subplots(
-    rows=3, cols=1, shared_xaxes=True, vertical_spacing=0.02
-)
+fig = make_subplots(rows=3, cols=1,
```
nicolaskruchten: this feels like a one-liner to me

doc/python/subplots.md
```diff
@@ -354,7 +355,7 @@ fig.show()
 ```
 
 #### Subplots Types
-By default, the `make_subplots` function assumes that the traces that will be added to all subplots are 2-dimensional cartesian traces (e.g. `scatter`, `bar`, `histogram`, `violin`, etc.).  Traces with other subplot types (e.g. `scatterpolar`, `scattergeo`, `parcoords`, etc.) are supporteed by specifying the `type` subplot option in the `specs` argument to `make_subplots`.
+By default, the `make_subplots()` function assumes that the traces that will be added to all subplots are 2-dimensional cartesian traces (e.g. `scatter`, `bar`, `histogram`, `violin`, etc.).  Traces with other subplot types (e.g. `scatterpolar`, `scattergeo`, `parcoords`, etc.) are supporteed by specifying the `type` subplot option in the `specs` argument to `make_subplots`.
```
nicolaskruchten: in general, I don't really agree with putting the `()` after the function name everywhere (this comment applies to this whole PR btw). The function is properly named `make_subplots` and calling it with no parameters like `make_subplots()` will probably result in an error. It's probably OK to use this for something like "the `.show()` method" as that actually will work with no arguments.

Also note: maybe we haven't used it super consistently, but the leading `.` in "the `.show()` method" is actually informative IMO in that it reminds the reader that `show` isn't a free function but a method of a figure. As a matter of style we could say "the `fig.show` method" or "the `.show` method of every `go.Figure` object" or something.

In any case, unless you're going to fully standardize the usage of this kind of thing throughout the docs, I find it pretty distracting and hard to review changes like this :)

doc/python/renderers.md
```diff
@@ -171,18 +175,17 @@ fig.show(renderer="png")
 ```
 
 ###### `pdf`
-This renderer displays figures as static PDF files. This is especially useful for notebooks that will be exported to PDF files using the LaTeX export capabilities of nbconvert.
+This renderer displays figures as static `.pdf` files. This is especially useful for notebooks that will be exported to `.pdf` files using the `LaTeX` export capabilities of `nbconvert`.
```
nicolaskruchten: here the usage of backticks is questionable also... "static PDF files" is acceptable, no? It's not clear why "static `.pdf` files" is clearer or better :)

doc/python/renderers.md
```diff
@@ -106,23 +110,23 @@ fig = go.Figure(
...
 
-This renderer is a good choice for notebooks that will be exported to HTML files (Either using [nbconvert](https://nbconvert.readthedocs.io/en/latest/) or the "Download as HTML" menu action) because the exported HTML files will work without an internet connection.
+This renderer is a good choice for notebooks that will be exported to HTML files (Either using [nbconvert](https://nbconvert.readthedocs.io/en/latest/) or the "Download as HTML" menu action) because the exported HTML files will work without an Internet connection.
 
-> Note: Adding the Plotly.js bundle to the notebook does add a few megabytes to the notebook size, so if you can count on having an internet connection you may want to consider the `notebook_connected` renderer.
+> Note: Adding the `plotly.js` bundle to the notebook adds a few megabytes to the notebook size. If you can count on always having an Internet connection, you may want to consider using the `notebook_connected` renderer if notebook size is a constraint.
```
nicolaskruchten: I don't think we should backtick "plotly.js" ...

doc/python/renderers.md
```diff
@@ -33,19 +33,21 @@ jupyter:
...
-### Displaying plotly figures
-This section covers the many ways to display plotly figures from Python.  At the highest level, there are three general approaches:
+Plotly's Python graphing library, `plotly.py`, gives you a wide range of options for how and where to display your figures. 
 
- 1. Using the renderers framework in the context of a script or notebook
- 2. Using Dash in a web app context
- 3. Using a `FigureWidget` in an ipywidgets context
+In general, there are three different approaches you can take in order to display figures:
+
+ 1. Using the `renderers` framework in the context of a script or notebook
```
nicolaskruchten: Here also I think the usage of backticks for `renderers` is going a bit too far IMO...

doc/python/creating-and-updating-figures.md
```diff
@@ -487,34 +559,48 @@ fig.for_each_trace(
...
 ```
 
-There are also `for_each_xaxis` and `for_each_yaxis` methods that are analogous to the `for_each_trace` method described above. For non-cartesian subplot types (e.g. polar), there are additional `update_{type}` and `for_each_{type}` methods (e.g. `update_polar`, `for_each_polar`).
+There are also `for_each_xaxis()` and `for_each_yaxis()` methods that are analogous to the `for_each_trace()` method described above. For non-cartesian subplot types (e.g. polar), there are additional `update_{type}` and `for_each_{type}` methods (e.g. `update_polar()`, `for_each_polar()`).
 
-### Other update methods
+### Other Update Methods
 
-`go` figures also support `update_layout_images` in order to [update background layout images](/python/images/), `update_annotations` in order to [update annotations](/python/text-and-annotations/#multiple-annotations), and `update-shapes` in order to [update shapes](/python/shapes/).
+Figures created with the `plotly.graph_objects` module also support:
```
nicolaskruchten: This isn't totally correct: all figures support this, including those created with make_subplots and PX...

nicolaskruchten:
As a general comment, I think the style rule around backticks could be something like "text that might reasonably end up being executed gets backticks"... this excludes `LaTeX` and "`x` axis" but includes `nbconvert` and `.show` and `go.Figure` and `px.scatter` ...  thoughts?

jdamiba:
@nicolaskruchten I agree with your style comments and think this is a good opportunity to standardize. 

To summarize:

`fig`
`fig.show()`
`.show()`
`make_subplots`

No backticks for file formats. 

What about package and library names like `plotly.py`, `plotly.js` or `htmlwidgets`? I think they should get backticks.

jdamiba:
@nicolaskruchten Can you please take another look at this.

doc/requirements.txt
```diff
@@ -23,5 +23,4 @@ recommonmark
 pathlib
 python-frontmatter
 datashader
-pyarrow
-cufflinks==0.17.3
+pyarrow
```
nicolaskruchten, line 26: this diff should not be in this PR: we need cufflinks installed

nicolaskruchten: I'll make the change and merge this

PR #2215, opened 2020-02-18
adding example for multiple disconnected lines

felixniemeyer:
I found this feature very helpful for plotting large networks with many many edges.
I think it might be valuable for others who stumble across this article, too.

### DIFF
--- a/doc/python/lines-on-maps.md
+++ b/doc/python/lines-on-maps.md
@@ -147,9 +147,15 @@ fig.add_trace(go.Scattergeo(
 flight_paths = []
 lons = []
 lats = []
-for i in range(len(df_flight_paths)):
-    lons += [df_flight_paths['start_lon'][i], df_flight_paths['end_lon'][i], None] 
-    lats += [df_flight_paths['start_lat'][i], df_flight_paths['end_lat'][i], None]
+import numpy as np
+lons = np.empty(3 * len(df_flight_paths))
+lons[::3] = df_flight_paths['start_lon']
+lons[1::3] = df_flight_paths['end_lon']
+lons[::3] = None
+lats = np.empty(3 * len(df_flight_paths))
+lats[::3] = df_flight_paths['start_lat']
+lats[1::3] = df_flight_paths['end_lat']
+lats[::3] = None
 
 fig.add_trace(
     go.Scattergeo(
--- a/doc/python/lines-on-maps.md
+++ b/doc/python/lines-on-maps.md
@@ -111,17 +111,43 @@ fig.update_layout(
 
 fig.show()
 ```
-### High performance US Flight Paths Map
-If you can relinquish having individual styles for the flight paths (e.g. opacity), you can put multiple paths into one trace, which makes the map render much faster. 
+### Performance improvement: put many lines in the same trace
+For very large amounts (>1000) of lines, performance may become critcal. If you can relinquish setting individual line styles (e.g. opacity), you can put multiple paths into one trace. This makes the map render faster and reduces the script execution time and memory consumption. 
 
 Use ```None``` between path coordinates to create a break in the otherwise connected paths. 
 
 ```python 
-# ... omitted code: look at the previous "US Flight Paths Map" example
+import plotly.graph_objects as go
+import pandas as pd
+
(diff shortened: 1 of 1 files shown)


jdamiba:
@felixniemeyer Thanks a lot for taking the time to submit this example! It's always awesome to see community members submitting pull requests. 

I only have a few minor comments. 

First, can you please update your PR so that the example you added has all the code needed to execute it? In our experience, users find it frustrating to not be able to copy/paste entire examples through the use of techniques like `# ... omitted code `. If you have seen any examples in the documentation that use this pattern please let me know!

Second, can you please provide a screenshot of what the example figure created by your example looks like and explain a little more the motivation behind it? I am having a hard time seeing what you mean by "multiple disconnected lines" when I am looking at the figure in my development environment. 

Thanks again!

emmanuelle:
@jdamiba I think the use case here is that if you have one trace per segment your figure will render more slowly than if you have a single trace of (here interrupted) lines. 

@felixniemeyer thanks for your contribution? Would you have the time to address @jdamiba 's comments about having a standalone code cell?

doc/python/lines-on-maps.md
```diff
@@ -111,6 +111,33 @@ fig.update_layout(
 
 fig.show()
 ```
+### High performance US Flight Paths Map
+If you can relinquish having individual styles for the flight paths (e.g. opacity), you can put multiple paths into one trace, which makes the map render much faster.
```
emmanuelle: About the wording I'm hesitating between `faster` and `much faster` here... Do you have orders of magnitude for the two solutions?

felixniemeyer:
Thanks for your feedback everybody. 

@jdamiba
I've added the omitted code in the example.

@emmanuelle  
I did an experiment to clarify the performance improvement 
I'm drawing 25,734 lines. 

Creating one trace for each line results in: 
5min 3sec execution time of the script. 
Frame time of ~1500ms
on my computer. 

Creating only 4 traces in total and adding multiple lines to them results in: 
0min 7sec execution time of the script. 
Frame time of ~700ms

--

Here are two people, for whom the "multiple lines in a trace" was the solution to their performance issues:  
https://community.plot.ly/t/multiple-lines-in-same-trace/14659

Another use case (besides performance improvement) for this feature of using "None" in the coordinates arrays to start multiple lines within the same trace may arise in combination with drawing a legend: 
On the legend, every trace has it's entry. If you have 100 lines each drawn as a separate trace, you would end up having 100 entries in your legend. 
If you draw them within one trace, they will have one entry in the legend.

doc/python/lines-on-maps.md
```diff
@@ -111,6 +111,74 @@ fig.update_layout(
...
+            color = 'rgba(68, 68, 68, 0)'
+        )
+    )))
+
+flight_paths = []
+lons = []
+lats = []
+for i in range(len(df_flight_paths)):
+    lons += [df_flight_paths['start_lon'][i], df_flight_paths['end_lon'][i], None] 
+    lats += [df_flight_paths['start_lat'][i], df_flight_paths['end_lat'][i], None]
```
emmanuelle:
```suggestion
import numpy as np
lons = np.empty(3 * len(df_flight_paths))
lons[::3] = df_flight_paths['start_lon']
lons[1::3] = df_flight_paths['end_lon']
lons[::3] = None
lats = np.empty(3 * len(df_flight_paths))
lats[::3] = df_flight_paths['start_lat']
lats[1::3] = df_flight_paths['end_lat']
lats[::3] = None
```

emmanuelle: Since this example is about performance, if you have a large number of lines using arrays will be much faster than the for loop (for 10000 lines the for loop takes almost one second on my machine).

emmanuelle:
Thank you @felixniemeyer for the update and the benchmark. I just left a last suggestion to improve performance even more :-).

emmanuelle:
@felixniemeyer did you have the time to look at my suggestion? I'd love to merge this cool example!

emmanuelle:
@felixniemeyer I merged my suggestion to your branch so that we can merge the PR, I hope you don't mind. Thank you very much for the PR!

PR #2221, opened 2020-02-21
add example showing how to customize figure subplot titles

jdamiba:
The purpose of this PR is to add an example which demonstrates how to customize figure subplot titles. 

see https://github.com/plotly/plotly_express/issues/36

### DIFF
--- a/doc/python/facet-plots.md
+++ b/doc/python/facet-plots.md
@@ -5,8 +5,8 @@ jupyter:
     text_representation:
       extension: .md
       format_name: markdown
-      format_version: "1.1"
-      jupytext_version: 1.1.1
+      format_version: '1.2'
+      jupytext_version: 1.3.4
   kernelspec:
     display_name: Python 3
     language: python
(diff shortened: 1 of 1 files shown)


doc/python/facet-plots.md
```diff
@@ -31,8 +31,8 @@ jupyter:
     page_type: u-guide
     permalink: python/facet-plots/
     redirect_from:
-      - python/trellis-plots/
-      - python/facet-trellis/
+    - python/trellis-plots/
+    - python/facet-trellis/
```
emmanuelle: did you change this yourself or was it Jupyter? From a quick `grep` I think we have the two syntaxes in the doc directory, so this should be fine, but just checking.

emmanuelle:
One little thing I was not sure of in the header, but otherwise :dancer:

doc/python/facet-plots.md
jdamiba: It was Jupyter, not me that edited the front matter.

emmanuelle: ok, :dancer:!

Issue #1965, opened 2019-12-03
Wishlist: examples wanted


emmanuelle:
Most items coming from questions on the community forum

- [ ] how to force an axis to be categorical, see https://community.plot.ly/t/how-to-plot-categorical-values-in-x/31982/3 or https://community.plot.ly/t/how-to-get-bars-located-to-each-other-in-bar-plot-with-float-x-axis-values/31554/2. To go to https://plot.ly/python/axes/ or to a new "meta" page explaining what the different types of axes are?
- [ ] how to remove the colorbar in a figure created by a `px.imshow` figure. https://community.plot.ly/t/hide-colorbar-in-px-imshow/31853. To go to `imshow.md` tutorial.
- [ ] improve explanations on `updatemenus` and changing animation speed in animations. See https://stackoverflow.com/questions/59070876/plotly-how-do-the-buttons-for-the-update-menus-really-work/59076340#59076340,  https://community.plot.ly/t/scatter-animation-doesnt-work/31944 https://community.plot.ly/t/need-animation-example-with-scattermapbox/9720/10 https://community.plot.ly/t/how-to-slow-down-animation-in-plotly-express/31309/7
- [ ] add example of using customdata in hovertemplate, in https://plot.ly/python/hover-text-and-formatting/. See https://community.plot.ly/t/can-hovertemplate-combine-information-from-multiple-traces/31960/4 or https://community.plot.ly/t/density-mapbox-tooltips/31932/2
- [ ] on histogram tutorial, mention that we cannot access the values of counts (it's asked a lot), and maybe mention numpy's function for this.
- [ ] example of shape or annotation covering several subplots thanks to `xref='paper'`. Also the shapes and annotations tutorial should link to each other. 
- [ ] add px example in 2d density page, links between 2d density and 2d density contour pages, remove gremlin.

emmanuelle:
I've tagged this issue as a good first issue since some examples are quite easy to add (especially if the forum posts have some code examples). Items which have already been ticked are already done. Feel free to ask questions here if you would like to contribute an example but are unsure about how to do it. Please also refer to the doc README https://github.com/plotly/plotly.py/blob/master/doc/README.md

emmanuelle:
Here is also a couple of searches with no result on the doc website, which should return something
- [ ] `lasso` and `box selection` (we do have a page for selection-events but it's for `FigureWidget`, we don't have anything for selections in plotly figures I believe, only in the config page maybe)
- [ ] `io`
[...]

PR #2235, opened 2020-02-27
Update facet-plots.md

SylwiaOliwia2:
Added section 'Zooming subplots to the same range'. A part of #1965

### DIFF
--- a/doc/python/facet-plots.md
+++ b/doc/python/facet-plots.md
@@ -118,3 +118,20 @@ fig.show()
 ```python
 
 ```
+
+### Zooming subplots to the same range
+
+Using `facet_col` from `plotly.express` let zoom each facet to the same range impliciltly. However, if the subplots are created with `make_subplots`, the axis needs to be updated with `matches` parameter to zoom all the subplots accordingly. Zoom in one trace below, to see the other subplots zoomed to the same x-axis range:
+
+```python
+import plotly.graph_objects as go
+from plotly.subplots import make_subplots
+import numpy as np
+N = 20
+x = np.linspace(0, 1, N)
+fig = make_subplots(1, 3)
+for i in range(1, 4):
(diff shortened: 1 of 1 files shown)


doc/python/facet-plots.md
```diff
@@ -118,3 +118,20 @@ fig.show()
 ```python
 
 ```
+
+### Zooming subplots to the same range
+
+Using `facet_col` from `plotly.express` let zoom each facet to the same range impliciltly. However, if the subplots are created with `make_subplots`, the axis needs to be updated with `matches` parameter to zoom all the subplots accordingly. Zoom in one trace below, to see the other subplots zoomed to the same x-axis range:
```
emmanuelle:
```suggestion
Using `facet_col` from `plotly.express` let zoom each facet to the same range implicitly. However, if the subplots are created with `make_subplots`, the axis needs to be updated with `matches` parameter to zoom all the subplots accordingly. Zoom in one trace below, to see the other subplots zoomed to the same x-axis range:
```

doc/python/facet-plots.md
```diff
@@ -118,3 +118,20 @@ fig.show()
 ```python
 
 ```
+
+### Zooming subplots to the same range
```
emmanuelle: `matches` does actually more than synchronizing zoom, you also have synchronized panning for example. How about a more general title like "Synchronizing axes in subplots with `matches`"? Also what do you think of mentioning pan and zoom in the text below (compared to just zoom at the moment?)

emmanuelle:
Thanks a lot @SylwiaOliwia2 for this nice example! The example is great, I left a suggestion on how to make the text a little more general, please tell me what you think.

doc/python/facet-plots.md
SylwiaOliwia2: Sure! I will edit the title and mention the pan.

emmanuelle:
Thank you very much! Merging :-).

PR #2242, opened 2020-03-03
Update histograms.md

SylwiaOliwia2:
Added `Accessing the y-axis values` section

### DIFF
--- a/doc/python/histograms.md
+++ b/doc/python/histograms.md
@@ -69,6 +69,32 @@ fig = px.histogram(df, x="total_bill", nbins=20)
 fig.show()
 ```
 
+#### Accessing the y-axis values
+
+JavaScript calculates the y-axis (count) values on the fly in the browser, so it's not accessible in the `fig`. You can manually calculate it using `pandas.cut` or `np.digitize`.
+
+```python
+import plotly.express as px
+import pandas as pd
+
+df = px.data.tips()
+
+# create the bins; use the `range` to get the same bin size as in the histogram in the previous section.
+df["total_bill_bins"] = pd.cut(df.total_bill, bins=range(0, 60, 5), right=False)
+
+# calculate counts
+df_counts = df.pivot_table(index="total_bill_bins", values="size", aggfunc='count').reset_index()
+df_counts.rename(columns={"size": "count"}, inplace=True)
+
+# sort, then convert to string
+df_counts = df_counts.sort_values(by="total_bill_bins")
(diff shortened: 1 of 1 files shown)


doc/python/histograms.md
```diff
@@ -69,6 +69,32 @@ fig = px.histogram(df, x="total_bill", nbins=20)
 fig.show()
 ```
 
+#### Accessing the y-axis values
```
emmanuelle: Maybe `Accessing the counts (y-axis) values`?

emmanuelle:
Thank you for the example @SylwiaOliwia2 ! A lot of people get confused about this, it will be very useful. I'm just curious to know why you used pandas functions instead of `np.histogram`, is it because some people working with pandas prefer to use only pandas functions? I post below a version using `np.histogram` so that we can compare and discuss about the two solutions.
```
import plotly.express as px
import numpy as np
df = px.data.tips()
# create the bins
counts, bins = np.histogram(df.total_bill, bins=range(0, 60, 5))
bins = 0.5 * (bins[:-1] + bins[1:])
fig = px.bar(x=bins, y=counts, labels={'x':'total_bill', 'y':'count'})
fig.show()
```

SylwiaOliwia2:
Indeed, I used to pandas, this solution was the most obvious solution to me.
However, `numpy.histogram` is much simpler and using continuous x-scale! If you agree I'll keep only the NumPy solution.

emmanuelle:
@SylwiaOliwia2 I agree that I find the numpy solution simpler.

SylwiaOliwia2:
@emmanuelle done :)

emmanuelle:
Thank you @SylwiaOliwia2 ! Now that you have contributed several doc PRs to plotly.py, do you have any feedback about the contribution process? Any way that we could improve it? Do you feel like trying to work on a bug or a part of the code for your next PR ;-) ? Great to have here anyway!

SylwiaOliwia2:
> Thank you @SylwiaOliwia2 ! Now that you have contributed several doc PRs to plotly.py, do you have any feedback about the contribution process? Any way that we could improve it? Do you feel like trying to work on a bug or a part of the code for your next PR ;-) ? Great to have here anyway!

Thank you, @emmanuelle :) Definitely I can try working out on the code the next time. Feel free to suggest any issue :)

Regarding contribution, using tags is very helpful (that's how I've found the issue), especially for newbies. So keep using them, please ;)

PR #2245, opened 2020-03-04
Support JupyterLab 2.0.0

### LINKED ISSUE #2244: Using plotly with jupyterlab 2.0
See https://community.plot.ly/t/jupyterlab-2-0-support/35715

consideRatio:
I think this is a backward compatible change that makes the jupyterlab extension `jupyterlab-plotly` support JupyterLab 2.0.0 by accepting also JupyterLab 2.0.0 required dependencies.

Fixes #2244

## JupyterLab 2.0.0 references
- [Change log for developers](https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html#for-developers)
- [A migration guide](https://jupyterlab.readthedocs.io/en/stable/developer/extension_migration.html#extension-migration)

### DIFF
--- a/packages/javascript/jupyterlab-plotly/package-lock.json
+++ b/packages/javascript/jupyterlab-plotly/package-lock.json
@@ -9,9 +9,9 @@
       "resolved": "https://registry.npmjs.org/3d-view/-/3d-view-2.0.0.tgz",
       "integrity": "sha1-gxrpQtdQjFCAHj4G+v4ejFdOF74=",
       "requires": {
-        "matrix-camera-controller": "2.1.3",
-        "orbit-camera-controller": "4.0.0",
-        "turntable-camera-controller": "3.0.1"
+        "matrix-camera-controller": "^2.1.1",
+        "orbit-camera-controller": "^4.0.0",
+        "turntable-camera-controller": "^3.0.0"
       }
     },
     "@choojs/findup": {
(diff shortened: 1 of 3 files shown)


nicolaskruchten:
Thanks very much for this PR! I'll QA it today and if all looks good, release as well.

This is intended to be backwards-compatible in the sense that the extension should run in JupyterLab 1.2 as well, correct?

consideRatio:
@nicolaskruchten yes correct it will installable with both JupyterLab 1 and 2, https://www.npmjs.com/package/jupyterlab-plotly. If we would remove the `||` parts and only use the newer version, it would result in someone wanting to use JupyterLab 1 would end up using the latest compatible released version of the plotly-extension instead of the latest available.

I opted to write in `||` making the old version usable alongside the new as there will be various JupyterLab 1 users straggling that may want to use the latest releases coming quite often from plotly-extension.

consideRatio:
@nicolaskruchten I bumped the node package version of this npm package itself, I'm not sure if that made sense to do.

nicolaskruchten:
Super, thanks for the clarifications. Bumping the package version is fine I think: normally I do it during the release process but I'll likely merge and release within a few minutes anyway so this is great :)

nicolaskruchten:
OK this all looks awesome, I should be able to merge and release tonight after the kids are in bed (famous last words!)

I should add that it's always a real pleasure to head to work thinking "man, gotta deal with this new problem that cropped up" and to arrive to find a PR that fixes it, ready to go, so thanks again!

nicolaskruchten:
This is now released! Unfortunately `FigureWidget`s don't appear to work in JLab 2.x... See https://github.com/plotly/plotly.py/issues/2249 for details if you're curious or have any ideas!

PR #2253, opened 2020-03-06
documentation PR checklist

nicolaskruchten:
feedback welcome!


emmanuelle:
This looks great! Maybe, in order to make the checklist look less compelling, you can break it in two or three with subtitles like "Version and branch", "Structure", "Style". :dancer:

PR #2255, opened 2020-03-06
tried to fix default renderer for ipython terminal

### LINKED ISSUE #2254: default renderer not always automatically set
if I open ipython on my Ubuntu machine, I would like the default renderer to be set automatically to 'browser'  (as it's intended to be I think) but it's not the case:
```
$ ipython3
Python 3.7.3 (default, Oct  7 2019, 12:56:13) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.8.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import plotly.io as pio                                                                                                                          

In [2]: pio.renderers.default                                                                                                                            
Out[2]: 'plotly_mimetype+notebook'
```

With jupyter notebook and jupyter lab it works fine. @jonmmease is this intended to work?

emmanuelle:
Closes #2254 

I'm not sure yet whether this is the right place to add the check for the terminal. Also need to check tests.

### DIFF
--- a/packages/python/plotly/plotly/io/_renderers.py
+++ b/packages/python/plotly/plotly/io/_renderers.py
@@ -452,7 +452,7 @@ environment variable: {env_renderer}""".format(
         )
 
     default_renderer = env_renderer
-elif ipython and ipython.get_ipython() and not (ipython.get_ipython().__class__.__name__ == 'TerminalInteractiveShell'):
+elif ipython and ipython.get_ipython():
     # Try to detect environment so that we can enable a useful
     # default renderer
     if not default_renderer:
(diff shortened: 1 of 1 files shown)


jonmmease:
Thanks for taking a look at this @emmanuelle.  As I recall, I didn't come up with a good way to distinguish the `ipython` console from a jupyter notebook kernel.  Have you checked that this codepath is not triggered in a notebook context?

emmanuelle:
Hi @jonmmease I gave it a try on my laptop, also on Colab and in Kaggle. 
```
from plotly import optional_imports
ipython = optional_imports.get_module("IPython")
print(ipython.get_ipython().__class__.__name__)
```
returns `ZMQInteractiveShell` with my version of the notebook and in Kaggle, and `Shell` in Colab.

Given the fact that the Ipython terminal is a small fraction of our marker (however, with a strong bias towards experimented developers!), I understand we need to be cautious here and be sure not to break anything for notebook-like environments. This is also why I put this test after all the ones checking environments variables. Do you have any suggestion on how to test systematically whether this change does not break anything?

jonmmease:
Thanks for investigating! That all sounds good. :dancer: from me

PR #2260, opened 2020-03-10
Added lazy loading to jupyterlab-plotly

### LINKED ISSUE #1913: Lazy load strategy for plotly bundle
While having the plotly bundle be deduplicated between this and other extensions (#1891) would help some of the issues users are seeing, the fact still remains that, minified or unminified, the plotly bundle is enormous: even minified, it's 4mb size is close to the entire minified lab bundle (5.5mb, last i checked).

Not familiar with this code base and whichever other ones use the plotly bundle, the example for another heavy widget https://github.com/plotly/plotly.py/pull/1891#issuecomment-555646304 would at least get the bundle out of the critical path. It is likely worth investigating if there is a way to do it at a finer-grained level (e.g. just plotly itself) but then the additional points of asynchrony might need to be found.

As suggested in https://github.com/plotly/plotly.py/pull/1891#issuecomment-555603317, lazy loading the bundle would provide several benefits to end users:
- less up-front time waiting for the "moons" animation for all users
- less CPU used during a local webpack rebuild for those that aren't using hosted/pre-built lab applications

The only downside is there would be a (potentially noticeable on slower connections) delay when the first widget is requested.

pragyagarg642:
With this change, plotly.js will be loaded on demand. This uses dynamic
imports to achieve this.

Fixes #1913


pragyagarg642:
@nicolaskruchten Can you please review this?

nicolaskruchten:
Yes, I will take a look. A couple of questions/concerns though: right now in JupyterLab we need to install both `jupyterlab-plotly` and `plotlywidget` and both need plotly.js loaded. Right now they both statically load the same bundle and webpack deduplicates it. Is there any way that your change could include `plotlywidget` as well? If not, I fear the bundle size will remain the same *and* we will lazy-load plotly.js.

It might be that we need to first merge the two extensions together and then make this change, but I'm not sure.

pragyagarg642:
@nicolaskruchten I'm working on the changes.

pragyagarg642:
@nicolaskruchten I've made changes according to your comment. Can you please review this?

nicolaskruchten:
Thanks for these changes! We're quite busy preparing for the 4.6.0 release of Plotly.py but I will review this during the 4.7.x cycle.

pragyagarg642:
Hi @nicolaskruchten , as the 4.6.0 release is complete, can you please review this?

nicolaskruchten:
OK, I've had a chance to take a look at this PR, thanks again for submitting it! The TypeScript doesn't actually compile for me with this PR as-is, but with a few adjustments it does, and sure enough, the bundle is nicely split, loads when needed, deduplicates between the two extensions etc. I'll need to fork this PR to add my changes such that things compile, and then the rest of my team can do some QA on Windows and Linux etc, but just wanted to let you know that unless anything weird crops up, we'll be able to merge this for 4.7.

cc @emmanuelle @jonmmease

nicolaskruchten:
Update: we need to release 4.7 on Monday to meet some internal deadlines, so I'm bumping this to 4.8, which is only about 10 days away!

pragyagarg642:
Hi @nicolaskruchten , prefetch option can also added to this change. With that option, module will be fetched just after the initial page load is completed and when required, module will be loaded from prefetch cache, making it faster. Without this option, it will take time to load. I can send PR for this change as well.

pragyagarg642:
Hi @nicolaskruchten , prefetch option can also added to this change. With that option, module will be fetched just after the initial page load is completed and when required, module will be loaded from prefetch cache, making it faster. Without this option, it will take time to load. I can update the PR with this change as well.

nicolaskruchten:
@pragyagarg642 sure, let's see what prefetching looks like! BTW I had to make these changes https://github.com/plotly/plotly.py/pull/2454 to get your branch to work, if you want to incorporate them... Otherwise TypeScript compilation failed.

I should add that in order to merge this, we'll have to make equivalent changes to https://github.com/plotly/jupyterlab-chart-editor as well, as people frequently install that one too, and we want to keep all the bundle-loading logic the same. I might move that code into this repo in the very near future to keep things in sync.

I'm sorry it's taking so long to get this merged in!

pragyagarg642:
@nicolaskruchten I updated the branch with required prefetch changes.

pragyagarg642:
Hi @nicolaskruchten, is there any update on this or is there anything pending on my end?

nicolaskruchten:
See my comment in https://github.com/plotly/plotly.py/pull/2454 ... the bottleneck here is just my time in being able to QA this to the point where we're willing to roll this change out to all of our users. I'm sorry I haven't been able to find the time but this is potentially quite a risky change.

nicolaskruchten:
I do apologize about how this PR was left unattended for so long, but it was superseded by https://github.com/plotly/plotly.py/pull/3142 which was just merged, so the functionality will make it into the next version of the library.

PR #2265, opened 2020-03-10
ipywidgets import bug

### LINKED ISSUE #2264: import plotly.graph_objects fails with ipywidget version is <= 7.0.0
Steps to reproduce:
<pre>
$ python3.7 -m venv test
$ . test/bin/activate
(test) $ pip install plotly
(test) $ pip install ipywidgets==6
(test) $ python -c "import plotly.graph_objects"
  File "<string>", line 1, in <module>
  File "/home/varoquau/tmp/clean_venv/test/lib/python3.7/site-packages/plotly/graph_objects.py", line 2, in <module>
    from plotly.graph_objs import *
AttributeError: module 'plotly.graph_objs' has no attribute 'FigureWidget'
</pre>
The problem is on line 100184 of plotly/graph_objs/__init__.py: the `__all__.append("FigureWidget")` that is a few lines below should be moved under the "if" clause.

Given that it is generated code, I don't know how to fix it, so I won't send a pull request.

emmanuelle:
Closes #2264

### DIFF
--- a/packages/python/plotly/codegen/__init__.py
+++ b/packages/python/plotly/codegen/__init__.py
@@ -259,9 +259,9 @@ try:
     from distutils.version import LooseVersion
     if LooseVersion(ipywidgets.__version__) >= LooseVersion('7.0.0'):
         from ._figurewidget import FigureWidget
+        __all__.append("FigureWidget")
     del LooseVersion
     del ipywidgets
-    __all__.append("FigureWidget")
 except ImportError:
     pass
 """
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
Hrm yes I see the problem. Only happens with `from plotly.graph_objs import *` I imagine?

💃

emmanuelle:
`from plotly.graph_objs import *` is a code line of `graph_objects.py` therefore `import plotly.graph_objects as go` also raises the error with an old ipywidgets.

nicolaskruchten:
Makes sense!

PR #2277, opened 2020-03-13
Update README.md

joydisette:
jupyter build times-out. Updated steps from this thread https://discourse.jupyter.org/t/installing-extensions-extremely-slow/2627

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,11 @@ jupyter labextension install jupyterlab-plotly@1.5.4 --no-build
 # Build extensions (must be done to activate extensions since --no-build is used above)
 jupyter lab build
 
+# Run the command below  instead if build times-out / is too slow but keep in mind it's only for the local use.
+#DO NOT RUN on a remote server
+jupyter lab build --minimize = False
+
+
 # Unset NODE_OPTIONS environment variable
 # (OS X/Linux)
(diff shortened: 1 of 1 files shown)


emmanuelle:
Thank you for your pull request @joydisette ! It's also very interesting for us to read the discussion on the jupyter discourse so thanks for sharing the link. Did this time-out happen to you? If this problem has happened many times I'll be in favor of adding this tip, but I'll let @nicolaskruchten or @jonmmease  comment first since they know more than me about the plotly bundles.

emmanuelle:
the CI failure has nothing to do with you so don't worry about this

joydisette:
@emmanuelle yes it did time-out for me too

emmanuelle:
Thank you @joydisette ! So we're trying to have quite a lot of stuff about Jupyterlab in the `README.md` (since several things can go wrong :-(), so probably I will refactor the README to move some instructions and your addition to the troubleshooting guide https://plot.ly/python/troubleshooting/. I'll make sure to pick your commit so that you are duly credited. Thanks again :-).

nicolaskruchten:
I'm not sure we want to recommend that people use dev builds, actually, as it could have unintended consequences for other extensions. I hope that this kind of thing will become unnecessary once #2441 is done.

PR #2278, opened 2020-03-13
unpinned versions of pandas (and scipy and numpy)

emmanuelle:
I did this unpinning mostly because no wheels exist for the version of pandas which was pinned, therefore building the binder container takes forever.


nicolaskruchten:
💃

PR #2280, opened 2020-03-15
debug

emmanuelle:
to test the new js build

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -4,7 +4,7 @@ Thank you for contributing to plotly.py!
 
 ## Code of Conduct
 
-Check out the [Code of Conduct](CODE_OF_CONDUCT.md). Don't tl:dr; it, but the general idea is to be nice.
+Check out the [Code of Conduct](CODE_OF_CONDUCT.md). Don't tl:dr; it, but the general idea is to be nice. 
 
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
FYI it's not necessary to make a PR or any commits to get a new dev build from the Plotly.js master... You can go into CircleCI under the master branch i.e. https://circleci.com/gh/plotly/plotly.py/tree/master and find the latest `dev_build` build and hit "rerun"

image

nicolaskruchten:
FWIW the build containing the rangebreaks and unified-hover stuff is here now:

`pip install https://33619-14579099-gh.circle-artifacts.com/0/packages/python/plotly/dist/plotly-4.5.4%2B27.g1c232fd4f.tar.gz` but it'll surely be out of date in a day or so as we squash some bugs :)

nicolaskruchten:
Ah also, given how sensitive the JupyterLab installation is, please note that you can only test this kind of build in Classic Notebooks I think.

emmanuelle:
ok, thanks for the tips for using the JS master, I had forgotten about this possibility. Did you see the percy changes? I'm a bit puzzled about where they come from.

PR #2289, opened 2020-03-18
Clarify `histnorm` description

potpath:
This is consistent with https://plot.ly/python-api-reference/generated/plotly.express.histogram.html#plotly.express.histogram and https://plot.ly/python/reference/#histogram-histnorm.


nicolaskruchten:
Thanks for this PR, sorry it took so long to merge it :)

PR #2293, opened 2020-03-19
pinning orca

emmanuelle:
Temporary solution to prevent CI break.


nicolaskruchten:
💃 even with the `plot_ly` failure

nicolaskruchten:
We should add this to README.md also for the time being

nicolaskruchten:
OK added it to the readme and cleared the file on chart studio... can merge when green :)

PR #2294, opened 2020-03-19
Remove unsupported opacity kwarg from pie-like functions

nicolaskruchten:
Whoops, `marker_opacity` isn't supported in `pie`-like traces! See https://community.plot.ly/t/possible-bug-color-discrete-map-and-opacity-in-px-pie-plotly-io-as-pio/36386/2


emmanuelle:
:dancer:

PR #2304, opened 2020-03-22
hide name from within hovertemplate

nicolaskruchten:
This should result in no change (i.e. the trace name still will not appear in at the edge of the hoverlabel) but makes it easier to use the upcoming "unified" `hovermode` with PX. With this PR, simply using `.update_traces(hovertemplate=None)` will result in a nice-looking unified hoverlabel, but without it, users will also have to set `namelength` back to some non-zero value.

Mind you, they might have to set it to some higher value than the default of 15 anyway, but at least the trace name will mostly/partly show up!


emmanuelle:
looks good, thank you! :dancer:

Issue #2301, opened 2020-03-20
failing tests on CI following Orca release 1.3.0


antoinerg:
With the release of Orca 1.3.0, some tests now fail on CircleCI.

As far as I know, here are the problems we're hitting:
- Sandboxing error due to electron version bump (tracked in https://github.com/plotly/orca/issues/294)
- EPS baselines used in test images will need to be updated

PR #2307, opened 2020-03-23
bump Orca to 1.3.1

antoinerg:
This PR closes https://github.com/plotly/plotly.py/issues/2301 by:

- we pin Orca's version to 1.3.1 to sync the releases (https://github.com/plotly/plotly.py/pull/2307/commits/44f4d6f7f7450c2d2cc38e316cbe60fa3003875f)
- update EPS baselines used in test images (https://github.com/plotly/plotly.py/pull/2307/commits/313ee4d3cd51904cad67ee68bf84c49262db0fec)


emmanuelle:
thank you very much @antoinerg for fixing all this! If we can keep the standard err as it was before by checking for an environment variable I think it would be better. The CI failure has nothing to do with your PR and will not block merging.

antoinerg:
I forced push to get rid of all the unnecessary cruft. This PR is now considerably simpler following a fix upstream to Orca (https://github.com/plotly/orca/pull/301) and the release of [Orca v1.3.1](https://github.com/plotly/orca/releases/tag/v1.3.1).

This will be ready for a review :smile_cat:

nicolaskruchten:
LGTM! 💃 Thanks @antoinerg

PR #2309, opened 2020-03-25
what about dash?

nicolaskruchten:
Closes https://github.com/plotly/documentation/issues/1654


nicolaskruchten:
merging, as I need these docs deployed :)

PR #2312, opened 2020-03-25
change .ly to .com

jdamiba:
The purpose of this PR is to change hyperlinks in the documentation from the `.ly` to the `.com` TLD.


emmanuelle:
Thanks @jdamiba looks good to me. The 2.7 failure is something else. Note that a large fraction of these changes could have been avoided if we had relative links only in the doc for markdown links, e.g. `/python/somepage` instead of the full URL. It would be an improvement to change everything to relative links, but probably we should merge this PR quickly so :dancer:

nicolaskruchten:
@jdamiba can you do a follow-up to this where you replace `https://plotly.com/python` with `/python` ?

PR #2317, opened 2020-03-26
Opacity in px.density_mapbox

### LINKED ISSUE #2316: Plotly Express density_mapbox, attribute 'opacity' raises ValueError
## Description
Hello,
When passing the `opacity` attribute different than `None` to `px.density_mapbox`, it raises a ValueError.

## Expected output
Taking the [example in the doc](https://plotly.com/python/mapbox-density-heatmaps/) and wanting to set the opacity to 0.5:
image

## Minimal reproducible code
```python
import pandas as pd
df = pd.read_csv('https://raw.githubusercontent.com/plotly/datasets/master/earthquakes-23k.csv')

import plotly.express as px
fig = px.density_mapbox(df, lat='Latitude', lon='Longitude', z='Magnitude', radius=10,
                        center=dict(lat=0, lon=180), zoom=0,
                        mapbox_style="stamen-terrain", opacity=0.5)
fig.show()
```

## Current workaround
It's possible to change the opacity by not adding the attribute in the constructor but adding it after using `fig.update_traces(opacity=0.5)`. But I'll open a PR with a fix so you don't have to use that.

tvaucher:
Hello,
This is a fix for #2316 that correctly map the opacity attribute in the `trace_path`.
It allows to correctly pass `opacity` as an attribute in `px.density_mapbox`. Check the issue for reproducible code.

***Limitations:*** I'm not aware of another mapbox figure that would require this change as they use proper markers (scatter, line), but if you are, then please comment and I'll add that.

Thanks for your time.
plotly version=4.5.4

### DIFF
--- a/packages/python/plotly/plotly/express/_chart_types.py
+++ b/packages/python/plotly/plotly/express/_chart_types.py
@@ -1399,9 +1399,7 @@ def funnel(
     rectangular sector of a funnel.
     """
     return make_figure(
-        args=locals(),
-        constructor=go.Funnel,
-        trace_patch=dict(opacity=opacity, orientation=orientation),
+        args=locals(), constructor=go.Funnel, trace_patch=dict(orientation=orientation),
     )
 
 
--- a/packages/python/plotly/plotly/express/_chart_types.py
+++ b/packages/python/plotly/plotly/express/_chart_types.py
@@ -1234,6 +1234,7 @@ def pie(
     template=None,
     width=None,
     height=None,
+    opacity=None,
     hole=None,
 ):
     """
@@ -1422,6 +1423,7 @@ def funnel_area(
     template=None,
     width=None,
     height=None,
+    opacity=None,
 ):
     """
     In a funnel area plot, each row of `data_frame` is represented as a
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1278,6 +1278,8 @@ def infer_config(args, constructor, trace_patch):
         if args["opacity"] is None:
             if "barmode" in args and args["barmode"] == "overlay":
                 trace_patch["marker"] = dict(opacity=0.5)
+        elif constructor in [go.Densitymapbox, go.Pie, go.Funnel, go.Funnelarea]:
+            trace_patch["opacity"] = args["opacity"]
         else:
             trace_patch["marker"] = dict(opacity=args["opacity"])
     if "line_group" in args:


emmanuelle:
Thank you very much for the fix @tvaucher ! Another possibility would be to use the `trace_patch` argument of `make_figure`, as in https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/express/_chart_types.py#L1403. This would reduce the amount of code in `_core.py` which is already huge. What do you think @nicolaskruchten ? Anyway we should merge a fix quickly since it concerns a bug so thank you again @tvaucher !

tvaucher:
Like this then. The if is still necessary as you otherwise end up with `markers` in the `trace_patch` which was originally responsable for the `ValueError`

emmanuelle:
thank you very much for your reactivity @tvaucher ! @nicolaskruchten you can choose between the two versions :-)

tvaucher:
So revert back to something similar from my first commit like:
```python
if args["opacity"] is None:
    if "barmode" in args and args["barmode"] == "overlay":
        trace_patch["marker"] = dict(opacity=0.5)
elif constructor in [go.Densitymapbox, go.Pie, go.Funnel, go.Funnelarea]:  # add more if existing
    trace_patch["opacity"] = args["opacity"]
else:
    trace_patch["marker"] = dict(opacity=args["opacity"])
```

As apparently both `go.Funnelarea` and `go.Pie`  have an `opacity`  property.
Additionally, revert #2294 and remove the `trace_patch(opacity=opacity, ...)` in [Funnel](https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/express/_chart_types.py#L1403).

I think it's probably easier to maintain if you add new Figures as you would need anyway to change the trace_patch if it doesn't support `marker` attributes as you would end up with an if anyway like in my second commit.

If you tell me, I'll make the necessary modifications for your review.

nicolaskruchten:
Yes, the changes you described above sound great to me! Thank you very much :)

PS: if you rebase onto master now, that CI failure should clear up.

tvaucher:
Done.

nicolaskruchten:
Thanks! Something is a bit odd about the rebase though, the diff looks like it includes parts of other pull requests...

tvaucher:
Yes indeed. Messed a bit with the previous git rebase. It should be much cleaner now (rebased on master and replayed the last 3 commits that contain the real changes).

nicolaskruchten:
Looks perfect, thanks! I'll do a bit more QA and then merge and it'll be part of the 4.6.0 release :)

PR #2318, opened 2020-03-26
use relative hyperlink for https://plotly.com/python

jdamiba:


doc/python/2D-Histogram.md
```diff
@@ -170,4 +170,4 @@ fig.show()
 ```
 
 #### Reference
-See https://plotly.com/python/reference/#histogram2d for more information and chart attribute options!
+See /python/reference/#histogram2d for more information and chart attribute options!
```
nicolaskruchten: Ah this won't work... let's just do this another time.

nicolaskruchten:
Let's leave this for another day: too many ways to break links this way.

PR #2329, opened 2020-03-28
Mention extra tag in hovertemplate tutorial

emmanuelle:
Following various interactions on the forum, including https://community.plotly.com/t/tooltip-in-dash-shows-trace-0-how-to-get-rid-of-that/36844/2


nicolaskruchten:
Thanks!

There's another way to control this, which is to set the `hoverlabel.namelength` attribute to 0 (to hide the name) or some big number (to show the whole name if it's long).

You could also mention that "anything goes" in the `<extra></extra>` it just happens to default to `%{name}` basically but you can put anything you like in there.

emmanuelle:
thanks for the ideas I will integrate them. Regarding the chart studio problem has this been fixed in master or is it something new?

nicolaskruchten:
I think it's fixed in master... let's ignore it for now

emmanuelle:
Turns out `hoverlabel.namelength` is not used when a hovertemplate is used, so there was no good place to fit this in, and I was not sure it was worth a whole example. Can I merge this one?

nicolaskruchten:
Does this conflict with the changes in your hover-data PR?

emmanuelle:
Hum probably thanks for noticing this (but the conflict should be easy to fix). So either we merge now into doc-prod and I take care of conflicts when merging doc-prod into master just after, or we wait post-release to merge this PR after master has been merged into doc-prod, and I fix conflicts within the PR. Both solutions are fine with me.

nicolaskruchten:
💃  to merge into `doc-prod` now, and if you have some time to merge `doc-prod` into master and resolve conflicts, that'd be great :)

emmanuelle:
done, there was indeed a small merge conflict. I've pushed to master after merging doc-prod.

PR #2330, opened 2020-03-29
Px special inputs

### LINKED ISSUE #2119: px.Constant and the "identity" problem
### `px.Constant`

In certain cases to get the kind of output one would want with PX we need to create a dummy column containing the same string or number in each row. I propose to create a special kind of object that we could pass into the `px` functions to do this like:

```python
px.bar(df, y="the_y_value", x=px.Constant(1, label="the_optional_label"))
```

which would be a shortcut for:

```python
px.bar(df, y="the_y_value", x=[1]*len(df), labels={"x": "the_optional_label"})
```

### `px.Identity` + `px.Direct` (identity option 1)

Relatedly, but on a different note, one might have a column that directly encodes something like the color, and we might want a similar mechanism for doing something like

```python
px.bar(x=[1,2,3], y=[1,2,3], color=px.Identity(["red", "green", "blue"]))
px.bar(x=[1,2,3], y=[1,2,3], color=px.Identity("column_with_colors"))
```

which would be a shortcut for:

```python
px.bar(x=[1,2,3], y=[1,2,3], color=["red", "green", "blue"], 
color_discrete_map={"red":"red", "blue":"blue", "green": "green"})
```

Using the two together would allow you to force a single series to a single color:

```python
px.bar(x=[1,2,3], y=[1,2,3], color=px.Identity(px.Constant("red")))
```

and this might warrant its own shortcut, maybe `px.Direct`?

```python
px.bar(x=[1,2,3], y=[1,2,3], color=px.Direct("red"))
```

which would be a more-intuitive shortcut for the following equivalent calls, in terms of their output:

```python
px.bar(x=[1,2,3], y=[1,2,3], color=px.Constant("red"),
    color_discrete_map={"red":"red"})
px.bar(x=[1,2,3], y=[1,2,3], color=["red"]*3, 
    color_discrete_map={"red":"red"})
px.bar(x=[1,2,3], y=[1,2,3], 
    color_discrete_sequence=["red"])
```

### `px.IDENTITY_MAP` (identity option 2)

A slightly different way of solving the "identity" problem would be to be able to pass in a magic value to `*_map` (edit: which per @emmanuelle could be just the string `"identity"`!) like this:

```python
px.bar(x=[1,2,3], y=[1,2,3], color=["red", "green", "blue"],  
color_discrete_map=px.IDENTITY_MAP)
```

which would also be the equivalent of

```python
px.bar(x=[1,2,3], y=[1,2,3], color=["red", "green", "blue"], 
color_discrete_map={"red":"red", "blue":"blue", "green": "green"})
```

This would avoid the need for nesting `px.Constant` and `px.Identity` when you want to use constant and identity together, and would mean no need for `px.Direct` as shortcut for that:

```python
px.bar(x=[1,2,3], y=[1,2,3], color=px.Constant("red"),  
color_discrete_map=px.IDENTITY_MAP)
```

(note that the foregoing could also apply to any groupable attrs beyond `color` like `symbol` or `line_dash`)

nicolaskruchten:
Closes #2119 ... @emmanuelle LMK what you think of the API and testing approach and if it makes sense I'll document it.

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -59,6 +59,12 @@ class IdentityMap(object):
         return self
 
 
+class Constant(object):
+    def __init__(self, value, label=None):
+        self.value = value
+        self.label = label
+
+
 MAPBOX_TOKEN = None
 
 
@@ -937,6 +943,8 @@ def build_dataframe(args, attrables, array_attrables):
         else:
             df_output[df_input.columns] = df_input[df_input.columns]
 
+    constants = dict()
+
     # Loop over possible arguments
     for field_name in attrables:
         # Massaging variables
@@ -968,8 +976,15 @@ def build_dataframe(args, attrables, array_attrables):
                     "pandas MultiIndex is not supported by plotly express "
                     "at the moment." % field
                 )
+            # ----------------- argument is a constant ----------------------
+            if isinstance(argument, Constant):
+                col_name = _check_name_not_reserved(
+                    str(argument.label) if argument.label is not None else field,
+                    reserved_names,
+                )
+                constants[col_name] = argument.value
             # ----------------- argument is a col name ----------------------
-            if isinstance(argument, str) or isinstance(
+            elif isinstance(argument, str) or isinstance(
                 argument, int
             ):  # just a column name given as str or int
                 if not df_provided:
@@ -1049,6 +1064,9 @@ def build_dataframe(args, attrables, array_attrables):
                 args[field_name] = str(col_name)
             else:
                 args[field_name][i] = str(col_name)
+
+    for col_name in constants:
+        df_output[col_name] = constants[col_name]
 
     args["data_frame"] = df_output
     return args
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1,6 +1,7 @@
 import plotly.graph_objs as go
 import plotly.io as pio
 from collections import namedtuple, OrderedDict
+from ._special_inputs import IdentityMap, Constant
 
 from _plotly_utils.basevalidators import ColorscaleValidator
 from .colors import qualitative, sequential
@@ -40,29 +41,6 @@ class PxDefaults(object):
 
 defaults = PxDefaults()
 del PxDefaults
-
-
-class IdentityMap(object):
-    """
-    `dict`-like object which can be passed in to arguments like `color_discrete_map` to
-    use the provided data values as colors, rather than mapping them to colors cycled
-    from `color_discrete_sequence`.
-    """
-
-    def __getitem__(self, key):
-        return key
-
-    def __contains__(self, key):
-        return True
-
-    def copy(self):
-        return self
-
-
-class Constant(object):
-    def __init__(self, value, label=None):
-        self.value = value
-        self.label = label
 
 
 MAPBOX_TOKEN = None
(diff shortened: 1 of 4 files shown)


packages/python/plotly/plotly/express/_core.py
```diff
@@ -41,6 +41,30 @@ def __init__(self):
 defaults = PxDefaults()
 del PxDefaults
 
+
+class IdentityMap(object):
```
emmanuelle: could these two classes be moved to a `_special_inputs.py` file to shorten a bit `_core.py`?

emmanuelle:
Thanks for the PR! I'm completely convinced by `px.Constant`, a bit less by `px.IdentityMap` which looks like a "dummy" class (it does not do much from what I understand of the code and tests). I'm curious to know why you chose a class instead of the `'identity'` string which we had mentioned. 

I'll do more QA on examples tomorrow, tonight I only had time for a quick pass reading the code.

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: sure, if needed

nicolaskruchten:
So `IdentityMap` isn't a dummy class, it's the actual implementation: each `IdentityMap` object returns the key on lookup, which is what we want in this case. I didn't implement the `"identity"` string thing yet just to see what this would feel like. I actually like this, personally, as it's a bit more Pythonic in a way, and doesn't have a "magic string" variable in a place where otherwise only a `dict`-like object is accepted. It's pretty trivial to implement the `"identity"` thing as well, just an `if` in `make_mapping()` somwhere to force `val_map` to be an `IdentityMap` when that string is encountered.

packages/python/plotly/plotly/express/_core.py
emmanuelle: as you wish, I just feel that the length of this file makes it a bit overwhelming.

emmanuelle:
Ah ok thanks for the explanations, I had not understood how the class worked. Maybe add a line with your explanation above in the docstring of the class?

emmanuelle:
So I've experimented a bit with your branch. It works well as expected. From a pure usability point of view, doing something like `fig = px.scatter(x=[1, 2], y=[3, 4], color=px.IdentityMap(['yellow', 'blue'])` would be slightly nicer than `fig = px.scatter(x=[1, 2], y=[3, 4], color=['yellow', 'blue'], color_discrete_map=px.IdentityMap())` but on the other hand the current implementation is simple and elegant (so simple I had originally thought that something was missing!). Thoughts?

nicolaskruchten:
Yep, `fig = px.scatter(x=[1, 2], y=[3, 4], color=px.IdentityMap(['yellow', 'blue'])` was one of the options I had listed in #2119 (as `px.Direct`) but after our conversation I thought that the second option seemed better...

nicolaskruchten:
OK I addressed your feedback, including the `"identity"` shortcut.

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: done!

emmanuelle:
> Yep, `fig = px.scatter(x=[1, 2], y=[3, 4], color=px.IdentityMap(['yellow', 'blue'])` was one of the options I had listed in #2119 (as `px.Direct`) but after our conversation I thought that the second option seemed better...

Oops... reading again our convo in #2119 I think I had preferred option 2 because it used a constant instead of a function or a class, but eventually it's not a constant...

nicolaskruchten:
> I had preferred option 2 because it used a constant instead of a function or a class, but eventually it's not a constant...

I added this in: you can use `"identity"` now also (under the hood it just swaps out an `IdentityMap` for you so you don't have to see it).

I can implement `px.Identity` and `px.Direct` still if you think it will help?

emmanuelle:
Thanks for implementing the `'identity'` thing. `px.Direct` could be useful indeed - I personally prefer using `fig.update_traces` but I understand that a fraction of users will want to do everything from `px`. As for `px.Identity` this is less necessary in my opinion, since when there are more than one colour people they will often exist already as a column in a DataFrame so people might as well do `color='my_color_column', color_discrete_map="identity"`. 

`px.Direct` can also come in another PR so feel free to merge this when you want. About docs: do you plan to add docs as part of this PR or together with one of the WIP PRs?

nicolaskruchten:
I was going to document everything in the last PR :)

nicolaskruchten:
Superseded by #2336

PR #2335, opened 2020-03-31
initial pass at PX auto-orientation

### LINKED ISSUE #2331: PX auto-detect orientation for box/violin/histogram
This shouldn't error out:

```python
import plotly.express as px
import numpy as np
px.histogram(y=np.random.rand(100))
```

And this shouldn't render a blank plot:

```python
import plotly.express as px
import numpy as np
px.box(x=np.random.rand(100))
```

Because in both cases PX should be smart enough to set `orientation`. In fact, even if both `x` and `y` are set, if one is categorical and the other isn't, we should be able to auto-detect the correct orientation (this would work well for `bar` also).

nicolaskruchten:
Closes #2331 

Docs and tests still missing, but this is the approach I was going to take... verbose but seems to work OK.


emmanuelle:
thanks for the PR! I'm curious to know why many things need to be specified and why one could not just pass `None` to plotly.js for the orientation? More a philosophical question I will dig more into the code later but I wonder why `px` needs to know about orientation. Is it for marginals / subplots for example?

nicolaskruchten:
Hmm good question. Early on in the PX development work this didn't work so well but maybe things have changed since then. I will try!

nicolaskruchten:
OK so Plotly.js will do a good job at orientations by default, but in order for PX to do stuff like setting marginals and bins on the correct axis, we need to set this explicitly for histograms, and to do the overlay trick in this PR, we need to know the orientation at the PX level for box/violin. So... yeah I think that doing this work in PX makes sense even though it's partly duplicating what Plotly.js is doing.

The overlay thing for box/violin is to avoid gross spacing problems if you have the base axis mapped to the same variable as colors... we have to set the mode to overlay explicitly right now which is really annoying:

image

image

nicolaskruchten:
Superseded by #2336

Issue #2224, opened 2020-02-24
Feature Request: Support list inputs for plotly express x, y arguments


samjett247:
First off, thanks for the awesome plotly.express package, its fulfilling a great need for interactive plotting capabilities. 

One problem I've encountered frequently is that I have variables I want to plot in separate columns in my dataframe, and I have to make a subplot to plot the distributions or correlations of all of these variables. For these purposes, it would be very helpful if `px.scatter`,  `px.box` , and the other px methods could accept the lists of column names as x or y variables. Note that this is different from faceting by column or row, because for the facet: (a) the yaxes and/or xaxes are typically shared, and (b) the "column" or row to be faceted on is a single column. Would love to hear the communities thoughts about this idea. Example code below.

```python
import plotly.express as px
from plotly.subplots import make_subplots
import pandas as pd

temp = [
    {"Clinic": "A", "Subject": "Bill", "Height(cm)": 182, "Weight(kg)": 101, "BloodPressureHigh": 128, "BloodPressureLow": 90},
    {"Clinic": "A", "Subject": "Susie", "Height(cm)": 142, "Weight(kg)": 67, "BloodPressureHigh": 120, "BloodPressureLow": 70},
    {"Clinic": "B", "Subject": "John", "Height(cm)": 202, "Weight(kg)": 89, "BloodPressureHigh": 118, "BloodPressureLow": 85},
    {"Clinic": "B", "Subject": "Stacy", "Height(cm)": 156, "Weight(kg)": 78, "BloodPressureHigh": 114, "BloodPressureLow": 76},
    {"Clinic": "B", "Subject": "Lisa", "Height(cm)": 164, "Weight(kg)": 59, "BloodPressureHigh": 112, "BloodPressureLow": 74} 
]
df = pd.DataFrame(temp)
# This is the call I want to be able to make
# px.box(data_frame=df, x='Clinic', y=['Height(cm)', 'Weight(kg)', "BloodPressureHigh", "BloodPressureLow"]) 

# Instead, this is what I have to do
fig = make_subplots(rows=1, cols=4)
for j,y in enumerate(['Height(cm)', 'Weight(kg)', "BloodPressureHigh", "BloodPressureLow"]):
    localfig = px.box(data_frame=df, x='Clinic', y=y) 
    # Or the below if you want to plot scatter
    # localfig = px.scatter(data_frame=df, x='Height(cm)', y=y)
    trace1 = localfig['data'][0]
[...]

PR #2336, opened 2020-03-31
PX Input upgrades - wide form, constant, identity, auto-orientation etc

### LINKED ISSUE #1735: Plotly for pandas
Hi Team,

Big thanks to bring such a good tool as plotly!

I have tried new plotly 4.1. It is excellent that you have split plotly and chartstudio.

Unfortunately I had to go back to version 3 as I could not run offline pandas.Dataframe.iplot() this was one of the best features plotly offered as you can with one line of code plot full dataframe additionally allowing to switch on/off each column in jupyter notebook.

I had to downgrade back to version 3 with conda command below:
conda install -c conda-forge python-cufflinks

Could you please advice if there is any way I can run iplot as in this notebook:
https://nbviewer.jupyter.org/gist/santosjorge/b278ce0ae2448f47c31d

### LINKED ISSUE #2005: Feature Request: Add plotly as `pandas.plotting.backend` option
Feature Request: Add plotly as `pandas.plotting.backend` option. 

We don't have to use Matplotlib as the plotting backend for pandas any longer. Holoviewsand Bokeh are options. Would love Plotly to be an option. 

See [this SO answer](https://stackoverflow.com/a/58440525/4590385) for a discussion.

Thank you for considering!

### LINKED ISSUE #2119: px.Constant and the "identity" problem

### LINKED ISSUE #2331: PX auto-detect orientation for box/violin/histogram

### LINKED ISSUE #2332: PX scatter, bar behaviour when only one of x or y is specified
Plotly.js defaults to assuming `x` or `y` are increasing integers if unspecified, but this is a bad fit for PX... consider the following case where coloring the markers results in their shifting position (note the x axis range!)

image

This case is even scarier, with individual bars have different heights based on position in the data frame:

image

For `bar`, `line`, `scatter` and `area` we should probably disallow specifying one but not both of `x` and `y`. Possibly in the 3d cases also? Anywhere else? Bar-like traces?

### LINKED ISSUE #2333: PX "wide-form" mode like Seaborn
Seaborn has a nice mode whereby in something like `px.line()`, if neither `x` nor `y` are specified (see also #2332!) it treats the data frame as "wide" rather than "long"/"tidy" and sets `x`=index, `y`=melted value and `color`=melted variable. We could do some variation on this as well.

nicolaskruchten:
Closes #2333 - wide mode
Closes #2119 - constant and identity
Closes #2331 - auto-orientation
Closes #2332 - `x` XOR `y` behaviour
Closes #2005 - Pandas backend
Closes #1735 - multiple columns for x, y & Pandas backend

Also addresses previously-resolved issues in a better way:
  * #2224 & https://github.com/plotly/plotly_express/issues/127- multiple columns for x, y
  * https://github.com/plotly/plotly_express/issues/126 - defaulting `x` to the index
  * https://github.com/plotly/plotly_express/issues/141 - simple single-column inputs
  * https://stackoverflow.com/questions/55822042/how-to-draw-a-multiple-line-chart-using-plotly-express
  * https://community.plotly.com/t/how-to-plot-multiple-lines-on-the-same-y-axis-using-plotly-express/29219
  * https://community.plotly.com/t/multiple-traces-plotly-express/23360

Note: the scope of "wide mode" has been broadly expanded to include all cartesian-2d PX functions other than `imshow`, and to include selection of `y=[col1, col2]` and optional overriding of the perpendicular `x` value to be something other than the index as well. This seems like an odd fit for `density_contour`/`density_heatmap` until you think of them as a drop-in replacements for `scatter` to deal with overplotting ;)

Todo:
- write medium post
- write forum post
- write readable docs
  - identity in the color section
  - pandas backend dedicated page
  - update px inputs page
  - remove "tidy" references everywhere


packages/python/plotly/plotly/express/_core.py
```diff
@@ -1031,6 +1064,32 @@ def build_dataframe(args, attrables, array_attrables):
...
             else:
                 args[field_name][i] = str(col_name)
+            wide_id_vars.add(str(col_name))
+
+    if wide_mode:
+        # TODO multi-level index
+        # TODO multi-level columns
+        index_name = df_output.index.name or "index"
+        wide_id_vars.add(index_name)
+        if index_name not in df_output.columns:
```
emmanuelle, line 1074: so if there is a column with name "index" it will be used for `x`, right?

emmanuelle: because otherwise this statement is always false I think

emmanuelle:
Thank you for an awesome PR. I'll give it another round of review with the docs and tests on wide format, but all this is going to be super useful and I think this integrates really well with the existing px code, in particular the "wide" part.

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: If something like `hover_name=df.index` happens this is needed to prevent the `melt()` from failing.

nicolaskruchten:
I think I will switch from `value` and `variable` to `_value_` and `_column_` to reduce collisions and make it clear these are “special” names.

nicolaskruchten:
Ask the last change I made stops something like `symbol=“variable”` from working so I will fix that too (during the `variable` to `_column_` rename)

nicolaskruchten:
@emmanuelle LMK if it's OK to approve the Percy mock change... basically with this PR we automatically set `histfunc` to `sum` if both x and y are provided to histogram. This is less surprising/more automatic IMO.

emmanuelle:
Also I'm not convinced by `px.Range`: to me it's a bit like reinventing `pd.RangeIndex`, the only difference (I think) being that we protect its name. I remember that we introduced the possibility to have arguments which were not column names, we decided not to keep the names of series if they were not a column of the dataframe argument. We could at some point revisit this decision. But anyway I don't think the convenience of `pd.RangeIndex` is worth the API clutter. Quite a matter of taste of course...

emmanuelle:
Regarding the "ugly names", could we disable the legend title and axis label when the names are `_value_` or `_column_`? This would be also closer to the behavior of `df.plot` (I just tested the matplotlib backend).

emmanuelle:
I'll read the doc when it's ready but I think I'm done for now. Very excited about the wide form :-).

packages/python/plotly/plotly/express/_core.py
```diff
@@ -15,6 +16,28 @@
...
+    "color",  # renamed to marker.color or line.color in infer_config
+    "symbol",  # renamed to marker.symbol in infer_config
+    "line_dash",  # renamed to line.dash in infer_config
+]
+all_attrables = (
+    direct_attrables + array_attrables + group_attrables + renameable_group_attrables
+)
+
+cartesians = [go.Scatter, go.Scattergl, go.Bar, go.Funnel, go.Box, go.Violin]
+cartesians += [go.Histogram, go.Histogram2d, go.Histogram2dContour]
```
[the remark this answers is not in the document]
nicolaskruchten: to avoid `black` making a bunch of sparse lines ;)

nicolaskruchten:
I'm glad you like it! I'm very pleased about how powerful and expressive this system is, even though there are a lot of corner cases internally for the various trace types which make it a less-than-elegant PR ;)

Some notes:
* `px.Constant()` was mostly useful for the `px.bar` case when you wanted all the `y` values to be set to 1... This is now happening internally but it'll still come up occasionally. It mostly replaces `x=[value]*len(arr)`
* `px.Range()` is mostly used internally when there is no data frame index present, but it also replaces something like `x=range(len(arr))` if you just want a sequence of numbers... 
* the "ugly names" are pretty important to display IMO for discoverability... if we don't show something in the axis/legend titles, people won't really ever know that you can relabel them with `labels` or that you can reassign them to other fields like `facet_col` etc, so I would not want to just hide them from the figure. Also, if we hide them, then people will have to figure out how to use `.update_xaxes(title="whatever")` in addition to messing with `hover_data` instead of using `labels=dict(_column_="whatever")` and hitting all cases...
* the "ugly names" themselves were chosen to suggest that they're "special" in some way and to avoid collisions with existing data frame columns called `value` and `column` but I could see us removing the underscores and adding checks for collisions instead. I think we would lose the "these are special values" feeling that we have now, however, which I would consider a downside...

Thoughts?

emmanuelle:
Regarding `_value_` and `_column_` I understand the arguments but I worry that people will think that the defaults of px figures are not aesthetically pleasing, and that we'll have a lot of figures on the Internet with `_value_` and `_column_` since most people don't change the defaults.

nicolaskruchten:
would you have the same perspective without the `_`underscores`_` ?

emmanuelle:
Oh I'd be much happier without the underscores :-).

emmanuelle:
But don't hesitate to take other opinions (looking at you @jonmmease) since it's a matter of taste.

nicolaskruchten:
OK, I'll see what it takes to handle collisions. I do think there is a positive aspect to the ugliness of the underscores to indicate that they're autogenerated, but I do take your point about people using defaults. OTOH we can work around that with docs a bit.

jonmmease:
That's tough. But I think I would lean toward omitting the underscores if you can work out something reasonable to do on collisions. I agree with @emmanuelle that it would be a bit annoying to see the underscore labels all over the place :slightly_smiling_face:

nicolaskruchten:
OK, well in any case I should probably build some collision-detection into this thing, so I might as well get that working.

nicolaskruchten:
OK @jonmmease @emmanuelle I've cleared up the `_uglynames_` ... we now use:
* `value` unless there is a column named value at which point we use `_value`
* if there's an index name we use that, unless there's a column with the same name, at which point we use `index`, and if there's no index name we use `index` and if there is also a column named `index` we use `_index`
* if there's column-index name we use that, unless there's a column with the same name, at which point we use `variable`, and if there's no column-index name we use `variable` and if there is also a column named `variable` we use `_variable`

I think that sorts out most of the possible collision cases, and in fact clears up some pre-existing corner cases outside of wide-mode when you were using `df.index` but also had a column named `index` 🤯

packages/python/plotly/plotly/express/_core.py
```diff
@@ -869,46 +910,42 @@ def _get_reserved_col_names(args, attrables, array_attrables):
     return reserved_names
 
 
-def build_dataframe(args, attrables, array_attrables):
-    """
-    Constructs a dataframe and modifies `args` in-place.
+def _is_col_list(df_input, arg):
```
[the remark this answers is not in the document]
nicolaskruchten: done

packages/python/plotly/plotly/data/__init__.py
```diff
@@ -102,7 +102,31 @@ def carshare():
...
+
+
+def experiment():
+    """
+Each row in this wide dataset represents the results of 100 simulated participants
+on three hypothetical experiments, along with their gender and smoker status.
+
+Returns:
+    A `pandas.DataFrame` with 100 rows and the following columns:
+    `['experiment_1', 'experiment_2', 'experiment_3', 'gender', 'smoker']`.
```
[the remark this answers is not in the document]
nicolaskruchten: done

doc/python/px-arguments.md
```diff
@@ -144,6 +176,16 @@ fig = px.line(x=[1, 2, 3, 4], y=[3, 5, 4, 8])
 fig.show()
 ```
 
+```python
+import plotly.express as px
+
+# List arguments in wide form
+series1 = [3, 5, 4, 8]
+series2 = [5, 4, 8, 3]
+fig = px.line(x=[1, 2, 3, 4], y=[series1, series2])
```
chriddyp, line 185: Wow, a bold new era!

doc/python/radar-chart.md
```diff
@@ -35,11 +35,11 @@ jupyter:
 
 A [Radar Chart](https://en.wikipedia.org/wiki/Radar_chart) (also known as a spider plot or star plot) displays multivariate data in the form of a two-dimensional chart of quantitative variables represented on axes originating from the center. The relative position and angle of the axes is typically uninformative. It is equivalent to a [parallel coordinates plot](/python/parallel-coordinates-plot/) with the axes arranged radially.
 
-For a Radar Chart, use a [polar chart](/python/polar-chart/) with categorical angular variables, with `px.line_polar` for data available as a tidy pandas DataFrame, or with `go.Scatterpolar` in the general case. See more examples of [polar charts here](/python/polar-chart/).
+For a Radar Chart, use a [polar chart](/python/polar-chart/) with categorical angular variables, with `px.line_polar`, or with `go.Scatterpolar`. See more examples of [polar charts here](/python/polar-chart/).
```
chriddyp, line 38:
```suggestion
For a Radar Chart, use a [polar chart](/python/polar-chart/) with categorical angular variables, with `px.line_polar`, or with `go.Scatterpolar`. See [more examples of polar charts](/python/polar-chart/).
```

doc/python/wide-form.md
```diff
@@ -0,0 +1,260 @@
...
+Plotly Express figures created using wide-form data can be [styled just like any other Plotly Express figure](/python/styling-plotly-express/):
+
+```python
+import plotly.express as px
+wide_df = px.data.short_track_wide()
+
+fig = px.bar(wide_df, x="nation", y=["gold", "silver", "bronze"], 
+             title="Wide-Form Input, styled",
+             labels={"value": "Medal Count", "variable": "Medal", "nation": "Olympic Nation"},
+             color_discrete_map={"gold":"gold", "silver": "silver", "bronze": "#c96"},
```
chriddyp, line 107:
```suggestion
             color_discrete_map={"gold": "gold", "silver": "silver", "bronze": "#c96"},
```

packages/python/plotly/plotly/data/__init__.py
```diff
@@ -102,6 +102,76 @@ def carshare():
...
+    `['experiment_1', 'experiment_2', 'experiment_3', 'gender', 'group']`.
+    If `indexed` is True, the data frame index is named "participant"
+"""
+    df = _get_dataset("experiment")
+    if indexed:
+        df.index.name = "participant"
+    return df
+
+
+def short_track_wide(indexed=False):
```
emmanuelle, line 139: How about changing `short_track` to `skating`? Shorter, and easier to understand when you don't know much about the different flavours of skating.

doc/python/wide-form.md
```diff
@@ -0,0 +1,260 @@
...
+```python
+import plotly.express as px
+mixed_df = px.data.experiment(indexed=True)
+
+fig = px.violin(mixed_df, y=["experiment_1", "experiment_2", "experiment_3"], 
+                color="gender", facet_col="group", hover_data=[mixed_df.index])
+fig.show()
+```
+
+And of course, we can reassign `variable` to another argument as well, in this case we'll assign it to `x` and facet by the wide variable, and we'll switch to a [box plot](/python/box-plots/) for variety.
```
emmanuelle, line 236: I don't understand this sentence ("we can reassign `variable` to another argument as well, in this case we'll assign it to `x`").

packages/python/plotly/plotly/__init__.py
```diff
@@ -75,3 +75,75 @@
         ],
         [".version.__version__"],
     )
+
+
+def plot(data_frame, kind, **kwargs):
+    """
+    Pandas plotting backend function, not meant to be called directly.
+    To activate, set pandas.options.plotting.backend="plotly.express.pandas_backend"
```
emmanuelle, line 83:
```suggestion
    To activate, set pandas.options.plotting.backend="plotly"
```

packages/python/plotly/plotly/__init__.py
```diff
@@ -75,3 +75,75 @@
...
+    if kind == "barh":
+        return bar(data_frame, orientation="h", **kwargs)
+    if kind == "box":
+        new_kwargs = {k: kwargs[k] for k in kwargs if k not in ["by"]}
+        return box(data_frame, **new_kwargs)
+    if kind in "hist":
+        new_kwargs = {k: kwargs[k] for k in kwargs if k not in ["by", "bins"]}
+        return histogram(data_frame, **new_kwargs)
+    raise NotImplementedError(
+        "The plotly.express backend doesn't yet support kind='%s'" % kind
```
emmanuelle, line 106:
```suggestion
        "The plotly backend doesn't yet support kind='%s'" % kind
```

packages/python/plotly/plotly/__init__.py
```diff
@@ -75,3 +75,75 @@
...
+    )
+
+
+def boxplot_frame(data_frame, **kwargs):
+    """
+    Pandas plotting backend function, not meant to be called directly.
+    To activate, set pandas.options.plotting.backend="plotly.express.pandas_backend"
```
emmanuelle, line 113:
```suggestion
    To activate, set pandas.options.plotting.backend="plotly"
```

emmanuelle: and in the other functions as well

packages/python/plotly/plotly/data/__init__.py
nicolaskruchten: Good feedback, thanks! renamed to `medals`

doc/python/wide-form.md
nicolaskruchten: yeah that sentence made no sense... fixed!

emmanuelle:
:woman_dancing: for an epic PR!

PR #2342, opened 2020-03-31
minimal example with customdata for sankey nodes

emmanuelle:


nicolaskruchten:
Thanks! Merged with some additions

PR #2347, opened 2020-04-02
Create announcement_4.6.md

nicolaskruchten:


announcement_4.6.md
```diff
@@ -0,0 +1,107 @@
...
+if you run into trouble, check out our [Troubleshooting Guide](https://plot.ly/python/troubleshooting/).
+
+# What's new in Plotly.py 4.6
+
+Our [changelog](https://github.com/plotly/plotly.py/releases/tag/v4.6.0) has links to individual pull requests, 
+but here are the highlights:
+
+### Unified Hover Labels
+
+Until today hovering on plots resulted in either one more multiple hover labels, each always attached to a single point.
```
emmanuelle, line 13: one or multiple?

nicolaskruchten: hehe ok that sentence is correct I think but hard to parse. It's intended to be read as `Until today hovering on plots resulted in either (one more multiple hover labels), (each always attached to a single point).`

PR #2355, opened 2020-04-03
clean up ci jobs

nicolaskruchten:
as discussed in the meeting, plus uncommenting something that seems like it should work.


nicolaskruchten:
OK, got this passing. Needed to find a set of dependencies that would install correctly with both 2.7 and 3.7, and I did.

emmanuelle:
Thank you! Great that this shapely thing is not bothering us any more. :dancer:

PR #2357, opened 2020-04-04
fix bug in trendline in the case of missing values

emmanuelle:
See https://community.plotly.com/t/plotly-trendlines-not-displaying-correctly-due-to-np-nan-values/37202

The case of nan values was not handled properly for trendlines:
- for lowess, the default value of statsmodels is to drop nans which is fine, but then the x values were not properly aligned with the y
- for ols, the default value is to do nothing with nans, resulting in all nans in the estimation.

This PR fixes the two cases and adds a test.

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -241,18 +241,22 @@ def make_trace_kwargs(args, trace_spec, trace_data, mapping_labels, sizeref):
                     sorted_trace_data = trace_data.sort_values(by=args["x"])
                     y = sorted_trace_data[args["y"]]
                     x = sorted_trace_data[args["x"]]
-                    trace_patch["x"] = x
+                    # trace_patch["x"] = x
 
                     if x.dtype.type == np.datetime64:
                         x = x.astype(int) / 10 ** 9  # convert to unix epoch seconds
 
                     if attr_value == "lowess":
-                        trendline = sm.nonparametric.lowess(y, x)
+                        # missing ='drop' is the default value for lowess but not for OLS (None)
+                        # we force it here in case statsmodels change their defaults
+                        trendline = sm.nonparametric.lowess(y, x, missing='drop')
+                        trace_patch["x"] = trendline[:, 0]
                         trace_patch["y"] = trendline[:, 1]
                         hover_header = "<b>LOWESS trendline</b><br><br>"
                     elif attr_value == "ols":
-                        fit_results = sm.OLS(y.values, sm.add_constant(x.values)).fit()
+                        fit_results = sm.OLS(y.values, sm.add_constant(x.values), missing='drop').fit()
                         trace_patch["y"] = fit_results.predict()
+                        trace_patch["x"] = x[np.logical_not(np.logical_or(np.isnan(y), np.isnan(x)))]
                         hover_header = "<b>OLS trendline</b><br>"
                         hover_header += "%s = %g * %s + %g<br>" % (
                             args["y"],
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -249,14 +249,18 @@ def make_trace_kwargs(args, trace_spec, trace_data, mapping_labels, sizeref):
                     if attr_value == "lowess":
                         # missing ='drop' is the default value for lowess but not for OLS (None)
                         # we force it here in case statsmodels change their defaults
-                        trendline = sm.nonparametric.lowess(y, x, missing='drop')
+                        trendline = sm.nonparametric.lowess(y, x, missing="drop")
                         trace_patch["x"] = trendline[:, 0]
                         trace_patch["y"] = trendline[:, 1]
                         hover_header = "<b>LOWESS trendline</b><br><br>"
                     elif attr_value == "ols":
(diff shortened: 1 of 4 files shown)


emmanuelle:
@nicolaskruchten I think this PR is ready

packages/python/plotly/plotly/express/_core.py
```diff
@@ -241,18 +241,26 @@ def make_trace_kwargs(args, trace_spec, trace_data, mapping_labels, sizeref):
                     sorted_trace_data = trace_data.sort_values(by=args["x"])
                     y = sorted_trace_data[args["y"]]
                     x = sorted_trace_data[args["x"]]
-                    trace_patch["x"] = x
+                    # trace_patch["x"] = x
```
nicolaskruchten, line 244:
```suggestion
```

packages/python/plotly/plotly/express/_core.py
```diff
@@ -241,18 +241,26 @@ def make_trace_kwargs(args, trace_spec, trace_data, mapping_labels, sizeref):
...
 
                     if x.dtype.type == np.datetime64:
                         x = x.astype(int) / 10 ** 9  # convert to unix epoch seconds
 
                     if attr_value == "lowess":
-                        trendline = sm.nonparametric.lowess(y, x)
+                        # missing ='drop' is the default value for lowess but not for OLS (None)
+                        # we force it here in case statsmodels change their defaults
+                        trendline = sm.nonparametric.lowess(y, x, missing="drop")
+                        trace_patch["x"] = trendline[:, 0]
```
nicolaskruchten, line 253: why do we not need to filter the x data here?

nicolaskruchten: Oh, I see, LOWESS will just pass those through in the estimates... hmm that means that with NaNs there are gaps in the line. OK.

packages/python/plotly/plotly/tests/test_core/test_px/test_trendline.py
```diff
@@ -0,0 +1,12 @@
+import plotly.express as px
+import numpy as np
+
+
+def test_trendline_nan_values():
```
nicolaskruchten, line 5: can we expand this test to cover both trendline types, and assert that the lengths of x and y are the same? I think that was not true before but should always be.

emmanuelle: Done ! Indeed x and y could have different lengths before this PR.

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: so `missing="drop"` means "carry NaNs through" in LOWESS?

emmanuelle: So for lowess
```
missing : str
    Available options are 'none', 'drop', and 'raise'. If 'none', no nan
    checking is done. If 'drop', any observations with nans are dropped.
    If 'raise', an error is raised. Default is 'drop'.
```
nans are just removed, and 'drop' is the default value (so I did not change anything for this function call for lowess, I just made the option explicit in case `sm` changes the default, since I was quite surprised that the default was different for `ols`). But I had to make sure we take the `trace_patch["x"]` from the `sm` returned object where NaNs have been dropped, because we used to take the original x-data which could have NaNs and then x and y did not have the same length.

For ols
```
missing : str
    Available options are 'none', 'drop', and 'raise'. If 'none', no nan
    checking is done. If 'drop', any observations with nans are dropped.
    If 'raise', an error is raised. Default is 'none'.
```
so I chose `drop`. It's weird that it's not the same default... (statsmodels version 0.11.1).

Does this answer your question? No we're not carrying NaNs (as we were previously for ols) through but I don't think we should.

packages/python/plotly/plotly/tests/test_core/test_px/test_trendline.py
```diff
@@ -0,0 +1,15 @@
...
+    df = px.data.gapminder().query("continent == 'Oceania'")
+    start_date = 1970
+    df["pop"][df["year"] < start_date] = np.nan
+    modes = ["ols", "lowess"]
+    for mode in modes:
+        fig = px.scatter(df, x="year", y="pop", color="country", trendline=mode)
+        country_numbers = len(fig["data"]) // 2
```
nicolaskruchten, line 12: what does this line do?

packages/python/plotly/plotly/tests/test_core/test_px/test_trendline.py
nicolaskruchten: awesome, thanks!

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: yep, all clear now!

nicolaskruchten:
💃

packages/python/plotly/plotly/tests/test_core/test_px/test_trendline.py
emmanuelle, line 12:
```suggestion
```

emmanuelle: Nothing :-). So I removed it. Thanks!

PR #2372, opened 2020-04-11
Update sunburst-charts.md

emmanuelle:
@nicolaskruchten please confirm that this was a typo and my change is correct.

### DIFF
--- a/doc/python/sunburst-charts.md
+++ b/doc/python/sunburst-charts.md
@@ -196,7 +196,7 @@ fig.show()
 
 With branchvalues "total", the value of the parent represents the width of its wedge. In the example below, "Enoch" is 4 and "Awan" is 6 and so Enoch's width is 4/6ths of Awans. With branchvalues "remainder", the parent's width is determined by its own value plus those of its children. So, Enoch's width is 4/10ths of Awan's (4 / (6 + 4)).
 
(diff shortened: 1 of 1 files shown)


doc/python/sunburst-charts.md
```diff
@@ -196,7 +196,7 @@ fig.show()
 
 With branchvalues "total", the value of the parent represents the width of its wedge. In the example below, "Enoch" is 4 and "Awan" is 6 and so Enoch's width is 4/6ths of Awans. With branchvalues "remainder", the parent's width is determined by its own value plus those of its children. So, Enoch's width is 4/10ths of Awan's (4 / (6 + 4)).
 
-Note that this means that the sum of the values of the children cannot exceed the value of their parent when branchvalues "total". When branchvalues "relative" (the default), children will not take up all of the space below their parent (unless the parent is the root and it has a value of 0).
+Note that this means that the sum of the values of the children cannot exceed the value of their parent when branchvalues "total". When branchvalues "remainder" (the default), children will not take up all of the space below their parent (unless the parent is the root and it has a value of 0).
```
nicolaskruchten, line 199:
```suggestion
Note that this means that the sum of the values of the children cannot exceed the value of their parent when branchvalues is set to "total". When branchvalues is set to "remainder" (the default), children will not take up all of the space below their parent (unless the parent is the root and it has a value of 0).
```

nicolaskruchten:
Yes, this is correct! Can we apply my change and make sure the treemap docs are corrected as well if they have these same errors please?

emmanuelle:
Change applied, treemap was already ok.

nicolaskruchten:
:dancer:

PR #2374, opened 2020-04-11
Fixed code block in "Displaying Figures Using Dash" on Renderers help page

supRy:
No code was changed. Just edited the markdown to show the code block correctly.

Since this is a minor change, I am not including the checklist. I didn't run any build, I wasn't sure if that was needed with this minor change.  Thanks!

### DIFF
--- a/doc/python/renderers.md
+++ b/doc/python/renderers.md
@@ -235,7 +235,6 @@ It is important to note that Dash does not use the renderers framework discussed
 
 Instead, pass your figure as the `figure` parameter to the [`dcc.Graph`](https://dash.plot.ly/dash-core-components/graph) component, which is part of the [Dash Core Components](https://dash.plot.ly/dash-core-components) library. The code below demonstrates how to do this. 
 
-```python
 import dash_core_components as dcc
 import plotly.graph_objs as go
 
--- a/doc/python/renderers.md
+++ b/doc/python/renderers.md
@@ -235,6 +235,7 @@ It is important to note that Dash does not use the renderers framework discussed
 
 Instead, pass your figure as the `figure` parameter to the [`dcc.Graph`](https://dash.plot.ly/dash-core-components/graph) component, which is part of the [Dash Core Components](https://dash.plot.ly/dash-core-components) library. The code below demonstrates how to do this. 
 
+```
 import dash_core_components as dcc
 import plotly.graph_objs as go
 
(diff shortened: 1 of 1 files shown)


emmanuelle:
Hi @supRy thank you for your interest in plotly.py and its documentation! Unfortunately, this formatting was made on purpose since we do not have a Dash server running in the documentation so we did not want this block to be executed when the notebook is executed by the documentation build.

Did you find this formatting misleading? Maybe we can improve this example and its presentation.

supRy:
Yeah, it a bit jarring to see code not formatted when everything else is.  Is there a way to exclude that codeblock from the CI?

doc/python/renderers.md
```diff
@@ -235,7 +235,7 @@ It is important to note that Dash does not use the renderers framework discussed
 
 Instead, pass your figure as the `figure` parameter to the [`dcc.Graph`](https://dash.plot.ly/dash-core-components/graph) component, which is part of the [Dash Core Components](https://dash.plot.ly/dash-core-components) library. The code below demonstrates how to do this. 
 
-<!-- #raw -->
+```python
```
emmanuelle, line 238:
```suggestion
```
```

Comment (emmanuelle) [inline doc/python/renderers.md:238]:
```suggestion
```
```

File: doc/python/renderers.md [hunk 2]
```diff
@@ -235,7 +235,6 @@ It is important to note that Dash does not use the renderers framework discussed
 
 Instead, pass your figure as the `figure` parameter to the [`dcc.Graph`](https://dash.plot.ly/dash-core-components/graph) component, which is part of the [Dash Core Components](https://dash.plot.ly/dash-core-components) library. The code below demonstrates how to do this. 
 
-<!-- #raw -->
 import dash_core_components as dcc
```
emmanuelle, line 238:
```suggestion
```
import dash_core_components as dcc
```

emmanuelle:
Hi @supRy I found out that if you just use backticks without specifying a language it is rendered as code but not executed which is exactly what we were looking for. I took the liberty to push to your branch, I hope you don't mind (in fact I hadn't realized first that the formatting was so bad, I thought it was just a bit different but hadn't seen that there were no line breaks). Thank you very much for reporting the problem and paving the way to a solution.

supRy:
Great! Thanks so much for figuring that out!

PR #2377, opened 2020-04-13
Hover format or skip

### LINKED ISSUE #1774: Formatting and hiding hover_data in PX
Let's implement the API from https://github.com/plotly/plotly_express/issues/115 and support `False` as a value so as to support https://github.com/plotly/plotly_express/issues/120.

emmanuelle:
Closes #1774 

Still need to add doc (and probably more tests), but early feedback is welcome.


emmanuelle:
I'm a bit surprised by the test failure on python 2.7 (why would a hovertemplate be different with python 2.7 and python 3?)

nicolaskruchten:
The Python 2.7 error is just an ordering problem due to `dict` order (in)stability... 

image

emmanuelle:
ahhh thank you :-). I'll change the test then. Python 3 users won't have the problem since dicts are insertion-ordered, and py2 users can always use an OrderedDict instead.

nicolaskruchten:
So this approach looks pretty good. I don't love the string manipulations, although they seem pretty safe!

One wrinkle: we support two different kinds of formatting... d3-format for numbers and d3-time-format for dates, and the delimiter between the attribute name and the formatting string is different (`:` vs `|`) ... how can we account for this? I think we should force the user to include the first character, at least in the d3-time-format case.

nicolaskruchten:
Example for date axes:

image

emmanuelle:
Sure, let's add the delimiter. As for string manipulations, maybe there's a cleaner way to do it, this was a first pass to see if it worked.

packages/python/plotly/plotly/express/_doc.py
```diff
@@ -180,8 +180,12 @@
         "Values from this column or array_like appear in bold in the hover tooltip.",
     ],
     hover_data=[
-        colref_list_type,
-        colref_list_desc,
+        "list of str or int, or Series or array-like, or dict",
+        "Either a list of names of columns in `data_frame`, or pandas Series,",
+        "or array_like objects",
+        "or a dict with column names as keys, with values True (for default formatting)",
```
nicolaskruchten, line 186: A few questions:

1. The intent here is that for a scatter, `hover_data=dict(y=False)` would still leave the x in the hovertemplate, right? The implied default is `True`?
2. I imagine if that I wanted a non-data-frame column with custom formatting, I could do something like `custom_data=[ [1,2,3] ], hover_data=dict(custom_data_0 = ".2f"), labels=dict(custom_data_0="my thing")`?

I wonder if we shouldn't allow a tuple instead of just a string value, so people could provide named data in hover like `hover_data={"my thing": (".2f", [1,2,3])}`...

Here are the cases I can imagine:

1. I want to hide a column: `False`
2. I want to format an existing, named column: `".2f"`
3. I want to provide a new named column: not possible now with just `hover_data`
3. I want to provide a new named column and formatting data: not possible now with just `hover_data`

emmanuelle: interesting, I'll explore the tuple idea.

nicolaskruchten: Does it seem like a too-complex signature to accept:

1. Booleans for on/off
2. strings for formatting
3. list-like for data
4. tuples for formatting and data

?

emmanuelle:
@nicolaskruchten I think this one is ready for review.

nicolaskruchten:
Can we also accept something like `hover_data = dict(new_col=[1,2,3,4,5])`, to be treated the same as `hover_data=dict(new_col=(True, [1,2,3,4,5]))` ?

emmanuelle:
> Can we also accept something like `hover_data = dict(new_col=[1,2,3,4,5])`, to be treated the same as `hover_data=dict(new_col=(True, [1,2,3,4,5]))` ?

I could maybe be convinced, but I liked the idea that the value is a formatting information, except when it's a tuple, in which case it's (formatting, data). It's also the logics which I have used in the code, converting everything to tuples internally.

nicolaskruchten:
Right, but I feel like if we've already got a `dict` here, we might as well give the user a shortcut to `hover_data=[ <list> ], labels=dict(hover_data_0="name")` in the form of `hover_data={"name": <list> }`. We can internally convert `<list>` to `(True, <list>)` for them.

packages/python/plotly/plotly/tests/test_core/test_px/test_px_hover.py
```diff
@@ -0,0 +1,88 @@
...
+            x="a",
+            y="b",
+            hover_data={"d": ":.1f"},
+        )
+    with pytest.raises(ValueError):
+        fig = px.scatter(
+            {"a": [1, 2], "b": [3, 4], "c": [2, 1]},
+            x="a",
+            y="b",
+            hover_data={"d": (True, [3, 4, 5])},
```
nicolaskruchten, line 87: this fails because of the length of the list, right?

nicolaskruchten:
Not quite done the review, but it looks great so far! We need to merge this in before #2336, as I think I'll have an easier time rebasing :)

packages/python/plotly/plotly/tests/test_core/test_px/test_px_hover.py
emmanuelle: yes !

packages/python/plotly/plotly/express/_doc.py
nicolaskruchten: awesome!

doc/python/hover-text-and-formatting.md
```diff
@@ -138,6 +138,35 @@ fig = px.scatter(df_2007, x="gdpPercap", y="lifeExp", log_x=True,
 fig.show()
 ```
 
+### Disabling or customizing hover of columns in plotly express
```
nicolaskruchten, line 141: Since we're in the neighbourhood, could you please add a new section saying something like "PX uses `hovertemplate` to format its hovers, and you may disable or overwrite this with `.update_traces(hovertemplate='whatever')` and provide a short example showing the actual `hovertemplate` from `fig.data.[0]` and then showing how to change it?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -387,7 +390,20 @@ def make_trace_kwargs(args, trace_spec, trace_data, mapping_labels, sizeref):
         go.Parcoords,
         go.Parcats,
     ]:
-        hover_lines = [k + "=" + v for k, v in mapping_labels.items()]
+        # Modify mapping_labels according to hover_data keys
+        # if hover_data is a dict
+        mapping_labels_copy = OrderedDict(mapping_labels)
```
nicolaskruchten, line 395: nitpick: we can just let this be called `mapping_labels` as in this context it's already a copy

packages/python/plotly/plotly/express/_core.py
```diff
@@ -869,6 +885,10 @@ def _get_reserved_col_names(args, attrables, array_attrables):
     return reserved_names
 
 
+def _isinstance_listlike(x):
+    return isinstance(x, list) or isinstance(x, np.ndarray) or isinstance(x, pd.Series)
```
nicolaskruchten, line 889: a slightly more permissive approach would be something like `False` if it's a string, then `True` if `iter(x)` doesn't raise an exception. This would allow through more stuff like `Series` from non-Pandas libraries etc. Of course it would allow through stuff like `dict`s but those would fail later on anyway.

nicolaskruchten:
Huh, it took me a while to understand how the dataframe loop was working, and then I realized that this happens:

image

nicolaskruchten:
So when we iterate through an "array attrable" and it's a dict, it sort of "just works" :) nice!

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: Ah but we do a `tuple` check later on. Hmm maybe we can check for a length-2-tuple first and then do this ?

emmanuelle:
> So when we iterate through an "array attrable" and it's a dict, it sort of "just works" :) nice!

Yes when you iterate through a dict it iterates through its keys. Simple and elegant ;-).

packages/python/plotly/plotly/express/_core.py
emmanuelle: I was trying to understand which objects can be added in a column of a dataframe, and it occurred to me that the only reason why we need `px.Constant` is that we error out on
```
px.scatter(x=[1, 2, 3], y=1])
```
or 
```
px.bar(x='Monday', y=[1, 2, 3])
```
because if we did not catch this we could just the let the value be added for all rows. No regrets here @nicolaskruchten ? I think I'm fine with raising and having px.Constant since this is more robust to typos, but it could also be a lighter way of calling functions so I'd like to be sure we discuss this.

doc/python/hover-text-and-formatting.md
emmanuelle: done

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: well `y=1` is used to refer to column names that are `1` already...

nicolaskruchten: I suppose we could be "smart" about it and automatically wrap `y=1` in a `px.Constant()` if there is no data frame or if there is no df column named `1` ... ?

nicolaskruchten: if we start going down that road we might be able to make `identity` be automatic too :)

emmanuelle: well these are suggestions, but personally I'm happier if px is not too magical, ie if it's not the viz equivalent of pandas (where almost everything is possible but you don't really understand why).

emmanuelle: regarding your initial suggestion I implemented the more permissive version of the function.

nicolaskruchten: Cool, this works for me. I also worry about too much magic :)

nicolaskruchten:
OK, this is awesome :)

:dancer:

emmanuelle:
Thanks for the :woman_dancing: let me just add the changelog entry and I'll merge.

Issue #2389, opened 2020-04-16
ImportError with tab completion


emmanuelle:
How to reproduce: 
* fresh pip env with python 3.7
* pip install ipython and jupyter
* install development version of plotly (master)
Inside `ipython`, do `import plotly.` and hit tab

@jonmmease probably related to the import optimization PR?
```
In [2]: import plotly.                                                           
Traceback (most recent call last):
  File "/home/emma/sandbox/venv_new/bin/ipython3", line 10, in <module>
    sys.exit(start_ipython())
  File "/home/emma/sandbox/venv_new/lib/python3.7/site-packages/IPython/__init__.py", line 126, in start_ipython
    return launch_new_instance(argv=argv, **kwargs)
  File "/home/emma/sandbox/venv_new/lib/python3.7/site-packages/traitlets/config/application.py", line 664, in launch_instance
    app.start()
  File "/home/emma/sandbox/venv_new/lib/python3.7/site-packages/IPython/terminal/ipapp.py", line 356, in start
    self.shell.mainloop()
  File "/home/emma/sandbox/venv_new/lib/python3.7/site-packages/IPython/terminal/interactiveshell.py", line 558, in mainloop
    self.interact()
  File "/home/emma/sandbox/venv_new/lib/python3.7/site-packages/IPython/terminal/interactiveshell.py", line 541, in interact
    code = self.prompt_for_code()
  File "/home/emma/sandbox/venv_new/lib/python3.7/site-packages/IPython/terminal/interactiveshell.py", line 469, in prompt_for_code
    **self._extra_prompt_options())
  File "/home/emma/sandbox/venv_new/lib/python3.7/site-packages/prompt_toolkit/shortcuts/prompt.py", line 994, in prompt
    return self.app.run(set_exception_handler=set_exception_handler)
  File "/home/emma/sandbox/venv_new/lib/python3.7/site-packages/prompt_toolkit/application/application.py", line 812, in run
    self.run_async(pre_run=pre_run, set_exception_handler=set_exception_handler)
  File "/usr/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "/home/emma/sandbox/venv_new/lib/python3.7/site-packages/prompt_toolkit/application/application.py", line 778, in run_async
    return await _run_async2()
  File "/home/emma/sandbox/venv_new/lib/python3.7/site-packages/prompt_toolkit/application/application.py", line 766, in _run_async2
[...]

PR #2390, opened 2020-04-16
Remove plotly express from delayed submodule import.

jonmmease:
Follow on to https://github.com/plotly/plotly.py/pull/2368 to address https://github.com/plotly/plotly.py/issues/2389.

The `express` module was mistakenly added as a submodule to auto-import on demand. This module was not previously auto-imported because it requires pandas.  It seems that requesting tab completion in ipython from an environemnt without pandas installed triggers an error here.

### DIFF
--- a/packages/python/plotly/plotly/__init__.py
+++ b/packages/python/plotly/plotly/__init__.py
@@ -72,7 +72,6 @@ else:
             ".io",
             ".data",
             ".colors",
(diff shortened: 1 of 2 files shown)


emmanuelle:
problem fixed in my env without pandas :-). :dancer:

jonmmease:
Great!

PR #2396, opened 2020-04-20
Add CHANGELOG step to PR template

nicolaskruchten:
@emmanuelle @jonmmease what do think of this process change?

If you agree, could you please add such an entry to the in-progress https://github.com/plotly/plotly.py/pull/2377 hover_data PR @emmanuelle  and @jonmmease can you please PR one in for your #2368 (optimization) changes?


nicolaskruchten:
Note: we would just have a new section at the top called "Not yet released" :)

emmanuelle:
Looks good to me! It will indeed make the release process less tedious. :dancer:

PR #2397, opened 2020-04-20
short-circuit instead of bypass flag

nicolaskruchten:
Just a little refactoring proposal to be merged into #2377 before it hits master.


emmanuelle:
good idea, thanks! Merging so that I can finish the work on #2377 :dancer:

PR #2402, opened 2020-04-21
title.text everywhere in PX

nicolaskruchten:
PX should use `title.text` instead of `title` everywhere... This should be a no-op as the validators handle this internally, but it's cleaner for PX not to rely on that so much IMO.


emmanuelle:
:dancer: thanks!

PR #2409, opened 2020-04-24
Update animations.md

romesc-CMU:
In the 'gapminder' example, the current code snippets appear to assign a list to `fig_dict["layout"]["sliders"]` twice. Once when the figure layout is initially created, and then again when more information is added to `sliders_dict`. This dictionary entry is overwritten before the figure is constructed. My proposed change just removes the initial assignment which seems to be of no consequence. Striving for clarity!


nicolaskruchten:
Thanks very much for catching this! :)

PR #2419, opened 2020-04-28
Sg scraper

emmanuelle:
This is an improvement of the sphinx-gallery scraper, which crawls only the directory of the example and now works for different structures of galleries (I was trying to add a plotly example to the gallery of scikit-image and it did not work with the scraper, with this change it is working).

A more sustainable mid-term change would be to add a `_repr_html_` method to the `BaseFigure` class (since the html would then be captured by sphinx-gallery) but for now the scraper works.

### DIFF
--- a/packages/python/plotly/plotly/io/_sg_scraper.py
+++ b/packages/python/plotly/plotly/io/_sg_scraper.py
@@ -1,7 +1,7 @@
 # This module defines an image scraper for sphinx-gallery
 # https://sphinx-gallery.github.io/
 # which can be used by projects using plotly in their documentation.
-import inspect, os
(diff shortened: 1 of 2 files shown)


emmanuelle:
oops I need to fix the tests!

nicolaskruchten:
💃 when the tests pass, so long as this is confirmed to be a backwards-compatible change ... ?

emmanuelle:
yes this is backward-compatible! Just a less convoluted (and more robust for some galleries) way of scraping png and html files.

nicolaskruchten:
Oops forgot to ask for a change log item :)

PR #2421, opened 2020-04-29
Repr html

emmanuelle:
This PR proposes to add a `_repr_html_` method to `go.Figure` and `go.FigureWidget`, which is a hook used by some notebook-like environments and other packages such as sphinx-gallery. For notebook environments using an ipython kernel, `_ipython_display_` has precedence over `_repr_html_`  (see https://ipython.readthedocs.io/en/stable/config/integrating.html) so this will not change anything.

When/if this PR is accepted I will submit a pull request to sphinx-gallery to show a plotly figure on their website (not possible at the moment) since they don't want to user scrapers in their configuration.

@jonmmease

### DIFF
--- a/packages/python/plotly/plotly/basedatatypes.py
+++ b/packages/python/plotly/plotly/basedatatypes.py
@@ -422,7 +422,26 @@ class BaseFigure(object):
         """
         Customize html representation
         """
-        return self.to_html(full_html=False, include_plotlyjs="cdn", auto_play=False)
+        bundle = self._repr_mimebundle_()
+        if 'text/html' in bundle:
+            return bundle['text/html']
+        else:
+            return ''
+            #return self.to_html(full_html=False, include_plotlyjs="cdn")
+
+    def _repr_mimebundle_(self, include=None, exclude=None, validate=True, **kwargs):
+        """
+        Return mimebundle corresponding to default renderer.
+        """
+        import plotly.io as pio
+        renderer_str = pio.renderers.default
+        renderers = pio._renderers.renderers
+        from plotly.io._utils import validate_coerce_fig_to_dict
+        fig_dict = validate_coerce_fig_to_dict(self, validate)
+        # Mimetype renderers
+        renderer = renderers[renderer_str]
+        bundle = renderer.to_mimebundle(fig_dict)
+        return bundle
 
     def _ipython_display_(self):
         """
--- a/packages/python/plotly/plotly/basedatatypes.py
+++ b/packages/python/plotly/plotly/basedatatypes.py
@@ -418,13 +418,11 @@ class BaseFigure(object):
 
         return repr_str
 
-
     def _repr_html_(self):
         """
-        Customizes html representation
+        Customize html representation
         """
-        return self.to_html(full_html=False, include_plotlyjs='cdn')
-
+        return self.to_html(full_html=False, include_plotlyjs="cdn", auto_play=False)
 
     def _ipython_display_(self):
         """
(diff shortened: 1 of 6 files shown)


jonmmease:
Hi @emmanuelle, thanks for working on this.  Since there are a lot of variations in how to produce an HTML representation of a plotly plot (e.g. in particular there is the choice of `div` tags or `html` tags, loading plotly.js in script or using require.js, using a CDN or the local bundle).

We've worked through a lot of this already in the `renderers` framework, so I'd like to propose we take advantage of that here.  Take a look at the implementation of `show` in `plotly/io/_renderers.py`. This uses the `_build_mime_bundle` renderer method to build a mime bundle for the figure.  Could `_repr_html_` call this same method and then extract the HTML representation from the mime bundle?  And/or I think we could have a `_repr_mimebundle_` method return this as-is.

This would allow the user to configure the representation using the renderers system.  I haven't looked at what you have here in detail, but it might be that we need a new HTML-based renderer to replicate this functionality. The existing HTML renderer implementations are in `plotly/io/_base_renderers.py` if you want to compare.

Let me know if any of that didn't make sense!

emmanuelle:
@jonmmease thanks a lot for your input. I tried to follow your input and 
- defined a `_repr_mimebundle_` method of `go.BaseFigure` using the `to_mimebundle` method of the default renderer, when it exists.
- defined a `_repr_html_` method using `text/html` from `_repr_mimebundle_` if the key exists, and if not I used `fig.to_html()` with some options which are quite arbitrary. Maybe we could pass options to `_repr_html_` to override these options.

I also had to define a new renderer for my sphinx-gallery application because a renderer can't be both an `ExternalRender` (I needed the `render` method to generate the png files together with the html files) and a `MimetypeRenderer`. So there wouldl be a renderer for people who can install orca (in their CI) and another one for people who can't.

packages/python/plotly/plotly/basedatatypes.py
```diff
@@ -418,6 +418,37 @@ def __repr__(self):
...
 
+    def _repr_html_(self):
+        """
+        Customize html representation
+        """
+        bundle = self._repr_mimebundle_()
+        if "text/html" in bundle:
+            return bundle["text/html"]
+        else:
+            return self.to_html(full_html=False, include_plotlyjs="cdn")
```
jonmmease, line 429: Nice, I like the fallback here.

packages/python/plotly/plotly/tests/test_io/test_renderers.py
```diff
@@ -287,3 +287,90 @@ def test_reject_invalid_renderer(renderer):
...
+                assert bundle[key].replace(id1, "") == ref_bundle[key].replace(id2, "")
+    else:
+        assert bundle == {}
+
+
+def test_repr_mimebundle_mixed_renderer(fig1):
+    pio.renderers.default = "notebook+plotly_mimetype"
+    assert set(fig1._repr_mimebundle_().keys()) == set(
+        {"application/vnd.plotly.v1+json", "text/html"}
+    )
```
jonmmease, line 376: Great, thanks for the tests!

nicolaskruchten:
Let's merge this a bit later so it will be part of 4.8 and not 4.7.1 ?

emmanuelle:
Can I merge this one now that 4.7.1 is out?

nicolaskruchten:
I was actually kind of thinking of doing the lazy-loaded extensions as v4.7.2, along with Plotly.js 1.54.2, this week. Is this a change that can go into a patch?

emmanuelle:
I'd rather not to. It is a new functionality, I hope it will not break anything for anybody but one can never be 100% sure so I'd prefer this feature to be released in a .0. No problem for waiting for 4.7.2 first, I just wanted to be sure I would not forget :-).

PR #2427, opened 2020-05-01
upgrade to plotly.js 1.54

emmanuelle:
I will also add a changelog comment.


emmanuelle:
I will rather add the changelog with the doc PR so that I can mention the doc.

nicolaskruchten:
:dancer:

PR #2436, opened 2020-05-05
bump plotly.js to 1.54.1

nicolaskruchten:


emmanuelle:
There is no change in codegened files but I believe you run the `updateplotlyjs`? :dancer: once the CI is green.

nicolaskruchten:
Yep, there are no changes to the API or schema descriptions, only to the JS code.

Issue #2443, opened 2020-05-06
Figure factory broken in release 4.7.0


cliffckerr:
**Describe the bug**

Related to https://github.com/plotly/plotly.py/issues/1111, `import plotly.figure_factory` fails with the most recent Plotly release (4.7) without a helpful error message. 

**To reproduce**

First noticed with

```
conda create -n plotly_test python=3.7.7
conda activate plotly_test
pip install plotly_express
python
>>> import plotly.figure_factory
```

(which worked until this morning's release), but also reproducible via:

```
conda create -n plotly_test python=3.7.7
conda activate plotly_test
pip install numpy==1.18.4 plotly==4.7.0
python
>>> import plotly.figure_factory
```

**Expected behavior**

It works, or at least gives a helpful error message.

**Actual behavior**

```
>>> import plotly.figure_factory
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/software/anaconda3/envs/plotly_test/lib/python3.7/site-packages/plotly/figure_factory/__init__.py", line 14, in <module>
    from plotly.figure_factory._2d_density import create_2d_density
  File "/software/anaconda3/envs/plotly_test/lib/python3.7/site-packages/plotly/figure_factory/_2d_density.py", line 8, in <module>
    from plotly.graph_objs import graph_objs
  File "/software/anaconda3/envs/plotly_test/lib/python3.7/site-packages/plotly/graph_objs/graph_objs.py", line 1, in <module>
    from plotly.graph_objs import *
  File "/software/anaconda3/envs/plotly_test/lib/python3.7/site-packages/plotly/graph_objs/__init__.py", line 290, in __getattr__
    return orig_getattr(import_name)
  File "/software/anaconda3/envs/plotly_test/lib/python3.7/site-packages/_plotly_utils/importers.py", line 41, in __getattr__
    name=import_name, __name__=parent_name
AttributeError: module 'plotly.graph_objs' has no attribute 'FigureWidget'
```

**Workaround**

Manually install `ipywidgets` (e.g., `pip install ipywidgets plotly_express` works).

**Suggested solution**

Add `ipywidgets` back to the `plotly_express` requirements, and make an `ipywidgets` error message similar to the one that exists if you try to use `figure_factory` without having `numpy` installed.

**Platform:**
 - OS: Ubuntu 18.04
- Python: 3.7.7 (observed on Python 3.8 as well, but not on Python 3.6)

jonmmease:
[...]

Issue #1111, opened 2018-08-14
Add informative error message when FigureWidget not available due to missing dependencies


vlizanae:
I'm running an alpine python3 container and for some reason FigureWidget is not available. Jupyter is not installed because it's meant to be a Dash app, are there any dependencies I might have overlooked?
![screenshot from 2018-08-14 17-56-03](https://user-images.githubusercontent.com/31910455/44118202-ba01f8c6-9feb-11e8-8f70-547b4c572d75.png)

jonmmease:
Hi @vlizanae ,

The `FigureWidget` class won't be available unless `ipywidgets` is installed. If you're making a Dash app, then you'll want to use the `graph_objs.Figure` class instead of `FigureWidget`.  `FigureWidget` is essentially a subclass of `Figure` with extra logic and dependencies for integrating with the Jupyter Notebook.

Thanks for reporting this because it reminds me that it would be nice to do something more user friendly when someone tries to access a `FigureWidget` when the dependencies aren't available.

Alright if I change the title of this issue to "Raise informative error if FigureWidget dependencies are not installed"?

vlizanae:
sure, thank you!

Juanlu001:
I was thinking about this a bit and I don't see a quick & straightforward solution. This is the way `FigureWidget` is being imported in the `graph_objs` namespace:

```python
try:
    import ipywidgets
    from distutils.version import LooseVersion
    if LooseVersion(ipywidgets.__version__) >= LooseVersion('7.0.0'):
        from ._figurewidget import FigureWidget
    del LooseVersion
    del ipywidgets
except ImportError:
    pass
```

The only options I see:

* Keep this code, but instead of silencing the `ImportError`, display a warning. The problem is that lots of users that are not interested in `FigureWidget`s will see it.
* Do not import it in `graph_objs`, and instead make the user do `from plotly.graph_objs.widgets import FigureWidget`. This is an API breaking change, and also annoying to have this object in a different module.
* Add `ipywidgets>7` to the plotly.py requirements and get rid of the conditional import altogether.

Other ideas?

nicolaskruchten:
Closed by #2445

PR #2445, opened 2020-05-07
Fix FigureWidget attribute error on wildcard import with ipywidgets not installed

jonmmease:
Fixes the regression in 4.7.0 reported in https://github.com/plotly/plotly.py/issues/2443.  Also takes care of https://github.com/plotly/plotly.py/issues/1111 by producing a more informative error message.

The problem is that with Python 3.7+, we add `"FigureWidget"` to the `__all__` list in the `plotly.graph_objs`/`plotly.graph_objects` modules regardless of whether `ipywidgets` is installed.  But when `FigureWidget` is lazily imported, an exception is raised if a supported version of `ipywidgets` is not installed.  This results in the reported error when a user, or library, calls `from plotly.graph_objs import *`.

One option would be to check for the proper version of `ipywidgets` when `plotly.graph_objs` is imported and only add `FigureWidget` to `__all__` if it is found. But this results in a performance hit on every import of `plotly.graph_objs`, whether or not `FigureWidget` is used.

Another option would be to always allow `FigureWidget` to be imported, but raise an exception when a user tries to construct a `FigureWidget` if `ipywidgets` is not found.  Unfortunately, this isn't possible because `ipywidgets` is required in order to define the `FigureWidget` class itself so we can't import it without `ipywidgets`.

Instead, this PR introduces a new `plotly.missing_ipywidgets.FigureWidget` class. This class is a subclass of `BaseFigure` (just like the standard `Figure`), but all it does is raise an informative exception in the constructor informing the user that they need to install `ipywidgets` in order to use `FigureWidget`.  With this approach, `"FigureWidget"` is still always added to `__all__`, but on lazy import we check whether `ipywidgets` is installed. If it is, the current `plotly.graph_objs._figurewidget.FigureWidget` class is returned, otherwise the new `plotly.missing_ipywidgets.FigureWidget` class is returned.

Tests added in `plotly/tests/test_core/test_figure_widget_backend/test_missing_ipywigets.py`.

## Code PR
  modified existing tests.

### DIFF
--- a/packages/python/plotly/plotly/missing_ipywidgets.py
+++ b/packages/python/plotly/plotly/missing_ipywidgets.py
@@ -1,0 +1,15 @@
+from .basedatatypes import BaseFigure
+
+
+class FigureWidget(BaseFigure):
+    """
+    FigureWidget stand-in for use when ipywidgets is not installed. The only purpose
+    of this class is to provide something to import as
+    `plotly.graph_objs.FigureWidget` when ipywidgets is not installed. This class
+    simply raises an informative error message when the constructor is called
+    """
+
+    def __init__(self, *args, **kwargs):
+        raise ImportError(
+            "Please install ipywidgets>=7.0.0 to use the FigureWidget class"
+        )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [4.7.1] - ???
+
+### Fixed
+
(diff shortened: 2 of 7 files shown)


packages/python/plotly/plotly/missing_ipywidgets.py
```diff
@@ -0,0 +1,15 @@
+from .basedatatypes import BaseFigure
+
+
+class FigureWidget(BaseFigure):
```
nicolaskruchten, line 4: why even bother subclassing `BaseFigure` here? :)

nicolaskruchten:
:dancer:

packages/python/plotly/plotly/missing_ipywidgets.py
jonmmease: It seemed safer to me for it to still pass `isinstance(fig, BaseFigure)` checks, but it probably doesn't make a difference.

cliffckerr:
Awesome, thank you for addressing this so quickly @jonmmease !

leo-smi:
the error still occur https://github.com/plotly/plotly.py/issues/2443#issuecomment-1003778894

PR #2454, opened 2020-05-09
Lazyload

### LINKED ISSUE #1913: Lazy load strategy for plotly bundle

nicolaskruchten:
Supersedes and closes #2260 and closes #1913

### DIFF
--- a/packages/javascript/jupyterlab-plotly/package.json
+++ b/packages/javascript/jupyterlab-plotly/package.json
@@ -31,7 +31,11 @@
     "typescript": "~3.7.0"
   },
   "dependencies": {
+<<<<<<< HEAD
     "plotly.js": "^1.54.1",
+=======
+    "plotly.js": "^1.53.0",
+>>>>>>> 1c80d7555... Resolving merge conflicts
     "@types/plotly.js": "1.44.28",
     "@jupyterlab/rendermime-interfaces": "^1.3.0 || ^2.0.0",
     "@lumino/messaging": "^1.2.3",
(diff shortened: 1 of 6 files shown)


pkvprakash:
I see this fix keeps on moving from milestone to milestone. Waiting for this functionality to be available for our use.

nicolaskruchten:
Yes, I'm sorry about this. The challenge is that this is potentially quite a risky change that could impact thousands of users, and I haven't had the time to fully test it and convince myself that it's a safe change that we're willing to support moving forward. I'm not sure when I'll have the time to do this, to be honest, although I really would like to.

nicolaskruchten:
I do apologize about how this PR was left unattended for so long, but it was superseded by https://github.com/plotly/plotly.py/pull/3142 which was just merged, so the functionality will make it into the next version of the library.

PR #2472, opened 2020-05-14
Fixed command for installing ipywidgets using pip

SubodhDahal:
There's a minor typo on the README for installing `ipywidgets` using `pip`, which I've fixed now.


nicolaskruchten:
Ah thanks!

PR #2491, opened 2020-05-22
order fixups

jdamiba:
The purpose of this PR is to fixup the order of examples in the `advanced_opt` and `chart_events` sections.


nicolaskruchten:
ok they're green!

PR #2493, opened 2020-05-22
added links to px.data and px.colors in API doc

emmanuelle:
This PR add links to `px.data` and `px.colors` at the end of the API doc page on plotly.express.

For `px.data`

Screenshot_2020-05-22 plotly data package — 4 7 0 documentation

For `px.colors`, the objects which we want to list are not functions but lists, and they are not listed in an `__all__` somewhere already so I figured out it would be easier to link to the corresponding doc tutorials.

image

In the main `px` page:
image


nicolaskruchten:
In your screenshot it's listed as `plotly.data` but it's actually `plotly.express.data`.

nicolaskruchten:
Re colors, can we list the submodules of `plotly.express.colors` like `qualitative` and `sequential` so at least people know that those are namespaces?

Maybe we could list the `px.colors.named_colorscales()` function?

nicolaskruchten:
If I had a function that outputted the color scales and sequences by namespace, could we execute that function and inline the output here?

emmanuelle:
> In your screenshot it's listed as `plotly.data` but it's actually `plotly.express.data`.

So if you want it to appear as `plotly.express.data`, `data` should be a sub-package of `plotly.express` which it is not at the moment (ie there should be a `data` directory with an `__init__.py`). This is what I did with `colors` but then I had to change some imports which apparently cause problem for py 3.6. Should I create the `plotly.express.data` package? This would be a cleaner organization of the code.

emmanuelle:
So this CI failure is very frustrating. I created subpackages `data` and `colors` in `plotly.express` in order to be able to document them in the API (and it's also a cleaner way to organize code). Everything works fine locally but on the CI `build-doc` build all notebooks where `px` is imported fail at the line 58 of the `__init__.py`: `from . import data, colors`. Everything works well on my laptop. I thought it could come from the version of python so I tried py 3.6 and 3.7 locally and both work. Any idea?

emmanuelle:
I removed the re-direction for failed builds of notebooks because the output is full of unicode characters which are hard to read but I'll put this back.

emmanuelle:
ok I can  reproduce the error if I try to install the package with `python setup.py install` instead of `python setup.py develop`, I'll investigate from here.

emmanuelle:
ok, silly me, it was just a question of adding the new subpackages to the `packages` list in `setup.py`. All green now!

doc/apidoc/Makefile
```diff
@@ -22,7 +22,7 @@ help:
 	sed -i 's/:class:`plotly.graph_objects/:class:`plotly.graph_objs/g' ../../packages/python/plotly/plotly/graph_objs/*/*.py
 	sed -i 's/:class:`plotly.graph_objects/:class:`plotly.graph_objs/g' ../../packages/python/plotly/plotly/graph_objs/*/*/*.py
 	sed -i 's/:class:`plotly.graph_objects/:class:`plotly.graph_objs/g' ../../packages/python/plotly/plotly/graph_objs/*/*/*/*.py
-	sphinx-apidoc -o generated ../../packages/python/plotly/plotly ../../packages/python/plotly/plotly/validators ../../packages/python/plotly/plotly/tests ../../packages/python/plotly/plotly/matplotlylib/ ../../packages/python/plotly/plotly/offline ../../packages/python/plotly/plotly/api
+	sphinx-apidoc -E -M -o generated ../../packages/python/plotly/plotly ../../packages/python/plotly/plotly/validators ../../packages/python/plotly/plotly/tests ../../packages/python/plotly/plotly/matplotlylib/ ../../packages/python/plotly/plotly/offline ../../packages/python/plotly/plotly/api
```
emmanuelle, line 25: These changes:
- remove titles such as "List of modules" which I deemed to be verbose and not helpful.
- put the module content before the list of submodules

doc/apidoc/plotly.express.rst
```diff
@@ -49,4 +49,13 @@ plotly's high-level API for rapid figure generation. ::
     imshow
 
 
+`plotly.express` submodules
```
emmanuelle, line 52:
```suggestion
`plotly.express` subpackages
```

emmanuelle:
I updated the screenshots above.

emmanuelle:
ok, I hacked the Makefile to get the architecture of `colors` right but this make me cringe a bit. Since we have generated code in `graph_objects` maybe we could also generate code in `colors` subpackages in the future? Anyway here is the screenshot for now
image

nicolaskruchten:
@jonmmease could you give these package manipulations a once-over plz? They seem fine to me but I’m always wary of breaking something :)

jonmmease:
It all looks reasonable to me!

doc/apidoc/index.rst
```diff
@@ -17,6 +17,7 @@ Submodules
 * :ref:`Subplots: helper function for layout out multi-plot figures <subplots>`
 * :ref:`Figure Factories: helper methods for building specific complex charts <ff>`
 * :ref:`I/O: low-level interface for displaying, reading and writing figures <io>`
+* :mod:`plotly.colors`: colorscales and utility functions
```
nicolaskruchten, line 20: do we want to add `plotly.data` here too?

packages/python/plotly/plotly/colors/__init__.py
```diff
@@ -0,0 +1,23 @@
+"""For a list of colors available in `plotly.colors`, please see
+
+* the `tutorial on discrete color sequences <https://plotly.com/python/discrete-color/#color-sequences-in-plotly-express>`_
```
nicolaskruchten, line 3: let's also add the tutorial on continuous color: https://plotly.com/python/colorscales/

packages/python/plotly/plotly/colors/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+* the `list of built-in continuous color scales <https://plotly.com/python/builtin-colorscales/>`_
+
+Color scales are available within the following namespaces
```
nicolaskruchten, line 6:
```suggestion
Color scales and sequences are available within the following namespaces
```

packages/python/plotly/plotly/express/colors/__init__.py
```diff
@@ -0,0 +1,24 @@
+"""For a list of colors available in `plotly.express.colors`, please see
+
+* the `tutorial on discrete color sequences <https://plotly.com/python/discrete-color/#color-sequences-in-plotly-express>`_
```
nicolaskruchten, line 3: same changes as above I guess?

packages/python/plotly/plotly/express/data/__init__.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import absolute_import
+from plotly.data import *
+
```
nicolaskruchten, line 3: should we add a short docstring here like in the top-level package?

```
"""
Built-in datasets for demonstration, educational and test purposes.
"""
```

nicolaskruchten:
Should be good to merge tomorrow right before the release... thanks @emmanuelle !

packages/python/plotly/plotly/express/data/__init__.py
emmanuelle: done

packages/python/plotly/plotly/colors/__init__.py
emmanuelle: ok

doc/apidoc/index.rst
emmanuelle: yes

packages/python/plotly/plotly/express/colors/__init__.py
emmanuelle: yes

emmanuelle:
I made the small requested changes, please tell me if I can merge!

nicolaskruchten:
💃

PR #2502, opened 2020-05-26
Bar fix

nicolaskruchten:
This is needed so we can pass in column names as e.g. `bool` or `Timestamp`, which anyway fail in the `else` branch. The Pandas visualization docs have this example, which doesn't run without this fix and does with it:

```python
import pandas as pd
import numpy as np
pd.options.plotting.backend="plotly"
ts = pd.Series(np.random.randn(1000), index=pd.date_range('1/1/2000', periods=1000))
df = pd.DataFrame(np.random.randn(1000, 4), index=ts.index, columns=list('ABCD'))
df.iloc[5].plot.bar()
```


emmanuelle:
:dancer: when all green! Thanks!

PR #2507, opened 2020-05-27
Enforce sequential order

jdamiba:


checkorder.py
```diff
@@ -0,0 +1,107 @@
+import frontmatter as fm
```
nicolaskruchten, line 1: this file should not be committed

doc/python/2D-Histogram.md
```diff
@@ -31,9 +31,10 @@ jupyter:
     page_type: u-guide
     permalink: python/2D-Histogram/
     redirect_from:
-      - python/2d-histogram/
-      - python/2d-histograms/
+    - python/2d-histogram/
+    - python/2d-histograms/
     thumbnail: thumbnail/histogram2d.jpg
+order: 5
```
nicolaskruchten, line 37: so the way this was run actually breaks things I think... the `order` is at the wrong level of indentation. Remember there was a mode for running the enforce order script in Python?

jdamiba: Looking closely at the `check-or-enforce-order.py` script, it looks like it was not updated to take into account the fact that the plotly.py docs are not in their own repo any more but are a part of the plotly.py repo. 

I'm working to update the script so I can run it successfully in this repo.

Issue #2484, opened 2020-05-19
utility function to interpolate N colors within a colorscale


nicolaskruchten:
This is the kind of thing that happens all the time: give me 4 colors within Viridis, or give me 5 colors within the firs 50% of Plasma or the 20%-65% range of Magma or whatever

CarlAndersson:
Is anyone working on this already? I had to implement this for my paper anyhow, so I could open a pull request in a few days if there's interest.

nicolaskruchten:
We'd be happy to accept a pull request, no one else is working on this as far as I know :)

PR #2510, opened 2020-05-27
complete the API of plotly.colors and plotly.express.colors

emmanuelle:
API exposed:

image


jonmmease:
Thanks @emmanuelle!  In terms of the imports, I think we also need to add these to the `__all__` lists:

 - `plotlyjs`
 - `DEFAULT_PLOTLY_COLORS`
 - `PLOTLY_SCALES`

I don't feel strongly that these need to be in the documentation, just that imports keep working.

emmanuelle:
@jonmmease done! Lists do not appear in the sphinx documentation and I did not do the copy thing for the `plotlyjs` module so this does not change the API doc. But of course now the objects can be imported both from `plotly.colors` and `px.colors`.

nicolaskruchten:
:dancer: !

It would be nice to clean up the docstrings in these functions a bit, maybe when we do https://github.com/plotly/plotly.py/issues/2484 ... I can't tell if any of these functions actually already do this!

PR #2524, opened 2020-06-01
fixing hover_data and hover_name bugs in path API

nicolaskruchten:
This approach is a bit less brittle I think.

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1432,11 +1432,12 @@ def process_dataframe_hierarchy(args):
     _check_dataframe_all_leaves(df[path[::-1]])
     discrete_color = False
 
+    new_path = []
     for col_name in path:
-        series_to_copy = df[col_name]
         new_col_name = col_name + "_path_copy"
-        path = [new_col_name if x == col_name else x for x in path]
-        df[new_col_name] = series_to_copy
+        new_path.append(new_col_name)
+        df[new_col_name] = df[col_name]
+    path = new_path
     # ------------ Define aggregation functions --------------------------------
 
     def aggfunc_discrete(x):
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1432,23 +1432,11 @@ def process_dataframe_hierarchy(args):
     _check_dataframe_all_leaves(df[path[::-1]])
     discrete_color = False
 
-    if args["color"] and args["color"] in path:
-        series_to_copy = df[args["color"]]
-        new_col_name = args["color"] + "additional_col_for_color"
-        path = [new_col_name if x == args["color"] else x for x in path]
+    for col_name in path:
+        series_to_copy = df[col_name]
+        new_col_name = col_name + "_path_copy"
+        path = [new_col_name if x == col_name else x for x in path]
         df[new_col_name] = series_to_copy
-    if args["hover_data"]:
-        for col_name in args["hover_data"]:
-            if col_name == args["color"]:
-                series_to_copy = df[col_name]
-                new_col_name = str(args["color"]) + "additional_col_for_hover"
-                df[new_col_name] = series_to_copy
-                args["color"] = new_col_name
-            elif col_name in path:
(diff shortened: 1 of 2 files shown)


packages/python/plotly/plotly/tests/test_core/test_px/test_px_functions.py
```diff
@@ -171,6 +171,17 @@ def test_sunburst_treemap_with_path_and_hover():
...
     assert "smoker" in fig.data[0].hovertemplate
 
+    df = px.data.gapminder().query("year == 2007")
+    fig = px.sunburst(
+        df, path=["continent", "country"], color="lifeExp", hover_data=df.columns
+    )
+    assert fig.layout.coloraxis.colorbar.title.text == "lifeExp"
+
+    df = px.data.tips()
+    fig = px.sunburst(df, path=["sex", "day", "time", "smoker"], hover_name="smoker")
```
nicolaskruchten, line 181: this actually errors out on master

packages/python/plotly/plotly/tests/test_core/test_px/test_px_functions.py
```diff
@@ -171,6 +171,17 @@ def test_sunburst_treemap_with_path_and_hover():
     )
     assert "smoker" in fig.data[0].hovertemplate
 
+    df = px.data.gapminder().query("year == 2007")
+    fig = px.sunburst(
+        df, path=["continent", "country"], color="lifeExp", hover_data=df.columns
+    )
+    assert fig.layout.coloraxis.colorbar.title.text == "lifeExp"
```
nicolaskruchten, line 178: on master this appears as `lifeExpadditional_col_for_hover`

packages/python/plotly/plotly/tests/test_core/test_px/test_px_functions.py
```diff
@@ -171,6 +171,23 @@ def test_sunburst_treemap_with_path_and_hover():
...
+
+    df = px.data.tips()
+    fig = px.sunburst(df, path=["sex", "day", "time", "smoker"], hover_name="smoker")
+    assert "smoker" not in fig.data[0].hovertemplate  # represented as '%{hovertext}'
+    assert "%{hovertext}" in fig.data[0].hovertemplate  # represented as '%{hovertext}'
+
+    df = px.data.tips()
+    fig = px.sunburst(df, path=["sex", "day", "time", "smoker"], custom_data=["smoker"])
```
nicolaskruchten, line 186: this also errors out on master (because `smoker` is in `path` but isn't accounted for in `custom_data`)

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1432,23 +1432,11 @@ def process_dataframe_hierarchy(args):
...
     discrete_color = False
 
-    if args["color"] and args["color"] in path:
-        series_to_copy = df[args["color"]]
-        new_col_name = args["color"] + "additional_col_for_color"
-        path = [new_col_name if x == args["color"] else x for x in path]
+    for col_name in path:
+        series_to_copy = df[col_name]
+        new_col_name = col_name + "_path_copy"
+        path = [new_col_name if x == col_name else x for x in path]
```
nicolaskruchten, line 1438: the new approach is just to make a full copy of everything in `path` even if they're not in `color` or whatever and not touch any of the other args, instead of spot-checking `color` and `hover_data`... Maybe there's a case here I'm missing but it's not in the tests and this approach seems more future-friendly.

emmanuelle: probably we don't need the `if x == col_name` here? Since `x` is in `path`.

emmanuelle: it's an elegant solution. I had not thought that column names in `path` were disposable since they would not appear in the final hovertemplate.

emmanuelle:
Thanks for the improvement! I left a small comment since I think one `if` check can be removed but otherwise it's good to merge! :woman_dancing:

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: you're right that comprehension is weird... check out the new version?

PR #2536, opened 2020-06-04
Changed opacity to 0.2 in sankey docs in order to obtain hover effect

lrob:
The sankey documentation contains a paragraph showing how to customize color of links. Currently, the example proposes to change the opacity to 0.4 which is the value at which the opacity is set during hover. This implies that no hover effect is obtained in the resulting plot. By using opacity equal to 0.2 the hover effect works properly.

### DIFF
--- a/doc/python/sankey-diagram.md
+++ b/doc/python/sankey-diagram.md
@@ -73,7 +73,7 @@ response = urllib.request.urlopen(url)
 data = json.loads(response.read())
 
 # override gray link colors with 'source' colors
-opacity = 0.4
+opacity = 0.2
 # change 'magenta' to its 'rgba' value to add opacity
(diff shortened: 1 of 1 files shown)


emmanuelle:
Hi @lrob thanks for your PR. Could you please explain which "hover data" you are looking after? I executed the two versions (before your change and with the change) and I don't really get what you gain with the decreased intensity of links. Thanks :-)

lrob:
When you place your mouse over a bar of the sankey if opacity is set to
0.2, all the flows incoming and outgoing from the bar become darker.
If opacity is set to 0.4 nothing happens.

lrob:
here the pictures of what I'm speaking about
https://drive.google.com/file/d/1gH-nWYI9pLrr5jMyr16UjMdbjZepyOA8/view?usp=sharing
https://drive.google.com/file/d/1xyLAjVGrFbCqs9LoJz0SdF1nq71v01XS/view?usp=sharing

PR #2540, opened 2020-06-05
export colors only once

nicolaskruchten:
This should be a no-op but still good to clean up, right?


jonmmease:
Doesn't this change mean that this won't work any more?

```
import plotly
plotly.colors
```

nicolaskruchten:
No I don't think so... if you look two lines up from each of the deleted lines we're already importing/exporting these. I assume this is a copy-paste thing rather than intentional?

PR #2554, opened 2020-06-08
make trendlines more robust

### LINKED ISSUE #2549: Error when using OLS trendline on single-point data
This could be an issue with statsmodels rather than Plotly, but when trying to fit trendlines to a line which only contains one repeated point, only one parameter is returned from the statsmdels fit and an error occurs when trying to build the hover header for the trendline.

```python
import pandas as pd
import numpy as np
import plotly.express as px

df = pd.DataFrame({'y': np.zeros(2), 'x': np.ones(2)})
fig = px.scatter(df, x='x', y='y', trendline='ols')
```

```
  File "C:\Users\slishak\AppData\Local\Programs\Python\Python38\lib\site-packages\IPython\core\interactiveshell.py", line 3331, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-17-ef61e6ef2734>", line 1, in <module>
    fig = px.scatter(df, x='y', y='x', trendline='ols')
  File "C:\Users\slishak\AppData\Local\Programs\Python\Python38\lib\site-packages\plotly\express\_chart_types.py", line 62, in scatter
    return make_figure(args=locals(), constructor=go.Scatter)
  File "C:\Users\slishak\AppData\Local\Programs\Python\Python38\lib\site-packages\plotly\express\_core.py", line 1907, in make_figure
    patch, fit_results = make_trace_kwargs(
  File "C:\Users\slishak\AppData\Local\Programs\Python\Python38\lib\site-packages\plotly\express\_core.py", line 298, in make_trace_kwargs
    fit_results.params[1],
IndexError: index 1 is out of bounds for axis 0 with size 1
```

Versions:
```
plotly==4.8.0
statsmodels==0.11.1
```

nicolaskruchten:
Closes #2549 (trendlines with zero slope) as well as a couple of other issues I found while testing:

* trendlines over timeseries data was just broken 🙈  https://stackoverflow.com/questions/61644142/plotly-how-to-show-trendline-for-time-series-data-using-plotly-express
* although we check for length > 1, we don't check for non-None/non-NaN length > 1
* we didn't have a lot of testing in general on trendlines/results, so I added a bunch

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -277,17 +277,24 @@ def make_trace_kwargs(args, trace_spec, trace_data, mapping_labels, sizeref):
                     attr_value in ["ols", "lowess"]
                     and args["x"]
                     and args["y"]
-                    and len(trace_data) > 1
+                    and len(trace_data[[args["x"], args["y"]]].dropna()) > 1
                 ):
                     import statsmodels.api as sm
 
                     # sorting is bad but trace_specs with "trendline" have no other attrs
                     sorted_trace_data = trace_data.sort_values(by=args["x"])
-                    y = sorted_trace_data[args["y"]]
-                    x = sorted_trace_data[args["x"]]
+                    y = sorted_trace_data[args["y"]].values
+                    x = sorted_trace_data[args["x"]].values
 
+                    x_is_date = False
                     if x.dtype.type == np.datetime64:
                         x = x.astype(int) / 10 ** 9  # convert to unix epoch seconds
+                        x_is_date = True
+                    elif x.dtype.type == np.object_:
+                        x = x.astype(np.float64)
+
+                    if y.dtype.type == np.object_:
+                        y = y.astype(np.float64)
 
                     if attr_value == "lowess":
                         # missing ='drop' is the default value for lowess but not for OLS (None)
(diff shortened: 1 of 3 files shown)


packages/python/plotly/plotly/tests/test_core/test_px/test_trendline.py
```diff
@@ -1,14 +1,103 @@
...
+def test_trendline_enough_values(mode):
+    fig = px.scatter(x=[0, 1], y=[0, 1], trendline=mode)
+    assert len(fig.data) == 2
+    assert len(fig.data[1].x) == 2
+    fig = px.scatter(x=[0], y=[0], trendline=mode)
+    assert len(fig.data) == 2
+    assert fig.data[1].x is None
+    fig = px.scatter(x=[0, 1], y=[0, None], trendline=mode)
+    assert len(fig.data) == 2
+    assert fig.data[1].x is None
```
emmanuelle, line 40: should we raise an error when there are not enough points to compute a trendline?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -265,17 +265,24 @@ def make_trace_kwargs(args, trace_spec, trace_data, mapping_labels, sizeref):
...
-                    x = sorted_trace_data[args["x"]]
+                    y = sorted_trace_data[args["y"]].values
+                    x = sorted_trace_data[args["x"]].values
 
+                    x_is_date = False
                     if x.dtype.type == np.datetime64:
                         x = x.astype(int) / 10 ** 9  # convert to unix epoch seconds
+                        x_is_date = True
+                    elif x.dtype.type == np.object_:
+                        x = x.astype(np.float64)
```
emmanuelle, line 282: Here if we have strings like the `data` column of `px.data.stocks()` this will raise an error message `ValueError: could not convert string to float: '2018-01-01'`. This is not a regression since the error message is better than the current one deep in statsmodels, but this line could be in a try except statement with an error message written by us.

packages/python/plotly/plotly/tests/test_core/test_px/test_trendline.py
nicolaskruchten: No, because it's easy to get into a situation where you facet on a variable, and then within a given facet there is only one point of a given color, so we don't want to prevent ppl from being able to draw trendlines on all the other colors. This happens e.g. with tips when you facet on day/time and color by sex: there's a day/time/sex with only one data point.

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: agreed, will fix

packages/python/plotly/plotly/tests/test_core/test_px/test_trendline.py
emmanuelle: makes sense, thanks for the explanation

PR #2558, opened 2020-06-10
add_shape, add_layout_image, add_annotation for multiple facets at once, add_hline and add_vline (+add_vrect/add_hrect)

### LINKED ISSUE #2140: add_shape, add_layout_image, add_annotation for multiple facets at once
Right now if I want to add the same shape (i.e. a horizontal line) to every facet in a PX figure, I have to loop through and do it myself. It seems like having a way of saying `row="*"` or something might be handy.

### LINKED ISSUE #2141: add_hline and add_vline (+add_vrect/add_hrect)
We should add these convenience methods, as the ref=x/ref=paper trick is pretty hard for people to understand.

~We'd want to be able to add these to all subplots, one subplot, or possibly a single row or a single column.~ Moved to #2140

We'd also want add_vspan and add_hspan (or vrect/hrect, not set on the names)

nicholas-esterer:
Closes #2141 and closes #2140 

Making it possible to specify multiple rows and columns when using `add_shape`, `add_layout_image` or `add_annotation`. 
This pull request is to share the work, so a lot is missing still.

## Code PR

  modified existing tests.


packages/python/plotly/plotly/basedatatypes.py
```diff
@@ -1090,42 +1145,51 @@ def _add_annotation_like(
                 "Received col parameter but not row.\n"
                 "row and col must be specified together"
             )
+        grid_ref = self._validate_get_grid_ref()
 
         # Get grid_ref if specific row or column requested
         if row is not None:
-            grid_ref = self._validate_get_grid_ref()
-            refs = grid_ref[row - 1][col - 1]
+            # TODO It is assumed that grid_ref has an equal number of columns in
```
nicolaskruchten, line 1152: Yes, this can be not-true, you can have really odd layouts: https://plotly.com/python/subplots/#multiple-custom-sized-subplots

nicholas-esterer: Hey so I checked this out and it turns out even in this case `grid_ref` is still a rectangle, just some of the entries are None. So we could still easily iterate through the `grid_ref` using something like
```python
for r in range(1,len(grid_ref)+1):
    for c in range(1,len(grid_ref[0])+1):
        sub_plot=fig.get_subplot(row=r,col=c)
        if sub_plot is not None:
            ....
```
We just check if it is not None before doing anything to it :wrench:

nicholas-esterer:
I greatly refactored the old approach and added basically this one function (it calls some helpers which you can check out in the code):

`BaseFigure._select_subplot_coordinates(self, rows, cols, product=False)`
This function will allow `add_shape, add_trace, ...` to select subsets of the subplots using the syntax described in the issue (e.g., `rows='all', col=[1,2]` selects all the rows in columns 1 and 2). However I've abandoned the (list,'all') syntax in favour of an extra option `product` that if `True` will form the cartesian product of the `row` and `col` lists and if `False` will just `zip` them together to form the coordinates of the subplots to address.

The idea is to change `add_shape, add_trace, ...` etc. so that if they receive the "magic" `row` or `col` arguments, they call `_select_subplot_coordinates` to get the subplots to address, and then call the same function again but on all the coordinates.

nicholas-esterer:
`_select_subplot_coordinates` only has to be put in `BaseFigure.add_trace` and `BaseFigure._add_annotation_like`.

nicholas-esterer:
We have decided to not support the cartesian product mode for `row` and `col`, so after adding some documentation this PR can be reviewed for merging :partying_face:

nicholas-esterer:
Hmm seeing now that the `update_*` functions also take `row` and `col` arguments. Need to see if this has already been taken care of...

nicholas-esterer:
Ok so the PR doesn't affect `update_layout` nor the `select_*` but for some reason the documentation is not updated for `add_trace`...

nicholas-esterer:
For reviewers:
This PR allows specifying `row='all'` and or `col='all` for all the `Figure` methods that use the functions:
`BaseFigure._add_annotation_like` and `BaseFigure.add_trace`.
Then from this functionality, you can add vertical or horizontal lines or boxes that span the paper so that they remain fixed at the edges of a plot when you pan it.
The new functions are `BaseFigure.add_{vline,hline,vrect,hrect}`.
Most changes happened in `packages/python/plotly/plotly/basedatatypes.py` but some documentation updates went into `packages/python/plotly/codegen/figure.py`.

packages/python/plotly/plotly/basedatatypes.py
```diff
@@ -8,6 +8,8 @@
 import warnings
 from contextlib import contextmanager
 from copy import deepcopy, copy
+import itertools
+import pdb
```
nicolaskruchten, line 12: this probably shouldn't be here?

nicolaskruchten:
Could you please revert the changes to package-lock.json? and also try to get all CI tests passing please :)

packages/python/plotly/plotly/basedatatypes.py
nicolaskruchten: (the `pdb` part I mean)

nicholas-esterer: itertools yes, pdb no

nicolaskruchten:
Ah also: the new `add_*` methods should return `self` when they're done, so we can chain them together

nicholas-esterer:
why is package-lock.json not in the .gitignore?

nicolaskruchten:
So here's a case we need to think about... without the `add_hrect` call, there are only 6 subplots, but here we add rects to all subplots.

```python
import plotly.express as px
df = px.data.tips()
fig = px.scatter(df, x="total_bill", y="tip", facet_row="day", facet_col="time")
fig.add_hrect(4, 6, row="all", col=2, line_width=0, fillcolor="red", opacity=0.2)
fig.show()
```

image


There are two options I can think of:

1. We have an option to only add shapes to subplots with traces
2. We modify PX to generate only the subplots it needs

nicolaskruchten:
We could probably just go with this... Would be nice to come up with a way of easily removing those though. We don't have `.delete_shapes` :)

nicholas-esterer:
Ah ok, so for package-lock we'll just merge whatever package-lock is in `master`?
I think it'd be straightforward to ignore the subplots without a trace, have to check though.

nicholas-esterer:
when we merge of course :) not now

nicolaskruchten:
OK nice! I think we'll need to expose that as a new parameter though, no? Like `exclude_invisible_subplots` or something? And then default should be `False` I think just for consistency. For complete consistency, unfortunately, we probably also need to check that these subplots don't have other shapes/annotations/layout-images attached to them.

jonmmease:
@nicholas-esterer The implementation looks solid. Great job!  Once the tests are passing and @nicolaskruchten is happy with the behavior this looks good to me.

nicolaskruchten:
Thanks for taking a look @jonmmease ! What do you think of the idea of an `exclude_invisible_subplots` argument? A bit gross but probably necessary...

nicholas-esterer:
Sure, we can expose a parameter, right now I called it `none_if_no_trace` in the `BaseFigure._make_paper_spanning_shape` function and by default it has the behaviour that it doesn't put shapes on subplots without traces. I could expose a parameter in the `add_vline` etc. functions, but what about the name `exclude_subplots_without_traces` or `exclude_subplots_without_data` because, at least for me, invisible implies that data are there but is just not shown? Whereas the implementation literally checks if data are in that subplot and only draws the shape if that's the case.

nicholas-esterer:
I called it `exclude_subplots_without_data` for now but I can `sed` and change it when we decide on the parameter's name :)

nicolaskruchten:
Thanks! So @emmanuelle and I chatted about this and our conclusion was `exclude_empty_subplots` :)

Next step: get all the CI steps to pass!

nicholas-esterer:
I added a bunch of tests.
One last test I thought of is for the custom-sized subplots as in the above example
https://plotly.com/python/subplots/#multiple-custom-sized-subplots
Right now if the subplot is None, you get the reasonable error message:
ValueError: No subplot found at position (4, 1) 
This error message I got when passing `row='all', col=1`.
Do we just require users to add shapes one at a time in this case (because we can't really predict how weird their subplot sizes might be)? Or do we check to see if no plot is at that subplot position and not draw a shape there? Note this is a different problem from checking if data exists in a subplot.

nicolaskruchten:
Could you please rename this PR too? It's hard to know when scanning a list of PRs what this refers to ;)

nicholas-esterer:
This is now working with the development branch of Plotly.js introducing domain referencing shapes, annotations and images, so it should be just a matter of getting that merged and this can tag along. Also see https://github.com/plotly/plotly.js/pull/5014

nicolaskruchten:
OK great! Time to write some docs :) Probably you'll need to expand on the existing shapes/annotations/images docs here, and create a new page for the hline/vline/hrect/vrect stuff. Basically just editing the files under `doc/python`

nicholas-esterer:
Hmm I'm trying to review the PR and after making the venv and installing the requirements, running with `python app.py` or `gunicorn app:server`, the app looks like this:
Screenshot from 2020-09-24 12-28-13
Maybe something is missing? :thinking:

nicholas-esterer:
Ok so I have it checking to see if the user specified 'domain' anywhere in the `xref` or `yref` string. The reason I have it like this is because the validator will not let me just specify `domain` for `xref` or `yref`. Is there a way to add to the validator after it has been generated from the plot-schema?

jonmmease:
> Is there a way to add to the validator after it has been generated from the plot-schema?

There are a couple of ways to handle this, but I don't *think* we want to actually change the validator here because, if I understand correctly, "domain" is only valid when used in the `add_*` methods.  Changing the validator would also make this acceptable anywhere it's used in the figure.

So I think you would probably want to add a special case to the relevant `add_*` methods to intercept this before it gets validated.  Does that make sense?

nicholas-esterer:
Installation instructions:
```bash
git checkout issue-2140
git pull
virtualenv venv
source venv/bin/activate
pip install -r packages/python/plotly/requirements.txt
pip install -r packages/python/plotly/optional-requirements.txt
(cd packages/python/plotly && python setup.py updateplotlyjsdev --devbranch issue-4958)
pip install -e packages/python/plotly/
```

nicholas-esterer:
Yes maybe we just add `xrefdomain`, `yrefdomain` boolean function parameters to these functions.

nicolaskruchten:
I think reusing `xref` and `yref` would be preferrable to adding new kwargs, given that we also accept all the kwargs from the underlying `add_shape`, anything we add here cannot be used in the future in `shape`... We're already adding `row`,`col` and `annotation` and ideally we'd keep it to that minimum :)

nicholas-esterer:
hline, vline, hrect, vrect, can be annotated. To see an example of what this looks like, you can run from the root of this repo:
```bash
VISUALIZE=1 python packages/python/plotly/plotly/tests/test_core/test_autoshapes/test_annotated_shapes.py
```

nicholas-esterer:
But if we say `fig.add_vline(x=5,annotation_text="some text")` then the annotation position defaults to `top right`?

nicolaskruchten:
That sounds good

nicolaskruchten:
how close are the docs to being done here? Need any help or guidance?

nicholas-esterer:
Just have to change them to reflect the behaviour where specifying no row or col puts the shapes on all subplots. And I have to document the `annotation_*` keywords. I was going to also put them as possible kwargs in the doc string of add_hline etc. so it shows up on calling errors.

nicholas-esterer:
Actually I was mistaken, it is sufficient to have the documentation appear when calling `help(fig.add_hline)` + the online docs?

nicolaskruchten:
No, we need to have docs written in the MD files :) 

Also: the build output from `doc/apidoc` should not be committed here please... I'm surprised it's not already in the `.gitignore` but that stuff is built in CI :)

nicolaskruchten:
OK, so once we release plotly.js 1.57.0, i'll update `master` to that branch and you can prepare a new PR that branches off of `master` and includes only the changes you need to make for this PR... that'll make it a lot easier to properly review :)

PR #2591, opened 2020-06-23
Correct bug when color and values correspond to the same color in treemap or sunburst

### LINKED ISSUE #2567: sunburst bug when color and values correspond to the same column
Reported in the community forum https://community.plotly.com/t/plotly-sunburst-node-colors/41037
```
import plotly.express as px
import numpy as np
df = px.data.gapminder().query("year == 2007")
fig = px.sunburst(df, path=['continent', 'country'], values='lifeExp',
                  color='lifeExp',
                  color_continuous_scale='RdBu')
fig.show()
```
image
The color of a sector should be the average of its children but it seems to be the sum for this case.

emmanuelle:
Closes #2567


packages/python/plotly/plotly/express/_core.py
```diff
@@ -473,12 +473,16 @@ def make_trace_kwargs(args, trace_spec, trace_data, mapping_labels, sizeref):
             for k, v in mapping_labels.items():
                 # We need to invert the mapping here
                 k_args = invert_label(args, k)
+                print(k, k_args, args["hover_data"])
```
emmanuelle, line 476:
```suggestion
```

nicolaskruchten:
Note: this also fixes a new problem introduced in #2544

packages/python/plotly/plotly/tests/test_core/test_px/test_px_functions.py
```diff
@@ -329,28 +331,6 @@ def test_parcats_dimensions_max():
 
     # default behaviour
     fig = px.parallel_categories(df)
-    assert [d.label for d in fig.data[0].dimensions] == [
```
nicolaskruchten, line 332: why this deletion?

emmanuelle: hum good catch, this is not wanted. Let me put it back.

packages/python/plotly/plotly/tests/test_core/test_px/test_px_functions.py
```diff
@@ -149,10 +149,6 @@ def test_sunburst_treemap_with_path():
     fig = px.sunburst(df, path=path, values="values")
     assert fig.data[0].branchvalues == "total"
     assert fig.data[0].values[-1] == np.sum(values)
-    # Continuous colorscale
```
nicolaskruchten, line 152: should we keep this one also just for completeness? it still passes right?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -474,11 +474,14 @@ def make_trace_kwargs(args, trace_spec, trace_data, mapping_labels, sizeref):
                 # We need to invert the mapping here
                 k_args = invert_label(args, k)
                 if k_args in args["hover_data"]:
-                    if args["hover_data"][k_args][0]:
-                        if isinstance(args["hover_data"][k_args][0], str):
-                            mapping_labels_copy[k] = v.replace(
-                                "}", "%s}" % args["hover_data"][k_args][0]
-                            )
+                    formatter = (
```
nicolaskruchten, line 477: could we get a test to lock down whatever bug this is correcting?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1576,6 +1581,9 @@ def aggfunc_continuous(x):
     if args["color"]:
         if not args["hover_data"]:
             args["hover_data"] = [args["color"]]
+        elif isinstance(args["hover_data"], dict):
```
nicolaskruchten, line 1584: could we get a test to lock down whatever bug this is correcting?

packages/python/plotly/plotly/tests/test_core/test_px/test_px_functions.py
emmanuelle: no, interestingly it does not because this one tests the behaviour we don't want any more (that `color` has the same values as `values`, since we used to have the same `sum` aggregation function but now we have two different functions.)

nicolaskruchten: got it, makes sense.

packages/python/plotly/plotly/tests/test_core/test_px/test_px_hover.py
```diff
@@ -151,3 +151,15 @@ def test_fail_wrong_column():
         "Ambiguous input: values for 'c' appear both in hover_data and data_frame"
         in str(err_msg.value)
     )
+
+
+def test_sunburst_hoverdict_color():
```
emmanuelle, line 156: this test fails on master because it tries to append to a dict

packages/python/plotly/plotly/express/_core.py
emmanuelle: test added

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1576,6 +1577,9 @@ def aggfunc_continuous(x):
...
+            if not args["hover_data"].get(args["color"]):
+                args["hover_data"][args["color"]] = (True, None)
```
emmanuelle, line 1582: in a first version I had put only `True` this is why I needed to modify the code later on but by putting a tuple directly there's no need for this anymore.

nicolaskruchten: ok great!

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1499,7 +1498,9 @@ def aggfunc_discrete(x):
 
         if args["color"]:
             if args["color"] == args["values"]:
-                aggfunc_color = "sum"
+                new_value_col_name = args["values"] + "_total"
+                df[new_value_col_name] = df[args["values"]]
+                args["values"] = new_value_col_name
```
nicolaskruchten, line 1503: so like this, the `<col>_total` appears in the hoverlabel as the value. This is what we want? Not sure what would be better...

nicolaskruchten: Maybe we want `<col>_sum` instead? or `<col>_weighted_average` for the other one? not sure

nicolaskruchten: `<col>_color` for the other one maybe?

packages/python/plotly/plotly/tests/test_core/test_px/test_px_hover.py
nicolaskruchten: 👍

packages/python/plotly/plotly/express/_core.py
emmanuelle: I'm fine changing `total` to `sum`. I put `total` like `branchvalues='total'`. We could also put a suffix for the `colors` column but if we want to minimize the occurrence of such suffixes, we'd better put it on the `values` column (only in the case where it's the same as the `colors` column).

nicolaskruchten: ok let's do `sum` and merge this thing :)

nicolaskruchten:
💃 when changed to `sum`

Oh and a changelog entry please?

emmanuelle:
will merge after the CI runs

PR #2598, opened 2020-06-25
accept integer and float columns in wide mode

### LINKED ISSUE #2561: wide-mode type checking too strict
As reported in https://community.plotly.com/t/df-plot-kind-box-with-plotly-returns-error-of-wide-form-data/40729/5?u=nicolaskruchten it would be nice if a mix of floats and ints were still accepted for wide-mode processing, rather than strictly saying the types don't match.

nicolaskruchten:
Closes #2561

## Code PR

  modified existing tests.

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1390,9 +1390,11 @@ def build_dataframe(args, constructor):
         del args["wide_cross"]
         dtype = None
         for v in wide_value_vars:
+            v_dtype = df_output[v].dtype.kind
+            v_dtype = "number" if v_dtype in ["i", "f"] else v_dtype
             if dtype is None:
-                dtype = df_output[v].dtype
-            elif dtype != df_output[v].dtype:
+                dtype = v_dtype
+            elif dtype != v_dtype:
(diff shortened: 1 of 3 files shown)


packages/python/plotly/plotly/express/_core.py
```diff
@@ -1390,9 +1390,11 @@ def build_dataframe(args, constructor):
         del args["wide_cross"]
         dtype = None
         for v in wide_value_vars:
+            v_dtype = df_output[v].dtype.kind
+            v_dtype = "number" if v_dtype in ["i", "f"] else v_dtype
```
emmanuelle, line 1394:
```suggestion
            v_dtype = "number" if v_dtype in ["i", "f", "u"] else v_dtype
```

emmanuelle: "u" is for unsigned integer it's used in images for example

emmanuelle:
:woman_dancing: (I suggested to add `"u"` in the number types)

nicolaskruchten:
Awesome, thanks for catching that... I'm not as familiar with the types here as I should be :)

PR #2613, opened 2020-07-03
Add Kaleido image export support

jonmmease:
This PR adds preview support for performing static image export using [Kaleido](https://github.com/plotly/Kaleido) instead of [Orca](https://github.com/plotly/orca). See the Kaleido README for discussion of project motivations and how it compares to Orca.  There has not yet been a stable release of Kaleido, so the API may still evolve a little.

With this PR, Kaleido support is enabled by installing the `kaleido` package from PyPI...
```
$ pip install kaleido
```
and setting the new `engine` kwarg in `plotly.io.to_image` and `plotly.io.write_image` to `"kaleido"`.  `"kaleido"` is also the default value when `kaleido` is installed.

TODO:
 - [x] Add `engine` kwarg to image renderers to make it possible to override default engine
 - [x] Add tests

----
### Original Approach

With this PR, Kaleido support is enabled by installing the `kaleido` package from PyPI...
```
$ pip install kaleido
```
and importing the `kaleido_export` future flag before importing plotly
```python
from _plotly_future_ import kaleido_export
import plotly.io as pio
pio.to_image
```
```
 <function plotly.io._kaleido.to_image(fig, format=None, width=None, height=None, scale=None, validate=True)>
```
When Kaleido mode is enabled, the `to_image` and `write_image` functions use Kaleido rather than Orca.

There is no change in default behavior if the `kaleido_export` future flag is not imported

There are still open questions about how we will roll out the transition from Orca to Kaleido. For example, do we wait for a version 5 for the switch? Do we support both Orca and Kaleido for a time and design a mechanism to dynamically select between them? What should we do with the `plotly.io.orca` module after orca support is dropped? Should we remove it, or remap compatible options to `plotly.io.kaleido` even though only a subset of the options are compatible?

This PR doesn't make any assertions about how we should do that going forward except that the Kaleido approach will still be invoked using the current `plotly.io.to_image` and `plotly.io.write_image` functions. I'd like to release this preview support in version 4.9.0 so we can start asking people to try it out and gathering feedback (especially when people have problems getting Orca working).


jonmmease:
As of `kaleido==0.0.1rc5`, Kaleido supports EPS and EMF formats using poppler (for EPS) and inkscape (for EMF) the same way Orca does.

This brings Kaleido to feature parity with Orca, which would make it more feasible to do a complete Orca to Kaleido switchover in a version 5.

packages/python/plotly/plotly/io/_kaleido.py
```diff
@@ -0,0 +1,183 @@
...
+    fig:
+        Figure object or dict representing a figure
+
+    format: str or None
+        The desired image format. One of
+          - 'png'
+          - 'jpg' or 'jpeg'
+          - 'webp'
+          - 'svg'
+          - 'pdf'
```
emmanuelle, line 29: add eps?

nicolaskruchten:
How about we add a new `engine` flag to `to_image` and `write_image` which defaults to `auto` in which case it uses Kaleido if installed, otherwise Orca. People could then force it to `orca` if they happen to have both installed and prefer the old way? I would feel comfortable doing this without bumping to 5.x :)

emmanuelle:
Would it be possible that `fig.write_image?` displays the docstring of the corresponding `pio` function, mentioning kaleido? This is not the case at the moment if I'm not mistaken.

jonmmease:
> Would it be possible that fig.write_image? displays the docstring of the corresponding pio function, mentioning kaleido? This is not the case at the moment if I'm not mistaken.

@emmanuelle I'm not sure I understand what you mean. Are you talking about auto-generating the docstrings?  We used to do this, but I removed it during the import optimization work because there was a non-trivial import delay.

jonmmease:
> How about we add a new engine flag to to_image and write_image which defaults to auto in which case it uses Kaleido if installed, otherwise Orca.

This works for `to_image`, but there would also need to be a solution for the image renderers.  We could add an engine option to the image renderers as well I suppose.  I'll take a look.

jonmmease:
Ok, I removed the future flag and added an `engine` kwarg to the `plotly.io.to_image` and `plotly.io.write_image` functions. Still need to add `engine` to the image renderes, but I think this approach will work well.  I also still need to port some of the image tests from orca to run with Kaleido as well.

jonmmease:
@nicolaskruchten and @emmanuelle this is ready for another review. I removed the future flag, added an `engine` kwarg to the `to_image` and `write_image`  functions, and to the `Image` renderer subclasses.

For testing, I took a different approach than with orca. I'm basically just using mocks to make sure that the `PlotlyScope.transform` method is called with the expected arguments in various situations.I think this is a bit cleaner since testing against image baselines is going to be handled by the kaleido project.

jonmmease:
The `python-2.7-optional` tests are failing due to a compatibility issue in `kaleido`. I'm building a new version that should be compatible with 2.7.

The `python-3.7-orca` test is failing due to that dendrogram error. Has anyone worked out the cause of that?

nicolaskruchten:
The 3.7 orca test failure is logged as https://github.com/plotly/plotly.py/issues/2618 and I should get to it this week :)

nicolaskruchten:
Thanks for adding the `engine` stuff! So if you removed the future flag, the default is "Kaleido if it's present" basically right? I kind of thought we would keep the future flag to control the polarity of the "auto" setting but I'm on the fence. In the short run, no one will have Kaleido installed for any reason other than to use it with Plotly, but in the future, there may be some situation we can't yet foresee wherein folks have Kaleido installed for some other library support and yet there's some issue such that they want to use Orca for Plotly and they'd consider this a breaking change.

I think I'm ok with that though... thoughts?

nicolaskruchten:
Are you planning on updating the docs in this PR as well? I think I'd like that if possible, including changes to the README and getting-started.md please :)

jonmmease:
I think I'd be fine bringing the future flag back to control the default.  But when would you picture changing the default, not till a v5?

Yeah, I'll do the docs and README. So far I've held off because I wasn't sure how much to present Kaleido as experimental with Orca as stable, vs Kaleido as recommended and Orca as legacy.

nicolaskruchten:
> I think I'd be fine bringing the future flag back to control the default. But when would you picture changing the default, not till a v5?

I'm not dead-set on adding the future flag... just wanted to bring up/discuss this outside chance of some future where someone is annoyed about this. I think I'm OK with running this risk.

jonmmease:
I updated the README and image export documentation to recommend Kaleido, while still describing how to use orca.

Note that the documentation for installing Kaleido using conda is dependent on the acceptance of https://github.com/conda-forge/staged-recipes/pull/12093 into conda-forge.

I'll merge in the dendrogram fix as soon as https://github.com/plotly/plotly.py/pull/2627 is merged, but otherwise this is ready for review.

jonmmease:
@nicholas-esterer @emmanuelle @nicholas-esterer 

Tests are passing now and this is ready for re-review.

README.md
```diff
@@ -133,12 +133,27 @@ jupyter labextension install @jupyter-widgets/jupyterlab-manager plotlywidget@4.
...
+```
+
+or conda.
+```
+$ conda install -c plotly python-kaleido
+``` 
+
+### Static Image Export with Orca
+
+While Kaleido is now the recommended approach, image export can also be supported
```
nicolaskruchten, line 154: I would mention that Orca is the older, harder to install option.

doc/python/orca-management.md
```diff
@@ -34,13 +34,65 @@ jupyter:
 ---
 
 ### Overview
-This section covers the lower-level details of how plotly.py uses orca to perform static image generation. Please refer to the [Static Image Export](/python/static-image-export/) section for general information on creating static images from plotly.py figures.
+This section covers the lower-level details of how plotly.py can use orca to perform static image generation. Please refer to the [Static Image Export](/python/static-image-export/) section for general information on creating static images from plotly.py figures.
```
nicolaskruchten, line 37: I would add a section above this saying "Orca is no longer the recommended way to do static export, and we now recommend Kaleido" with a link back to the static image export page.

doc/python/static-image-export.md
```diff
@@ -41,31 +41,20 @@ Plotly figures are interactive when viewed in a web browser: you can hover over
 
 
 <!-- #region -->
-#### Install Dependencies
-Static image generation requires the [orca](https://github.com/plotly/orca) commandline utility and the [psutil](https://github.com/giampaolo/psutil) and [requests](https://2.python-requests.org/en/master/) Python libraries. There are 3 general approach to installing these dependencies.
-
-##### conda
-Using the [conda](https://conda.io/docs/) package manager, you can install these dependencies in a single command:
+#### Install Dependencies (Kaleido)
+Static image generation requires the [`kaleido`](https://github.com/plotly/Kaleido) package which can be installed using pip...
```
nicolaskruchten, line 45: strictly speaking this isn't correct, it requires "either Kaleido (recommended) or Orca (deprecated)" or something like that, no? Just to avoid confusing people who already know Orca? (same comment applies in other places in this PR where we've said Kaleido is required)

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Added image export support using Kaleido. The image export backend can be configured using the new `engine` argument to `plotly.io.to_image` and `plotly.io.write_image`. The `engine` argument may be set to `"kaleido"`, `"orca"`, or `"auto"`. The default is `engine="auto"`, in which case the Kaleido backend is enabled if the `kaleido` package from PyPI is installed, otherwise Orca is used. ([#2613](https://github.com/plotly/plotly.py/pull/2613)).
```
nicolaskruchten, line 10:
```suggestion
- Added image export support using [Kaleido](https://github.com/plotly/Kaleido). The image export backend can be configured using the new `engine` argument to `plotly.io.to_image` and `plotly.io.write_image`. The `engine` argument may be set to `"kaleido"`, `"orca"`, or `"auto"`. The default is `engine="auto"`, in which case the Kaleido backend is enabled if the `kaleido` package from PyPI is installed, otherwise Orca is used. ([#2613](https://github.com/plotly/plotly.py/pull/2613)).
```

nicolaskruchten:
:dancer: modulo the documentation feedback

doc/python/static-image-export.md
nicholas-esterer: I can't select it but on line 40 JEPG should be JPEG.

doc/python/static-image-export.md
```diff
@@ -41,31 +41,20 @@ Plotly figures are interactive when viewed in a web browser: you can hover over
...
-If you are unable to install conda or npm, you can install orca as a precompiled binary for your operating system. Follow the instructions in the orca [README](https://github.com/plotly/orca) to install orca and add it to your system `PATH`. Then use pip to install `psutil`.
+#### Install Dependencies (Orca)
+While Kaleido is now the recommended approach, image export can also be supported by the [orca](https://github.com/plotly/orca) command line utility. See the [Orca Management](/python/orca-management/) section for instructions on installing, configuring, and troubleshooting orca.
 
-```
-$ pip install psutil requests
-```
 <!-- #endregion -->
 
 ### Create a Figure
```
nicholas-esterer, line 60: line 61 should be "Now let's create a simple scatter plot with 100 random points of varying color and size."

doc/python/static-image-export.md
```diff
@@ -215,7 +177,45 @@ img_bytes = fig.to_image(format="png", width=600, height=350, scale=2)
...
+```python
+import plotly.io as pio
+pio.kaleido.scope.default_format = "svg"
+```  
+
+Here is a complete listing of the available image export settings:
+
+ - **`default_width`**: The default pixel width to use on image export.
+ - **`default_height`**: The default pixel height to use on image export.
+ - **`default_scale`**: The default image scale facor applied on image export.
```
nicholas-esterer, line 209:
```suggestion
 - **`default_scale`**: The default image scale factor applied on image export.
```

doc/python/orca-management.md
```diff
@@ -34,13 +34,65 @@ jupyter:
...
+
+```python
+from google.colab import files
+files.download('fig1.svg')
+files.download('fig1.png')
+```
+<!-- #endregion -->
 
 ### Create a Figure
 Now let's create a simple scatter plot with 100 random points of variying color and size.
```
nicholas-esterer, line 98:
```suggestion
Now let's create a simple scatter plot with 100 random points of varying color and size.
```

nicholas-esterer, line 98:
```suggestion
Now let's create a simple scatter plot with 100 random points of varying color and size.
```

doc/python/orca-management.md
```diff
@@ -207,4 +259,4 @@ In addition to the `executable` property, the `plotly.io.orca.config` object can
```
nicholas-esterer, line 259: line 253 should be `- **`default_scale`**: The default image scale factor applied on image export.`

jonmmease: Fixed in 8731ecfe3402002d1eac53e74e7d7ce821b2c821

doc/python/static-image-export.md
jonmmease: Fixed in 70c11ac67c02f4fc0cbe4003f99b75d0941c2ab1

doc/python/static-image-export.md
jonmmease: Thanks:
 - JPEG change (a033df0c3e10f550d7d8df0448e43bacd1eb49e6)
 - "Write Image File" changes (2bdcfe210865695b9db89e64c5f55e6c2c039be8)
 - Merged install dependency sections and better explain that orca is legacy and kaleido is new (d1fc9a8a98830abdef33bbb15ef2b9ee56228023)

doc/python/orca-management.md
jonmmease: Update in a25f5b27c710c7b3635e2c3657a8536304f5e598

README.md
jonmmease: Update in 00987c05999b9b587a69b6644468cc807e5b2cb8

doc/python/orca-management.md
jonmmease: done

jonmmease:
Thanks for the reviews @nicholas-esterer and @nicolaskruchten, updates applied. Merging!

PR #2627, opened 2020-07-08
KeyError when running the create_dendrogram example

### LINKED ISSUE #2618: Error when running the create_dendrogram example
I am running plotly.py version 4.8.2 in a jupyter notebook.
I ran the following code, copied from the [documentation page](https://plotly.com/python/dendrogram/):
```
import numpy as np
np.random.seed(1)

X = np.random.rand(10,10)
fig = ff.create_dendrogram(X)
fig.update_layout(width=800, height=500)
fig.show()
```
and getting the following error message:
```
KeyError                                  Traceback (most recent call last)
<ipython-input-1-386e391a19fb> in <module>
      4 
      5 X = np.random.rand(10,10)
----> 6 fig = ff.create_dendrogram(X)
      7 fig.update_layout(width=800, height=500)
      8 fig.show()

~/miniconda3/envs/jupyterlab_new/lib/python3.8/site-packages/plotly/figure_factory/_dendrogram.py in create_dendrogram(X, orientation, labels, colorscale, distfun, linkagefun, hovertext, color_threshold)
     89         distfun = scs.distance.pdist
     90 
---> 91     dendrogram = _Dendrogram(
     92         X,
     93         orientation,

~/miniconda3/envs/jupyterlab_new/lib/python3.8/site-packages/plotly/figure_factory/_dendrogram.py in __init__(self, X, orientation, labels, colorscale, width, height, xaxis, yaxis, distfun, linkagefun, hovertext, color_threshold)
    143             distfun = scs.distance.pdist
    144 
--> 145         (dd_traces, xvals, yvals, ordered_labels, leaves) = self.get_dendrogram_traces(
    146             X, colorscale, distfun, linkagefun, hovertext, color_threshold
    147         )

~/miniconda3/envs/jupyterlab_new/lib/python3.8/site-packages/plotly/figure_factory/_dendrogram.py in get_dendrogram_traces(self, X, colorscale, distfun, linkagefun, hovertext, color_threshold)
    335                 y=np.multiply(self.sign[self.yaxis], ys),
    336                 mode="lines",
--> 337                 marker=dict(color=colors[color_key]),
    338                 text=hovertext_label,
    339                 hoverinfo="text",

KeyError: 'C1'
```
I've tried downgrading to 4.8.1, but got the same error.
My working env is created with miniconda3, using the yml file:
```
name: jupyterlab_new
channels:
  - conda-forge
  - bioconda
dependencies:
  - jupyterlab
  - pandas
  - numpy
  - seaborn
  - plotly
  - biopython
```

Thanks!

nicholas-esterer:
closes #2618 

With the new color names ('C0','C1', 'C2', ...) I only added 5 colours whereas with the old names 8 colours were added. Looking at `matplotlib.rcParams['axes.prop_cycle']` it seems up to `'C9'` is possible, so do we add mappings for colours 'C5' to 'C9'?

### DIFF
--- a/packages/python/plotly/plotly/figure_factory/_dendrogram.py
+++ b/packages/python/plotly/plotly/figure_factory/_dendrogram.py
@@ -198,7 +198,7 @@ class _Dendrogram(object):
         default_colors = OrderedDict(sorted(d.items(), key=lambda t: t[0]))
 
         if colorscale is None:
-            colorscale = [
+            rgb_colorscale = [
                 "rgb(0,116,217)",  # blue
                 "rgb(35,205,205)",  # cyan
                 "rgb(61,153,112)",  # green
@@ -206,13 +206,37 @@ class _Dendrogram(object):
                 "rgb(133,20,75)",  # magenta
                 "rgb(255,65,54)",  # red
                 "rgb(255,255,255)",  # white
-                "rgb(255,220,0)",
-            ]  # yellow
+                "rgb(255,220,0)",  # yellow
+            ]
+        else:
+            rgb_colorscale = colorscale
 
         for i in range(len(default_colors.keys())):
             k = list(default_colors.keys())[i]  # PY3 won't index keys
-            if i < len(colorscale):
-                default_colors[k] = colorscale[i]
+            if i < len(rgb_colorscale):
+                default_colors[k] = rgb_colorscale[i]
+
+        # add support for cyclic format colors as introduced in scipy===1.5.0
+        # before this, color_list, from which the color_key is obtained was:
+        # ['g', 'r', 'b', 'c', 'm', 'b', 'b', 'b', 'b'], now it is
+        # ['C1', 'C2', 'C0', 'C3', 'C4', 'C0', 'C0', 'C0', 'C0'], so to keep the
+        # colors consistent regardless of the version of scipy, 'C1' is mapped
+        # to 'rgb(61,153,112)' (what 'g' was mapped to before), 'C2' is mapped
+        # to 'rgb(255,65,54)', etc.
+        cyclic_color_names = ["C%d" % (n,) for n in range(5)]
+        if colorscale is None:
+            cyclic_color_rgb = [
+                "rgb(0,116,217)",
+                "rgb(61,153,112)",
+                "rgb(255,65,54)",
+                "rgb(35,205,205)",
+                "rgb(133,20,75)",
+            ]
+        else:
+            cyclic_color_rgb = colorscale
+
+        for k, c in zip(cyclic_color_names, cyclic_color_rgb):
+            default_colors[k] = c
 
         return default_colors
 
--- a/packages/python/plotly/plotly/figure_factory/_dendrogram.py
+++ b/packages/python/plotly/plotly/figure_factory/_dendrogram.py
@@ -25,36 +25,22 @@ def create_dendrogram(
     color_threshold=None,
 ):
     """
-    Function that returns a dendrogram Plotly figure object.
+    Function that returns a dendrogram Plotly figure object. This is a thin
+    wrapper around scipy.cluster.hierarchy.dendrogram.
 
     See also https://dash.plot.ly/dash-bio/clustergram.
 
     :param (ndarray) X: Matrix of observations as array of arrays
     :param (str) orientation: 'top', 'right', 'bottom', or 'left'
     :param (list) labels: List of axis category labels(observation labels)
-    :param (list) colorscale: Optional colorscale for dendrogram tree. To
-                              totally replace the default colorscale, a custom 
-                              colorscale must contain 8 colors, corresponding 
-                              to when the underlying 
-                              scipy.cluster.hierarchy.dendrogram specifies 
-                              'b', 'c', 'g', 'k', 'm', 'r', 'w', 'y', in that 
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
So with scipy 1.4.1, if I made a plotly dendrogram with 10 colors, which colours would they be?

nicolaskruchten:
(small request: could you please name your branches and your PRs a little more descriptively? a good branch name here would be `fix2618_dendrogram` and a good PR name would be "Fix ff.create_dendrogram with scipy 1.5+" or something like this :)

nicolaskruchten:
To answer my own question, here's what the output looks like today with scipy 1.4.1:

image

nicolaskruchten:
This same code fails on this branch with `scipy` 1.5.1 with the error `KeyError: 'C5'`, but ideally it would have the same output as above.

nicolaskruchten:
Final note: let's slightly improve the docstring and doc page for this thing so that it's clearer that this is a thin thin wrapper around https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.dendrogram.html :)

I might actually say this just visualizes the output of 

```python
from scipy.cluster.hierarchy import dendrogram
dendrogram(
    Z=linkagefun(distfun(X)),
    orientation=orientation,
    labels=labels,
    color_threshold=color_threshold,
)
```

nicholas-esterer:
So sadly I don't think we can get exactly the same color sequence with scipy==1.5.1, because the Cn color sequence and 'r','g','b' etc. are actually not isomorphic, but it's pretty close:

image
image

(number shown is the scipy version number)

nicolaskruchten:
That is indeed pretty close! I'm surprised that it's *that* close without being identical :)

nicolaskruchten:
Oh, I see, you're repeating "red, green, cyan" at the end of the cycle?

nicholas-esterer:
I don't think so, I just tried to find a mapping that made the change subtle. It was really hard to figure out exactly the relationship between the 1.4.1 colors and the 1.5.1 colors, but if you compare the sequences for a large dendrogram (in the old version a list of 'r','g', etc., in the new version  a list of 'c1', 'c2', etc), there are cns that map to multiple old colors. But I think the mapping I chose keeps the color change pretty minimized between version updates. I'll submit a PR in a few minutes and you can see exactly what I did.

nicholas-esterer:
I also improved the documentation for the colorscale argument. You might think the documentation is weird, but this is literally what the argument does :O !

packages/python/plotly/plotly/figure_factory/_dendrogram.py
```diff
@@ -32,7 +32,29 @@ def create_dendrogram(
     :param (ndarray) X: Matrix of observations as array of arrays
     :param (str) orientation: 'top', 'right', 'bottom', or 'left'
     :param (list) labels: List of axis category labels(observation labels)
-    :param (list) colorscale: Optional colorscale for dendrogram tree
+    :param (list) colorscale: Optional colorscale for dendrogram tree. To
```
nicolaskruchten, line 35: this is probably too detailed an explanation... most people will likely not want to follow the indirection and go read up on what scipy does here. I think just saying "should be 8 colors" is sufficient :)

nicolaskruchten: And we should mention that no matter the version of `scipy`, the 7th color in the scale is ignored (note: this is because the `w` key isn't outputted by scipy), and in `scipy` 15+, colors 1,2 and 3 are used twice as often as the others. 🤦

nicolaskruchten:
OK, let's just update the docstring with my terser version and merge this.

packages/python/plotly/plotly/figure_factory/_dendrogram.py
```diff
@@ -25,14 +25,22 @@ def create_dendrogram(
...
     See also https://dash.plot.ly/dash-bio/clustergram.
 
     :param (ndarray) X: Matrix of observations as array of arrays
     :param (str) orientation: 'top', 'right', 'bottom', or 'left'
     :param (list) labels: List of axis category labels(observation labels)
-    :param (list) colorscale: Optional colorscale for dendrogram tree
+    :param (list) colorscale: Optional colorscale for the dendrogram tree. With
+                              scipy<=1.4.1 requires 8 colors to be specified,
+                              the 7th of which is ignored.  With scipy>=1.5.0,
+                              requires 10 colors. In this case the 8th color is
```
nicolaskruchten, line 39: so actually we only ever want 8 here, and the 7th is always ignored.

nicolaskruchten: right?

nicholas-esterer: Right what I wrote this time is actually wrong. Yes we always want 8, and the 7th is ignored. How about this:
```
:param (list) colorscale: Optional colorscale for the dendrogram tree.                          Requires 8 colors to be specified, the 7th of
                          which is ignored.  With scipy>=1.5.0, the 2nd, 3rd
                          and 6th are used twice as often as the others.
                          Given a shorter list, the missing values are
                          replaced with defaults and with a longer list the
                          extra values are ignored.
```

nicolaskruchten: Looks great! 💃  once done

PR #2631, opened 2020-07-10
Fixed make_docstring's override_dict 

RenaudLN:
For arguments that are not in `docs`, the override_dict only displayed the first argument of the docstring (the type) but not the rest of the description. This was the case for `px.pie` with the `hole` argument for instance.


nicolaskruchten:
Yikes, thanks for catching that :)

PR #2662, opened 2020-07-23
Update time-series.md

Nermin-Ghith:
Hi,
There is a missing library for figures 6-7 and 9-10.
So, I added (import plotly.express as px) into the script for these figures. 
BW,
Nermin


nicolaskruchten:
Nice catch, thank you :)

PR #2687, opened 2020-08-05
Sphinx fixes (against doc-prod)

emmanuelle:
Follow-up of #2681, but against doc-prod (I had to create a new PR because of the complicated merge history). I will of course squash before merging.


emmanuelle:
Merging since #2681 has been approved.

PR #2711, opened 2020-08-17
Fix selection behaviour of histograms

### LINKED ISSUE #2698: The selectedpoints attribute for Histogram plots is being set incorrectly
I am working in a Jupyter notebook and when accessing the `selectedpoints` attribute of a `go.Histogram` or the points list in the `on_selection` callback, the list does not contain integer indices of the selected points but instead just a list of objects.

```python
import numpy as np

np.random.seed(1)
x = np.random.randn(500)

def on_selection_print(obj, points, selector):
    print(points.point_inds)

figure_widget = go.FigureWidget(data=[go.Histogram(x=x)])
figure_widget.update_layout(dragmode="select")    
figure_widget.data[0].on_selection(on_selection_print)
figure_widget
```

When selecting points in the histogram, the output looks something like:
`[<object object at 0x7ff062c497f0>, <object object at 0x7ff062c497f0>, <object object at 0x7ff062c497f0>, <object object at 0x7ff062c497f0>, <object object at 0x7ff062c497f0>, <object object at 0x7ff062c497f0>]`

A call to `figure_widget.data[0].selectedpoints` returns the same list.

The expected output would be something like:
`[0, 12, 13, 26, 31, ..., 2534]`

As you can see, all objects are identical (memory address `0x7ff062c497f0`) and the objects in the list do not have any attributes. The length of the list corresponds to the number of bars selected if that helps.

I am working with plotly version `4.9.0`, ipywidgets version `7.5.1` and notebook version `6.0.1`.

**UPDATE**:
I checked how Plotly.js handles selection for histograms and found that it creates an object for each selected bar, which in turn contains all the points for that bar.
e.g.
```javascript
{
    points: [
        0: {..., pointIndices: [1,2,3,4], ...},
        1: {..., pointIndices: [5,6,7,8,9], ...},
        ...
    ],
    range: [...]
}
```
Codepen can be found here: https://codepen.io/meffmadd/pen/zYqGmaw

So the problem seems to be that the points are not correctly serialised for histograms. Without understanding the whole codebase I think there seem to be two options:
1. Update the setter to `_js2py_pointsCallback` when the figure is a histogram in https://github.com/plotly/plotly.py/blob/efc66509526bdde772e8a8caaf1bcee865ccec3a/packages/javascript/plotlywidget/src/Figure.js#L1134
2. Updating the model for the histogram on the Python side to reflect the behaviour of the JavaScript implementation. I don't know how exactly this would work yet.

Any pointers in the right direction are highly appreciated!

meffmadd:
This change addresses the selection behaviour in histograms. 

A selection now returns the point indices in a single list like scatter plots or others. In the past, the selection was just a list of empty objects due to histograms having a different structure of the selected points. More information can be found in my [comment](https://github.com/plotly/plotly.py/issues/2698#issuecomment-674785363) on the original issue.

Now the indices of the selection are passed correctly:
histogram selection

Fixes #2698.

### DIFF
--- a/packages/javascript/plotlywidget/src/Figure.js
+++ b/packages/javascript/plotlywidget/src/Figure.js
@@ -928,7 +928,6 @@ var FigureView = widgets.DOMWidgetView.extend({
         }
       }
 
-
       // Add z if present
       var hasZ =
         pointObjects[0] !== undefined && pointObjects[0].hasOwnProperty("z");
--- a/packages/javascript/plotlywidget/src/Figure.js
+++ b/packages/javascript/plotlywidget/src/Figure.js
@@ -889,10 +889,17 @@ var FigureView = widgets.DOMWidgetView.extend({
       var pointObjects = data["points"];
       var numPoints = pointObjects.length;
 
-      var hasNestedPointObjects = pointObjects.every(value => value.hasOwnProperty("pointNumbers"));
+      var hasNestedPointObjects = true;
+      for (let i = 0; i < numPoints; i++) {
+        hasNestedPointObjects = (hasNestedPointObjects && pointObjects[i].hasOwnProperty("pointNumbers"));
+        if (!hasNestedPointObjects) break;
+      }
       var numPointNumbers = numPoints;
       if (hasNestedPointObjects) {
-        numPointNumbers = pointObjects.reduce((acc, v) => acc + v["pointNumbers"].length, 0)
+        numPointNumbers = 0;
+        for (let i = 0; i < numPoints; i++) {
+          numPointNumbers += pointObjects[i]["pointNumbers"].length;
+        }
       }
       pointsObject = {
         trace_indexes: new Array(numPointNumbers),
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
Thanks for this pull request! @alexcjohnson and/or @jonmmease can you take a look please?

jonmmease:
Hi @meffmadd, thanks for the PR. I'll look it over soon. Sorry it slipped through cracks initially!

jonmmease:
Looks and works great @meffmadd, thanks!

jonmmease:
BTW, https://github.com/meffmadd/pandas-visual-analysis looks really cool!

meffmadd:
> BTW, https://github.com/meffmadd/pandas-visual-analysis looks really cool!

Thank you so much! 😊

I'm happy I could contribute to this awesome project!

iadjali:
According to the plotly.py documentation clickmode='select' should also set selectedpoints with the points indices in the clicked bar of the histogram. However this is not working for me in with figurewidget in a Jupyter notebook; the only way to get the points in a specific bar of the histogram is to to use dragmode (not ideal if you only want the data in one bar), clickmode doesn't seem to do anything to the points argument in the callback function. Am I missing something?

PR #2724, opened 2020-08-24
Clarified where to find higher-level functions on figures

nicholas-esterer:
I think it is important to clarify which code needs to be generated and which code needs to be written in Python. For example, I think in the contributing guide, the `packages/python/plotly/codegen/figure.py` file is being referred to, but there also might be cases where one needs to contribute to `packages/python/plotly/plotly/basedatatypes.py`. My change is not the final one, but I wanted to get the conversation going.


nicolaskruchten:
thanks!

PR #2728, opened 2020-08-26
Fix warning regarding ABC import from collections

tirkarthi:
Similar to https://github.com/plotly/plotly.py/pull/1417


nicolaskruchten:
Thanks for the fix!

PR #2730, opened 2020-08-28
distplot: only compute traces that will be shown

akbo:
In this pull request I make a small change to ff.create_distplot. ff.create_distplot produces a figure showing the distribution of some data. It can create three traces (histogram, kernel density estimation (KDE) and rug) and it has switches that let users select, which of the aforementioned traces to include in the resulting figure.

Independent of the user choice (boolean parameters: show_hist, show_curve and show_rug), it will compute all three traces whether they will be included in the resulting figure or not. I changed the code so that only the traces that will be included in the resulting figure get computed. 

This will make the function more efficient in situations where a user doesn't need all three traces. 

Also (this is my use case), it allows one corner case that currently doesn't work: The function currently doesn't support being called with a dataset (hist_data parameter) that only contains a single entry. What happens in this case is that scipy throws an exception when trying to compute the KDE. However, in some use cases, there is no need for a KDE (show_curve=False). It should be possible to create a distplot with a single-entry dataset without a KDE.

## Code PR

  modified existing tests.

### DIFF
--- a/packages/python/plotly/plotly/figure_factory/_distplot.py
+++ b/packages/python/plotly/plotly/figure_factory/_distplot.py
@@ -183,61 +183,66 @@ def create_distplot(
     if isinstance(bin_size, (float, int)):
         bin_size = [bin_size] * len(hist_data)
 
-    hist = _Distplot(
-        hist_data,
-        histnorm,
-        group_labels,
-        bin_size,
-        curve_type,
-        colors,
-        rug_text,
-        show_hist,
-        show_curve,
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
Seems straightforward enough! Thanks :)

nicolaskruchten:
Released as part of 4.10! https://community.plotly.com/t/announcing-plotly-py-4-10-date-axis-improvements-a-faster-px-imshow-and-full-figures-for-development/44685

Issue #1967, opened 2019-12-04
Access _fullLayout, _fullData and JS errors from Python


nicolaskruchten:
Just to record my pie-in-the-sky idea: if [Orca](https://github.com/plotly/orca) had a mode where  instead of returning an image, it returned a JSON object that contained `_fullLayout`, `_fullData` and some warning objects, then we could make it much easier for users to understand the defaulting logic and various other-wise opaque error messages. From Python this could be accessed via a new renderer as `fig.show("full_json")` or something?

nicolaskruchten:
@antoinerg heads-up!

nicolaskruchten:
The JSON bit is going to be pretty awesome, based on Kaleido... basically something as simple as:

```python
import plotly.io as pio
import json
import plotly.graph_objects as go

def coerce(fig):
    return go.Figure(json.loads(pio.kaleido.scope.transform(fig, format="json")), skip_invalid=True)
```

gvwilson:
Hi - we are trying to tidy up the stale issues and PRs in Plotly's public repositories so that we can focus on things that are still important to our community. Since this one has been sitting for several years, I'm going to close it; if it is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. Thanks for your help - @gvwilson

PR #2737, opened 2020-09-01
full_figure_for_development

nicolaskruchten:
Addresses the non-error portion of #1967

@jonmmease I'm not sure how to best test this or if such a trivial function is worth a test... ?


packages/python/plotly/plotly/io/_kaleido.py
```diff
@@ -260,4 +261,58 @@ def write_image(
...
+    -------
+    plotly.graph_objects.Figure or dict
+        The full figure
+    """
+
+    # Raise informative error message if Kaleido is not installed
+    if scope is None:
+        raise ValueError(
+            """
+Image export using the "kaleido" engine requires the kaleido package,
```
jonmmease, line 294: Should probably refer to full_figure_for_development instead of Image export here

nicolaskruchten: thanks! done

nicolaskruchten:
OK added a test and some docs!

nicolaskruchten:
OK, all done! I had to add `.decode("utf-8")` to get bytes into strings for the Python 3.5 optional tests to pass ... normal? that'll work on 2.7 also right?

nicolaskruchten:
also I added a doc page... feedback welcome :)  @emmanuelle if you want to take a look that'd be cool too :)

jonmmease:
> OK, all done! I had to add .decode("utf-8") to get bytes into strings for the Python 3.5 optional tests to pass ... normal? that'll work on 2.7 also right?

Not sure about the Python 3.5 thing specifically, but decoding explicitly to utf-8 is the right thing to do.

Issue #2311, opened 2020-03-25
[BUG] Plotly Express (px) pandas datetime with timezone converted to UTC


m0wer:
Similar to what happened in <https://github.com/plotly/plotly.py/issues/209> Plotly Express converts the Pandas datetime object to UTC instead of respecting its timezone. If the datetime is a string or is converted to `pydatetime` it works as expected. A reproducible example is available at <https://gist.github.com/m0wer/74d5a8901b9499795d13992d29cc2f0c>.

A workaround (shown in the example) is to convert the datetime column of the Pandas DataFrame to `pydatetime` with:

```python
df['date'].dt.to_pydatetime()
```

amadeuspzs:
I can confirm this behaviour and also perceive it to be a bug

nicolaskruchten:
Hmm, it is really unfortunate that Plotly Express behaves differently than Graph Objects here, as `px.scatter(df, x='date', y='value')` does what is expected. I'll have to think about how big a breaking change this is.

nicolaskruchten:
Update: `px.scatter(df, x='date', y='value')` does *not* do what is expected per the changes we made in the 4.0 release (see https://plotly.com/python/v4-migration/#timezones), but `go.Figure(go.Scatter(x=df["date"], y=df["value"]))` does in this particular case... digging into why and what we can do about it.

nicolaskruchten:
I suspect this was introduced in https://github.com/plotly/plotly.py/pull/1934 because pulling the `.values` out of a series seems to strip the timezone information:

![image](https://user-images.githubusercontent.com/203523/91625705-95185000-e977-11ea-8eab-2b295fa1c301.png)

@jonmmease @alexcjohnson @emmanuelle should we fix this? The bug has been in the wild now much longer than the non-buggy code :(

nicolaskruchten:
Also cc @chriddyp would love your input on this one.

nicolaskruchten:
I'm thinking we should fix it... the only plots which will break are those where folks made plots on an hourly timescale and fed in data with timezones but really wanted the plot to be in UTC and relied on the fact that it's automatic, which feels like a small set to me.

alexcjohnson:
I think I agree we should fix it - px and non-px need to behave the same, and if / when we implement real timezone support in plotly.js the fixed version will still be compatible.

nicolaskruchten:
[...]

PR #2749, opened 2020-09-05
fix PX timezone treatment

nicolaskruchten:
Closes https://github.com/plotly/plotly.py/issues/2311 essentially by removing calls to `pd.Series.values`, which seems to always convert datetimes to UTC.

  modified existing tests.

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -222,7 +222,6 @@ def make_trace_kwargs(args, trace_spec, trace_data, mapping_labels, sizeref):
     trace_patch = trace_spec.trace_patch.copy() or {}
     fit_results = None
     hover_header = ""
-    custom_data_len = 0
     for attr_name in trace_spec.attrs:
         attr_value = args[attr_name]
         attr_label = get_decorated_label(args, attr_value, attr_name)
@@ -243,7 +242,7 @@ def make_trace_kwargs(args, trace_spec, trace_data, mapping_labels, sizeref):
                 )
             ]
             trace_patch["dimensions"] = [
-                dict(label=get_label(args, name), values=column.values)
+                dict(label=get_label(args, name), values=column)
                 for (name, column) in dims
             ]
             if trace_spec.constructor == go.Splom:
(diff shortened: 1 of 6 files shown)


nicolaskruchten:
Ok this all works on my machine but fails all over CI. Looks like my approach is too brittle :)

nicolaskruchten:
New approach succeeded: do everything in pandas-land instead of transiting through numpy-land at all :) 

Ready for review!

packages/python/plotly/tox.ini
```diff
@@ -81,25 +81,25 @@ deps=
...
 commands=
     python --version
-    pytest {posargs} -x plotly/tests/test_core
+    pytest {posargs} plotly/tests/test_core
 
 [testenv:py36-core]
 basepython={env:PLOTLY_TOX_PYTHON_36:}
 commands=
     python --version
-    pytest {posargs} -x plotly/tests/test_core
```
emmanuelle, line 96: why did you remove the `-x` ? Exiting on first failing test is good for the environment :-).

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1029,6 +1020,16 @@ def _escape_col_name(df_input, col_name, extra):
...
 
 
+def to_unindexed_series(x):
+    """
+    assuming x is list-like or even an existing pd.Series, return a new pd.Series with
+    no index, without extracting the data from an existing Series via numpy, which
+    seems to mangle datetime columns. Stripping the index from existing pd.Series is
+    required to get things to match up right in the new DataFrame we're building
+    """
+    return pd.Series(x).reset_index(drop=True)
```
emmanuelle, line 1030: this makes a copy of the series because of the `reset_index` step, so it's suboptimal for non datetime columns. Could a solution be to return `x.values` in other cases?

packages/python/plotly/tox.ini
nicolaskruchten: I needed to be able to see if I fixed a test even if an "earlier" one failed. Looking at the timings, it seems to me that the container setup etc takes more time than the test suites proper, so while we have the container spun up we may as well extract as much information as possible about the run, is my feeling :)

packages/python/plotly/plotly/express/_core.py
nicolaskruchten: Hmmm yeah we weren't copying before but we are now. After spending like 8 hours trying various permutations here I think I'm OK with the copy on the principle of "make it correct, then make it fast". We can optimize it later if this really causes problems.

PR #2761, opened 2020-09-10
Use x and y parameters for Image trace in imshow (for RGB or binary_string=True)

### LINKED ISSUE #2758: `x` and `y` arguments are not taken into account when using `go.Image` inside `px.imshow`
For example
```python
from skimage import data
import numpy as np
import plotly.express as px
img = data.chelsea()
x = 10 * np.arange(img.shape[1])
fig = px.imshow(img, x=x)
fig.show()
```

image

RGB images (and single-channel images as well with `binary_string`=True) use `go.Image`, which does not accept `x` and `y` arguments. For `xarrays` in particular, it would be convenient to be able to use `x` and `y`.

emmanuelle:
Closes #2758 

I could only get it to work for numerical data types, I don't think the `x0` and `dx` parameters of `go.Image` are meant to work with str or datetime types (could you please confirm @antoinerg?)

For example for the `airtemps` dataset used in the [imshow tutorial](https://plotly.com/python/imshow/#display-an-xarray-image-with-pximshow), this now works
image


nicolaskruchten:
What does the output look like if you try this with an xarray with dates?

nicolaskruchten:
or should we raise in that case instead of trying to render?

nicolaskruchten:
@emmanuelle *bump* on my last comment :)

emmanuelle:
@nicolaskruchten looks like 
image
(dates are converted to time intervals I guess)

I'll look into accepting only numerical types for `x` and `y`.

nicolaskruchten:
thanks! I think the dates are just ignored and the axis just counts pixels actually... not all that graceful :(

packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -403,6 +397,27 @@ def imshow(
...
             )
+        x0, y0, dx, dy = (None,) * 4
+        if x is not None:
+            x = np.asanyarray(x)
+            if np.issubdtype(x.dtype, np.number):
+                x0 = x[0]
+                dx = x[1] - x[0]
+            else:
+                raise ValueError(
+                    "Only numerical values are accepted for the `x` parameter "
```
nicolaskruchten, line 408: this is correct but I think a bit opaque for users? in the xarray case they've maybe not specified `x` (as it's implicit) and they don't know they're using an `Image` trace (all they know is they set `binary_string=True`

emmanuelle: ok I can think of a better error message, but do we agree that we should also error when `x` is passed explicitly and does not have a numerical dtype? (otherwise it would just be ignored, which I agree is not a good thing).

emmanuelle: I changed the error message, is it better now?

nicolaskruchten:
LGTM! :dancer:

PR #2769, opened 2020-09-14
Fix wrong slicing in performance improvement example

danieleldjarn:
Fix wrong slicing in performance improvement example and remove unused variable


nicolaskruchten:
Nice catch, thank you :)

PR #2771, opened 2020-09-15
Fix selection of histograms with multiple traces

meffmadd:
This change addresses the selection behaviour of histograms with more than one trace and fixes a bug introduced in my last change #2711.

In the previous change, the point_indexes were sorted before being returned. This is not a problem when there is only one trace, however, with multiple traces, the indices of the values get mixed and result in incorrect behaviour.

So for example consider two traces with a random distribution and both traces are being selected. Firstly, all the indices are combined into one large array with the corresponding trace_indices [0,0,…,0,0,1,1 …,1,1].
Then the index-array is sorted and all the indices get reordered while the trace information is unchanged, which causes indices being assigned to the incorrect traces later on.

A quick solution is not sorting the array, which results in correct behaviour. This has the effect that the results are also not sorted in Python in each trace. Please let me know if this is an issue.

Screenshot 2020-09-15 at 12 57 24

### DIFF
--- a/packages/javascript/plotlywidget/src/Figure.js
+++ b/packages/javascript/plotlywidget/src/Figure.js
@@ -919,9 +919,6 @@ var FigureView = widgets.DOMWidgetView.extend({
             pointsObject["trace_indexes"][flatPointIndex] = pointObjects[p]["curveNumber"];
           }
         }
-        pointsObject["point_indexes"].sort(function(a, b) {
-          return a - b;
-        });
       } else {
         for (var p = 0; p < numPoints; p++) {
           pointsObject["trace_indexes"][p] = pointObjects[p]["curveNumber"];
--- a/packages/javascript/plotlywidget/src/Figure.js
+++ b/packages/javascript/plotlywidget/src/Figure.js
@@ -1,7 +1,7 @@
 var widgets = require("@jupyter-widgets/base");
 var _ = require("lodash");
 
-window.PlotlyConfig = { MathJaxConfig: "local" };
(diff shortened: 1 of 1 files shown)


meffmadd:
Are there any updates on this issue, @nicolaskruchten, @alexcjohnson and/or @jonmmease? 

In the meantime I also implemented sorting for each trace individually, however, this comes with a performance penalty so let me know if this is necessary for multiple traces.

nicolaskruchten:
@jonmmease if you have a sec to look at this one I'd appreciate it :)

jonmmease:
Not sorting the indices is fine. This looks good to me. Thanks, @meffmadd!

meffmadd:
Thank you for reviewing the changes! I never pushed the version where the indices are sorted so the PR could be merged.

meffmadd:
Hey @nicolaskruchten, I think this PR fell through the cracks but as far as I understand it, it was approved and could be merged immediately without causing problems. It would really help my project to have this functionality. Thanks!

nicolaskruchten:
Indeed, we've not done a good job of pushing this PR along, and I apologize. The next version of Plotly.py will be v5.0 and I expect it will come out in about a month, and I'll try to get this merged and QA'ed for that release.

meffmadd:
Thank you very much and no problem! I was busy either way the last couple of months but recently found the time to work on my stuff again.

jonmmease:
Finally merging this, now that we're pulling version 5 changes into master. Thanks again @meffmadd!

meffmadd:
Thanks, always happy to contribute!

PR #2813, opened 2020-10-08
Import problem does not work in Jupyter lab

reservoirinvest:
Should this be 'uninstall' instead of 'remove'?

`$ pip remove plotly` gives:
```
ERROR: unknown command "remove"
```
in JupyterLab -> Terminal v2.2.8 on Windows 10.


nicolaskruchten:
Fixing downstream... thanks for the heads-up, sorry for the long delay :)

Issue #2072, opened 2020-01-15
magic underscore error messages


nicolaskruchten:
When doing something like `.update_layout(geo_lataxis_nonexistent=False)` the validation error I get today is from `go.Layout` but it would be better to have it come from `go.Layout.geo.lataxis` or however far in the magic_underscore_path is correct

nicholas-esterer:
When a property is set via an argument using the `__init__` method, all the known arguments are processed first, then `BasePlotlyType._process_kwargs` in `packages/python/plotly/plotly/basedatatypes.py` is called on the arguments that didn't match any known arguments. This function uses the `in` keyword to see if the argument is valid, which calls `BasePlotlyType.__contains__`. This function actually goes down the tree of magic underscores and finds the last object that was found using a valid specifier. But when the function fails it simply returns `False` (which it probably needs to do in order to be compatible with `in`). So one way to do this would be to make `__contains__` into a function that wraps another function called say `walk_magic_underscore_tree`. This way if `__contains__` returns `False`, we can call `walk_magic_underscore_tree` on the argument it returned `False` on to see how far down the tree it got before failing. Of course it'd be "more optimal" to not call the same function twice, but I worry that changing the `__contains__` function to something else that returns a boolean and the last object it found would require a lot of changes elsewhere in the code (we'd have to change `in` calls to something else expecting 2 return values).

nicholas-esterer:
Below I've tried different ways of using magic underscores to see where I need to tackle the problem and if I am missing any use cases Here it is clear the change needs to be made both in `BaseFigure` and `BasePlotlyType`, maybe there are other places I missed? Also are there ways to use magic underscores that I missed?

```
import plotly.graph_objects as go

fig=go.Figure()
fig.add_trace(go.Scatter(x=[],y=[]))

try:
    print(fig['layout_flakes'])
except KeyError as e:
    # This works properly but it could be improved by saying that it's trying to
    # assign a key called "flakes" in layout and failing.
    print("<<<Accessing property of figure using key")
    print(e.args)

try:
    fig['layout_flakes']=(go.layout.Shape(),)
except ValueError as e:
    # This works properly, maybe it could print where it got to before failing, e.g,:
    #
    # Bad Property Path: layout_flakes
    #                           ^
    # so users can see exactly where the problem is.
    print("<<<Setting property of figure using key")
    print(e.args)

try:
    fig2 = go.Figure(
[...]

PR #2824, opened 2020-10-09
Deep magic underscore error messages

nicholas-esterer:
Addresses https://github.com/plotly/plotly.py/issues/2072
Not finished yet, just putting up so others can review and comment.


packages/python/plotly/_plotly_utils/utils.py
```diff
@@ -2,6 +2,8 @@
 import json as _json
 import sys
 import re
+from functools import reduce
+from numpy import cumsum
```
nicolaskruchten, line 6: not sure we want to import `numpy` here ... it's not a hard dependency of `plotly`

packages/python/plotly/plotly/tests/test_core/test_errors/test_dict_path_errors.py
```diff
@@ -0,0 +1,154 @@
...
+            )
+        )
+
+    # finds bad part when using the path as a keyword argument to a subclass of
+    # BaseFigure and throws the error for the last good property found in
+    # the path
+    try:
+        fig2 = go.Figure(layout_title_txt="two")
+    except ValueError as e:
+        assert (
```
nicolaskruchten, line 117: for these assertions, it would be a lot more readable and future-proof to assert that the error message from `go.Figure(layout_title_txt="two")` is simply the same as the error message from `go.Figure(layout_title = dict(txt="two"))`, rather than doing assertions on the string itself, that way if and when we change the string we have fewer places to check :)

nicholas-esterer:
The one failure of `plotlyjs_dev_build` is solved in my hline vline PR, so once that's merged I can merge that into this and that test should pass.

nicholas-esterer:
Other than the one failure, which should be resolved once plotly.js 1.57 is out, this PR is ready for review :+1:

nicolaskruchten:
Please merge master into this PR or rebase this one onto master :)

nicolaskruchten:
There are still 800 modified files here, so probably more merge/rebase/squash work is needed :)

packages/python/plotly/_plotly_utils/utils.py
jonmmease: Agreed, numpy shouldn't be a hard dependency here. Here's an alternative approach you can use for cumsum if you need it (https://realpython.com/python-reduce-function/#comparing-reduce-and-accumulate).

packages/python/plotly/plotly/tests/test_core/test_errors/test_dict_path_errors.py
jonmmease: This is another case for `with pytest.raisese`: https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions

nicholas-esterer: Hey Jon! I should close this PR, it's been superseded by this one: https://github.com/plotly/plotly.py/pull/2843

jonmmease: Thanks!

PR #2836, opened 2020-10-15
`select_traces` `selector` argument can now be a function

### LINKED ISSUE #2629: select traces doesn't do partial matches on `meta`
Today, the output of 

```python

fig = go.Figure(go.Scatter(meta=dict(a=1,b=2)))

print(len([f for f in fig.select_traces(selector=dict(meta=dict(a=1,b=2)))]))
print(len([f for f in fig.select_traces(selector=dict(meta=dict(a=1)))]))
```

is `1, 0`, meaning that we can't use `selector` to do partial matches on `meta`.

It would be really convenient if PX could tag all of its traces with some structured data in `meta` so that users could target these traces with updates, but for that to be useful we would need to do partial matches.

@jonmmease any major reason not to do this?

nicholas-esterer:
closes #2629

### DIFF
--- a/packages/python/plotly/plotly/basedatatypes.py
+++ b/packages/python/plotly/plotly/basedatatypes.py
@@ -804,10 +804,19 @@ class BaseFigure(object):
                     continue
 
             # Filter by selector
-            if not self._selector_matches(trace, selector):
-                continue
-
-            yield trace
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
Superseded by #2844 ?

PR #2838, opened 2020-10-16
Make plotly.py's validator compatible with plotly.js 1.57.0 which now allows domain referenced axis references

nicholas-esterer:
This just involves adding ' domain' to `EnumeratedValidator.build_regex_replacement`

### DIFF
--- a/packages/python/plotly/_plotly_utils/basevalidators.py
+++ b/packages/python/plotly/_plotly_utils/basevalidators.py
@@ -354,14 +354,14 @@ class BaseValidator(object):
 
 class DataArrayValidator(BaseValidator):
     """
-        "data_array": {
-            "description": "An {array} of data. The value MUST be an
-                            {array}, or we ignore it.",
-            "requiredOpts": [],
-            "otherOpts": [
-                "dflt"
-            ]
-        },
+    "data_array": {
+        "description": "An {array} of data. The value MUST be an
+                        {array}, or we ignore it.",
+        "requiredOpts": [],
+        "otherOpts": [
+            "dflt"
+        ]
+    },
     """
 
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
Looks great! are you able to avoid the whitespace changes here at all? Would be nice to include just the changes you made to the code.

:dancer:

nicholas-esterer:
`black` did that I think. There's always `git diff -b` ;)

nicolaskruchten:
Can you ensure you're using the same version of `black` as is in the repo?

nicolaskruchten:
ah I guess we don't pin it. I have `19.10b0`, how about you?

nicholas-esterer:
black, version 20.8b1
In general, why don't we pin stuff always? It seems to break stuff often when pip repos are upgraded.

nicholas-esterer:
All it "requires" (ROFL) is a little `pip freeze > requirements.txt`

nicolaskruchten:
well, it's helpful to get a warning when a newer version of pandas is incompatible with our stuff... but yeah let's pin black to `19.10b0` for now. Can you do that in this PR and rerun it on your changes?

nicholas-esterer:
Ah, I see it's nice that it shows that automatically. What if releases are pinned but master isn't? Then the warnings happen in master but the releases always work.
Anyhow I can try pinning black :+1:

nicholas-esterer:
Unfortunately running `black` version `19.10b0` doesn't change the file `packages/python/plotly/_plotly_utils/basevalidators.py`, but I pinned it in `optional-requirements.txt`.

nicolaskruchten:
OK, can you please undo the whitespace changes and merge then?

PR #2843, opened 2020-10-19
Deep magic underscore error messages rebase

nicholas-esterer:
See https://github.com/plotly/plotly.py/issues/2072

Rebase of 

https://github.com/plotly/plotly.py/pull/2824


packages/python/plotly/plotly/graph_objs/_deprecations.py
```diff
@@ -7,25 +7,25 @@
...
 class Data(list):
     """
-    plotly.graph_objs.Data is deprecated.
-Please replace it with a list or tuple of instances of the following types
-  - plotly.graph_objs.Scatter
-  - plotly.graph_objs.Bar
-  - plotly.graph_objs.Area
-  - plotly.graph_objs.Histogram
-  - etc.
+        plotly.graph_objs.Data is deprecated.
```
nicolaskruchten, line 10: please revert these whitespace-only changes ... I think your editor must be doing this?

packages/python/plotly/plotly/graph_objs/layout/_annotation.py
```diff
@@ -319,6 +319,14 @@ def ay(self, val):
...
     def ayref(self):
         """
+        Indicates in what terms the tail of the annotation (ax,ay)  is
+        specified. If `pixel`, `ay` is a relative offset in pixels
+        from `y`. If set to a y axis id (e.g. "y" or "y2"), `ay` is
+        specified in the same terms as that axis. This is useful  for
+        trendline annotations which should continue to indicate  the
+        correct trend when zoomed.
+
+=======
```
nicolaskruchten, line 329: some sort of merge gremlin here, also this whole file contains nothing but whitespace changes :)

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_layout_subplots.py
```diff
@@ -39,7 +39,7 @@ def test_initial_access_subplot2(self):
             self.layout.xaxis2
 
     def test_initial_access_subplot2(self):
-        with pytest.raises(KeyError):
```
nicolaskruchten, line 42: why this change? I think `KeyError` is the appropriate error here, no?

packages/python/plotly/plotly/tests/test_core/test_errors/test_dict_path_errors.py
```diff
@@ -0,0 +1,223 @@
...
+    fig.add_shape(type="rect", x0=10, x1=20, y0=30, y1=40)
+    fig.add_trace(go.Scatter(x=[1, 2, 3], y=[4, 5, 6]))
+    return fig
+
+
+def test_raises_on_bad_index(some_fig):
+    # Check indexing errors can be detected when path used as key to go.Figure
+    try:
+        x0 = some_fig["layout.shapes[2].x0"]
+    except IndexError as e:
```
nicolaskruchten, line 25: Ah, so every test here needs to be rewritten to use `with pytest.raises(...)` because otherwise these tests can pass if no error is raised! We need to guarantee that the error is raised here, and check its contents.

packages/python/plotly/plotly/tests/test_core/test_errors/test_dict_path_errors.py
```diff
@@ -0,0 +1,223 @@
...
+    assert len(e_correct_substr) > 0
+    try:
+        some_fig["data[0].line_colr"] = "blue"
+    except ValueError as e:
+        e_substr = error_substr(
+            e.args[0],
+            """
+Bad property path:
+data[0].line_colr
+             ^""",
```
nicolaskruchten, line 96: wow, nice job on this little arrow thing!

nicolaskruchten:
So the error message for this call has changed for the worse IMO:

```python
go.Figure(go.Scatter()).update_traces(text_yo="hey")
```

It used to say basically `text_yo` isn't valid and now it drills into `text` and says "I can't do anything with `yo`". Maybe we need specific handling for leaves here?

nicolaskruchten:
Also needs a changelog entry

packages/python/plotly/plotly/tests/test_core/test_errors/test_dict_path_errors.py
nicholas-esterer: I tried using pytest.raisesRegex because I wanted to check the error message, but had trouble making a regex matching multiple lines. I can show you the error when using the multiline flag for `re`, it kind of looked like a bug with `re`. Also I don't need a full-blown regex, just a string comparison would suffice.

packages/python/plotly/plotly/graph_objs/layout/_annotation.py
nicholas-esterer: Yes I think so, I have a line in my editor's config that removes trailing whitespace. Also it seems there's a botched merge in this file so I'm just checking out the file in master (it's generated anyways) :+1:

packages/python/plotly/plotly/graph_objs/_deprecations.py
nicholas-esterer: This is weird because I never opened this file, so maybe `black` did it. I'll just check out the master version.

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_layout_subplots.py
nicholas-esterer: The way the better error messages work for subclasses of `BasePlotlyType` is they recursively call `BasePlotlyType.__getitem__(last_prop,next_prop_name)` where `last_prop` is the object resulting from the prop just looked up in the list of properties (the split magic underscores list) and `next_prop_name` is the next property in the list to look up. If we raise KeyError at the end of this list, we don't get the nice print out of all the possible properties like you do with `BasePlotlyType._raise_on_invalid_property_error`, which raises `ValueError`. Now we could change that to `KeyError` but I wouldn't recommend it because: a) we will break a lot more tests (my change to make it throw `ValueError` resulted in only this test breaking :+1:) and b) we won't get a nice print out of the error message, because the Python interpreter calls `repr` on the messages of `KeyError` so that `{}['']` prints `'KeyError: ""` and not `KeyError: ` (see e.g., https://stackoverflow.com/questions/34051333/strange-error-message-printed-out-for-keyerror). That's probably why `BasePlotlyType._raise_on_invalid_property_error` raises `ValueError` in the first place :-)

nicolaskruchten:
OK so this seems to add the shape to all subplots, even if they don't have traces on them:

```python
import plotly.express as px

df = px.data.tips()
fig = px.scatter(df, x="total_bill", y="tip", color="sex", facet_row="time", facet_col="day")
fig.add_shape(x0=0, x1=1, y0=0, y1=1, type="line", row="all", col="all")
fig.show()
```

but the equivalent `.hline()` call does not... why?

nicholas-esterer:
> So the error message for this call has changed for the worse IMO:
> 
> ```python
> import plotly.graph_objects as go
> go.Figure(go.Scatter()).update_traces(text_yo="hey")
> ```
> 
> It used to say basically `text_yo` isn't valid and now it drills into `text` and says "I can't do anything with `yo`". Maybe we need specific handling for leaves here?

Do you mean this error message is not so great?:
`TypeError: 'NoneType' object is not subscriptable`
I agree but I guess I don't understand why it doesn't work like for the other cases. For example, this doesn't fail:
```python
go.Figure(go.Scatter()).update_traces(text="hey")
```
Could it be because `text` doesn't have a corresponding `go.layout.Text` object? I'll have to look.

nicolaskruchten:
This is an improvement, but the error message is still a little opaque. `fig.update_layout(width_x=1)` gives:

```
Property does not support subscripting:
width_x
~~~~~
```

I wonder if it shouldn't be similar to the error we get if we try `fig.update_layout(width="a")` which gives:

```
    Invalid value of type 'builtins.str' received for the 'width' property of layout
        Received value: 'a'

    The 'width' property is a number and may be specified as:
      - An int or float in the interval [10, inf]
```

Perhaps the `width_x` case could return:

```
    Invalid value received for the 'width' property of layout

    The 'width' property is a number and may be specified as:
      - An int or float in the interval [10, inf]

Property does not support subscripting:
width_x
~~~~~
```

nicolaskruchten:
@jonmmease mind taking a quick look here to see if everything makes sense?

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_layout_subplots.py
jonmmease: This would be a breaking change. As I mentioned above, we should be able to subclass `KeyError` to customize the displayed string and this would not be breaking.

jonmmease (changes requested):
Looks really nice! I left a couple of comments on the exception class hierarchy. I still want to check out the branch and play around with it a little. I'll do that today.

jonmmease:
Right now it looks like trailing underscores are ignored, e.g. this is valid:

```python
fig.update_layout(width_=300)
```

I'm torn on whether this should raise an error (looks like it did before).

jonmmease:
I agree with @nicolaskruchten that the `width_x` error message is a bit awkward. But I don't have a better suggestion at the moment:

This one is also a little off IMO:

```python
fig.update_layout(title_text_asdf=300)
```
```
TypeError: string indices must be integers
Bad property path:
title_text_asdf
```

nicholas-esterer:
Do you have the latest commit? For me I get this, which was desired:
```python
TypeError: 'NoneType' object is not subscriptable

Invalid value received for the 'text' property of layout.title

    The 'text' property is a string and must be specified as:
      - A string
      - A number that will be converted to a string

Property does not support subscripting:
title_text_asdf
      ~~~~
```

nicholas-esterer:
But I'm also open to that being problematic : )

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_layout_subplots.py
nicholas-esterer: Ok that works :+1:. Does this work for the subclass:
```python
class PlotlyKeyError(KeyError):
    def __str__(self):
        return Exception.__str__(self)
```

packages/python/plotly/plotly/basedatatypes.py
```diff
@@ -3932,6 +4066,9 @@ def __contains__(self, prop):
         -------
         bool
         """
+        # TODO: We don't want to throw an error in __contains__ because any code that
+        # relies on it returning False will have to be changed (it will have to
+        # have a try except block...).
```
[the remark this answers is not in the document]
nicholas-esterer: This TODO is obsolete, good catch :+1:

nicholas-esterer:
Thanks @jonmmease and @nicolaskruchten  for the reviews!

How important is it to maintain `KeyError` over `ValueError`?
Before this PR, this is the error that was thrown when calling `__getitem__` on a subclass of `BasePlotlyType` with a property of length 1 (e.g., `go.Scatter()['doesntexist']`). The reason I changed this to call `self._raise_on_invalid_property_error(prop)` is because this latter function gives a more verbose error message which lists the valid properties. That way, after walking down the path described by the magic underscores, the last error encountered would be this error and it would list the valid properties.

Now we could change  `self._raise_on_invalid_property_error(prop)` to raise `KeyError`, because that's actually a more relevant `Exception`, but then I reckon we're changing the API in more places (and breaking more tests).
The way to keep everything the same is to add an argument that causes it to raise the desired exception, say `self._raise_on_invalid_property_error(prop,error_type=KeyError)`. The difficulty of that is knowing when to pass that argument in, because currently this function can't know if it was called with a property with or without underscores because in the underscore case, the property is split into single properties.

You may ask: "But how did the more verbose error get thrown before this PR?"
It got thrown only on property assignment and not on property lookup. The original desire of this PR was to have
```
go.Scatter(x=[1,2],y=[3,4],line_colr='blue')
```
show the possible values for `plotly.graph_objs.scatter.Line` and not scatter. The way that's done is properties are **looked up** in order to descend the property tree (here `'line'` is looked up in `go.Scatter` and `colr` is looked up in the resulting `plotly.graph_objs.scatter.Line` object: this last lookup fails). So I wanted to get that failure to give a better error message and so now lookup calls `self._raise_on_invalid_property_error(prop)` and doesn't just throw `KeyError`.
Before this PR,
```
go.Scatter(x=[1,2],y=[3,4],line_colr='blue')
```
tried to do this basically
```
go.Scatter(x=[1,2],y=[3,4])['line_colr']='blue'
```
and so that's why the error was reported relative to `go.Scatter` and not `plotly.graph_objs.scatter.Line`.

nicolaskruchten:
@nicholas-esterer I think I agree with Jon that changing the `KeyError` type is a breaking change. We do want the nicer error *message* but without changing the *type*.  Can we not catch the `ValueError` internally right before throwing in in the `[]` case and rewrap it in a `KeyError` with the nicer message?

nicholas-esterer:
By the `[]` case, you mean when a user does:`some_obj['badprop']`? No, because the routine that walks down the property tree also has to use `[]`.

nicholas-esterer:
And so then every bad access would be changed to `KeyError.`

nicholas-esterer:
But maybe we want that? `ValueError` is not the correct error to be throwing. But what I'm saying is that also breaks the API :).

nicholas-esterer:
Seeing what happens if we change `BasePlotlyType._raise_on_invalid_property_error` to raise `PlotlyKeyError` :crossed_fingers:
But this also breaks the API.

nicolaskruchten:
> By the [] case, you mean when a user does:some_obj['badprop']? No, because the routine that walks down the property tree also has to use [].

Yes, that's the case I meant. OK, then how about we leave that one case to return a KeyError (albeit with a nicer error message), and catch that KeyError and rewrap it as a ValueError in the non-[] cases? 

Either way, I think Jon and I are aligned that we should not change the error types in this PR. I agree that the current situation is not ideal, with KeyError sometimes and ValueError sometimes, but it's too late to change this basically :)

nicholas-esterer:
I agree too that changing the error types is a bad idea. I guess I got to start digging :hole: :slightly_smiling_face:

nicholas-esterer:
OK! So I think I've managed to preserve the previous exception behaviour as I have not changed any of the already existing tests (the only test "changed" is `packages/python/plotly/plotly/tests/test_core/test_errors/test_dict_path_errors.py` and it's the tests I wrote for this PR).

nicholas-esterer:
Now subscripting errors are thrown for any type throwing TypeError when subscripting them. This is enough for Plotly.py purposes because other than `dict`s and `BasePlotlyType`-dervived objects (which behave much like `dict`s), the objects found at keys are `None`, `str` or numbers, which throw `TypeError` when you try and use `dict`-like lookup (e.g., `1['badkey']`). These errors are cast to `ValueError` for assignment and `PlotlyKeyError` for lookup.

nicolaskruchten:
This looks like it's getting close to mergeable! Can we switch to using the full-width `^` highlighting for both cases here rather than single-`^` and full-width `~` please?

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - The `add_trace`, `add_shape`, `add_annotation`, `add_layout_image`, `add_hline`, `add_vline`, `add_hrect`, `add_vrect` functions accept an argument `exclude_empty_subplots` which if `True`, only adds the object to subplots already containing traces or layout objects. This is useful in conjunction with the `row="all"` and `col="all"` arguments. ([#2840](https://github.com/plotly/plotly.py/pull/2840))
 - For all `go.Figure` functions accepting a selector argument (e.g., `select_traces`), this argument can now also be a function which is passed each relevant graph object (in the case of `select_traces`, it is passed every trace in the figure). For graph objects where this function returns true, the graph object is included in the selection. ([#2844](https://github.com/plotly/plotly.py/pull/2844))
 
+### Added
```
nicolaskruchten, line 22: please move this to the 4.13 block

nicolaskruchten:
The only outstanding item here seems to be the "trailing underscore" question... Right now this fails on `master` and succeeds in this branch: `fig.update_layout(title_text____="test")`

nicolaskruchten:
Ahh actually it's a bit worse... this also works: `fig.update_layout(__title__text__=300)`

nicolaskruchten:
Cool, just need some tests on the leading/trailing/multiple underscores :)

nicolaskruchten:
@jonmmease this is ready for another review I think. Check out the "did you mean...?" feature!!

jonmmease (approved):
Looks great @nicholas-esterer. I'm happy with the code, so if @nicolaskruchten is happy with the behavior (I didn't check out and play with the latest version) then lgtm.

PR #2844, opened 2020-10-19
Figure select function selector

### LINKED ISSUE #2629: select traces doesn't do partial matches on `meta`

nicholas-esterer:
Closes #2629 and more.
Any function calling `BaseFigure._selector_matches` can now pass a function for the `selector` argument. This function is called on each object and if it returns True, then that object is included in the selection.


nicolaskruchten:
Superseded by https://github.com/plotly/plotly.py/pull/2844 I imagine? Can close?

nicolaskruchten:
I meant: supersedes https://github.com/plotly/plotly.py/pull/2836 therefore we can close that one!

packages/python/plotly/plotly/basedatatypes.py
```diff
@@ -813,24 +813,33 @@ def _perform_select_traces(self, filter_by_subplot, grid_subplot_refs, selector)
...
+                if obj_val != selector_val:
+                    return False
+            return True
+        # If selector is a function, call it with the obj as the argument
+        elif type(selector) == type(lambda x: True):
+            return selector(obj)
+        else:
+            raise TypeError(
+                "selector must be dict or a function "
+                "accepting a trace returning a boolean."
```
nicolaskruchten, line 841: this is now more than just "a trace" right? maybe "a graph object" ?

nicolaskruchten:
Nicely done!

💃 when you address my comment about `trace` above and you add a changelog entry :)

packages/python/plotly/plotly/basedatatypes.py
nicholas-esterer: ah yes, copy paste error! good catch :+1:

nicolaskruchten:
Looks great, please merge!

PR #2849, opened 2020-10-20
Fix missing commit info in auto-generated archives

Wuestengecko:
When the repository was restructured to what it is now, the
`.gitattributes` file was not updated to reflect the new paths. As a
result, the current commit information is no longer inserted into
`ploty/_version.py` during automatic tarball exports (as done by
Github). With no information being provided implicitly by Git itself,
this means the version information is entirely unavailable to the build
system at `packages/python/ploty/setup.py`; this in turn results in a
reported version number of `0+unknown` for builds from these archives.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-plotly/_version.py export-subst
+packages/python/plotly/plotly/_version.py export-subst


nicolaskruchten:
Thanks!

PR #2855, opened 2020-10-22
Exclude totally empty subplots

### LINKED ISSUE #2854: `exclude_empty_subplots` argument of `_add_annotation_like` and `add_trace` should check other annotation-like objects
The goal is to not address totally empty subplots, meaning subplots without data or layout objects referring to them.

nicholas-esterer:
closes #2854 
We should be able to select subplots containing anything, and also added is the feature to select subplots containing any combination of traces, shapes, annotations or layout_images.

### DIFF
--- a/packages/python/plotly/plotly/basedatatypes.py
+++ b/packages/python/plotly/plotly/basedatatypes.py
@@ -1302,7 +1302,7 @@ because subplot does not have a secondary y-axis"""
             # if exclude_empty_subplots is True, check to see if subplot is
             # empty and return if it is
             if exclude_empty_subplots and (
-                not self._subplot_contains_trace(xref, yref)
+                not self._subplot_not_empty(xref, yref, selector=exclude_empty_subplots)
             ):
                 return self
             # in case the user specified they wanted an axis to refer to the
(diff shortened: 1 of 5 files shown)


packages/python/plotly/plotly/basedatatypes.py
```diff
@@ -1302,7 +1302,7 @@ def _add_annotation_like(
             # if exclude_empty_subplots is True, check to see if subplot is
             # empty and return if it is
             if exclude_empty_subplots and (
-                not self._subplot_contains_trace(xref, yref)
+                not self._subplot_not_empty(xref, yref, selector=exclude_empty_subplots)
```
nicolaskruchten, line 1305: does this line mean that one could do `fig.add_shape( ... , row="all", col="all", exclude_empty_subplots=["traces", "shapes"])` ?

nicolaskruchten: to be clear: i don't think that's necessary, I would just do `True`/`False` here :)

nicholas-esterer: You could but by default True means "all". I also don't think it's necessary but you get this functionality for free, all that needs to happen is the list of objects to check is made into the keyword argument of the function (and coerced to "all" if it is True). But I can disable that if you want and have it just accept True/False.

nicolaskruchten: Yeah, let's just accept truthy/falsey for now. I think it's easier to not support certain exotic use-cases that then some people depend upon later :)

nicolaskruchten:
Also: could you please add a `CHANGELOG` entry in this PR that covers the previous one's features?

nicholas-esterer:
OK! should be ready for review.

PR #2859, opened 2020-10-28
[WIP] Testing iframe in ML docs

xhlulu:
PLEASE DO NOT MERGE


nicolaskruchten:
No, the solution we'll use will live in the graphing-library-docs repo and will be a little more involved. I have a rough design but we'll need to iterate on it a couple of times

xhlulu:
Sounds good!

Issue #2863, opened 2020-10-30
Make `_array_to_b64str` part of the public API (and find a better name)


emmanuelle:
The function `_array_to_b64str` defined in the `_imshow.py` module (https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/express/_imshow.py#L27) is quite useful in Dash apps with image processing. We could make it part of the public API (where ? probably in `plotly.utils` ?). What do you think @nicolaskruchten ? This was brought up recently in Almar's work (https://github.com/plotly/dash-3d-viewer/pull/2).

gvwilson:
Hi - we are trying to tidy up the stale issues and PRs in Plotly's public repositories so that we can focus on things that are still important to our community. Since this one has been sitting for several years, I'm going to close it; if it is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. Thanks for your help - @gvwilson

PR #2879, opened 2020-11-06
move array_to_imagestr function to be part of public API

emmanuelle:
Solves #2863.

Now you can do 
`from plotly.utils import array_to_imagestr`

However, `plotly.utils` is not a subpackage but just a module so it does not appear in the API doc. It'd be easy to change this to use a subpackage like what was done for `plotly.colors` when working on the API doc. One the other hand it's enough for me if the function appears in image-related docs of plotly.py and dash so I won't push hard for this change.


packages/python/plotly/_plotly_utils/data_utils.py
```diff
@@ -0,0 +1,75 @@
...
+
+try:
+    from PIL import Image
+
+    pil_imported = True
+except ImportError:
+    pil_imported = False
+
+
+def array_to_imagestr(img, backend="pil", compression=4, ext="png"):
```
nicolaskruchten, line 13: the name here is maybe something to think about? technically what we are doing here is representing an image array as a [data URI](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) so I would vote for something like `image_array_to_data_url` ?

packages/python/plotly/_plotly_utils/data_utils.py
```diff
@@ -0,0 +1,75 @@
...
+    """Converts a numpy array of uint8 into a base64 png string.
```
nicolaskruchten, line 14: docstring nit: this isn't always PNG now :)

nicolaskruchten:
💃

Issue #1842, opened 2019-10-22
Potential performance improvements?


pfbuxton:
I have profiled a simple heatmap here:

**profile_code.py**
```python
from werkzeug.contrib.profiler import ProfilerMiddleware
from app import server

server.config['PROFILE'] = True
server.wsgi_app = ProfilerMiddleware(server.wsgi_app, restrictions=[30])
server.run(debug = True)
```

**app.py**
```python
import numpy as np

import dash
import dash_core_components as dcc
import dash_html_components as html
import plotly.graph_objects as go

import flask

# Heatmap
Z = np.random.rand(1000,1000)

server = flask.Flask(__name__)
app = dash.Dash(__name__, server=server)

app.layout = html.Div(children=[
    dcc.Graph(
        id='example-graph',
        figure=dict(
			data=[go.Heatmap(
				z=Z
			)],
			layout=dict()
		)
    )
])
```

Result (Python 3.7 Windows):
```
         1068 function calls (1060 primitive calls) in 4.032 seconds

   Ordered by: cumulative time
   List reduced from 243 to 30 due to restriction <30>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    4.032    4.032 C:\Python37-32\lib\site-packages\werkzeug\contrib\profiler.py:95(runapp)
        1    0.000    0.000    4.032    4.032 C:\Python37-32\lib\site-packages\flask\app.py:2262(wsgi_app)
        1    0.000    0.000    4.032    4.032 C:\Python37-32\lib\site-packages\flask\app.py:1801(full_dispatch_request)
        1    0.000    0.000    2.473    2.473 C:\Python37-32\lib\site-packages\flask\app.py:1779(dispatch_request)
        1    0.002    0.002    2.472    2.472 C:\Python37-32\lib\site-packages\dash\dash.py:467(serve_layout)
      2/1    0.013    0.006    2.462    2.462 C:\Python37-32\lib\json\__init__.py:183(dumps)
        1    0.002    0.002    2.451    2.451 C:\Python37-32\lib\site-packages\_plotly_utils\utils.py:35(encode)
        2    0.000    0.000    1.995    0.998 C:\Python37-32\lib\json\encoder.py:182(encode)
        2    1.856    0.928    1.980    0.990 C:\Python37-32\lib\json\encoder.py:204(iterencode)
[...]

PR #2880, opened 2020-11-06
accelerate plotly JSON encoder for numpy arrays without nans

emmanuelle:
This is a proof of concept PR, I need to add a bunch of tests to be sure it behaves well for numpy arrays of different types, but the idea is to remove the encoding-decoding-reencoding process when JSON-serializing a numpy array without nans or infs (which is the reason why there is this 3-step process).

For large data arrays (for example a 1000x1000 array) there is a large performance gain. For example for
```python
from time import time
import numpy as np
import plotly.express as px
from plotly.utils import PlotlyJSONEncoder
import json
img = np.random.randint(255, size=(1000, 1000)).astype(np.uint8)
fig = go.Figure(go.Image(z=img))
t1 = time()
_ = json.dumps(fig, cls=PlotlyJSONEncoder)
t2 = time()
print(t2 - t1)
```
the proposed change results in a 2.5x performance gain on my machine (0.11 s instead of 0.27). The performance gain is interesting in Dash apps in particular.

@jonmmease what do you think of this?

### DIFF
--- a/packages/python/plotly/_plotly_utils/utils.py
+++ b/packages/python/plotly/_plotly_utils/utils.py
@@ -46,7 +46,7 @@ class PlotlyJSONEncoder(_json.JSONEncoder):
         # We catch false positive cases (e.g. strings such as titles, labels etc.)
         # but this is ok since the intention is to skip the decoding / reencoding
         # step when it's completely safe
-        if not ("Infinity" in encoded_o or "NaN" in encoded_o):
+        if not ("NaN" in encoded_o or "Infinity" in encoded_o):
             return encoded_o
         # now:
         #    1. `loads` to switch Infinity, -Infinity, NaN to None
--- a/packages/python/plotly/_plotly_utils/utils.py
+++ b/packages/python/plotly/_plotly_utils/utils.py
@@ -41,9 +41,9 @@ class PlotlyJSONEncoder(_json.JSONEncoder):
 
         """
         # this will raise errors in a normal-expected way
-        self.unsafe = False
+        self.hasinfnans = False
         encoded_o = super(PlotlyJSONEncoder, self).encode(o)
-        if self.unsafe:
+        if not self.hasinfnans:
             return encoded_o
         # now:
         #    1. `loads` to switch Infinity, -Infinity, NaN to None
@@ -97,13 +97,20 @@ class PlotlyJSONEncoder(_json.JSONEncoder):
 
         """
         numpy = get_module("numpy", should_load=False)
+        # Try to detect any nans of infs by aggressively converting to numpy
+        # (catching any errors resulting from this conversion)
+        # and checking with np.isfinite
         if numpy:
-            if isinstance(obj, numpy.ndarray) and numpy.issubdtype(
-                obj.dtype, numpy.number
-            ):
-                if numpy.all(numpy.isfinite(obj)):
-                    self.unsafe = True
-
+            try:
+                obj_as_numpy = numpy.asanyarray(obj)
+                try:
+                    is_finite =  numpy.all(numpy.isfinite(obj_as_numpy))
+                    if not is_finite:
+                        self.hasinfnans = True
+                except TypeError:
+                    pass
+            except:
+                pass
         # TODO: The ordering if these methods is *very* important. Is this OK?
         encoding_methods = (
             self.encode_as_plotly,
--- a/packages/python/plotly/_plotly_utils/utils.py
+++ b/packages/python/plotly/_plotly_utils/utils.py
@@ -98,7 +98,9 @@ class PlotlyJSONEncoder(_json.JSONEncoder):
         """
         numpy = get_module("numpy", should_load=False)
         if numpy:
-            if isinstance(obj, numpy.ndarray):
+            if isinstance(obj, numpy.ndarray) and numpy.issubdtype(
+                obj.dtype, numpy.number
+            ):
                 if numpy.all(numpy.isfinite(obj)):
                     self.unsafe = True
 
--- a/packages/python/plotly/_plotly_utils/utils.py
+++ b/packages/python/plotly/_plotly_utils/utils.py
@@ -47,7 +47,7 @@ class PlotlyJSONEncoder(_json.JSONEncoder):
         # We catch false positive cases (e.g. strings such as titles, labels etc.)
         # but this is ok since the intention is to skip the decoding / reencoding
         # step when it's completely safe
-        if not ('Infinity' in encoded_o or 'NaN' in encoded_o):
+        if not ("Infinity" in encoded_o or "NaN" in encoded_o):
             return encoded_o
         # now:
         #    1. `loads` to switch Infinity, -Infinity, NaN to None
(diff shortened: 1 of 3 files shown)


jonmmease:
Is the basic idea here that you're skipping the recursive encoding of the elements of the array?  If so, that makes a lot of sense.

emmanuelle:
I understand (but maybe this is not correct) that there is an additional cycle of decoding-reencoding [here](https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/_plotly_utils/utils.py#L51) in order to take care of nans and infs. If this is the case, we can check if there are no nans and infs and skip these steps (the "recursive encoding") I think.

But maybe I don't understand completely the purpose of the recursive encoding. The CI failure of one of the builds is definitely linked to this PR, I need to investigate.

jonmmease:
ohh, wow, I didn't realize (remember?) that we did that double encoding. That would definitely slow things down.  It would be great to get rid of that all together, but it would take some investigation to work through why it's there in the first place.

What I meant by "recursive" is that JSON serialization handles lists of objects (like fig["data"]) by recursively encoding each element of the list.  I'm not certain, but I think this recursive encoding happens for each element of numpy arrays as well, even though it wouldn't be strictly necessary.

nicolaskruchten:
It'd be nice to get rid of the double-encoding in general but in the meantime, if there are special cases where we feel safe to skip it e.g. arrays with every entry a finite value, then that would be a short-term win also for many cases :)

nicolaskruchten:
I'll note that by playing string games with `numpy.savetxt` I was able to get around a 10% speedup, but it was pretty gross :)

packages/python/plotly/_plotly_utils/utils.py
```diff
@@ -95,6 +96,14 @@ def default(self, obj):
         Therefore, we only anticipate either unknown iterables or values here.
 
         """
+        numpy = get_module("numpy", should_load=False)
+        if numpy:
+            if isinstance(obj, numpy.ndarray) and numpy.issubdtype(
+                obj.dtype, numpy.number
+            ):
+                if numpy.all(numpy.isfinite(obj)):
+                    self.unsafe = True
```
almarklein, line 105: Not sure if I am missing something. We can test all the numpy arrays that are in the dict being encoded, but we'd not catch nans and infs in any sublists, right?

emmanuelle: Yes this will work only for numpy arrays but for other kinds of objects the existing process (encoding - decoding - reencoding) will still be executed. Or at least this is the intended behaviour :-), maybe I am the one missing something! I need to write tests anyway.

emmanuelle:
@jonmmease I wonder if you could help me understanding the py3.7-orca CI failure, which I cannot reproduce locally in a conda environment with (I think) the same packages installed and py3.7. The problem occurs in the doctest of the `create_dendogram` figure factory, there is both an ImportError and a ValueError. Does it make sense to you?

packages/python/plotly/_plotly_utils/utils.py
almarklein: >  but for other kinds of objects the existing process (encoding - decoding - reencoding) will still be executed

I don't think this is the case, since as I understand this class, `encode()` is not called recursively, or is it?

almarklein:
FWIW - just in case anyone was looking to go down this route: Iast Friday I did a brief stab at making encoding `nan` and `inf` as `null` in the first place (instead of replacing such values after the fact by the extra decode-encode pass).

I figured we could overload the [JSONEncoder.iterencode method](https://github.com/python/cpython/blob/96aeaec64738b730c719562125070a52ed570210/Lib/json/encoder.py#L204-L243). It defines a function `floatstr()`, which seems exactly the thing we'd want to change. Except ... it's only used when the Python-based low-level encoder is used. It looks like the c-version brings it's own version of `floatstr()` :(

pfbuxton:
Hi @emmanuelle,

A while ago I profiled plotly/dash and found the numpy to JSON encoding to be one of the bottlenecks and also found that by not doing the encoding twice the code can be sped up by approximately a factor 2:
https://github.com/plotly/plotly.py/issues/1842#issuecomment-549401190

The way I implemented it I was able to handle both +/- inf's and nan's without the need to re-encode. But some care is needed as numpy has quite a lot of different variable types.

Hope this helps.

emmanuelle:
@almarklein you were right of course I had not understood in which order the different methods were called.

My workaround with `np.isfinite` did not work since some arguments might be something like `x=[1, float("nan"), "platypus"]` (this is actually a variable used in the tests!) and it's not possible to call `np.isfinite` on this kind of beast.

Therefore I resorted to a brute-force approach of testing whether the json string encoded the first time has the substrings "NaN" or "Infinity" and if not, skip the additional decoding and re-encoding step. For large data arrays the cost of checking these two patterns if small compared to the encoding time (~2% of the time on a couple of examples I tested). This is a slight cost burden to be added to figures with NaN or Infinity either in their data or in text strings (for example titles, labels etc.) but I think we can live with this, given the large speed-up (x2.5) for non-problematic cases.

packages/python/plotly/_plotly_utils/utils.py
```diff
@@ -40,10 +40,15 @@ def encode(self, o):
         Note that setting invalid separators will cause a failure at this step.
 
         """
-
         # this will raise errors in a normal-expected way
+        self.hasinfnans = False
```
jonmmease, line 44: Are you using `hasinfnans` anywhere else now?

packages/python/plotly/_plotly_utils/utils.py
```diff
@@ -40,10 +40,15 @@ def encode(self, o):
...
         encoded_o = super(PlotlyJSONEncoder, self).encode(o)
-
+        # Brute force guessing whether NaN or Infinity values are in the string
+        # We catch false positive cases (e.g. strings such as titles, labels etc.)
+        # but this is ok since the intention is to skip the decoding / reencoding
+        # step when it's completely safe
+        if not ("Infinity" in encoded_o or "NaN" in encoded_o):
+            return encoded_o
```
jonmmease, line 51: I'm fine with this for now given your description of the performance characteristics.

jonmmease (approved):
One question, but otherwise lgtm

packages/python/plotly/_plotly_utils/utils.py
emmanuelle: no you were right :-). Thanks!

almarklein:
I like this solution! I find it actually quite elegant given the limitations of the json encoder that we have to work with, and I'm pretty sure that any solution that somehow checks the input beforehand would be slower than this proposed approach.

The only downside is that the speed of encoding depends on the presence of NaN and Inf in the encoded object, which might be an unexpected effect. But these cases are hopefully rare, e.g. the [plotly example](https://plotly.com/python/line-charts/) about gaps in line charts uses `None`, which becomes `null` in json.

packages/python/plotly/_plotly_utils/utils.py
nicolaskruchten: maybe check for `NaN` first, as it feels like that'd appear more frequently?

nicolaskruchten:
:dancer:

nicolaskruchten:
@pfbuxton looks like you went down this path before, I'm sorry no one replied to you more than a year ago :( Thanks for bringing up your idea again in this context!

I think that your idea would layer well onto this one, i.e. now that we only re-encode when we need to, modifying `encode_as_list` in the way you suggest would safely exploit this fast path.

PR #2890, opened 2020-11-10
Add Dash Snippet to discrete-color

xhlulu:
A lower impact addition compared to https://github.com/plotly/plotly.py/pull/2884


nicolaskruchten:
Let's do it!

nicolaskruchten:
I didn't catch this in the review, but you need to remember to put the `hide_code=true` bit on the ```python line :)

Issue #2187, opened 2020-02-12
Annotated heatmaps: font colour blends with the colour of the square when using zmin and zmax


PlatonB:
Since there's no reaction to the new message in #1300, I'll write it as a separate issue.

Bug can be reproduced when using `zmin` and `zmax` arguments. Plotly 4.5.0.

```
from random import triangular
import plotly.figure_factory as ff

z = [[round(triangular(0.01, 0.1), 3) for j in range(10)] for i in range(10)]
ff.create_annotated_heatmap(z, colorscale='greens').show()
ff.create_annotated_heatmap(z, colorscale='greens', zmin=0, zmax=1).show()
```

Without `zmin` and `zmax`:
![without_zmin_zmax](https://user-images.githubusercontent.com/25541767/73575713-ce441c80-4489-11ea-8c08-a89bc45bf564.png)

With `zmin=0` and `zmax=1`:
![with_zmin_zmax](https://user-images.githubusercontent.com/25541767/73575733-d9974800-4489-11ea-8108-9223e0baa3f3.png)

emmanuelle:
Thanks @PlatonB this is indeed a bug.

bensdm:
I have the same issue with colorscale='Rdbu'
![image](https://user-images.githubusercontent.com/29406515/93321716-ca93b900-f812-11ea-9ac3-b83ca3481a77.png)

nicolaskruchten:
Agreed, this is a bug, and is probably fixable in this repo rather than Plotly.js, for once :)

bensdm:
Any update on this? should be quite easy to fix. Also was wondering why RdBu is in both sequential and diverging color scales https://plotly.com/python/builtin-colorscales/

nicolaskruchten:
No update, we'd be happy to accept a PR for it if it's easy :) 

The `RdBu` inclusion in sequential was a mistake, as noted in the docs you linked:

![image](https://user-images.githubusercontent.com/203523/98828473-1f644200-2406-11eb-858f-06ebce17502e.png)

bensdm:
I submit a PR, (this is my first one), not sure why one requirement failed https://github.com/plotly/plotly.py/pull/2892

PR #2892, opened 2020-11-11
fix annotations font color

bensdm:
Correct the font annotation color when using zmin and zmax or zmid https://github.com/plotly/plotly.py/issues/2187
image
image
image


nicolaskruchten:
Thanks for this pull request! I'll try to review and merge it soon. The build failure is odd and I've seen it before but it doesn't make a ton of sense. I'll look into it :)

packages/python/plotly/plotly/figure_factory/_annotated_heatmap.py
```diff
@@ -264,34 +262,48 @@ def get_text_color(self):
...
         """
         min_text_color, max_text_color = _AnnotatedHeatmap.get_text_color(self)
-        z_mid = _AnnotatedHeatmap.get_z_mid(self)
+        if zmin is None or zmax is None:
+            zmin, zmax = _AnnotatedHeatmap.get_z_min_max(self)
+        if zmid is None:
+            zmid = (zmax + zmin) / 2
+        if min_text_color == max_text_color:
+            # diverging colorscale
+            mid_text_color = "#000000"
```
nicolaskruchten, line 293: I'm not sure this is a great default... this is assuming the diverging scale is light in the middle?

packages/python/plotly/plotly/figure_factory/_annotated_heatmap.py
```diff
@@ -264,34 +262,48 @@ def get_text_color(self):
...
         """
         Get annotations for each cell of the heatmap with graph_objs.Annotation
 
         :rtype (list[dict]) annotations: list of annotations for each cell of
             the heatmap
         """
         min_text_color, max_text_color = _AnnotatedHeatmap.get_text_color(self)
-        z_mid = _AnnotatedHeatmap.get_z_mid(self)
+        if zmin is None or zmax is None:
+            zmin, zmax = _AnnotatedHeatmap.get_z_min_max(self)
```
nicolaskruchten, line 288: here it seems like we overwrite both if either is missing? probably best not to do that I think.

packages/python/plotly/plotly/figure_factory/_annotated_heatmap.py
nicolaskruchten: I wonder if we shouldn't accept a third color in `font_colors` for this and only set this one here if it's not present?

nicolaskruchten:
Thanks for this PR! It's prompted me to make this one https://github.com/plotly/plotly.py/pull/2921 which is a bit more targeted and doesn't change any method signatures etc. I'd love your thoughts on the other one, and if we merge that then you can try to build on it to address the color problem in diverging colorscales?

packages/python/plotly/plotly/figure_factory/_annotated_heatmap.py
bensdm: From what I understand from the doc https://plotly.github.io/plotly.py-docs/generated/plotly.graph_objects.Heatmap.html "zmax – Sets the upper bound of the color domain. Value should have the same units as in z and if set, zmin must be set as well." we can not pass only one of these 2 thats why i overwrite both

packages/python/plotly/plotly/figure_factory/_annotated_heatmap.py
bensdm: Yes can be an option but is there any diverging colorscale where the it's not the case?

nicolaskruchten: Well, users can pass in their own colorscale, they're not limited to just the built-in ones.

packages/python/plotly/plotly/figure_factory/_annotated_heatmap.py
nicolaskruchten: This is true of the underlying heatmap trace, yes, but this figure-factory already does its own bounds-computation, so we may as well reuse it. I've already addressed this comment in #2921.

Issue #2653, opened 2020-07-17
.show() docstring incomplete


nicolaskruchten:
would be good for the docstring to at least hint that `config`, `height`, `width` exists :)

PR #2893, opened 2020-11-11
add width, height, and config to .show docstring

bandersen23:
For #2653 - added width, height, and config to docstring of `figure.show()`.

@nicolaskruchten - I think I edited the docstrings properly, but let me know if not! I saw in one file the note about the automatic docstring generation, but couldn't find the file referenced.


nicolaskruchten:
Thanks!

Issue #2061, opened 2020-01-13
go.Figure() does not render in JupyterLab, only go.FigureWidget()


prncevince:
`go.Figure()` does not render in JupyterLab, only `go.FigureWidget()`.

Chrome Console Error: `Uncaught ReferenceError: require is not defined`

Example:
```python
import plotly
import plotly.io as pio
import plotly.graph_objects as go

figwidget = go.FigureWidget(data=go.Bar(y=[1, 2, 3]))
fig = go.Figure(data=go.Bar(y=[3, 2, 1]))

display(pio.renderers)
display(figwidget)
display(fig)
```

<img width="1280" alt="image" src="https://user-images.githubusercontent.com/17718313/72232439-2d032e00-358f-11ea-8abb-4c32804256f9.png">

Jupyter & Plotly specifics: 

```bash
%load_ext watermark
%watermark -d -iv -v -m -u -n -t -z
!jupyter labextension list plotly
```

```
plotly 4.4.1
last updated: 2020-01-12 22:40:00 EST 

CPython 3.7.6
IPython 7.11.1

compiler   : Clang 4.0.1 (tags/RELEASE_401/final)
system     : Darwin
release    : 19.2.0
machine    : x86_64
processor  : i386
CPU cores  : 4
interpreter: 64bit
JupyterLab v1.2.4
Known labextensions:
   app dir: ~/anaconda3/share/jupyter/lab
        @jupyter-widgets/jupyterlab-manager v1.1.0  enabled  OK
        jupyterlab-plotly v1.4.0  enabled  OK
        plotlywidget v1.4.0  enabled  OK
        qgrid v1.1.1  enabled  OK
```

emmanuelle:
Does the figure display correctly when you call `fig.show()`?

prncevince:
@emmanuelle negative
![image](https://user-images.githubusercontent.com/17718313/72299653-730fce80-362f-11ea-9f16-aa184cae2c87.png)

emmanuelle:
ok, thanks. Did you take a look at the Jupyterlab section in the troubleshooting guide https://plot.ly/python/troubleshooting/? In particular the fact that the extension version needs to be the correct one.

prncevince:
My plotly & extension versions (displayed above) match exactly as shown in https://plot.ly/python/getting-started/ . 
plotly.py v4.4.1 and both jupyterlab-plotly & plotlywidget v1.4.0

prncevince:
[...]

Issue #2039, opened 2020-01-07
ModuleNotFoundError: No module named 'plotly'


karthikGandiban:
I am a beginner here , as part of my learning module which requires  Plotly , i followed the standard method of using pip install plotly and pip install cufflinks - after successful installation when i call them on my jupyter note book , i am getting this error .

ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-d863167bda64> in <module>()
      1 import pandas as pd
      2 import numpy as np
----> 3 from plotly import __version__
      4 print(__verison__)

ModuleNotFoundError: No module named 'plotly'

Tried few of the options mentioned like downgrading Plotly etc etc but of no use , request your support . sorry if the question is lame.

emmanuelle:
Hi @karthikGandiban are you using a virtual environment when installing plotly? One common issue is when the jupyter notebook is not launched in the same environment as the one where plotly is installed. 
Could you please paste here the output of
```
pip list | grep plotly
```
if you're working on Mac or linux, or just
```
pip list
```
if you're on Windows.

karthikGandiban:
C:\Users\gak2ban>pip list
Package            Version
------------------ ----------
attrs              19.3.0
backcall           0.1.0
beautifulsoup4     4.8.1
bleach             3.1.0
certifi            2019.11.28
chardet            3.0.4
chart-studio       1.0.0
colorama           0.4.3
colorlover         0.3.0
cufflinks          0.17.0
cycler             0.10.0
decorator          4.4.1
defusedxml         0.6.0
entrypoints        0.3
html5lib           1.0.1
idna               2.8
importlib-metadata 1.3.0
ipykernel          5.1.3
ipython            7.11.1
ipython-genutils   0.2.0
ipywidgets         7.5.1
jedi               0.15.2
Jinja2             2.10.3
json5              0.8.5
jsonschema         3.2.0
jupyter-client     5.3.4
jupyter-core       4.6.1
jupyterlab         1.2.0
jupyterlab-server  1.0.6
kiwisolver         1.1.0
lxml               4.4.2
MarkupSafe         1.1.1
matplotlib         3.1.2
mistune            0.8.4
more-itertools     8.0.2
nbconvert          5.6.1
nbformat           4.4.0
notebook           6.0.2
numpy              1.17.4
pandas             0.25.3
pandocfilters      1.4.2
parso              0.5.2
pickleshare        0.7.5
pip                19.3.1
plotly             4.4.1
[...]

PR #2906, opened 2020-11-16
add note on jupyter multikernel

surchs:
Very small doc change:
- explained that in multi-kernel jupyter setups, the default kernel needs plotly jupyterlab extension (as per existing docs), and the processing kernel needs plotly + nbformat
- noted two failure scenarios if this isn't done correctly (blank figure or error asking to install `nbformat`)

### DIFF
--- a/doc/python/getting-started.md
+++ b/doc/python/getting-started.md
@@ -174,7 +174,7 @@ fig = go.FigureWidget(data=go.Bar(y=[2, 3, 1]))
 fig
 ```
 
-Please check out our [Troubleshooting guide](/python/troubleshooting/) if you run into any problems with JupyterLab.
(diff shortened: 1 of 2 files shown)


surchs:
Going through some of the open issues, I think multi-kernel jupyter instances are the source of several hot-topic issues like #2039 and #2061 . Maybe it would be worth it to actually expand on this PR with a dedicated section on "Using plotly in jupyter with more than one python kernel" or something like this.

nicolaskruchten:
I've just pushed a set of changes to the troubleshooting docs which I think captures the info you wanted to add here... take a look at the conflict and LMK ?

surchs:
@nicolaskruchten yeah, I think that covers most of it. I would still add a worked example on which environment needs the extension (the "server" environment) and that the "processing environment/kernel" needs plotly + `nbformat`.

What do you think?

PR #2921, opened 2020-11-20
honor zmin/zmid/zmax in annotated_heatmap font colors

nicolaskruchten:
Alternative to https://github.com/plotly/plotly.py/pull/2892 to close https://github.com/plotly/plotly.py/issues/2187, cc @bensdm.

This is a more targeted fix for the zmin/zmax problem specifically, and doesn't address the difficulty of light colors in the middle range of a diverging colorscale.

### DIFF
--- a/packages/python/plotly/plotly/figure_factory/_annotated_heatmap.py
+++ b/packages/python/plotly/plotly/figure_factory/_annotated_heatmap.py
@@ -196,14 +196,14 @@ class _AnnotatedHeatmap(object):
             self.zmin = min([v for row in self.z for v in row])
             self.zmax = max([v for row in self.z for v in row])
 
-        if "zmin" in kwargs:
+        if kwargs.get("zmin", None) is not None:
             self.zmin = kwargs["zmin"]
-        if "zmax" in kwargs:
+        if kwargs.get("zmax", None) is not None:
             self.zmax = kwargs["zmax"]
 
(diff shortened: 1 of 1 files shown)


packages/python/plotly/plotly/figure_factory/_annotated_heatmap.py
```diff
@@ -189,6 +189,23 @@ def __init__(
         self.reversescale = reversescale
         self.font_colors = font_colors
 
+        if np and isinstance(self.z, np.ndarray):
+            self.zmin = np.amin(self.z)
+            self.zmax = np.amax(self.z)
```
bensdm, line 194: I think we should use np.nanmin and np.max

packages/python/plotly/plotly/figure_factory/_annotated_heatmap.py
```diff
@@ -287,11 +289,10 @@ def make_annotations(self):
             the heatmap
         """
         min_text_color, max_text_color = _AnnotatedHeatmap.get_text_color(self)
-        z_mid = _AnnotatedHeatmap.get_z_mid(self)
         annotations = []
         for n, row in enumerate(self.z):
             for m, val in enumerate(row):
-                font_color = min_text_color if val < z_mid else max_text_color
+                font_color = min_text_color if val < self.zmid else max_text_color
```
bensdm, line 295: what about diverging colorscale?

PR #2953, opened 2020-12-03
sphinx-gallery fix: don't save plotlyjs in every figure

emmanuelle:


emmanuelle:
I figured out that it was not a good idea to have every example page weighing 3+ Mo if we want more people to use plotly in docs :-).

packages/python/plotly/plotly/io/_base_renderers.py
```diff
@@ -861,7 +861,7 @@ def render(self, fig_dict):
         filename_html = filename_root + ".html"
         filename_png = filename_root + ".png"
         figure = return_figure_from_figure_or_data(fig_dict, True)
-        _ = write_html(fig_dict, file=filename_html)
+        _ = write_html(fig_dict, file=filename_html, include_plotlyjs=False)
```
nicolaskruchten, line 864: hmm so how does Plotly.js get included?

nicolaskruchten: the first time I mean

emmanuelle: Hum good question. I had checked that with `include_plotlyjs=False` the plotly cdn was included in the html files (which is weird, but it did include the cdn). I switched to `include_plotlyjs='cdn'` instead.

nicolaskruchten:
:dancer:

Issue #2944, opened 2020-11-29
ENH: Optional dependencies for accelerating JSON serialization 


jonmmease:
On top of https://github.com/plotly/plotly.py/pull/2943, I investigated a couple of interesting libraries we could potentially use as optional dependencies to further accelerate JSON serialization

## pybase64
I played with [`pybase64`](https://pybase64.readthedocs.io/en/v0.2.0/) a little, and it looks like an easy way to get a decent speedup over the built-in python `base64` module for performing the numpy base64 encoding step being introduced in https://github.com/plotly/plotly.py/pull/2943.  This wouldn't require any refactoring or anything, and can drop the base64 encoding time (which is a substantial portion of the total json encoding time for figures that contain large numpy arrays) by something liek 20% to 40%.

## orjson
[`orjson`](https://github.com/ijl/orjson) is a really impressive alternative JSON encoder that, in playing with a little bit, I've seen it be 2x to 5x times faster than the built-in Python `json` encoder.

`orjson` doesn't support custom JSON encoder classes (like `PlotlyJSONEncoder`), so supporting this as an optional dependency would require a total refactor of the current json encoding process.

Basically, we would need to switch to an architecture where we would preprocess the figure dictionary recursively to perform any conversions we need, and then feed that dictionary through the JSON encoder.

Another nice thing about `orjson` is that it automatically converts `nan` and `infinity` values to JSON `null` values, so the JSON re-endcoding stuff we were working through in https://github.com/plotly/plotly.py/pull/2880 wouldn't be needed (cc @emmanuelle ).

sdementen:
Some complement on the performance of orjson: https://python-rapidjson.readthedocs.io/en/latest/benchmarks.html#tables

I have also been digging into the JSON serialization performance in plotly, and noticed that, on a large plot `px.line(df)` with `df` a pandas Dataframe (17520 rows x 8 columns of random float, index is a DatetimeIndex) that takes 1.8s to generate (`px.line(df)` only),
- there is an **extra 0.4s** due to the `deepcopy` in line https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/basedatatypes.py#L3289 which is not necessary I guess as the JSON serialization will not change the data
- for serializing one column of the Dataframe (pandas Series), `json.dumps(sr, cls=plotly.utils.PlotlyJSONEncoder)` take **10x more time** than serializing via `sr.to_json(orient="values")`
[...]

PR #2911, opened 2020-11-17
Replaced 'retrying' dependency with 'tenacity' in plotly package

jmsmdy:
The plotly dependency 'retrying' is no longer being maintained. The currently-maintained fork, 'tenacity', is compatible with both Python 2 and 3, and is feature-compatible (although not API compatible). This pull request changes the few instances where the retrying package is used in plotly to use tenacity instead. 

One motivation for this is to get plotly working in Pyodide (a means developed by Firefox to run Python in the browser, which currently supports numpy, pandas, matplotlib, and a number of other datascience packages). 'Retrying' seems to be incompatible with Pyodide.

closes https://github.com/plotly/plotly.py/issues/2907

PR #2955, opened 2020-12-05
JSON encoding refactor and orjson encoding

jonmmease:
## Overview
Initial implementation of the idea from https://github.com/plotly/plotly.py/issues/2944 of refactoring the JSON encoding pipeline and optionally performing JSON encoding with orjson.

orjson is impressively fast, and it includes built-in support for numpy arrays which is many times faster than the current approach of converting them to lists before encoding.

Also, orjson automatically converts all non-finite values to JSON `null` values, so we don't need workarounds like re-encoding as discussed in https://github.com/plotly/plotly.py/pull/2880.

## JSON config object
To configure the JSON encoding engine, this PR adds a `plotly.io.json.config` object that mirrors `plotly.io.orca.config` and `plotly.io.kaleido.config`.  Currently the only option is `default_engine` which can be set to "json" for the current encoder based on `PlotlyJSONEncoder` or "orjson" which is pretty much always much faster.

The `to_json`/`write_json` also provide an engine argument to override the default.

To try it out, install orjson with pip
```
$ pip install orjson
```
or conda
```
$ conda install -c conda-forge orjson
```
Then configure plotly to use it with
```python
import plotly.io as pio
import numpy as np
```

## Quick timing example
Then time the encoding speed
```python
N = 1000000
dtype = "float32"
x = np.random.randn(N).astype(dtype)
y = np.random.randn(N).astype(dtype)
size = np.random.rand(N).astype(dtype) * 10
opacity = np.random.rand(N).astype(dtype)
fig = go.Figure(data=[go.Scatter(x=x, y=y, marker_size=size, marker_opacity=opacity)])
```

```python
%%timeit
res1 = pio.to_json(fig, engine="legacy")
```
```
2.06 s ± 18.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```python
%%timeit
res1 = pio.to_json(fig, engine="orjson")
```
```
169 ms ± 2.12 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

In this large figure case of encoding a figure with four one-million element arrays, the orjson encoding is 12x faster on my machine!

## Relationship to base64 encoding
This approach is fully compatible with the base64 encoding work in https://github.com/plotly/plotly.py/pull/2943. I think we should focus on this approach first because there are substantial performance gains to be had without changing the schema of the resulting JSON and (hopefully) without requireing changes in Plotly.js

After this is merged, we can add base64 encoding on top of it for additional performance improvements.

## Correctness testing
In addition to adding a new test suite, this PR has been tested against all of the documentation figures using the slightly modified instrumentation branch at https://github.com/plotly/plotly.py/pull/3012.  This branch executes all json encoding requests using both the `json` and `orjson` encoders, and checks that the encoded string are identical.

There are two documentation examples that fail this test: `imshow` and `ml-tsne-umap-projections`. Both of these fail due to difference in how the orjson encoding handles floating-point numbers with precision less than 64 bits (see next section)

### Numpy floating point preceision
The "json" encoder handles numpy arrays by first converting them to lists, and then encoding the lists.  All numpy floating point types are converted to 64-bit python float values.  For floating point numpy arrays with less than 64 bit precision, converting them to 64 bits before encoding artificially increases the precision of the values in the array, but this is really the only option.

The orjson encoder accepts numpy arrays directly, and it will output the appropriate amount of decimal places for precision of the input array. 

So the encoded JSON values between the legacy `json` encoder and the `orjson` encoder will not agree when encoding floating point numpy arrays with less than 64-bit precision.  This is the only known discrepancy.

## Benchmarking
To compare the performance of the `orjson` encoder against the legacy `json`, https://github.com/plotly/plotly.py/pull/3012 records the encoding time for both encoders and writes the results to a file.  Here are plots of the relative timing results across all of the figures in the plotly.py documentation

Note that ``length` here is the number of characters is the encoded JSON string

newplot (11)

newplot (12)

the `orjson` encoder is almost always faster (up to 40x in one case).  The handful of cases that have equivalent or slower performance are cases that include values that are not natively supported by orjson (e.g. `pd.Timestamp` or `PIL.Image` objects), and that don't have sizable numpy arrays.

All of the cases where it's slower run in less than half a *millisecond*.

My conclusion is that defaulting to `orjson` when the package is installed is a safe default that will almost always improve performance.

## TODO
 - [x] The current tests are pretty much passing with the new encoder, but before merging I want to add some more tests specifically around JSON encoding, especially with dates and datetimes. And make sure that all of the encoding engines are tested.


nicolaskruchten:
Cool :) 

Why not just swap this out right now and make it a hard dependency? Is there a specific risk or subdependency we don't want to pull in or something?

chriddyp:
Looking at https://github.com/ijl/orjson#why-cant-i-install-it-from-pypi, we should verify that we can install this on <=4.0.1 versions of DE with older `pip`

jonmmease:
> Why not just swap this out right now and make it a hard dependency? Is there a specific risk or subdependency we don't want to pull in or something?

I haven't looked into it deeply, but it has native components and isn't available in the main anaconda channel yet (it is on conda-forge though).  If we did make it a hard dependency for 5.0, then it probably would get added to the main anaconda channel as well 🙂 

So, adding it as a hard dependency carries the risk of breakage for some folks.

> we should verify that we can install this on <=4.0.1 versions of DE with older pip

Yeah, it'll have the same problem we've hit elsewhere with older versions of pip.

jonmmease:
Tests written and ready for review

jonmmease:
Renamed `pio.json.to_plotly_json` to `pio.json.to_json_plotly` because we, unfortunately, use `fig.to_plotly_json` as a method that converts a graph_object (or Dash component) to a dictionary.

For symmetry, also renamed `from_plotly_json` to `from_json_plotly`.

Feedback welcome on the names

nicolaskruchten:
I wonder if there's some way we could instrument `fig.show()` to capture and assert that the JSON outputs are identical, and then run all the doc and test examples, and then do something similar with dash-docs ?

jonmmease:
> I wonder if there's some way we could instrument fig.show() to capture and assert that the JSON outputs are identical, and then run all the doc and test examples, and then do something similar with dash-docs ?

Done for the plotly.py docs in https://github.com/plotly/plotly.py/pull/3012. And this did shake loose a couple more things around datetimes, so it was definitely a good thing to do!

Dash won't pick this up automatically since it uses `PlotlyJSONEncoder` directly. So I'm going to defer checking the Dash docs until after this release, as part of https://github.com/plotly/dash/pull/1514.

nicolaskruchten:
I'm not sure I understand why this change and what the implications are...?

> Renamed pio.json.to_plotly_json to pio.json.to_json_plotly because we, unfortunately, use fig.to_plotly_json as a method that converts a graph_object (or Dash component) to a dictionary.

Is this to avoid the legacy encoder from calling the new encoder while recursing through a structure that contains a figure?

jonmmease:
> Renamed pio.json.to_plotly_json to pio.json.to_json_plotly because we, unfortunately, use fig.to_plotly_json as a method that converts a graph_object (or Dash component) to a dictionary.

This naming change was all internal to this PR. Before this PR, there was only `to_json` which both did figure validation and performed JSON encoding.

I extracted the JSON encoding into `to_json_plotly`, which is now called by `to_json`.  `to_json_plotly` doesn't perform figure validation, and is what Dash will use for JSON encoding.

jonmmease:
I think the plotlyjs_dev_build  failure is due to the `role` removal on plotly.js master

jonmmease:
Up updated the overview comment with a description of the correctness testing, and the benchmarking results, obtained using the https://github.com/plotly/plotly.py/pull/3012 branch on the plotly.py documentation examples.

jmsmdy:
> Why not just swap this out right now and make it a hard dependency? Is there a specific risk or subdependency we don't want to pull in or something?

Just wanted to add my opinion to not to make orjson a hard dependency. orjson is written in Rust (apparently depending on some recent version to build), which is a barrier for running plotly.py on platforms that Rust has trouble compiling to.

- There were recently issues with orjson on the new apple silicon (see: https://github.com/ijl/orjson/issues/155). These have been resolved, but these issues would have held up people trying to use plotly on the new macs had orjson been a hard dependency.

- orjson does not compile to WebAssembly. After the "retrying" dependency is replaced by "tenacity" (when this pull request is merged: https://github.com/plotly/plotly.py/pull/2911), all (hard) dependencies of plotly.py will be available as universal wheels on PyPi, which enables plotly.py to run on Pyodide (a version of CPython compiled to run in WebAssembly). This would break if orjson were made a hard dependency. 

This is not an objection to this pull request. In fact, the preferred solution is to get orjson working in Pyodide (since there is already a need for fast serialization to communicate between JS and Python).

jonmmease:
Thanks for the feedback and perspective @jmsmdy.  That all makes a lot of sense.  This PR did end up making orjson optional, and that will be the case going forward.

nicolaskruchten:
Just to refresh my memory: with this PR in non-`orjson` mode, do we still do the trick from https://github.com/plotly/plotly.py/pull/2880 that got us a nice performance boost in some cases?

nicolaskruchten:
Also note to self to re-think-about the comment in https://github.com/plotly/plotly.py/pull/2880#issuecomment-726860782

mherrmann3:
Regarding the numpy floating point precision and that `PlotlyJSONEncoder` always casts those to float64 due to using `tolist()`...

This had always bugged me, as it resulted in much larger exports (i.e. html / ipynb file sizes) than necessary (when float16 or float32 is sufficient) and affected not only coordinate data, but also marker sizes, meta info, etc.

Just in case the plotly.py devs or others are interested: I had found a way to avoid this number inflation by modifying (& monkey patching) the [`encode_as_list` method](https://github.com/plotly/plotly.py/blob/340aed33bf116530a574badde64e867175c067ad/packages/python/plotly/_plotly_utils/utils.py#L147):

```python
@staticmethod
def encode_as_list_patch(obj):
    """Attempt to use `tolist` method to convert to normal Python list."""
    if hasattr(obj, "tolist"):

        numpy = get_module("numpy")
        try:
            if isinstance(obj, numpy.ndarray) \
               and obj.dtype == numpy.float32 or obj.dtype == numpy.float16 \
               and obj.flags.contiguous:
                return [float('%s' % x) for x in obj]
        except AttributeError:
            raise NotEncodable

        return obj.tolist()
    else:
        raise NotEncodable
```
It's about 30-50x slower than `.tolist()`, but - being in the order of a few μs - still much faster than the json encoding, with the benefit of ~3x smaller exports.

I always wanted to report this, and this PR revived the topic. Could this be relevant for a new issue (especially since orjson will not become the default)?

FYI: for reference, a quick search revealed that a patch of encode_as_list was already suggested before: https://github.com/plotly/plotly.py/issues/1842#issuecomment-549401190, in the context of treating inf & NaN, which got brought up again in https://github.com/plotly/plotly.py/pull/2880#issuecomment-726860782.

dhirschfeld:
I'm super keen to give this a go as I've got medium sized data and am having performance issues :(

It sounds like `orjson` will be used automatically if installed:

image

Assuming that gives a good speedup, is there a way to configure `plotly` to automatically use a different json parser - e.g. [`pysimdjson`](https://github.com/TkTech/pysimdjson)?

jonmmease:
> is there a way to configure plotly to automatically use a different json parser 

Not right now, the logic around the JSON library needs to be customized a bit (e.g. for handling datetime formatting). That said, the refactoring that went into the orjson support, and the switchable JSON engines, would make it a lot easier to add support for additional JSON libraries in the future.

PR #2958, opened 2020-12-07
Update kaleido instructions to install from conda-forge

jonmmease:
Now that Kaliedo is on conda-forge, new versions will only be published there. This PR updates the installation instructions in the README and docs to install from conda-forge.

cc @nicolaskruchten because I don't remember the doc deployment sequence :slightly_smiling_face:

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ $ pip install -U kaleido
 or conda.
 
 ```
-$ conda install -c plotly python-kaleido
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
This is great, thanks! I'm releasing 4.14 today so PRs to `master` like this will come out today but ordinarily you should target these to the `doc-prod` branch to get auto-deployed right away :)

PR #2967, opened 2020-12-10
Add Dash Snippet to t-sne page

xhlulu:


xhlulu:
Seems like the t-sne app is breaking the gallery. I'm removing it for now and i'll wait until pip is upgraded on gallery before re-adding it. Closing this for now.

PR #2996, opened 2020-12-29
Add missing whitespaces in ImportError message.

mkcor:
## Minor PR

I stumbled across the error myself and it read "ImportError: orca and psutil are required to use the `sphinx-gallery-orca` renderer. See https://plotly.com/python/static-image-export/ for instructions **onhow** to install orca. Alternatively, you can use the **`sphinx-gallery`renderer** (note that png thumbnails can only be generated **withthe** `sphinx-gallery-orca` renderer)."


nicolaskruchten:
Thanks!

PR #3016, opened 2021-01-11
support JupyterLab 3 in extensions

nicolaskruchten:
Closes https://github.com/plotly/plotly.py/issues/2994

### DIFF
--- a/packages/javascript/jupyterlab-plotly/package.json
+++ b/packages/javascript/jupyterlab-plotly/package.json
@@ -33,7 +33,7 @@
   "dependencies": {
     "plotly.js": "^1.58.2",
     "@types/plotly.js": "1.44.28",
-    "@jupyterlab/rendermime-interfaces": "^1.3.0 || ^2.0.0",
+    "@jupyterlab/rendermime-interfaces": "^1.3.0 || ^2.0.0 || ^3.0.0",
     "@lumino/messaging": "^1.2.3",
     "@lumino/widgets": "^1.8.1",
     "lodash": "^4.17.4"
--- a/packages/javascript/plotlywidget/package.json
+++ b/packages/javascript/plotlywidget/package.json
@@ -38,6 +38,12 @@
     "lodash": "^4.17.4"
   },
   "jupyterlab": {
-    "extension": "src/jupyterlab-plugin.js"
+    "extension": "src/jupyterlab-plugin.js",
+    "sharedPackages": {
+      "@jupyter-widgets/base": {
+        "bundled": false,
+        "singleton": true
+      }
+    }
   }
 }


jasongrout:
Don't forget to bump the `@jupyter-widgets/base` requirement at https://github.com/plotly/plotly.py/pull/3016/files#diff-f4a5512d7079ea169117ae30665e2b6b10fea530afebaa1e53fe0f5c6384cb9fR37

jasongrout:
(Technically it will still work since you've said it was a singleton and not bundled, but it will complain in the console nonetheless about the version not matching what you've required)

jasongrout:
@nicolaskruchten ^

nicolaskruchten:
Ah, didn't see this comment until now... Thanks! This will cause Javascript console warnings only basically?

jasongrout:
> This will cause Javascript console warnings only basically?

For now. In the future we plan to turn on strict versioning to be more accurate with our versions, which would mean this would fail then.

nicolaskruchten:
OK, thanks! I'll get this fixed in the next version, which I expect will come soon, as this upgrade seemed "too easy" :P

Alexboiboi:
Hi @nicolaskruchten,

Thanks for the updated, works nicely ;)

In the readme file I think the ipywidgets version **should be >=7.6**, otherwise it doesn't work in jlab v3

jasongrout:
Actually, @nicolaskruchten - you should update the version number of `@jupyter-widgets/base` now. You'll have problems now if they install `@jupyter-widgets/jupyterlab-manager` as a source extension if you don't update. Things will work (with complaining) only if they install the widget manager as a prebuilt extension.

jasongrout:
> In the readme file I think the ipywidgets version **should be >=7.6**, otherwise it doesn't work in jlab v3

ipywidgets can technically be < 7.6 - all 7.6 basically does is install widgets for jlab 3 by default.

You can have ipywidgets 7.5 and install `jupyterlab_widgets` 1.0 to get widgets in jlab 3, for example.

nicolaskruchten:
@jasongrout I was able to get it to work with all extensions installed as source extensions...

jasongrout:
> @jasongrout I was able to get it to work with all extensions installed as source extensions...

Without updating your `@jupyter-widgets/base` requirement to include `^4`?

nicolaskruchten:
yeah, I just did `jupyter labextension install @jupyter-widgets/jupyterlab-manager` then `jupyter labextension install plotlywidget@4.14.2` and it worked out. Did that somehow install a prebuilt version of the manager?

nicolaskruchten:
(my only reticence is that it's late and it's about a 90-minute process to do a whole release BTW! I'll surely do it in the next few days)

jasongrout:
Yep, I just checked, and it luckily works out when we merge dependency configs to give the base 4.0 version. It may not be so lucky if we have two widgets installed as source extensions and the manager as prebuilt?

In any case, there certainly is a technical issue (the build is pulling in two copies of widgets base, version 3 and version 4), though it seems that for some reason things seem to collapse to the right version in my test setup too.

If you're releasing again in the next few days, I won't trace it further for now. It is a bit puzzling to me that it works, though.

nicolaskruchten:
Hehe ok, thanks for digging! Sorry to distract you with a juicy puzzle ;)

nicolaskruchten:
Thanks both! I've just released 4.14.3 with the `package.json` changes requested and I've lightly edited the readme/getting-started docs so as to reduce the likelihood that someone will have a bad/incompatible version of `ipywidgets`.

Issue #2518, opened 2020-05-30
DateTime object not rendered correctly in plot tooltip


samjett247:
I noticed that DateTime objects in Python do not render correctly in the plot tooltips when they are included as `hover_data` arguments in plotly express functions. See example below. 
![image](https://user-images.githubusercontent.com/30481194/83315754-fd1f9480-a1d6-11ea-8d62-6455ce07e002.png)

Desired behavior would be the way that IPython renders the DateTime objects using the `display` function. See below: 
![image](https://user-images.githubusercontent.com/30481194/83315800-3a842200-a1d7-11ea-9492-4cb5e4d57402.png)

The exact datatype of the Timestamp Series is `dtype('<M8[ns]')`, which is a numpy specific dtype.

dimucciojonathan:
Where you input your date to the tooltip, I'm assuming you're doing something like:

`'Timestamp=' + df['Timestamp']`

Try converting the series to a normal array with '.values' like:

`'Timestamp=' + df['Timestamp'].values`

samjett247:
I was plotting it via plotly express. However, I tried passing in the Timestamp.values  (instead of `hover_data =['Timestamp']` I used `hover_data=[df['Timestamp'].values]`) but got the same result

ennosigaeus:
Plotly allows you passing in a formatting string when using `hover_data` (see: https://plotly.com/python/hover-text-and-formatting/):

```
data = px.scatter(df, x='utc_time', y='name', color='name', hover_data={'utc_time': '|%Y-%m-%d %H:%M:%S.%LZ', ...})
```

However, I noticed this does not work if `df['utc_time']` is an actual DateTime object, which leads to a tooltip with the following output:

![image](https://user-images.githubusercontent.com/649330/85324231-bec18080-b4c9-11ea-8250-2f9dd242464d.png)

I needed to "cast" the DateTime object to string, and only then it worked:
```df['utc_time'] = df['utc_time'].dt.strftime('%Y-%m-%d %H:%M:%S.%LZ')
data = px.scatter(df, x='utc_time', y='name', color='name', hover_data={'utc_time': '|%Y-%m-%d %H:%M:%S.%LZ', ...})
```

![image](https://user-images.githubusercontent.com/649330/85324395-0f38de00-b4ca-11ea-8462-320858ea479d.png)

It took me a while to figure this out as, intuitively, I would have assumed this transformation to be done implicitly. Maybe this could be added as a feature?

[...]

PR #3022, opened 2021-01-13
Pandas and Numpy datetime serialization fixes

jonmmease:
Should close https://github.com/plotly/plotly.py/issues/2518
 - Add support to `PlotlyJSONEncoder` for `numpy` `datetime64` arrays (1 or more dimensions) (050dd8a)
 - Add support to `DataArrayValidator` for homogeneous pandas DataFrames of datetime values (46d6a59)

cc @nicolaskruchten

### DIFF
--- a/packages/python/plotly/_plotly_utils/basevalidators.py
+++ b/packages/python/plotly/_plotly_utils/basevalidators.py
@@ -104,7 +104,7 @@ def copy_to_readonly_numpy_array(v, kind=None, force_numeric=False):
                 # DatetimeIndex
                 v = v.to_pydatetime()
     elif pd and isinstance(v, pd.DataFrame) and len(set(v.dtypes)) == 1:
-        dtype = v.dtypes[0]
+        dtype = v.dtypes.tolist()[0]
         if dtype.kind in numeric_kinds:
             v = v.values
         elif dtype.kind == "M":
--- a/packages/python/plotly/_plotly_utils/basevalidators.py
+++ b/packages/python/plotly/_plotly_utils/basevalidators.py
@@ -103,6 +103,13 @@ def copy_to_readonly_numpy_array(v, kind=None, force_numeric=False):
             else:
                 # DatetimeIndex
                 v = v.to_pydatetime()
+    elif pd and isinstance(v, pd.DataFrame) and len(set(v.dtypes)) == 1:
+        dtype = v.dtypes[0]
+        if dtype.kind in numeric_kinds:
+            v = v.values
+        elif dtype.kind == "M":
+            v = [row.dt.to_pydatetime().tolist() for i, row in v.iterrows()]
+
     if not isinstance(v, np.ndarray):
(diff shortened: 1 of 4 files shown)


packages/python/plotly/_plotly_utils/utils.py
```diff
@@ -184,8 +184,13 @@ def encode_as_numpy(obj):
 
         if obj is numpy.ma.core.masked:
             return float("nan")
-        else:
-            raise NotEncodable
+        elif isinstance(obj, numpy.ndarray):
+            try:
+                return numpy.datetime_as_string(obj).tolist()
```
nicolaskruchten, line 189: this doesn't try to coerce large ints into dates does it?

packages/python/plotly/_plotly_utils/basevalidators.py
```diff
@@ -103,6 +103,13 @@ def copy_to_readonly_numpy_array(v, kind=None, force_numeric=False):
             else:
                 # DatetimeIndex
                 v = v.to_pydatetime()
+    elif pd and isinstance(v, pd.DataFrame) and len(set(v.dtypes)) == 1:
```
nicolaskruchten, line 106: this wouldn't handle a `customdata` case where all the custom data isn't the same type, right?

packages/python/plotly/_plotly_utils/utils.py
jonmmease: The traceback's claim that it rejects non datetime values
```
np.datetime_as_string([1, 2, 3])

  File "/media/jmmease/SSD2/anaconda/envs/plotly_dev36/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3331, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-6-52e13b098ed9>", line 1, in <module>
    np.datetime_as_string([1, 2, 3])
  File "<__array_function__ internals>", line 6, in datetime_as_string
TypeError: input must have type NumPy datetime
```

but we could be more conservative and check the dtype kind first

packages/python/plotly/_plotly_utils/basevalidators.py
jonmmease: Right, but that was already working (in my testing)

jonmmease:
@nicolaskruchten tests green when you have another chance to take a look

packages/python/plotly/_plotly_utils/utils.py
nicolaskruchten: I think that would be good, yes.

jonmmease: Done in https://github.com/plotly/plotly.py/pull/3022/commits/7d5023bbfa2fd5b0f08933828aad06c42f50bb10

nicolaskruchten:
LGTM! Nice to nail down yet another odd edge case, and a nice fix in the context of the encoder work :) 💃

PR #3061, opened 2021-02-04
Update _dendrogram.py

tschm:
Little bug


nicolaskruchten:
Thanks! I'll take a look and include this as part of our next release, in a few weeks.

PR #3073, opened 2021-02-16
fixing documents

c-chaitanya:


nicolaskruchten:
Thanks very much for this big sweep! I'll merge this and it should hit our site in 10-20 minutes :)

nicolaskruchten:
@c-chaitanya how hard would it be to run this process on the function docstrings? It would be nice to batch-fix issues like https://github.com/plotly/plotly.py/issues/3074 if possible rather than doing them one at a time ;)

jmonticolo:
not only docstrings, but functions arguments too like [this one](https://github.com/plotly/plotly.py/blob/03979d105c65dda3df3a155322eaff18f203b03f/packages/python/plotly/_plotly_utils/colors/__init__.py#L419).

nicolaskruchten:
🙈 yikes, that's awkward :) well, that's been in our public API for so long I'm not sure it would be all that helpful to change it now but maybe we can do it in the v5 bump

c-chaitanya:
Hi @nicolaskruchten
I have also tried this approach for the docstring part(the part between ''' ''' and """ """). The issue I faced was that there were several entries in my output file (over 5000) , and almost 99% of them(basic visual understanding) were from the tox folder. So I just left them for a future date.

PR #3079, opened 2021-02-18
Doc prod

c-chaitanya:
In continuation with the previous pull request #3073, I have fixed the iteration logic of my "spelling mistakes detection" script to iterate over all directories and its subdirectories. Fixing a few more documents which were missed in the previous sweep. I guess all documents are completely covered.

For my next step I'm modifying my script to fix docstrings within python files. The regex to extract comments between """ """ is in place but not for ''' ''' and # ...

Let me know if the pull request for docstring correction within python files goes to doc-prod or master

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,7 +135,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   faster image rendering and smaller figure size. Additional optional arguments
   `binary_backend`, `binary_format` and `binary_compression_level` control
   how to generate the b64 string ([#2691](https://github.com/plotly/plotly.py/pull/2691)
-- `px.imshow` has a new `constrast_rescaling` argument in order to choose how
+- `px.imshow` has a new `contrast_rescaling` argument in order to choose how
   to set data values corresponding to the bounds of the color range
   ([#2691](https://github.com/plotly/plotly.py/pull/2691)
 
@@ -239,7 +239,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- The `hover_data` parameter of `px` functions can now be a dictionary. This makes it possible to skip hover information for some arguments or to change the formatting of hover informatiom [#2377](https://github.com/plotly/plotly.py/pull/2377).
+- The `hover_data` parameter of `px` functions can now be a dictionary. This makes it possible to skip hover information for some arguments or to change the formatting of hover information [#2377](https://github.com/plotly/plotly.py/pull/2377).
 - It's now possible to build a development version of Plotly.py against the build artifacts from a non-`master` branch of Plotly.js, which makes for faster QA and development cycles [#2349](https://github.com/plotly/plotly.py/pull/2349). Thanks [@zouhairm](https://github.com/zouhairm) for this Pull Request!
 
 ### Fixed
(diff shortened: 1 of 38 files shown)


CHANGELOG.md
```diff
@@ -439,7 +439,7 @@ section [#1969](https://github.com/plotly/plotly.py/pull/1969).
  - The width of a figure produced by the `create_gantt` figure factory now resizes responsively ([#1724](https://github.com/plotly/plotly.py/pull/1724))
 
 ### Fixed
- - The name of the steps property of `graph_objects.indicator.Guage` has been renamed from `stepss` to `steps`
+ - The name of the steps property of `graph_objects.indicator.Gauge` has been renamed from `stepss` to `steps`
```
nicolaskruchten, line 442: I'm very impressed that the spell-checking finds this mistake but ignores the `stepss` on the same line! How does this work? Are you manually reviewing the changes before submitting the PRs?

doc/python/imshow.md
```diff
@@ -231,7 +231,7 @@ fig.show()
 
 ### Automatic contrast rescaling in `px.imshow`
 
-When `zmin` and `zmax` are not specified, the `contrast_rescaling` arguments determines how `zmin` and `zmax` are computed. For `contrast_rescaling='minmax'`, the extrema of the data range are used. For `contrast_rescaling='infer'`, a heuristic based on the data type is used:
+When `zmin` and `zmax` are not specified, the `contrast_rescaling` arguments determines how `zmin` and `zmax` are computed. For `contrast_rescaling='minmax'`, the extreme of the data range are used. For `contrast_rescaling='infer'`, a heuristic based on the data type is used:
```
nicolaskruchten, line 234: this is the only one I disagree with... "extrema" is the correct latin word here although "extremes" would be less confusing to some ;)

doc/python/linear-fits.md
```diff
@@ -23,7 +23,7 @@ jupyter:
     version: 3.6.8
   plotly:
     description: Add linear Ordinary Least Squares (OLS) regression trendlines or
-      non-linear Locally Weighted Scatterplot Smoothing (LOEWSS) trendlines to scatterplots
+      non-linear Locally Weighted Scatterplot Smoothing (LOWESS) trendlines to scatterplots
```
nicolaskruchten, line 26: again, very impressive catch!

nicolaskruchten:
Just one word to fix and I can merge this, thank you :)

Any changes to Python files/files not in `doc` should target `master` please!

CHANGELOG.md
c-chaitanya: Appreciate the comments @nicolaskruchten. No advanced NLP used here, but very basic NLP stuff. Firstly I have a corpus of english words with me. In my second step I create a corpus from all the documents in the project (lets call this plotly corpus). Then I count all the words in the plotly corpus. Then I add all words in plotly corpus having word count greater than a fixed threshold to the english ones. This step ensures that the programming jargon is not counted as a spelling error.

Now all words below the threshold are the errors. If any document contains these words, my script just prints out the filename and the errored word. I go through this and fix them manually.

c-chaitanya:
@nicolaskruchten Done with the changes you suggested.

CHANGELOG.md
nicolaskruchten: Interesting! `stepss` is definitely not anywhere else, or shouldn't be. Its inclusion in this line of the changelog is specifically to point out that we made a mistake and had to fix it ;)

Either way, thanks again for the contribution :)

c-chaitanya: There is lot of manual work involved. I go through the present sentence understand the context and then read the correction suggestion and take my call.

PR #3094, opened 2021-02-27
Kaleido v5 updates

jonmmease:
Two small kaleido updates that I'd like to get into v5

1) When neither kaleido nor orca is installed, display the kaleido installation instructions instead of orca. 92b8954

2) Don't override kaleido's MathJax path (with the CDN url) if kaleido was successful in autodetecting a local MathJax installation. Local MathJax support will be released in Kaleido 0.2 (See https://github.com/plotly/Kaleido/pull/75). 2b66c73


jonmmease:
Looks like there are some unrelated test failures in `plotly/tests/test_core/test_px/test_px_functions.py`

nicolaskruchten:
💃

PR #3143, opened 2021-04-09
Matplotlylib fixes

fdion:
## Code PR

  modified existing tests.

This PR fixes compatibility of matplotlylib with matplotib `3.3` and above and specifically tested with the latest release (`3.4.1`). Before fixing the compatibility, I fixed any matplotlylib tests that were not passing. Then I fixed the code.

Specifically, this addresses changes to `matplotlib`:
- axis._gridOnMajor moved
- spine.is_frame_like() deprecated
- FuncFormatter instead of FixedFormatter
- changes in dash patterns
- number of points for bounding boxes

It also renders the shapes for the matplotib legend (previously, only the text would display in a legend). This is mainly done through the method `draw_legend_shapes` which is called by `draw_marked_line` and handles lines, markers and lines+markers. It doesn't handle the legend bounding box.

Tested with matplotlib 3.2,2 (last known working with matplotlylib, with warnings) and 3.4.1 (current), from notebooks with various charts / markers. Also tested with `hotelling` package with `matplotlib 3.4.1.
newplot(1)


packages/python/plotly/plotly/matplotlylib/renderer.py
```diff
@@ -312,10 +312,84 @@ def draw_bar(self, coll):
                 "assuming data redundancy, not plotting."
             )
 
+    def draw_legend_shapes(self, mode, shape, **props):
```
nicolaskruchten, line 315: I'm not sure I understand what this function does... could you please explain or provide an example pair of matplotlib/plotly figures that uses this system?

nicolaskruchten:
Thanks for this PR, and apologies for the long time before reviewing it... I've been focused on the changes to Plotly.js for the upcoming 2.0 release, but now that that's mostly done I'm looking at merging a bunch of Python PRs :)

If you could please rebase this onto master or merge master into it, I'd appreciate it, and if you could answer my question about your second commit, I should be able to get this merged shortly!

packages/python/plotly/plotly/matplotlylib/renderer.py
fdion: Without this function which is called further down on line 583 when coordinates are "axes" (plotly only handles "data"), the plots legends have no markers or lines, just text. This is why we get the warning: "Bummer! Plotly can currently only draw Line2D objects from matplotlib that are in 'data' coordinates!"

See example notebook here:

https://colab.research.google.com/drive/1-MNmqy_DeDxw8yPX0U7M9E8AEMtfIobP?usp=sharing

Matplotlib plot:
index


Plotly version:
newplot(2)


When coordinates == axes I call my new function that draws the shapes. go.layout.Shape cant handle circles, so I handle this special case, handle lines, and everything else i pass to go.layout.shape.

Result:
newplot(3)

It still is left with one warning: "I found a path object that I don't think is part of a bar chart. Ignoring."

This is due to the bounding box that in matplotlib is (now) a fancy box with curved corners. This would have to be addressed in a future PR.

fdion:
@nicolaskruchten Merged master and documented the reason for the draw_legend_shapes function

nicolaskruchten:
Thanks very much for the fixes :)

Issue #3158, opened 2021-04-21
Drop Python 2.7 and 3.5 support


nicolaskruchten:
- [ ] drop CI builds
- [ ] mess with setup.py in the appropriate way

nicolaskruchten:
https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

Issue #3159, opened 2021-04-21
Unify/stagger CI requirements.txt


nicolaskruchten:
We could:
* drop tox
* have different requirements.txt files per Python version
* merge "optional" in there
* unify the docs-build requirements.txt with that
* pull PX tests out of core and move those tests to optional
* run the docs as part of optional
* rename `plot_ly` to `chart-studio`

nicolaskruchten:
Good enough! Not sure if the docs unification is super-worth it ATM :)

PR #3160, opened 2021-04-21
Require python 3.6 and CI maintenance

jonmmease:
This PR removes support for Python versions prior to 3.6 and does some clean up of the CI test suite.  In particular, it removes the use of tox and instead relies on a separate requirements.txt file for each Python version.  This makes it possible to test different versions of dependencies.

Still to do in follow-on PR(s)
 - Remove use of six for Python 2/3 compatibility
 - Unify optional requirements and doc requirements
 - Unify dependencies for conda+orca test job with optional requirements

Addresses parts of https://github.com/plotly/plotly.py/issues/3159 and closes https://github.com/plotly/plotly.py/issues/3158


nicolaskruchten:
This PR looks fine to me, but I wonder if we shouldn't merge https://github.com/plotly/plotly.py/pull/2911 first and then bubble the `retrying`->`tenacity` changes through this PR prior to merge, to avoid making it harder to merge #2911

jonmmease:
Yeah, that makes sense.  I'll go ahead and merge #2911 and update this PR.

jonmmease:
ping @nicolaskruchten, I think this is ready to go now.  I added a few notes above for some follow ups that we can do next.

nicolaskruchten:
💃

PR #3164, opened 2021-04-22
CircleCI updates: Unify requirements, drop conda, refactor to reduce duplication

jonmmease:
This PR is a follow-on to https://github.com/plotly/plotly.py/pull/3160 that inlcudes a bit more cleanup to the CI/CD architecture.

 - Previously, the orca test job used conda to install orca and the other python dependencies, and the python dependencies were not the same as those used in the optional environment.  Now, the orca tests use an identical environment to the optional tests, and they install orca using npm (no more conda).
 - Previously the percy tests used an environment configured with a custom set of python dependencies that didn't match anything else. Now it uses the same optional requirements.txt file as optional.
 - I updated the circleci configuration to version 2.1, and refactored it to use commands and parameters (https://circleci.com/docs/2.0/reusing-config/#using-the-parameters-declaration) to significantly reduce configuration duplication.

Not done:
 - This PR does not unify the requirements of the documentation build with the other CI test jobs. I played with it a little bit, but ran into some documentation build failures and so I decided not to lump that into this PR.


nicolaskruchten:
This looks really great, thank you! I'm curious about why the `orca` build is on 3.8? Has it always been that way?

jonmmease:
> I'm curious about why the orca build is on 3.8?

It was 3.7 before. No particular reason for the change, and happy to switch it back to 3.7.

nicolaskruchten:
No need, just curious. :dancer:

Re the docs build, we can probably deal with that another day... the current set of improvements is already huge IMO :)

nicolaskruchten:
We're probably losing a little bit of configuration-diversity in terms of what we're testing on a regular basis, but we're gaining in reliability and maintainability, which is a good tradeoff ATM IMO.

PR #3231, opened 2021-06-08
Add prebuilt assets when installing from sdist

### LINKED ISSUE #3230: [v5.0.0rc2] Source tarball installation failed
@fcollonval question: I tried to `pip install <tarball>` and the labextension wasn't automatically installed. Is this by design/a known limitation?

_Originally posted by @nicolaskruchten in https://github.com/plotly/plotly.py/issues/3142#issuecomment-856117527_

fcollonval:
Fixes #3230

### DIFF
--- a/packages/python/plotly/setup.py
+++ b/packages/python/plotly/setup.py
@@ -33,6 +33,26 @@ if not os.path.exists(labstatic):
     # Ensure the folder exists when we will look for files in it
     os.makedirs(labstatic)
 
+prebuilt_assets = [
+    (
+        "share/jupyter/labextensions/jupyterlab-plotly",
+        ["jupyterlab_plotly/labextension/package.json",],
+    ),
+    (
+        "share/jupyter/labextensions/jupyterlab-plotly/static",
+        [os.path.join(labstatic, f) for f in os.listdir(labstatic)],
+    ),
+    (
+        "share/jupyter/nbextensions/jupyterlab-plotly",
+        [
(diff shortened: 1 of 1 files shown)


fcollonval:
I tested it locally by changing the setup.py in the tar.gz.

It may be worth adding some tests.

nicolaskruchten:
Thank you again! I'll get this rolled into the next RC

nicolaskruchten:
A couple of additional questions for you in terms of expected behaviour:

1. is it normal that `pip install -e .` does not install the extension?
2. what does a user need to restart after installing? refresh the frontend? restart the kernel? clear notebook outputs? restart JupyterLab?

fcollonval:
> * is it normal that `pip install -e .` does not install the extension?

To achieve that, the setup command will need to be modified (by default the editable mode does not copy any files in the python system tree). But I'm not sure this is the best idea. I would recommend running these two commands instead:

```
pip install -e .
jupyter labextension develop . --overwrite
```

See [documentation](https://jupyterlab.readthedocs.io/en/stable/extension/extension_tutorial.html#build-and-install-the-extension-for-development)

> * what does a user need to restart after installing? refresh the frontend? restart the kernel? clear notebook outputs? restart JupyterLab?

If plotly was not installed, they need to restart the kernel and refresh the frontend.

nicolaskruchten:
Thanks! I'm hoping to release this this week :)

Issue #2651, opened 2020-07-17
Static renderers ignore height/width


nicolaskruchten:
This always renders at the default size for me:

```python
import plotly.express as px
fig = px.scatter(y=[1,2,3], height=300, width=300)
fig.show("png")
```

I would consider this a bug, although there may be some reasoning behind this behaviour ... @jonmmease ?

jonmmease:
Yeah, this behavior is not intuitive.  The issue is that the `height` and `width` here are overriding the `height` and `width` properties of the Image renderer.  So this is roughly equivalend to 
```
import plotly.io as pio
pio.renderers["png"].height = 300
pio.renderers["png"].width = 300

import plotly.express as px
fig = px.scatter(y=[1,2,3])
fig.show("png")
```

So here, `width` and `height` are really the default width and height. And here I think it makes sense for the figure's own width/height to take precedence.

But in any case, we do need to put some more thought into precedence.

nicolaskruchten:
Right so this is the same as `fig.show("png", width=300, height=300)`

nicolaskruchten:
See also #2653

nicolaskruchten:
This is addressed in #3240 and will be part of version 5.0

Issue #2933, opened 2020-11-24
Image export ignores figure layout (using Kaleido)


sergiocalde94:
Hi!

Thank you very much for your awesome library 🥇 

I am having some issues about figure sizing with svg render. I think a reproducible example is very visual:

With default render (plotly_mimetype+notebook) the layout correctly changes the figure size:

<img width="1402" alt="Screenshot 2020-11-24 at 19 01 40" src="https://user-images.githubusercontent.com/22750693/100133778-7eef3280-2e87-11eb-9425-baa16ab9f897.png">

But if I try it with svg render it ignores the layout 😢:

<img width="1402" alt="Screenshot 2020-11-24 at 19 02 51" src="https://user-images.githubusercontent.com/22750693/100133897-a8a85980-2e87-11eb-8163-d28c1df664b6.png">

Do you know how to avoid this problem? I need an uniform strategy, that with the same code returns same figure size, excepting the rendered type.

Thank you in advance 👍

sergiocalde94:
Update

**Plotly version**: 4.12.0

nicolaskruchten:
This is partly addressed in #3240 and will be part of version 5.0 ... if you don't specify a width, in Jupyter the plot will fill the window width, but when rendering with PNG you won't get the same width, you'll get the underlying default.

PR #3240, opened 2021-06-11
Remove default width/height for static image renderers

jonmmease:
This takes care  of https://github.com/plotly/plotly.py/issues/2651.

Now the image size precedence is:
 - width/height provided to fig.show() (e.g. `fig.show(width=700, height=400))
 - width/height specified in figure (e.g. `px.scatter(..., width=800, height=500))
 - default_width/default_height specified in the kaleido scope (e.g. pio.kaleido.scope.default_width)

### DIFF
--- a/packages/python/plotly/plotly/io/_renderers.py
+++ b/packages/python/plotly/plotly/io/_renderers.py
@@ -428,13 +428,12 @@ renderers["databricks"] = DatabricksRenderer()
 renderers["json"] = JsonRenderer()
 
 # Static Image
-img_kwargs = dict(height=450, width=700)
-renderers["png"] = PngRenderer(**img_kwargs)
-jpeg_renderer = JpegRenderer(**img_kwargs)
+renderers["png"] = PngRenderer()
+jpeg_renderer = JpegRenderer()
 renderers["jpeg"] = jpeg_renderer
 renderers["jpg"] = jpeg_renderer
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
Note to self: test against https://github.com/plotly/plotly.py/issues/2933


thanks Jon!

nicolaskruchten:
So we also need to set `pio.kaleido.scope.default_(width|height) = None` somewhere, right? because without explicitly doing that I'm still getting those defaults.

nicolaskruchten:
Err.. I guess what I mean is that for me, `pio.kaleido.scope.default_(width|height)` are still taking precedence over the other two.

jonmmease:
Huh, here's what I'm seeing:

**args to `show` take precedence over everything:**
Screenshot_20210614_071313

**Figure dimensions take precedence when nothing specified in `show`**
Screenshot_20210614_071254

**Scope defaults take precedence when no args to show and no figure dimensions specified**
Screenshot_20210614_071329

nicolaskruchten:
Hmmm that's not what I'm seeing... I pulled master and did `pip install -e .` and here is what I get:

image

nicolaskruchten:
Running `pip install -e .` on the tip of this branch gives me the same output as I have on master... is there an uncommitted file or something here?

PR #3242, opened 2021-06-11
Add missing double quotations

nzw0301:
The current lines do not work correctly as follows:
 
```zsh
# suppose zsh
$ conda install jupyterlab>=3 "ipywidgets>=7.6"
zsh: 3 not found
```

I think bash also does not work with these commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,13 +101,13 @@ For use in JupyterLab, install the `jupyterlab` and `ipywidgets`
 packages using `pip`:
 
 ```
-pip install jupyterlab>=3 "ipywidgets>=7.6"
+pip install "jupyterlab>=3" "ipywidgets>=7.6"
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
Thanks for this PR! Actually the state of the Readme was in a bit of a strange place: it contained parts of the instructions for the upcoming v5! I've reverted to the previous version of the Readme for now and I'll incorporate the feedback from your PR into the v5 Readme in a separate branch :)

nzw0301:
@nicolaskruchten Thank you for your comment and explanation! So, I think this PR is unnecessary anymore. Should I close this PR?

nicolaskruchten:
Yes, let's close it. v5 will be out next week I hope :)

Issue #2348, opened 2020-04-02
chrome/chromium renderer not detected correctly on Ubuntu (name is chromium-browser) 


emmanuelle:
On my Ubuntu machine I cannot configure a chrome renderer (my default browser is firefox). The reason is that Chromium is registered with the name `chromium-browser` as shown below, so that `webbrowser.get` throws an error when called with `chrome`.
```
>>> import plotly.io as pio                                                                                                      
>>> pio.renderers.default = 'chromium'                                                                                           
>>> import plotly.graph_objects as go                                                                                            
>>> go.Figure(go.Scatter(x=[1], y=[0])).show()                                                                                   
Traceback (most recent call last):
  File "<ipython-input-5-d0da89a1679b>", line 1, in <module>
    go.Figure(go.Scatter(x=[1], y=[0])).show()
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/basedatatypes.py", line 2794, in show
    return pio.show(self, *args, **kwargs)
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/io/_renderers.py", line 391, in show
    renderers._perform_external_rendering(fig_dict, renderers_string=renderer, **kwargs)
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/io/_renderers.py", line 340, in _perform_external_rendering
    renderer.render(fig_dict)
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/io/_base_renderers.py", line 736, in render
    open_html_in_browser(html, self.using, self.new, self.autoraise)
  File "/home/emma/code/plotly.py/packages/python/plotly/plotly/io/_base_renderers.py", line 681, in open_html_in_browser
    webbrowser.get(using).open(
  File "/usr/lib/python3.7/webbrowser.py", line 65, in get
    raise Error("could not locate runnable browser")
Error: could not locate runnable browser

>>> import webbrowser                                                                                                            
>>> print(webbrowser._browsers)                                                                                                  
[...]

PR #3278, opened 2021-06-25
fix for chromium renderer on linux systems (issue #2348)

c-chaitanya:
- [✓ ] I have read through the [contributing notes](https://github.com/plotly/plotly.py/blob/master/contributing.md) and understand the structure of the package. In particular, if my PR modifies code of `plotly.graph_objects`, my modifications concern the `codegen` files and not generated files.
- [ ✓] I have added tests (if submitting a new feature or correcting a bug) or
  modified existing tests.

This PR is a fix for issue https://github.com/plotly/plotly.py/issues/2348. I have read the documentation and the contributing notes. Test cases have been added for the same fix.  Im not sure if I need to add anything in the changelog. Let me know if anything needs to be changed.


c-chaitanya:
Fixed black formatting and reopening pull request

nicolaskruchten:
Thanks for this pull request! I'm rerunning the CI job to see if the orca thing was an isolated failure.

packages/python/plotly/plotly/io/_base_renderers.py
```diff
@@ -669,6 +669,11 @@ def open_html_in_browser(html, using=None, new=0, autoraise=True):
     if isinstance(html, six.string_types):
         html = html.encode("utf8")
 
+    if isinstance(using, tuple):
+        using = [i for i in webbrowser._browsers.keys() if any(j in i for j in using)][
+            0
+        ]
```
jonmmease, line 675: Is there a reason to use `if any(j in i for j in using)` here instead of `if i in using` here?  I think the way it is now, you would get partial string matches, which might be confusing.

Also, I think this will cause an index-out-of-bounds error if none of the browsers in the tuple are available.  I don't recall what error is currently raised, but it would be nice to report something a little more informative (e.g. A `ValueError` that explains that none of the elements in `using` were found in `list(if any(j in i for j in using))`).

jonmmease:
Thanks a lot @c-chaitanya. I left one small comment. It'll be great to get this in!

packages/python/plotly/plotly/io/_base_renderers.py
c-chaitanya: Yes @jonmmease using `if i in using` makes more sense than `if any(j in i for j in using)`, will get that change done. 

Regarding the second point I could do something like
```
if isinstance(using, tuple):
    try:
        using = [i for i in webbrowser._browsers.keys() if i in using][0]
    except TypeError:
         raise TypeError("""
Unable to detect any browsers in webbrowser._browsers.keys().
Check that the desired browser is installed on your system.
If problem still persists, try registering the browser using the python webbrowser module """)
```
Let me know if the TypeError is easy to understand or any suggestions are welcome

ValueError for invalid names like `chromee` is already handled and throws and error as follows `ValueError: 
Invalid named renderer(s) received: ['chromee']`

jonmmease: I think sticking with raising a `ValueError` makes sense here (to match to existing behavior).  And I think you'll need to trap an `IndexError` rather than a `TypeError` to handle the case where no matches are found (this is the exception raised when you evaluate `[][0]`).

For the error message, I pictured including the actual list of supported browsers (the result of evaluating `list(webbrowser._browsers.keys())`) rather than "`webbrowser._browsers.keys()`") rather than itself.

Thanks!

nicolaskruchten:
@c-chaitanya if you can address Jon's last comment I can try to get this one into the next release next week!

packages/python/plotly/plotly/io/_base_renderers.py
c-chaitanya: Hi @jonmmease can you please help me out with the content to put in the value error. Does this content seem ok? Any suggestions are welcome.
```
if isinstance(using, tuple):
    try:
        using = [i for i in webbrowser._browsers.keys() if i in using)][0]
    except IndexError:
        raise ValueError("""
Unable to detect the given browser.
Try one among the following "chrome", "chromium", "firefox" or "default" to use your default browser.
""")
```

c-chaitanya:
Hi @nicolaskruchten, I have addressed @jonmmease comments on the pull request. Let me know if anything is pending from my side. Ill try to get it fixed asap for the new release.

jonmmease:
Thanks @c-chaitanya. lgtm, but will let @nicolaskruchten do the merge

nicolaskruchten:
Thanks very much!

Issue #3292, opened 2021-07-07
minor regression: `customdata` is sometimes a tuple 


nicolaskruchten:
This returns a numpy array in 4.14.3 and a tuple in 5.1.0:

```python
import plotly.express as px
fig = px.scatter(px.data.tips(), x="total_bill", y="tip", custom_data=["sex"])
fig.data[0].customdata
```

Ideally we would get back to the old behaviour if possible.

nicolaskruchten:
Closed by #3293

PR #3293, opened 2021-07-08
Fix: customdata is sometimes a tuple

jonmmease:
This PR is a fix for https://github.com/plotly/plotly.py/issues/3292

The regression was that numpy/pandas arrays with Object `dtype` were getting converted into Python lists when added to a `Figure`.  This was caused by an unnecessary changed I made in the JSON serialization PR in commit https://github.com/plotly/plotly.py/pull/2955/commits/6cea61db9280b33d4870ff527be9dcafeed45bd3.

This PR is largely a manual revert of that commit, with an added test to make sure serialization is still behaving properly when object arrays are included in a figure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## UNRELEASED
 
+### Fixed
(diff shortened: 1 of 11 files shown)


nicolaskruchten:
Thanks! Any chance that this degrades performance somehow or anything?

jonmmease:
> Thanks! Any chance that this degrades performance somehow or anything?

No, I don't think so. The conversion to the tuple form would actually have been a performance hit.  Serialization now has to perform the conversion, but that's what it would have needed to do before.

nicolaskruchten:
OK, thanks!

Issue #3253, opened 2021-06-18
Plotly express not compatible with pandas NAType/pandas.NA/pd.NA values


devmcp:
Plotly express functions fall over when given dataframes containing `pd.NA` values. I think these should have the same behaviour as `None` values.

Example:
```
# This works
px.scatter(pd.DataFrame([1, 2, 3, None, 5]))

# This results in "TypeError: Object of type NAType is not JSON serializable"
px.scatter(pd.DataFrame([1, 2, 3, None, 5]).convert_dtypes())
```

nicolaskruchten:
Thanks for pointing that out! We'll try to get it fixed soon :)

nicolaskruchten:
Note to self: will need to look at

https://github.com/plotly/plotly.py/blob/b34bd6d2004b2b4c96d114011c82913573e50513/packages/python/plotly/_plotly_utils/utils.py#L168-L178 

https://github.com/plotly/plotly.py/blob/b34bd6d2004b2b4c96d114011c82913573e50513/packages/python/plotly/plotly/io/_json.py#L507-L526

astafan8:
This simple addition worked for me when i used ``px.timeline`` https://github.com/plotly/plotly.py/pull/3336. Note that I'm a plotly newbie.

avm19:
#3705 partially fixed this issue. The issue is fixed when Python's `json` is used, but the problem persists for `orjson`.

```python
import orjson  # assert that orjson is available

plotly.io.json.config.default_engine = 'json'
px.scatter(pd.DataFrame([1, 2, 3, None, 5]).convert_dtypes()).to_json()  # works
plotly.io.json.config.default_engine = 'auto'  # or `orjson`
px.scatter(pd.DataFrame([1, 2, 3, None, 5]).convert_dtypes()).to_json()  # fails
```

Coding-with-Adam:
thank you for taking a look at this @avm19 .

gvwilson:
Hi - we are tidying up stale issues and PRs in Plotly's public repositories so that we can focus on things that are still important to our community. Since this one has been sitting for a while, I'm going to close it; if it is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. If you'd like to submit a PR, we'd be happy to prioritize a review, and if it's a request for tech support, please post in our community forum. Thank you - @gvwilson

PR #3336, opened 2021-08-05
Initial stab at supporting pandas NAType with json serialization

astafan8:
Starts addressing https://github.com/plotly/plotly.py/issues/3253


nicolaskruchten:
Thanks for the push on this, I've merged #3705 which takes care of this.

PR #3340, opened 2021-08-09
Change URL from githubusercontent.com to github.io

xhluca:
Before:
```python
data = pd.read_csv("https://raw.githubusercontent.com/plotly/datasets/master/2014_usa_states.csv")
```

After:
```python
data = pd.read_csv("https://plotly.github.io/datasets/2014_usa_states.csv")
```

If we can set the URL for `plotly.github.io/datasets` to `data.plotly.com`, we could have something like:
```python
data = pd.read_csv("https://data.plotly.com/2014_usa_states.csv")
```

### DIFF
--- a/doc/python/line-and-scatter.md
+++ b/doc/python/line-and-scatter.md
@@ -208,7 +208,7 @@ fig.show()
 import plotly.graph_objects as go
 import pandas as pd
 
-data= pd.read_csv("https://raw.githubusercontent.com/plotly/datasets/master/2014_usa_states.csv")
+data= pd.read_csv("https://plotly.github.io/datasets/2014_usa_states.csv")
 
 fig = go.Figure(data=go.Scatter(x=data['Postal'],
                                 y=data['Population'],


emilykl:
@gvwilson If we're going to change this URL we should also change the [96 others](https://github.com/search?q=repo%3Aplotly%2Fplotly.py%20raw.githubusercontent.com&type=code) that appear in the Plotly.py docs.

I don't have a full understanding of how the plotly.github.io domain is configured or maintained -- @LiamConnors do you know? Is it via [this repo](https://github.com/plotly/plotly.github.io)?

LiamConnors:
Wasn't aware that we could access the datasets at that plotly.github.io URL, but yeah it would be that repo.
I think for now we should leave it as. We'd need to update all the links, as you mention @emilykl . Also, going to `https://raw.githubusercontent.com/plotly/datasets/master/2014_usa_states.csv` in a browser displays the data, but the other link downloads the file for some reason. I think the former behaviour is better for someone wanting to open a dataset while reviewing a docs example.

PR #3351, opened 2021-08-19
Fix error in orjson engine when a dict has a non-string value as a key

jonmmease:
## Background
Python's built-in json library can handle encoding dictionaries with keys that aren't strings by coercing them to strings implicitly.  The orjson library requires the keys to be strings, otherwise an error is raised.

This would very rarely come up in plotly figures. But now that we're updating Dash (for Dash 2) to use the active plotly.py json engine, it comes up on occasion.

## Change
All dict keys are converted to strings during the existing cleaning pass that is used prior to orjson encoding.

## Tests
Test added that encodes a dict with integer keys.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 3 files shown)


nicolaskruchten:
:dancer:

PR #3386, opened 2021-09-15
Do not sort keys during json serialization

### LINKED ISSUE #3380: Handle serializing dicts with ints and strings
The current `to_json_plotly` doesn't handle dicts with integers and strings as keys. This used to work before we set `sort_keys=True`.

This is relatively common in Dash.

Here is a MWE:
```python
import pandas as pd
from plotly.io.json import to_json_plotly

df = pd.DataFrame({'a': [0], 1: [0]})
to_json_plotly(df.to_dict('records'))
```

```
TypeError: '<' not supported between instances of 'int' and 'str'
```

Here is the issue in the Python tracker: https://bugs.python.org/issue25457 re `sort_keys=True`

Originally reported in https://community.plotly.com/t/dash-2-0-0-migration-error/56478/7

jonmmease:
Closes https://github.com/plotly/plotly.py/issues/3380

This PR removes the instruction to the json/orjson engines to sort dict keys during serialization.  This fixes #3380 and should improve serialization performance a little.

Also, while looking at the orjson option flags, I noticed that there is a [`OPT_NON_STR_KEYS`](https://github.com/ijl/orjson#opt_non_str_keys) option. 
> Serialize dict keys of type other than str. This allows dict keys to be one of str, int, float, bool, None, datetime.datetime, datetime.date, datetime.time, enum.Enum, and uuid.UUID. For comparison, the standard library serializes str, int, float, bool or None by default.

This would have been a better fix that what I did in https://github.com/plotly/plotly.py/pull/3351, so I made the change here.

### DIFF
--- a/packages/python/plotly/plotly/tests/test_optional/test_offline/test_offline.py
+++ b/packages/python/plotly/plotly/tests/test_optional/test_offline/test_offline.py
@@ -89,7 +89,6 @@ class PlotlyOfflineMPLTestCase(TestCase):
             # just make sure a few of the parts are in here
             # like PlotlyOfflineTestCase(TestCase) in test_core
             self.assertTrue(data_json in html)  # data is in there
-            self.assertTrue(layout_json in html)  # layout is in there too
             self.assertTrue(PLOTLYJS in html)  # and the source code
             # and it's an <html> doc
             self.assertTrue(html.startswith("<html>") and html.endswith("</html>"))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [next] - ???
+
+### Fixed
+  - Fixed error when serializing dict with mix of string and non-string keys [#3380](https://github.com/plotly/plotly.py/issues/3380)
+
+### Updated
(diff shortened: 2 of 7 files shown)


nicolaskruchten:
@jonmmease is this one ready for review?

jonmmease:
yes

packages/python/plotly/plotly/tests/test_optional/test_offline/test_offline.py
```diff
@@ -89,7 +89,6 @@ def test_default_mpl_plot_generates_expected_html(self):
             # just make sure a few of the parts are in here
             # like PlotlyOfflineTestCase(TestCase) in test_core
             self.assertTrue(data_json in html)  # data is in there
-            self.assertTrue(layout_json in html)  # layout is in there too
```
nicolaskruchten, line 92: why does this line need to go?

jonmmease: Oh, I should have mentioned that.  Without sorting, there is a property order difference in the layout string here.  I spent a half hour trying to track it down but didn't get all the way there, it looked like it was related to some combination of fig.to_dict() and later figure validation not always preserving the property order.  All of the properties are there, they just get reordered in the HTML output compared to the plain json serialization.

nicolaskruchten: hmmm ok.

PR #3387, opened 2021-09-16
support dataframe protocol (tested with Vaex)

maartenbreddels:
This allows plotly express to take in any dataframe that supports
the dataframe protocol, see:
https://data-apis.org/blog/dataframe_protocol_rfc/
https://data-apis.org/dataframe-protocol/latest/index.html

Test includes an example with vaex, which should work with
https://github.com/vaexio/vaex/pull/1509/
(not yet released)

This is only a POC, I think this needs to wait till Pandas implemented the `from_dataframe`, and if you'd like to keep this test, would require a Vaex version with above mentioned PR merged and released.

Usage:
image

Note that this does not speed up any aggregation/processing, although reading from hdf5/arrow/parquet might be faster.

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1304,14 +1304,17 @@ def build_dataframe(args, constructor):
     df_provided = args["data_frame"] is not None
     if df_provided and not isinstance(args["data_frame"], pd.DataFrame):
         if hasattr(args["data_frame"], "__dataframe__"):
-            # Pandas does not implement a `from_dataframe` yet
-            # $ wget https://raw.githubusercontent.com/data-apis/dataframe-api/main/protocol/pandas_implementation.py
-            # $ export PYTHONPATH=`pwd`
-            import pandas_implementation
-
-            args["data_frame"] = pandas_implementation.from_dataframe(
-                args["data_frame"]
-            )
+            try:
+                import pandas.api.interchange
+            except ModuleNotFoundError:
+                raise NotImplementedError(
+                    "The dataframe you provided supports the dataframe interchange"
+                    "protocol, "
+                    "but pandas 1.5.0 or greater is required to consume it."
+                )
+            df_not_pandas = args["data_frame"]
+            df_pandas = pandas.api.interchange.from_dataframe(df_not_pandas)
+            args["data_frame"] = df_pandas
         else:
             args["data_frame"] = pd.DataFrame(args["data_frame"])
     df_input = args["data_frame"]
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1303,7 +1303,17 @@ def build_dataframe(args, constructor):
     # Cast data_frame argument to DataFrame (it could be a numpy array, dict etc.)
     df_provided = args["data_frame"] is not None
     if df_provided and not isinstance(args["data_frame"], pd.DataFrame):
-        args["data_frame"] = pd.DataFrame(args["data_frame"])
+        if hasattr(args["data_frame"], "__dataframe__"):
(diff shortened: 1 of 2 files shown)


maartenbreddels:
credits go to @AlenkaF for doing the heavy lifting on the Vaex side (cc @rgommers)

nicolaskruchten:
Cool, thanks! I'll be happy to merge this once it's a bit more ready :)

nicolaskruchten:
See also https://github.com/plotly/plotly.py/pull/3901 for an alternative approach

maartenbreddels:
Since pandas 1.5.0 it has support for the protocol: https://pandas.pydata.org/docs/whatsnew/v1.5.0.html#dataframe-interchange-protocol-implementation

maartenbreddels:
Not sure if you'd like a vaex dependency for testing, but in case you're ok with it, where should that do?

nicolaskruchten:
Thanks Maarten! There should be a requirements_optional.txt somewhere that I can add a vaex dependency :)

Once we merge this, we can still offer https://github.com/plotly/plotly.py/pull/3901 as a fallback, for cases where someone has an old pandas or an old vaex without the interchange stuff but, say, a vaex that does still export itself to_pandas, right?

maartenbreddels:
Absolutely, although we have a to_pandas_df() method

labanyamukhopadhyay:
Hi @nicolaskruchten Is there an update regarding support for the dataframe exchange protocol? It would be useful for interoperability with Plotly and [Modin ](https://github.com/modin-project/modin)dataframes!

alexander-beedie:
As an FYI, this would also enable transparent [polars](https://www.pola.rs) integration (closing https://github.com/plotly/plotly.py/issues/3637) as we implemented support for the DataFrame Interchange Protocol back in `0.16.2` (end of Jan/beginning of Feb this year) 📈

MarcoGorelli:
Is this still active?

If so, I'd strongly suggest setting 2.0.2 as the minimum pandas version to try interchanging from, because there's some pretty basic mistakes in earlier versions:
```python
In [1]: df = pl.DataFrame({'a': [1,2,3]})

In [2]: pd.api.interchange.from_dataframe(df[1:])
Out[2]:
                 a
0                2
1                3
2  125822987010162
```
(😱 )

anmyachev:
> If so, I'd strongly suggest setting 2.0.2* as the minimum pandas version to try interchanging from, because there's some pretty basic mistakes in earlier versions:

Hi, 2.0.2 is out.

As an option, one can change the condition so that this protocol is used for those users who have already switched to pandas 2.0.2 and not wait until the plotly's minimum supported version of pandas will be 2.0.2.
```python
from packaging import version
if hasattr(args["data_frame"], "__dataframe__") and version.parse(pd.__version__) >= version.parse("2.0.2"):
```

MarcoGorelli:
thanks @anmyachev - yes, sorry, that's what I meant, rather than bumping the minimum pandas version for everything

anmyachev:
> Once we merge this, we can still offer #3901 as a fallback, for cases where someone has an old pandas or an old vaex without the interchange stuff but, say, a vaex that does still export itself to_pandas, right?

@nicolaskruchten using interchange protocol and also having a fallback as in #3901 looks the best option for now, since IIUC interchange protocol doesn't work for Series.

cc @MarcoGorelli

anmyachev:
> This is probably a good idea still yes, if someone wants to update this PR to implement this fallback :)

Good! In that case, I'll take care of it, if no one minds :)

anmyachev:
@nicolaskruchten I made a separate pull request, with the continuation of this work. https://github.com/plotly/plotly.py/pull/4244

alexcjohnson:
Yes, thanks @anmyachev!

PR #3409, opened 2021-10-03
dash-bio docs posts

olayway:
## added dash-bio docs posts for dash-graphing-library:
-  AlignmentChart
-  Circos
-  Clustergram
-  Molecular Visualizations
-  VolcanoPlot

Couldn't assign @emmeb as a reviewer, so I'm tagging instead. @smallstepman for visibility.


nicolaskruchten:
Unfortunately this straightforward approach will not work to display or document dash apps in this framework... let's discuss on Slack after the launch webinar on Wednesday :)

smallstepman:
We were worried about same thing, tho we thought we'll try simplistic approach first ^^. We saw other dash examples use iframe wrapped inside ` ```python no_display=true ... `. We're happy  to redo this PR to follow same convention, however we'd need some general direction (e.g. where to deploy each of the dash-bio examples - dash gallery or somewhere else)

PR #3411, opened 2021-10-05
change plotly.js dependencie by plotly.js-dist

FernandezMathieu:
Hi,

This PR refers to this issue : https://github.com/plotly/plotly.js/issues/5966

We are facing trouble to build jupyterlab-plotly because of the box-intersect package, which is not accessible through our company CI network.

PR #3412, opened 2021-10-05
use plotly.js-dist instead of plotly.js

archmoj:
Supersedes #3411
Addressing: https://github.com/plotly/plotly.js/issues/5966
cc: @FernandezMathieu

### DIFF
--- a/packages/javascript/jupyterlab-plotly/package-lock.json
+++ b/packages/javascript/jupyterlab-plotly/package-lock.json
@@ -4,24 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "3d-view": {
-      "version": "2.0.0",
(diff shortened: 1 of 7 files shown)


nicolaskruchten:
This approach won't work... the overall code-gen system needs the full build. I can look into this next week or the week after.

nicolaskruchten:
I'm curious about why you need to build the extension locally as opposed to downloading it from NPM or just letting the plotly.py installation process handle loading the extension?

nicolaskruchten:
So Plotly.js 2.6 is now out and should be buildable again without needing to connect to Github... Could you please see if your issue is resolved?

gvwilson:
@archmoj please let us know if this ought to be in Cycle 16

archmoj:
It looks like plotly.py at the moment needs rebuild plot-schema.
So this require extra work.
Closing for now until but we may revisit this.

PR #3418, opened 2021-10-09
add highlightjs on codeblock

congnd91:


nicolaskruchten:
Really, it seems like all you want to do is suppress the NBConvert syntax highlighting... maybe there's some option to do this?

PR #3423, opened 2021-10-12
Update 3d-scatter-plots.md

farzinghanbari:
Added fig.show() to the code snippet in "Style 3d scatter plot" section


LiamConnors (approved):
Thanks! I've changed the target branch to doc-prod so the update will go live after it's merged. 
:dancer:

PR #3446, opened 2021-10-31
UPDATE: consistent CLI installation instructions

mitchazj:
Removed the $ symbol on the few that had them, so that these commands are
a) consistent with the other ones already in the README
b) more compatible with GitHub's new feature allowing you to click the copy button on the side of a code snippet and paste to terminal without having to delete the "$"


nicolaskruchten:
Thanks!

PR #3454, opened 2021-11-06
Fix missing whitespaces in imshow error message.

mkcor:
I noticed that because I'm getting this error message in a CI test! :sob:


nicolaskruchten:
Thanks!

PR #3487, opened 2021-11-24
feat: make `to_html` deterministic

### LINKED ISSUE #3449: Plotly Diagram file is non-deterministic due to random UUID
We use Plotly in conjunction with DVC, where DVC tracks the html file that Plotly generates as the output of one of our stages. DVC does some basic equality checking on the output files based on checksums. If a stage is run, and the Plotly-generated HTML changes its contents, we'll see that reflected in a different checksum in `dvc.lock`.

We found that Plotly will generate two different files, even if the input is identical. This is likely caused by this line, which is later emitted into the HTML:

https://github.com/plotly/plotly.py/blob/9a550187ad8a6ff9b84fa26a90cb0c1544a72eaa/packages/python/plotly/plotly/io/_html.py#L138

What are the reasons that lead to using a random ID? Could this potentially be some non-changing value?

Skn0tt:
Closes #3449 

This PR makes `to_html` deterministic. It does so by using a static ID for the HTML div that the plots are rendered into, as opposed to a random UUID that changes on every run.
This is useful when HTML reports are tracked using tools like DVC, where a difference in checksum is used as an indicator for content change.

## Code PR

  modified existing tests.


nicolaskruchten:
Thanks for this PR! Unfortunately, I won't be able to accept it as-is, as it might break things for folks who are using this function to generate HTML fragments for multiple figures, and then injecting them into the same downstream HTML file.

I would recommend adding a new argument to the end of the argument list like `div_id` with a default value of `None` (which would preserve the current behaviour) and if it's a string, then that string would be used in place of the UUID.

Skn0tt:
Hi Nicolas, thanks for getting back so fast! That makes a lot of sense, didn't think of usecases where multiple HTML outputs are in the same file. Implemented in https://github.com/plotly/plotly.py/pull/3487/commits/3b461a7f70212e37c649b71043fbe08742437646!

nicolaskruchten:
Thanks!

PR #3510, opened 2021-12-11
update plotly.js to 2.8.0

nicolaskruchten:


nicolaskruchten:
@LiamConnors `master` here is now based on 2.8.0

PR #3518, opened 2021-12-15
add text_auto to px

nicolaskruchten:
@LiamConnors @alexcjohnson this is what I'm thinking of adding to PX.


packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -208,6 +209,10 @@ def imshow(
         since it uses lossless compression, but 'jpg' (lossy) compression can
         result if smaller binary strings for natural images.
 
+    text_auto: bool or str (default `False`)
+        If `True` or a string, single-channel `img` values will be displayed as text",
+        A string like `'.2f'`, it will be interpreted as a `texttemplate` numeric formatting directive.
```
alexcjohnson, line 214:
```suggestion
        If `True` or a string, single-channel `img` values will be displayed as text.
        A string like `'.2f'` will be interpreted as a `texttemplate` numeric formatting directive.
```

packages/python/plotly/plotly/express/_imshow.py
```diff
@@ -385,8 +390,21 @@ def imshow(
                 "The length of the x vector must match the length of the second "
                 + "dimension of the img matrix."
             )
+
+        if text_auto is True:
+            texttemplate = "%{z}"
+        elif text_auto is not False:
+            texttemplate = "%{z:" + text_auto + "}"
```
alexcjohnson, line 397: does this need an `else` since we're definitely using `texttemplate` below?

alexcjohnson (approved):
LGTM! 💃

packages/python/plotly/plotly/express/_imshow.py
nicolaskruchten: yes, thanks!

PR #3593, opened 2022-02-09
Clean install of dependencies

archmoj:
See https://docs.npmjs.com/cli/v6/commands/npm-ci#description


nicolaskruchten:
I've removed this section of the release process and the CI system now uses `npm ci` ... thanks for the push!

PR #3628, opened 2022-03-12
Smoothing being applied in the "unnoised" data

mateomontero01:
The smoothening was being applied in y, and not in y_noise. That makes no sense.


gvwilson:
@alexcjohnson can you please tell me if the commit comment makes sense?

alexcjohnson (approved):
Good catch @mateomontero01 - and apologies for not noticing your PR earlier! Yes of course, the whole point here is what this filter does to noisy data, not to the underlying smooth curve. 💃

PR #3649, opened 2022-03-30
Make pandas.NA JSON-serializable

GlebMorgan:
When plotting mapbox from pandas dataframe, if `hover_data` contains `pandas.NA` values, it throws `Object of type NAType is not JSON serializable` error. However for `datetime64` objects it handles `pandas.NaT` values fine. I consider these cases to be similar and propose to make `NA` objects serializable the same way.


nicolaskruchten:
Thanks for this! Do you know which version of pandas introduced `pandas.NA`? Should we check for its existence before trying to access it?

nicolaskruchten:
Thanks for the push on this, I've merged #3705 which takes care of this.

PR #3688, opened 2022-04-22
FigureWidget: Handle after-show Lumino message

martinRenou:
# Handle the `after-show` Lumino message

This fixes an issue when the Figure widget is embeded in a tab/accordion container that caused the figure to not have the right initial size.

Before this PR:

before_aftershow

After this PR:

aftershow


trungleduc (approved):

nicolaskruchten:
Thanks!

Issue #3702, opened 2022-05-01
build-doc workflow fails on CircleCI


maresb:
Looking at the checks for all the recent PRs, they are failing on the "make html" step with

```
ImportError: cannot import name 'contextfilter' from 'jinja2'
```

This seems to be a problem with a breaking change since jinja2 v3.1. Since the nbconvert version is currently [pinned](https://github.com/plotly/plotly.py/blob/470a4ac2650be1178c92fc92f94d64d3cb0bbb64/doc/requirements.txt#L34) to the old `v5.6.1` the version constraints of nbconvert don't account for the change to jinja2.

I can see three obvious potential ways to solve this:

1. Pin `jinja2<3.1`
2. Unpin `nbconvert=5.6.1`
3. Use [some tool](https://github.com/jupyter/nbconvert/issues/1742#issuecomment-1084367167) to lock dependencies to guarantee reproducible builds

At least one of 1. or 2. will probably lead to a quick fix, but will inevitably lead to similar eventual future breakage. So 3. is probably the better solution, but requires significantly more effort.

Reference: nbconvert [issue](https://github.com/jupyter/nbconvert/issues/1742) regarding ImportError

nicolaskruchten:
Thanks! I'm not sure why the jobs are not running on your PRs but the path of least resistance here is to pin `jinja` I think.

nicolaskruchten:
Closed by #3704

PR #3703, opened 2022-05-01
Unpin nbconvert in doc/requirements.txt

maresb:
Second proposed solution to #3702

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -31,6 +31,6 @@ kaleido
 umap-learn==0.5.1
 pooch
 wget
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
CI didn't run here for reasons I don't understand yet, but `nbconvert` is pinned because newer versions aren't compatible with the [NB template in the `doc` repo](https://github.com/plotly/plotly.py/blob/master/doc/nb.tpl). This is probably an hour of work to fix but it's an hour I never found :)

nicolaskruchten:
#3704 resolved the issue, thanks again!

Issue #3696, opened 2022-04-28
`Mathjax.Hub.config` used for plotly 5.7.0


AakashGfude:
Could not find a duplicate of this issue anywhere so, not sure if anyone else had an issue like this, but when using `plotly` in a sphinx project and creating HTML pages, I get the following js code included in my HTML page:

```
<div class="output text_html">        <script type="text/javascript">
        window.PlotlyConfig = {MathJaxConfig: 'local'};
        if (window.MathJax) {MathJax.Hub.Config({SVG: {font: "STIX-Web"}});}
        if (typeof require !== 'undefined') {
        require.undef("plotly");
        define('plotly', function(require, exports, module) {
            /**
* plotly.js v2.11.1
.....
```

When I have a math block in the same page as a plotly graph, then `window.MathJax` is defined and `MathJax.Hub.Config` gets called. Which throws an error as I am using mathjax v3, which is supported by v5.7.0.

The line `MathJax.Hub.Config({SVG: {font: "STIX-Web"}})` I see written in a bunch of files like:
https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/io/_base_renderers.py#L240

Is it a legit issue?

archmoj:
You must check the version of MathJax to be v2 e.g. here
```js
if (window.MathJax && (window.MathJax.version || '').split('.')[0] === '2') {MathJax.Hub.Config({SVG: {font: "STIX-Web"}});}
```

alexcjohnson:
@archmoj let's add this to the queue for after the coming plotly.js release, see if you can figure out the equivalent command to switch to when we detect v3.

@AakashGfude thanks for the report!

PR #3706, opened 2022-05-03
fix setting config for MathJax v2 - ensure version 2 Hub

archmoj:
Addressing #3696

cc: @alexcjohnson @nicolaskruchten

### DIFF
--- a/packages/javascript/jupyterlab-plotly/src/Figure.ts
+++ b/packages/javascript/jupyterlab-plotly/src/Figure.ts
@@ -826,9 +826,13 @@ export class FigureView extends DOMWidgetView {
     this.model.on("change:_py2js_update", this.do_update, this);
     this.model.on("change:_py2js_animate", this.do_animate, this);
 
-    // MathJax configuration
+    // MathJax v2 configuration
     // ---------------------
-    if ((window as any).MathJax) {
+    if (
+      (window as any).MathJax &&
+      (window as any).MathJax.Hub &&
+      (window as any).MathJax.Hub.Config
+    ) {
       (window as any).MathJax.Hub.Config({ SVG: { font: "STIX-Web" } });
     }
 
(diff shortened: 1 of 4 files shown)


packages/javascript/jupyterlab-plotly/src/Figure.ts
```diff
@@ -826,9 +826,13 @@ export class FigureView extends DOMWidgetView {
     this.model.on("change:_py2js_update", this.do_update, this);
     this.model.on("change:_py2js_animate", this.do_animate, this);
 
-    // MathJax configuration
+    // MathJax v2 configuration
```
alexcjohnson, line 829: Is there a v3 equivalent?

archmoj: Not at the moment it appears: http://docs.mathjax.org/en/latest/output/fonts.html

packages/javascript/jupyterlab-plotly/src/Figure.ts
```diff
@@ -826,9 +826,13 @@ export class FigureView extends DOMWidgetView {
...
     // ---------------------
-    if ((window as any).MathJax) {
+    if (
+      (window as any).MathJax &&
+      (window as any).MathJax.Hub &&
+      (window as any).MathJax.Hub.Config
```
alexcjohnson, line 834: As this is a more modern toolchain, we should be able to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining):
```js
if ((window as any)?.MathJax?.Hub?.Config)
```

And in fact, if we don't branch off for v3 you can even use the function call version:
```js
(window as any)?.MathJax?.Hub?.Config?.({ SVG: { font: "STIX-Web" } });
```

archmoj: Done in 13f7c91f7ba0c89ded4383b35ac0ffc46f563060.

alexcjohnson (approved):
LGTM! 💃

nicolaskruchten:
Thanks!

PR #3708, opened 2022-05-03
type annotations for chainable Figure methods

nicolaskruchten:


nicolaskruchten:
This PR lets you do things like `px.scatter(...).<tab>` and your IDE (VSCode, Jupyter, others) will be able to suggest all the `go.Figure` methods to you like `update_layout()`. Furthermore, those methods have type hints now too so you can do `px.scatter(...).update_layout(...).<tab>` and get suggestions.

For folks not into long chains, the following is nice too: `fig = px.scatter(...); fig.<tab>` and you get suggestions.

(N.B. the autocomplete key isn't always `<tab>` but hopefully my meaning is clear :)

PR #3746, opened 2022-05-25
Remove errroneous comma in SVG path

SeanMcCarren:
This is only a minor mistake, but slightly misleading because one may think that this is an example of how to draw partial edges.

### DIFF
--- a/doc/python/shapes.md
+++ b/doc/python/shapes.md
@@ -565,7 +565,7 @@ fig.update_layout(
         # filled Polygon
         dict(
             type="path",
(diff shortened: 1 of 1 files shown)


gvwilson:
@SeanMcCarren thank you for submitting this - we've moved the docs to another repo, so I've filed an issue there to make sure this is corrected in what we're now using.

LiamConnors:
Thanks for the PR @SeanMcCarren! Sorry for the delay in getting it merged

PR #3749, opened 2022-05-26
allow non-string extras in flaglist attributes

alexcjohnson:
To support https://github.com/plotly/plotly.js/pull/6193

  modified existing tests.

### DIFF
--- a/packages/python/plotly/_plotly_utils/basevalidators.py
+++ b/packages/python/plotly/_plotly_utils/basevalidators.py
@@ -1795,8 +1795,6 @@ class FlaglistValidator(BaseValidator):
         self.extras = extras if extras is not None else []
         self.array_ok = array_ok
 
-        self.all_flags = self.flags + self.extras
-
     def description(self):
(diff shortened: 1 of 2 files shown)


packages/python/plotly/setup.py
```diff
@@ -286,7 +286,7 @@ def get_latest_publish_build_info(repo, branch):
 
     url = (
         r"https://circleci.com/api/v1.1/project/github/"
-        r"{repo}/tree/{branch}?limit=10000\&filter=completed"
+        r"{repo}/tree/{branch}?limit=100&filter=completed"
```
alexcjohnson, line 289: After I remove the `\` that makes its way into the final URL but before changing the `limit` value, this returns:
```
{
  "message" : "limit must be between 0 and 100"
}
```

I don't know what set of jobs this was returning when the `\` was there, looked recent but somehow didn't include any `publish-dist` jobs. 🤷 anyway now it seems to work locally.

PR #3760, opened 2022-06-05
Update figure-structure.md

Gargeya-Grey:
Change Figure-wide to Figure-wise


nicolaskruchten:
Actually, "Figure-wide" is what we're trying to say here :)

Issue #1743, opened 2019-08-29
Slow performance with plotly chart builder


helloIAmPau:
Hello all,
I need to use `plotly` as the backend of a microservice who generates charts dynamically.
Unfortunately, after a little benchmarking, I found that the `plotly.express` framework is very slow (around 5 secs to generate a chart from 500 lines dataset).

Here is the script I use to generate a scatter matrix:
```python
import sys
import os
import traceback
import json
import time
sys.path.append('c:\\statwolf\\python\packages\Lib\site-packages')
input = json.loads('{\"file\":\"/tmp/data.tsv",\"color\":\"club_country\",\"dimensions\":[\"nolo\",\"tolo\",\"yolo\"]}')
def action():
    def run():
        import plotly.express as px
        from pandas import read_csv
    
        color = None if input['color'] == "" else input['color']
    
        first = time.time()
        d = read_csv(input['file'], sep='\t')
        second = time.time()
        fig = px.scatter_matrix(d, dimensions=input['dimensions'], color=color)
        third = time.time()
        j = fig.to_json()
        fourth = time.time()
    
        print('read: ' + str(second - first))
        print('plot: ' + str(third - second))
        print('json: ' + str(fourth - third))
        return j
    
    import time
    
    for i in range(0, 3):
        start = time.time()
        result = run()
        end = time.time()
    
        print('iteration: ' + str(i) + '\ntime: ' + str(end - start))
    
    return result
    
result = None
try:
    result = { 'outcome': action() }
except Exception as e:
    traceback.print_exc()
    result = { 'error': str(e) }
resultDir = os.path.dirname(os.path.realpath(__file__))
resultFile = open(resultDir + '/result.json', 'w')
json.dump(result, resultFile)
resultFile.close()
```
from the dataset:
https://www.dropbox.com/s/cm9i3pfv10exbba/data.tsv?dl=1

and this is the report with timing:
https://www.dropbox.com/s/l2x3jqzea4i4xqw/report.txt?dl=1

Now:
1. Is there any tweak I can implement to improve performances?
2. Do you plan to focus on speed for the following releases?

WenyinWei:
Hello all,
I am also disappointed at the performance of plotly.py because my dataset is not that small. plotly only utilized one of my CPUs to exectute its chart drawing. Does plotly plan to improve the performance by parallel processing or something else?

ievgennaida:
The same issue. 
18 seconds for 800 data points when data_frame is used. 
[...]

PR #3761, opened 2022-06-06
:zap: avoid expensive & unnecessary groupby in px

jvdd:
This PR enables significant speed-ups for `plotly.express` when no [groupby operation](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.groupby.html) is required. This PR stems from the observations I made in #1743; current behavior always performs at least one groupby, even when this is not necessary (e.g., `px.scatter` of a 1D array).

Performing such a groupby is a rather expensive operation + the `.get_group` and the `get_ordenings` operations are also rather time-consuming. This PR checks via the `all_same` variable whether grouping is required. If no grouping is necessary, no groupby will be performed, resulting in significant performance gains.

### Illustration of the speed-up

**plotly 5.8.0**
image

**this PR**
image

Can you validate that this speed-up is correct? 
(FYI: all pytest tests succeed locally)

## Code PR checklist

- [ ] I have added tests (if submitting a new feature or correcting a bug) or modified existing tests.
- [ ] For a new feature, I have added documentation examples in an existing or new tutorial notebook (please see the doc checklist as well).

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1957,14 +1957,15 @@ def make_figure(args, constructor, trace_patch=None, layout_patch=None):
     )
     grouper_ = [x.grouper or one_group for x in grouped_mappings] or [one_group]
 
-    all_same = True  # Variable indicating if grouping can be avoided
+    all_same_group = True  # variable indicating if grouping can be avoided
     for g in grouper_:
         if g is not one_group:
-            all_same &= args["data_frame"][g].nunique() == 1
-        if not all_same:
+            arr = args["data_frame"][g].values
+            all_same_group &= (arr[0] == arr).all(axis=0)
+        if not all_same_group:
             break  # early stopping if not all the same
 
-    if all_same:
+    if all_same_group:
         # Do not perform an expensive groupby operation when there are either
         # no groups to group by, or when the group has only one (i.e., the same) value
         grouper = [g for g in grouper_ if g is not one_group]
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
Thanks for this PR, it looks like a promising optimization!

I certainly like the idea of skipping the grouping when it's all `one_group`... When plotting, say, a 5000-row dataframe with 20 colors and 5 symbols or something, what is the performance hit here of the nunique call?

jvdd:
Hi @nicolaskruchten 

In the latest commit I did some minor refactoring (renaming variables) + improved the efficiency of the check I added (replaced `.nunique` with the suggestion from here https://stackoverflow.com/a/54405767

I checked the performance for the following plot; (is this what you intended?)

```py
import pandas as pd
import numpy as np
import plotly.express as px

df = pd.DataFrame(np.random.randn(5_000, 1), columns=["data"])
df["c"] = np.tile(np.arange(20), 5_000 // 20)  # 20 colors
df["s"] = np.tile(np.arange(5), 5_000 // 5)  # 5 symbols

fig = px.scatter(df, y="data", color="c", symbol="s")
```

The checks that I added for grouping (i.e., creating the `all_same_group` variable) take 0.1-0.2% of the total time of `make_figure`.
```py
%lprun -f px._core.make_figure px.scatter(df, y="data", color="c", symbol="s")
```

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1955,10 +1955,37 @@ def make_figure(args, constructor, trace_patch=None, layout_patch=None):
...
+            all_same_group &= (arr[0] == arr).all(axis=0)
+        if not all_same_group:
+            break  # early stopping if not all the same
+
+    if all_same_group:
+        # Do not perform an expensive groupby operation when there are either
+        # no groups to group by, or when the group has only one (i.e., the same) value
+        grouper = [g for g in grouper_ if g is not one_group]
+        assert len(grouper) <= 1
+        # -> create orders, sorted_group_names equivalent to those of get_ordings
```
nicolaskruchten, line 1973: I love the optimization where we don't group if the groupers are all `one_group` but I'm worried about the duplication of logic here... If we just do the optimization when no grouping is requested at all (i.e. leave out the case when we do have e.g. `symbol` but there's only one value) then this PR can be less invasive/have less duplication. I've done an implementation of that over here https://github.com/plotly/plotly.py/pull/3765 if you want to take a look at it.

I think that most of the value of this PR comes from the "all `one_group`" case, and that leaving out the case where there happens to only be one group would be OK for now as this is likely not all that common, but I'm not dead set on it.

nicolaskruchten, line 1973: I love the optimization where we don't group if the groupers are all `one_group` but I'm worried about the duplication of logic here... If we just do the optimization when no grouping is requested at all (i.e. leave out the case when we do have e.g. `symbol` but there's only one value) then this PR can be less invasive/have less duplication. I've done an implementation of that over here https://github.com/plotly/plotly.py/pull/3765 if you want to take a look at it.

I think that most of the value of this PR comes from the "all `one_group`" case, and that leaving out the case where there happens to only be one group would be OK for now as this is likely not all that common, but I'm not dead set on it.

jvdd: Do you mean like your implementation in #3765?

nicolaskruchten:
Superseded by #3765

PR #3767, opened 2022-06-09
:recycle: check for all same groups

jvdd:
Check for all same group instead of checking for all one group (see #3765)

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1916,10 +1916,17 @@ def get_orderings(args, grouper, grouped, all_same_group):
     of a single dimension-group
     """
     orders = {} if "category_orders" not in args else args["category_orders"].copy()
+    sorted_group_names = []
 
     if all_same_group:
-        sorted_group_names = [("",) * len(grouper)]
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
(note: we have "Percy" image tests that compare the graphical output of PRs to that of master, but for some reason these don't run on pull requests from folks outside the Plotly organization. The example I gave above would likely cause this test to fail... the first commit on my PR failed this test as well for similar reasons)

nicolaskruchten:
Note: https://github.com/plotly/plotly.py/pull/3761 does appear backwards-compatible because it independently computes `sorted_group_names`

jvdd:
Oh, I see! I believe it is because of the hard-coding of the orders being `{}` and the `sorted_group_names` being `[("",....)]`.
I'll update the logic of constructing `orders` and `sorted_group_names` to match the current master behavior

jvdd:
> Note: #3761 does appear backwards-compatible because it independently computes `sorted_group_names`

I guess we both identified the issue at the same time! :smile:

jvdd:
Great! And thank you as well @nicolaskruchten for maintaining this amazing library & the quick responses! :)

PR #3769, opened 2022-06-09
:broom: avoid unnecessary one_group groupby operations

jvdd:
Just a quick implementation


nicolaskruchten:
Wow, that's really funny: I was working out in my head how to do *exactly this* while waiting for my kids to fall asleep so I could go implement it afterwards! Nicely done, and thanks for taking the time :)

nicolaskruchten:
Now `one_group` can be something else than a function maybe ;)

PR #3779, opened 2022-06-15
remove duplicate FigureWidget code in graph_objects

jvdd:
This PR removes the logic that is used to import FigureWidget  in `__init__.py` from `graph_objects` and imports it from `..graph_objs` instead

(I might be missing smth in my understanding - but atm I can't think of a reason why you would want exactly the same code in those two files)

### DIFF
--- a/packages/python/plotly/plotly/graph_objects/__init__.py
+++ b/packages/python/plotly/plotly/graph_objects/__init__.py
@@ -53,6 +53,7 @@ if sys.version_info < (3, 7) or TYPE_CHECKING:
     from ..graph_objs import Layout
     from ..graph_objs import Frame
     from ..graph_objs import Figure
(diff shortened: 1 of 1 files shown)


ndrezn:
This seems like a reasonable change. I'm not sure why we handle `FigureWidget` differently than any other imports in this file. Python < 3.7 doesn't support `..` imports, which is why we have this special handler, but it's not clear to me why we use the special handler for _just_ `FigureWidget` and no other import.

In favour of this change to make this file easier to understand 👍 I'd be curious whether this breaks compatibility with <3.6 for some arcane reason though (although to be honest I'm open to requiring Python>=3.8 for Plotly.py in general).

robertclaus:
@camdecoster could you take a look at this? It's very old, but the low line count implies it could be a quick win.

PR #3805, opened 2022-07-07
fix: clean up resize listener when view gets removed

### LINKED ISSUE #3100: Bug - global resize handlers are not removed with FigureWidget()
To reproduce, execute the following code in a cell a few times:
`go.FigureWidget()`
Then try to resize the window. You will notice the following console errors: 
image

However, when doing the same with `go.Figure()`, the problem does not occur. The console logs shows that plotly does detect a notebook cell get destroyed.
image

I need to use the `go.FigureWidget()`, because I can update it later in a another notebook cell. I need this functionality since I use my notebook to render a web application using Voila, and I need the plots to be rendered in specific cells.

Is there an option to use `go.Figure()` and to update it in a another cell?  
Or to somehow make `FigureWidget()` to unbind resize handlers?

maartenbreddels:
Fixes #3100

If this is not cleaned up, we see a lot of stacktraces in the JS console:

index.js?v=20220623202309:2 Uncaught (in promise) Error: Resize must be passed a displayed plot div element.
    at index.js?v=20220623202309:2:2385247
    at new Promise (<anonymous>)
    at b.resize (index.js?v=20220623202309:2:2385199)
    at h.autosizeFigure (index.js?v=20220623202309:2:3736430)
    at index.js?v=20220623202309:2:3736222

Also, the destroy method is renamed to remove. Destroy will never be
called, but remove is.

### DIFF
--- a/packages/javascript/jupyterlab-plotly/src/Figure.ts
+++ b/packages/javascript/jupyterlab-plotly/src/Figure.ts
@@ -804,6 +804,7 @@ export class FigureModel extends DOMWidgetModel {
  */
 export class FigureView extends DOMWidgetView {
   viewID: string;
+  resizeEventListener: () => void;
 
   /**
    * The perform_render method is called by processPhosphorMessage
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
Thanks!

PR #3839, opened 2022-08-09
Add Dash Club snippet

LiamConnors:
@Coding-with-Adam, as requested, this adds the Dash Club snippet that we already have on the Dash docs. Adds it to the ten most popular pages here to get started.


Coding-with-Adam:
Hi @LiamConnors 
is the goal to start with only 10 pages and then put the dash club snippet in more pages if it works?

nicolaskruchten:
Can I see a screenshot/mockup of what you're trying to do with this? There may be an easier site-wide way of doing this.

LiamConnors:
@nicolaskruchten sure. It's the snippet that we already have in the Dash docs. Like on this page:
https://dash.plotly.com/introduction

image

Coding-with-Adam:
@LiamConnors can you please share a screenshot of what this looks like in the docs.

LiamConnors:
@Coding-with-Adam it looks like this:

image

We could also maybe _instead_ add it to the What About Dash section at the end of each page?

image

As **What About Dash** is just one Markdown file that's added to each page when it's being built, it would be much quicker to get this snippet appearing on all pages and would be much easier to manage when we want to make changes.

LiamConnors:
@Coding-with-Adam, apologies, I meant it would be easier to implement and maintain. I see your point about it being too far down the page. What do you think about having it just below the Dash app example?

If we were put it below the Dash app example, maybe we could just add it to the Dash Snippet app itself. Do you see any issues with that approach @nicolaskruchten? I couldn't see any other straightforward way to implement this across the site. But if we add it to the app container (assuming that placement is okay with @Coding-with-Adam) we'd only have to update it in one place to get it on all those pages that have a Dash app.

Coding-with-Adam:
Hi @LiamConnors 

1. Yes, I think so. Are you suggesting we add the snippet to the [code section itself or right above it](https://plotly.com/python/line-and-scatter/#scatter-plots-in-dash)? 

2. In the "join now" section, can you please use this link that Jiny created with the UTM: https://go.plotly.com/dash-club?utm_source=Dash+Club+2022&utm_medium=graphing_libraries&utm_content=inline

3. Can you please update the snippet to say: "Free cheat sheets plus..." instead of "Two free cheat sheets plus...". We just added another two cheat sheets to the list, so they will be getting multiple.

LiamConnors:
> Hi @LiamConnors
> 
> 1. Yes, I think so. Are you suggesting we add the snippet to the [code section itself or right above it](https://plotly.com/python/line-and-scatter/#scatter-plots-in-dash)?

@Coding-with-Adam, I was thinking right below the running example 

>  2. In the "join now" section, can you please use this link that Jiny created with the UTM: https://go.plotly.com/dash-club?utm_source=Dash+Club+2022&utm_medium=graphing_libraries&utm_content=inline

Updated

>  3. Can you please update the snippet to say: "Free cheat sheets plus..." instead of "Two free cheat sheets plus...". We just added another two cheat sheets to the list, so they will be getting multiple.

Updated

I've added the snippet to the pages that have Dash examples. It looks like this:

image

For pages that don't have a Dash example or mention of Dash, where do you think it should go?

Coding-with-Adam:
Thank you @LiamConnors . We're good to go. I just talked to Chris and he liked the placement and wording as well. 
💃🏻 
Could we also update the script from "2+ cheat sheets" to "multiple cheat sheets" in the Dash docs dash club snippet?

chriddyp (approved):
💃🏼

PR #3901, opened 2022-09-25
PX can accept non-pandas dataframes that can .to_pandas()

nicolaskruchten:


exactlyallan:
Regarding accepting RAPIDS cuDF dataframes, this is great at reducing friction, but we lose transparency about moving off GPU to system memory that an explicit `.to_pandas()` shows. For single charts (most of PX's usage?) that really isnt an issue, but I'm wondering how much slower a Dash + PX dashboard would be without the tighter cuDF integration.

nicolaskruchten:
There's not really any tighter integration to be had atm... PX doesn't do any aggregation anyway, it just splits and dumps dataframe columns into JSON basically so it's coming off the GPU no matter what.

The major inefficiency in my PR is that it takes the entire df to pandas, not just the columns it needs, which would be a better approach but potentially trickier/further down into the code.

kevinheavey:
The other drawback of relying on `to_pandas` is it cements Pandas as a dependency

nicolaskruchten:
Ah well right now the dependency is extremely cemented internally, which is part of what this change would try to hide from the user.

LiamConnors (approved):
:dancer:

nicolaskruchten:
@LiamConnors itll be important to mention in the docs my comment above re "copying the whole data frame"!

nicolaskruchten:
definitely! PX does all sorts of manipulation under the hood which is pretty tightly-coupled to The Way Pandas Does Things™ but only out of necessity. If it turns out all of these can be faithfully ported to The Standard Way™ then I'm all for it. To be clear though: none of this is doing any data processing that can be sped up by non-Pandas backends, it's just a bunch of slicing and column rearranging work. PX does essentially no aggregation/math of any kind with Pandas, so the idea of accepting non-Pandas input is just an ergonomic one at this point to avoid the user having to do the `.to_pandas()` themselves.

There's another PR in flight to leverage the interchange approach on top of the `to_pandas()` approach: https://github.com/plotly/plotly.py/pull/4244

MarcoGorelli:
thanks @nicolaskruchten - the interchange protocol  isn't necessarily zero-copy (it just tries to be when possible), so it may still be more efficient to write library-agnostic code

I'll try it out in plotly and see how far it goes

nicolaskruchten:
👍 

probably the biggest win right now would be to `.to_pandas()` or to use the interchange to copy over less than the entire set of columns, which is what's happening right now and can be super-inefficient in the case of e.g. a very wide cudf thing where we haul the entire thing off the GPU just for two columns!

CHANGELOG.md
```diff
@@ -17,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    this feature was anonymously sponsored: thank you to our sponsor!
     - Add `legend.xref` and `legend.yref` to enable container-referenced positioning of legends [[#6589](https://github.com/plotly/plotly.js/pull/6589)], with thanks to [Gamma Technologies](https://www.gtisoft.com/) for sponsoring the related development.
     - Add `colorbar.xref` and `colorbar.yref` to enable container-referenced positioning of colorbars [[#6593](https://github.com/plotly/plotly.js/pull/6593)], with thanks to [Gamma Technologies](https://www.gtisoft.com/) for sponsoring the related development.
+  - `px` methods now accept data-frame-like objects that support a `to_pandas()` method, such as polars, cudf, vaex etc
```
anmyachev, line 19: @nicolaskruchten Looks like `vaex` doesn't have `to_pandas` method, only [`to_pandas_df`](https://vaex.io/docs/_modules/vaex/dataframe.html#DataFrame.to_pandas_df).

Therefore, tests with vaex do not pass in https://github.com/plotly/plotly.py/pull/4244.

Issue #3686, opened 2022-04-22
Incompatible FigureWidget with ipywidgets 8.0.0rc0


JP-Ellis:
Plotly's `FigureWidget` does not appear to be compatible with the upcoming ipywidgets release.

```sh
python -m venv .venv
. ./.venv/bin/activate
pip install plotly jupyterlab ipywidgets==8.0.0rc0
jupyer-lab
```

If using directly the FigureWidget as the output of the cell, the figure is shown by with the following error:

```python
import plotly.graph_objects as go

go.FigureWidget(go.Scatter(x=[1, 2, 3], y=[1, 4, 9]))
```

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File /tmp/josh/bug-report/.venv/lib/python3.10/site-packages/IPython/core/formatters.py:921, in IPythonDisplayFormatter.__call__(self, obj)
    919 method = get_real_method(obj, self.print_method)
    920 if method is not None:
--> 921     method()
    922     return True

File /tmp/josh/bug-report/.venv/lib/python3.10/site-packages/plotly/basewidget.py:741, in BaseFigureWidget._ipython_display_(self)
    737 """
    738 Handle rich display of figures in ipython contexts
    739 """
    740 # Override BaseFigure's display to make sure we display the widget version
--> 741 widgets.DOMWidget._ipython_display_(self)

AttributeError: type object 'DOMWidget' has no attribute '_ipython_display_'
```

On the other hand, if the FigureWidget is used within other widgets, the figure does not appear at all (though there isn't any error message):

```python
import plotly.graph_objects as go
from ipywidgets import widgets

widgets.VBox([
    widgets.HTML("<p>Start</p>"),
    go.FigureWidget(go.Scatter(x=[1, 2, 3], y=[1, 4, 9])),
    widgets.HTML("<p>End</p>"),
])
```

![image](https://user-images.githubusercontent.com/3196162/164600861-658a95b2-6112-4583-a162-e14913114860.png)

This was generated using Plotly `5.7.0` and ipywidgets `8.0.0rc0` all within JupyterLab `3.3.4`.

fabratu:
I have same error for Plotly `5.8.0` and `5.90`.

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
[...]

Issue #3689, opened 2022-04-22
FigureWidget: resizing can only expand but cannot shrink


martinRenou:
When executing the following in a JupyterLab Notebook, increasing the window size results in the plot correctly expanding, but reducing the size of the window does not shrink the plot correctly (see following GIF):

```python
import ipywidgets as widgets

import plotly.graph_objects as go
import plotly.express as px


fig = px.line([1, 2, 3, 4, 6], template='plotly_dark')

fw = go.FigureWidget(fig)

fw
```

![resizing](https://user-images.githubusercontent.com/21197331/164736806-59d97ac4-4abf-436b-809d-ac11e21804fb.gif)

Alexboiboi:
@martinRenou @nicolaskruchten 
When building dashbords with resizable windows and FigureWidgets this is getting quite annoying. Is there any workaround ?

martinRenou:
I never took the time to look deeper into it unfortunately. And I cannot really think of a workaround.

Alexboiboi:
> I never took the time to look deeper into it unfortunately. And I cannot really think of a workaround.

Hi @martinRenou, this hack (https://github.com/plotly/plotly.py/issues/2062) is kinda working

Alexboiboi:
Hi @martinRenou,

looks like the latest release (5.12) solves this issue. 

Worth looking at  [bqplot/bqplot#1531] to adress a similiar issue when the FigureWidget is inside another container.

nicolaskruchten:
Ok great! Let's open a separate issue for any lingering problems plz. No one on the Plotly team is really actively working on the Widget side of things so PRs are very welcome :)

Issue #3669, opened 2022-04-05
5.7.0: pep517 build fails


kloczek:
Looks like 5.7.0 is not pep517 ready
```console
+ /usr/bin/python3 -sBm build -w --no-isolation
* Getting dependencies for wheel...
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/pep517/in_process/_in_process.py", line 363, in <module>
    main()
  File "/usr/lib/python3.8/site-packages/pep517/in_process/_in_process.py", line 345, in main
    json_out['return_val'] = hook(**hook_input['kwargs'])
  File "/usr/lib/python3.8/site-packages/pep517/in_process/_in_process.py", line 130, in get_requires_for_build_wheel
    return hook(config_settings)
  File "/usr/lib/python3.8/site-packages/setuptools/build_meta.py", line 177, in get_requires_for_build_wheel
    return self._get_build_requires(
  File "/usr/lib/python3.8/site-packages/setuptools/build_meta.py", line 159, in _get_build_requires
    self.run_setup()
  File "/usr/lib/python3.8/site-packages/setuptools/build_meta.py", line 281, in run_setup
    super(_BuildMetaLegacyBackend,
  File "/usr/lib/python3.8/site-packages/setuptools/build_meta.py", line 174, in run_setup
    exec(compile(code, __file__, 'exec'), locals())
  File "setup.py", line 4, in <module>
    raise Exception(
Exception:

    This file exists to set up the Github "Used By" counter properly and is not meant to be run.

    Please refer to ./packages/python/plotly/setup.py instead.


ERROR Backend subproccess exited when trying to invoke get_requires_for_build_wheel
```

nicolaskruchten:
Did you clone the repo and run `python setup.py` ? In order to install the `plotly` package you must `cd packages/python/plotly` first, as it suggest in the error message :)

kloczek:
Even with change directory to packages/python/plotly pep517 build fails.
```console
+ cd plotly.py-5.7.0
+ cd packages/python/plotly
+ /usr/bin/python3 -sBm build -w --no-isolation
* Getting dependencies for wheel...
/usr/lib/python3.8/site-packages/setuptools/dist.py:299: DistDeprecationWarning: use_2to3 is ignored.
  warnings.warn(f"{attr} is ignored.", DistDeprecationWarning)
running egg_info
running jsdeps
8.5.0
8.5.0
Installing build dependencies with npm.  This may take a while...
[...]

PR #3930, opened 2022-10-20
Compatibility with ipywidgets v8

abdelq:
Attempt at fixing https://github.com/plotly/plotly.py/issues/3686 while being compatible with the previous version.

## Code PR

- [x] I have added tests (if submitting a new feature or correcting a bug) or modified existing tests.
- [ ] For a new feature, I have added documentation examples in an existing or new tutorial notebook (please see the doc checklist as well).


vikramaditya91:
Could you please suggest how one can test this.
The standard install `python -m pip install 'git+https://github.com/abdelq/plotly.py.git@ipywidgets-8#egg=plotly'` did not work as it complained with this error: https://github.com/plotly/plotly.py/issues/3669

I thus did a `python -m pip install 'git+https://github.com/abdelq/plotly.py.git@ipywidgets-8#egg=plotly&subdirectory=packages/python/plotly'`
With this, I did not get the `type object 'DOMWidget' has no attribute '_ipython_display_'` error anymore and I was able to plot a very simple FigureWidget with:
```python

f = go.FigureWidget()
f
```

However, I do not think it could plot a FigureWidget with some scatter in it



I think I would also somehow need to install the changes that you made in javascript right?
Could you please let me know how I can do that? Thanks

abdelq:
This is what I did when building/installing the Jupyter extension:

```
cd packages/javascript/jupyterlab-plotly/
npm install
npm run build:dev
jupyter labextension install .
```

nicolaskruchten:
Thanks very much for this PR! I'll test it and try to do a release next week.

Question: is this PR intended to be compatible with v8 and also backwards-compatible with v7.6?

abdelq:
Correct. Should work for both.

abdelq:
I ran the example notebook and it works fine.

vikramaditya91:
I suppose the issue is about `plotly.FigureWidget` not working exactly as expected, and with your fix it does work even with ipywidgets set to 8. So, I would be happy to see this merged

vikramaditya91:
Hi. I would appreciate if this could be merged. 
Is there something we are waiting for before merging this?

nicolaskruchten:
Just need to do review and QA :)

packages/python/plotly/plotly/basedatatypes.py
```diff
@@ -819,18 +819,10 @@ def _repr_mimebundle_(self, include=None, exclude=None, validate=True, **kwargs)
 
         renderer_str = pio.renderers.default
         renderers = pio._renderers.renderers
-        renderer_names = renderers._validate_coerce_renderers(renderer_str)
```
nicolaskruchten, line 822: Can you please unpack what these changes are doing here? It's not immediately clear that the modified code has the same behaviour as the original...

nicolaskruchten:
Question re JupyterLab 2.x ... At the moment 2.x is supported with ipywidgets 7.x. Does this PR in principle change that? Have you been able to test it? It's a bit of a pain as you have to build the Javascript extension and then install it separately. I will try to do this on my end but I wanted to flag this as a concern in case you know ahead of time that it's not supported with this PR.

abdelq:
> Question re JupyterLab 2.x ... At the moment 2.x is supported with ipywidgets 7.x. Does this PR in principle change that? Have you been able to test it?

It looks like it works, but with console errors in the mix, namely:
`No provider for: jupyter.extensions.jupyterWidgetRegistry.`
and a bunch of:
`Exception opening new comm`
when running cells.

I think this is worth testing on your end too.

packages/python/plotly/plotly/basedatatypes.py
abdelq: One difference here is that `pio.renderers.default = None` will not return an exception and will just return the `repr`, which I think is the better behaviour.
Also, setting something like `pio.renderers.default = "plotly_mimetype+notebook_connected"` in the classic notebook will actually add the right snippet of JS (the whole activate stuff) to use the CDN, while the original version won't work.

nicolaskruchten:
Thanks! I need to do a release today but I will really try to get this PR merged and released next week.

aliallday:
@abdelq it looks like your modifications correctly allow a FigureWidget display but it doesn't seem like the widget redraws when the data changes. Here's a simple example to try

In one cell in jupyter lab run this:

f = go.FigureWidget()
f.add_scatter(y=[2, 1, 4, 3]);
display(f)

and then in another cell run this:
f.data[0]['y']=[1, 1, 1, 1]

The command in the second cell doesn't update the output of the first cell. I believe in earlier versions of ipywidgets it did.

abdelq:
After some debugging, `processMessage` in `RenderedPlotly` that is now used because of mime rendering is not calling `processLuminoMessage` in the view.

Alexboiboi:
Sorry to ask, but is there any progress going to be made soon? ;)

pmarcellino:
Just checking in here. I would be very excited to be able to use ipywidgets v8. Is this branch functional/going to be merged in?

I think this would also solve https://github.com/plotly/plotly.py/issues/3686

abdelq:
I'm using this as-is at work and it's working fine.

nvaytet:
Hi, checking in here also, would be great to see this fix released. Thanks!

maartenbreddels:
I tried this in notebook 6.4.8 and lab 3.2.9 with ipywidgets 7.7.2 to make sure it's backward compatible, and I see no issues!

packages/javascript/jupyterlab-plotly/src/Figure.ts
```diff
@@ -939,6 +942,21 @@ export class FigureView extends DOMWidgetView {
...
 
+  processPhosphorMessage(msg: any) {
+    this._processLuminoMessage(msg, super["processPhosphorMessage"]);
+
+    var that = this;
+    if (msg.type === "before-attach") {
+      window.addEventListener("resize", function () {
+        that.autosizeFigure();
+      });
+    }
```
maartenbreddels, line 953:
```suggestion
    var that = this;
    if (msg.type === "before-attach") {
      window.addEventListener("resize", function () {
        that.autosizeFigure();
      });
    }
```
I guess we can remove this?

abdelq: If we want compatibility, probably not? This is not done in `processLuminoMessage`, just in `processPhosphorMessage`

maartenbreddels: I see the same code in `_processLuminoMessage` right? (Modulo the code changes due to https://github.com/plotly/plotly.py/pull/3805)

Alexboiboi: Sorry to ask, but is this related also to 
- #3689

abdelq: I was referring to https://github.com/jupyter-widgets/ipywidgets/pull/3124, but I think with this current PR shrinking actually works fine (worth double checking).

maartenbreddels:
I don't wanna drag this PR too long, but for the future, using a strategy that @mariobuikhuizen used in https://github.com/bqplot/bqplot/pull/1531 we can get rid of the lumini/phosphor 'callbacks'.

packages/python/plotly/plotly/basewidget.py
```diff
@@ -733,12 +733,29 @@ def _handler_js2py_pointsCallback(self, change):
...
+            },
+        }
+
     def _ipython_display_(self):
         """
         Handle rich display of figures in ipython contexts
         """
-        # Override BaseFigure's display to make sure we display the widget version
-        widgets.DOMWidget._ipython_display_(self)
+        raise NotImplementedError  # Prefer _repr_mimebundle_
```
nicolaskruchten, line 758: Is this a backwards-compatible change?

packages/javascript/jupyterlab-plotly/src/Figure.ts
nicolaskruchten: Yes, let's please remove those lines, as they are indeed already executed in the `this._processLuminoMessage` call immediately above

nicolaskruchten: Otherwise we get two event listeners, right?

packages/python/plotly/plotly/basewidget.py
nicolaskruchten: OK, thanks!

nicolaskruchten:
OK, thanks everyone, especially @abdelq ! This'll get into the next release later this week :)

packages/javascript/jupyterlab-plotly/src/Figure.ts
Alexboiboi: The shrinking works fine for me too ;) 

But inside an ipywidgets container it stilll does not really resizes correctly. Maybe what @maartenbreddels suggested, applying the  [bqplot/bqplot#1531] strategy would solve the issues.

PR #3931, opened 2022-10-21
Docs oct

LiamConnors:

### DIFF
--- a/doc/python/legend.md
+++ b/doc/python/legend.md
@@ -6,7 +6,7 @@ jupyter:
       extension: .md
       format_name: markdown
       format_version: '1.3'
-      jupytext_version: 1.13.7
+      jupytext_version: 1.14.1
   kernelspec:
     display_name: Python 3 (ipykernel)
     language: python
@@ -20,7 +20,7 @@ jupyter:
     name: python
     nbconvert_exporter: python
     pygments_lexer: ipython3
-    version: 3.9.0
+    version: 3.8.0
   plotly:
     description: How to configure and style the legend in Plotly with Python.
     display_as: file_settings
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
Hey @LiamConnors, I've just merged your changes from `doc-prod` into `master` and also updated Plotly.js to 2.16.1 and approved the Percy snapshots, so if you want to merge the tip of `master` into this branch you'll have a passing build and an easier time :)

LiamConnors:
> Hey @LiamConnors, I've just merged your changes from `doc-prod` into `master` and also updated Plotly.js to 2.16.1 and approved the Percy snapshots, so if you want to merge the tip of `master` into this branch you'll have a passing build and an easier time :)

Thanks @nicolaskruchten. I've done that now.

doc/python/dumbbell-plots.md
```diff
@@ -0,0 +1,176 @@
...
+    order: 19
+    page_type: example_index
+    permalink: python/dumbbell-plots/
+    thumbnail: thumbnail/dumbbell-plot.jpg
+---
+
+## Basic Dumbbell Plot
+
+
+Dumbbell plots are useful for demonstrating change between two sets of data points, for example, the population change for a selection of countries for two different years
```
nicolaskruchten, line 39: This is a cool page, thanks for adding it!

Some feedback:

1. Neither plot has a legend, because you're using a single trace with hardcoded colors, so there's no way of knowing what green and blue mean. I would suggest a separate trace per year and then another one for the line but hide the line from the legend.
2. Along similar lines, it might be cool to have a third example (or a modified second example) with some lines that go the other way and that are colored red, say. Again I would make this a different trace so it can have a legend entry.

This is not blocking feedback, just something that's on my mind as I've always wanted a PX function to do this stuff :)

PR #3946, opened 2022-11-03
:recycle: avoid deepcopy of dict in validate_coerce

jvdd:
This PR avoids the `deepcopy` in the `_plotly_utils.basevalidators.BaseDataValidator` its `validate_coerce` method.

The reason I updated this code; I profiled plotly.py its graph construction time when adding large traces and +/- 70% of the time was spent in this `validate_coerce` method - of which +/- 33% of the time was spent at the `deepcopy` (see below).

```txt
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
  2626                                               def validate_coerce(self, v, skip_invalid=False, _validate=True):
  2627         2         26.0     13.0      0.0          from plotly.basedatatypes import BaseTraceType
  2628                                           
  2629                                                   # Import Histogram2dcontour, this is the deprecated name of the
  2630                                                   # Histogram2dContour trace.
  2631         2        321.0    160.5      0.1          from plotly.graph_objs import Histogram2dcontour
  2632                                           
  2633         2          5.0      2.5      0.0          if v is None:
  2634         1          2.0      2.0      0.0              v = []
  2635                                                   else:
  2636         1          4.0      4.0      0.0              if not isinstance(v, (list, tuple)):
  2637                                                           v = [v]
  2638                                           
  2639         1          3.0      3.0      0.0              res = []
  2640         1          1.0      1.0      0.0              invalid_els = []
  2641         2          5.0      2.5      0.0              for v_el in v:
  2642                                           
  2643         1          3.0      3.0      0.0                  if isinstance(v_el, BaseTraceType):
  2644                                                               # Clone input traces
  2645         1     138436.0 138436.0     44.4                      v_el = v_el.to_plotly_json()
  2646                                           
  2647         1          5.0      5.0      0.0                  if isinstance(v_el, dict):
  2648         1      99891.0  99891.0     32.0                      v_copy = deepcopy(v_el)
```

I believe it is unnecessary overhead to first create a dict (with `.to_plotly_json`) and then make a `deepcopy` of that dict to *possibly* `pop` one key (even when `v_el` is a dict and not a `BaseTraceType`, this still seems like unnecessary overhead to me).

Would appreciate any remarks!

~~Perhaps it might also be interesting to try avoid the `.to_plotly_json` call when it is a `BaseTraceTtype` - as this also takes a considerable amount of time (for large traces)?~~
-> @jonasvdd took a look at this and replaced the `.to_plotly_json` (which calls `deepcopy` on the `self._props`) with just `self._props`)

---

What I profiled:

```py
x = np.arange(5_000_000)
```
```py
%lprun -f plotly.graph_objects.Figure.add_trace plotly.graph_objects.Figure().add_trace(go.Scatter(x=x, y=x))
```

Before & after this PR:

image


packages/python/plotly/_plotly_utils/basevalidators.py
```diff
@@ -2645,30 +2640,28 @@ def validate_coerce(self, v, skip_invalid=False, _validate=True):
                     v_el = v_el.to_plotly_json()
 
                 if isinstance(v_el, dict):
-                    v_copy = deepcopy(v_el)
-
-                    if "type" in v_copy:
-                        trace_type = v_copy.pop("type")
-                    elif isinstance(v_el, Histogram2dcontour):
```
jvdd, line 2652: I believe this check never hit as in in line 2647 is checked whether `v_el` is a dict.

nicolaskruchten:
Thanks for this PR! Sorry I haven't had a chance to review it in detail yet, but the performance gains seem promising!

jonasvdd:
Hi @nicolaskruchten,

No worries, we both know how busy times can get!

Cheers,
Jonas & Jeroen

packages/python/plotly/_plotly_utils/basevalidators.py
```diff
@@ -2641,34 +2636,31 @@ def validate_coerce(self, v, skip_invalid=False, _validate=True):
             for v_el in v:
 
                 if isinstance(v_el, BaseTraceType):
-                    # Clone input traces
-                    v_el = v_el.to_plotly_json()
+                    v_el = v_el._props
```
alexcjohnson, line 2639:
```suggestion
                    if isinstance(v_el, Histogram2dcontour):
                        v_el = dict(type="histogram2dcontour", **v_el._props)
                    else:
                        v_el = v_el._props
                    
```
Seems like the check you called out [below](https://github.com/plotly/plotly.py/pull/3946/files#r1012659585) got broken at some point in the past, but it would work if we put it up here, right? And since this is a deprecation fallback I'm not worried about performance.

packages/python/plotly/_plotly_utils/basevalidators.py
```diff
@@ -2647,10 +2646,6 @@ def get_trace_class(self, trace_name):
     def validate_coerce(self, v, skip_invalid=False, _validate=True):
         from plotly.basedatatypes import BaseTraceType
```
alexcjohnson, line 2648:
```suggestion

        # Import Histogram2dcontour, this is the deprecated name of the
        # Histogram2dContour trace.
        from plotly.graph_objs import Histogram2dcontour
```
Looks like along with my previous comment we need to put this back in 😅

packages/python/plotly/_plotly_utils/basevalidators.py
```diff
@@ -2650,7 +2649,6 @@ def validate_coerce(self, v, skip_invalid=False, _validate=True):
         # Import Histogram2dcontour, this is the deprecated name of the
         # Histogram2dContour trace.
         from plotly.graph_objs import Histogram2dcontour
```
alexcjohnson, line 2651:
```suggestion
        from plotly.graph_objs import Histogram2dcontour

```

alexcjohnson (approved):
💃 Thanks very much @jvdd, and apologies for the delay!

PR #3972, opened 2022-11-30
Update splom.md

nelson123-lab:
The dimensions in the scatter matrix is not mentioned according to the data order. It gets very confused to see the scatter matrix in that format. The other scatter matrix are in the correct order as the dataset columns. t took me some time to realize that when the scatter matrix's style changed, the labels were mentioned in a new way.
The columns in the iris data set are in the following order: "sepal length," "sepal width," "petal length," and "petal width."


nicolaskruchten:
Thanks!

PR #3973, opened 2022-12-02
fix small typo in histograms doc

Kerybas:
I spotted a little typo where it was written `violin` instead of `histogram`.


nicolaskruchten:
Thanks!

PR #3995, opened 2022-12-23
Catch ImportError instead of Exception for FigureWidget

nvaytet:
`distutils` is deprecated. When using the `FigureWidget` in unit tests with `pytest` where warnings are promoted to errors, the warning becomes an error and the confusing error message is that we have to install `ipywidgets>=7.0.0`, even if `ipywidgets` is installed in the current environment.

This occurs because the `try/except` block is catching all exceptions, instead of just the `ImportError` which would be the error raised if `ipywidgets` was unavailable.

Alternatively, we could replace the `LooseVersion` with `packaging.version`, but I am not sure if this has wider implications.


nicolaskruchten:
Thanks!

PR #4063, opened 2023-02-09
Update ml-roc-pr.md

behloolsabir:
AUC for PR curve needs precision and recall as their input. fpr and tpr would be appropriate input for ROC curve.


LiamConnors (approved):
Thanks for the PR! Sorry for the delay in getting this merged

Issue #2192, opened 2020-02-13
CSS Named Colors for the Docs


emmanuelle:
Here is a list of colors matching nicely together, to be used for the tutorials.

MidnightBlue / LightSkyBlue
Aquamarine / LightSeaGreen
Orange / Gold
DarkBlue / LightPink
DarkBlue / RoyalBlue / Cyan
LightSalmon / PaleTurqoise / LightCyan
Gold / MediumTurqoise / LightGreen 
MediumSlateBlue / Lavender / DarkOrange
IndianRed / LightSalmon
RebeccaPurple / MediumPurple
LightSlateGray / LightSteelBlue
Crimson / PaleTurqoise

![image](https://user-images.githubusercontent.com/263366/74480146-6674d500-4e7e-11ea-8c2e-64056ca35be6.png)

gvwilson:
Hi - we are trying to tidy up the stale issues and PRs in Plotly's public repositories so that we can focus on things that are still important to our community. Since this one has been sitting for several years, I'm going to close it; if it is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. Thanks for your help - @gvwilson

PR #4083, opened 2023-02-25
Allowing hover_data and custom_data to pass string of column name instead of requiring a list.

### LINKED ISSUE #1683: Plotly Express strange hover_data ValueError
Scatter plotting with no problems, until I attempted to map a column of my data to hover_data.
Here's what my data looks like. I'm trying to map the `most_similar_kt` column to hover_data.

image

Here's the error message:
`ValueError: Element of value of 'hover_data' is not the name of a column in 'data_frame'. Expected one of ['x', 'y', 'label', 'freq', 'similarity', 'size', 'class', 'most_similar_kt'] but received: m`

And here's my code:
`fig = px.scatter(df, x='x',y='y', color='similarity', size='size', symbol='class',hover_name='label', hover_data='most_similar_kt', width=1500, height=1000, template='plotly_dark', render_mode='webgl')`

Apologies if this has already been addressed; I spent 10 minutes googling for errors and it doesn't appear anyone else had this issue.

If I `remove the hover_data assignment`, then the plot works:
image

Thanks for your assistance!

### LINKED ISSUE #2177: Better error message if passed in a column name instead of a list of column names into `custom_data`

lukefeilberg:
Howdy y'all! 

## Quick background
I often get tripped up when using the `hover_data` argument since nearly all arguments can be a single string of a column name except it (and `custom_data`). To make matters a bit worse the error message is quite confusing, example below.
```python
import plotly.express as px
df = px.data.tips()
px.scatter(
    df,
    x='total_bill',
    y='tip',
    color='sex',
    hover_data='day',
    title="Hover data won't work as string 😢"
)
```
This produces the following error
```bash
ValueError: Value of 'hover_data_0' is not the name of a column in 'data_frame'. Expected one of ['total_bill', 'tip', 'sex', 'smoker', 'day', 'time', 'size'] but received: d
```
which is pretty misleading since I did pass 'day'!

This has been brought up in the following issues:
- resolves #1683
- closes #2177
- closes https://github.com/plotly/plotly.py/pull/2185

Instead of making the error message more clear, I think we should just **allow a single string of a column name** to be consistent with nearly all other args.

---
## Source of problem
It appears the source of the problem is `hover_data` (and `custom_data`) are both in `array_attrables` and get cast as a list using `list(arg)` . And if `arg` is a string like `'my_col'` this turns it into `['m', 'y', '_', 'c', 'o', 'l']` which is undesirable. 

## Simple Solution
Thus my simple band-aid solution is checking if the field is in `["custom_data", "hover_data"]` and if `args[field] in args["data_frame"].columns` then we instead turn it into a list simply doing `args[field] = [args[field]]`. This suffices as shown in the screenshot below.
image

## Other ideas?
I'm open to ideas if folks prefer we do it in a different way (particularly the hardcoded list might be good define explicitly up top with an informative name) but I wanted to get a simple enough solution proposed that works.

I've also update the docs in this PR to reflect that this change would allow you to pass a string to both `hover_data` and `custom_data` -- I don't know if it's preferable or not for these to be in the same PR.

Happy to get any other feedback as well as this is my first attempt at an open source contribution. 😎

Thanks,
Luke Feilberg

---

### Documentation PR

- [x] I've [seen the `doc/README.md` file](https://github.com/plotly/plotly.py/blob/master/doc/README.md)
- [ ] This change runs in the current version of Plotly on PyPI and targets the `doc-prod` branch OR it targets the `master` branch
- [ ] If this PR modifies the first example in a page or adds a new one, it is a `px` example if at all possible
- [ ] Every new/modified example has a descriptive title and motivating sentence or paragraph
- [ ] Every new/modified example is independently runnable
- [ ] Every new/modified example is optimized for short line count	and focuses on the Plotly/visualization-related aspects of the example rather than the computation required to produce the data being visualized
- [ ] Meaningful/relatable datasets are used for all new examples instead of randomly-generated data where possible
- [ ] The random seed is set if using randomly-generated data in new/modified examples
- [ ] New/modified remote datasets are loaded from https://plotly.github.io/datasets and added to https://github.com/plotly/datasets
- [ ] Large computations are avoided in the new/modified examples in favour of loading remote datasets that represent the output of such computations
- [ ] Imports are `plotly.graph_objects as go` / `plotly.express as px` / `plotly.io as pio`
- [ ] Data frames are always called `df`
- [ ] `fig = <something>` call is high up in each new/modified example (either `px.<something>` or `make_subplots` or `go.Figure`)
- [ ] Liberal use is made of `fig.add_*` and `fig.update_*` rather than `go.Figure(data=..., layout=...)` in every new/modified example
- [ ] Specific adders and updaters like `fig.add_shape` and `fig.update_xaxes` are used instead of big `fig.update_layout` calls in every new/modified example
- [ ] `fig.show()` is at the end of each new/modified example
- [ ] `plotly.plot()` and `plotly.iplot()` are not used in any new/modified example
- [ ] Hex codes for colors are not used in any new/modified example in favour of [these nice ones](https://github.com/plotly/plotly.py/issues/2192)

### Code PR

  modified existing tests.


nicolaskruchten:
Thanks very much for this PR! I agree that the current behaviour is annoying enough that we should change it, and just accepting bare strings should work. I'll add some thoughts in the PR comments.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1295,11 +1295,17 @@ def build_dataframe(args, constructor):
...
-                dict(args[field])
-                if isinstance(args[field], dict)
-                else list(args[field])
-            )
+            if isinstance(args[field], dict):
+                args[field] = dict(args[field])
+            elif (
+                field in ["custom_data", "hover_data"]
+                and isinstance(args[field], str)
+                and "data_frame" in args.keys()
```
nicolaskruchten, line 1303: here I wouldn't check if the string is in the columns... there is other machinery elsewhere that does this kind of check and has better error messages etc, and simply checking here will disable this I think.

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1295,11 +1295,17 @@ def build_dataframe(args, constructor):
...
+                and args[field] in args["data_frame"].columns
+            ):
+                args[field] = [args[field]]
+            else:
+                list(args[field])
```
nicolaskruchten, line 1308: this needs to be `args[field] = list(args[field])` otherwise we're not actually using the copy we're making!

lukefeilberg: Ha good catch! I was failing tests on the main branch so didn't look too deep when I failed some on my branch 😅

packages/python/plotly/plotly/express/_core.py
lukefeilberg: 1. I took out the check for the string is in the columns which I agree was a good call. Now if I pass a single string that's not a column name I get the better error message `ValueError: Value of 'hover_data_0' is not the name of a column in 'data_frame'...`. 
2. All tests are now passing in `test_px` and I've added one new test and modified an existing one to account for both strings and dicts (and lists which wasn't present before).
    a. _Across all of test_optional I get `766 passed, 16 skipped, 79 warnings`, not too sure why some get skipped._
    b. _In test_core I fail 8 tests but all were `UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 291695` which I presume to be unrelated_
    c. I also needed to install various packages that weren't in the [optional-requirements.txt](https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/optional-requirements.txt) such as `statsmodels`, `scikit-image`, and `kaleido`. I believe there's others but I decided to focus on just test_optional and test_core. I may submit this as an additional issue and PR if that would be desirable!

image

nicolaskruchten: The tests look good to me on CI, thanks! To run the tests you need to install one of the testing requirements files from here https://github.com/plotly/plotly.py/tree/master/packages/python/plotly/test_requirements (which should probably be documented in the contributing guide!!)

lukefeilberg:
The build-doc is failing on make html. I made some real small changes to the documentation but just within strings. Can't tell if I've got some hard to spot or silly mistake or if this is just being funky. 🤔

nicolaskruchten:
`build-doc` is unfortunately flakey sometimes, and in this case it's not due to your PR :)

nicolaskruchten:
Thanks very much for this PR!

packages/python/plotly/plotly/express/_core.py
lukefeilberg: Ah thanks! It looks like the [contributing guide](https://github.com/plotly/plotly.py/blob/master/contributing.md) (assuming that's the right one to look at) only mentions optional-requirements. 
 
image

lukefeilberg:
Happy to contribute and I appreciate your help in the process!

Out of curiosity, what's the general cadence of releases or expected next release date? Not in any rush but will be excited to see my minor contribution in the wild 😎

nicolaskruchten:
We generally do minor releases timed with Plotly.js minor releases, or sooner if there are a batch of Python-driven features waiting for release. I'm not sure when the next Plotly.js minor is expected but likely within a few weeks.

PR #4091, opened 2023-03-03
keyword arguments comment in bio-manhantannplot documentation

thihanmoekyaw:
Added keyword arguments comment in the bio_dash source code so users can have more understanding of the changeable arguments.


LiamConnors:
@thihanmoekyaw let me know if you'd be interested in updating the PR to link out to the properties instead

PR #4147, opened 2023-04-05
Align an annotation with the center of a donut pie

yuji96:
## before
before

## after
after


LiamConnors (approved):
Thanks for the PR. This looks great! And sorry for the delay in getting it reviewed and merged.

PR #4148, opened 2023-04-05
Version changes for v5.14.1

LiamConnors:


alexcjohnson (approved):
💃 but please investigate the [instructions](https://github.com/plotly/plotly.py/blob/master/release.md#finalize-versions) that say to update `packages/python/plotly/README.md` - maybe that's what you did and it works, but in the PR it's the root README that got updated, because the package one is symlinked to the root.

PR #4161, opened 2023-04-15
Be more forgiving when anything with opening a webbrowser fails

bartbroere:
This (hopefully) fixes an issue I encountered with Starboard Notebook. I'll link the issue here.


jonmmease:
Hi @bartbroere, I'm not a maintainer of plotly projects any more, but FWIW these changes look reasonable to me!

packages/python/plotly/plotly/tests/test_io/test_renderers.py
```diff
@@ -381,3 +382,27 @@ def test_repr_mimebundle_mixed_renderer(fig1):
...
         {"application/vnd.plotly.v1+json", "text/html"}
     )
+
+
+def test_missing_webbrowser_module(fig1):
+    """
+    Assert that no errors occur if the webbrowser module is absent
+    """
+    removed_webbrowser_module = sys.modules["webbrowser"]
+    del sys.modules["webbrowser"]
```
alexcjohnson, line 392: Does this actually cause `import webbrowser` to fail? Let's add something to the test to validate this:
```py
with pytest.raises(ImportError):
    import webbrowser
```

And if (as I think may be the case) it's still importable, here's a solution that looks promising: https://stackoverflow.com/a/60229056/9188800

alexcjohnson:
@bartbroere the code change looks great, thanks! Just a couple of comments on the tests.

bartbroere:
Thanks for the commit bringing it up to date. I might have some time tonight (GMT+2) to finish this PR and apply the feedback.

packages/python/plotly/plotly/tests/test_io/test_renderers.py
bartbroere: This was helpful in pointing me in the right direction. Thanks! Landed on something similar buy slightly different in the end though.

bartbroere:
@alexcjohnson I think it's ready for merge now

alexcjohnson (approved):
💃 Thanks @bartbroere, looks great!

PR #4188, opened 2023-05-04
Update CI

alexcjohnson:
Bunch of CI config updates... the error in https://github.com/plotly/plotly.py/pull/4165 was actually fixed by *downgrading* Node from 18 to 16, somehow the Node 18 we're getting on CI can't handle [named unicode properties in regex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Unicode_property_escapes) even though Node 18 on my mac does that fine, but Node 16 on CI seems to work. I also upgraded to the `cimg/python` images, somewhere in there `black` noticed a couple more files, `pdftops` changed versions, and on Py36 now we need to explicitly install `packaging` (it was probably part of the base image before, and we weren't using a `venv` on CI, now we are).

@LiamConnors I'm getting an error still in [`build-doc`](https://app.circleci.com/pipelines/github/plotly/plotly.py/2340/workflows/3e23f610-e397-446b-a487-f59f8e74d24f/jobs/71293):
```
make: *** [Makefile:40: build/html/2019-07-03-static-image-export.html] Error 1
```
Do you know how to fix that? Otherwise all the tests are now passing on this branch.


alexcjohnson:
Hmph the Percy snapshots are also messed up... lots of images are way too narrow, others are unfinished, like we didn't wait for WebGL to render. That said some of the previous baselines also seem to have WebGL unrendered... and some of the narrow ones, the 375px image looks correct now but clipped previously, whereas the 1280px image looked correct previously. Did we actually intend to have Percy do both widths or is that a mistake? If we're only generating one width in the test we should ask Percy to only render at that same width, having Percy render multiple widths only makes sense for pages that can adapt purely via CSS.

alexcjohnson:
We're still losing WebGL from the Percy images, even when I try to use the trick from Dash of replacing the `<canvas>` elements with `` elements. Not sure if the wait I'm doing isn't long enough (or isn't working), or if we just aren't rendering WebGL at all in the updated container. Whatever, doesn't seem like this is actually that important a test, and it was already failing for some of the 3D WebGL plots. And I finally did succeed in dropping to a single width in Percy (and a single Browser).

the `build-doc` error was from pdf -> eps conversion failing, seems like I didn't get the right utils installed in that container. But we don't really need to run those code cells at all since they produce no output, so I disabled them by changing:
````
```python
fig.write_image("images/fig1.eps")
```
````
to:
````
~~~python
fig.write_image("images/fig1.eps")
~~~
````
Which should look the same on the docs site. It'll look different in Binder - rather than a separate code cell it'll just look like part of the markdown - but that's OK for now. These notebooks are broken in Binder anyway... something about our pinned numpy and Py3.10.

PR #4196, opened 2023-05-09
Safe JSON

alexcjohnson:
Fixing a possible XSS issue when our JSON is inserted into an HTML string. We could apply this just to `fig.to_html`, but to be even safer I'm applying it to all JSON serialization, in case users use a different mechanism to insert this into HTML.

Performance is a concern here, but in my testing except in really pathological situations these substitutions are a good deal faster than serialization even by `orjson`.

  modified existing tests.

### DIFF
--- a/packages/python/plotly/plotly/io/_json.py
+++ b/packages/python/plotly/plotly/io/_json.py
@@ -57,6 +57,23 @@ def coerce_to_strict(const):
         return const
 
 
+_swap = (
+    ("<", "\\u003c"),
+    (">", "\\u003e"),
+    ("/", "\\u002f"),
+    ("\u2028", "\\u2028"),
+    ("\u2029", "\\u2029"),
+)
+
+
+def _safe(json_str):
+    out = json_str
+    for unsafe_char, safe_char in _swap:
+        if unsafe_char in out:
+            out = out.replace(unsafe_char, safe_char)
+    return out
+
+
 def to_json_plotly(plotly_object, pretty=False, engine=None):
     """
     Convert a plotly/Dash object to a JSON string representation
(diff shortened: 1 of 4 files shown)


packages/python/plotly/plotly/io/_json.py
```diff
@@ -57,6 +57,25 @@ def coerce_to_strict(const):
...
 
 
+_swap_json = (
+    ("<", "\\u003c"),
+    (">", "\\u003e"),
+    ("/", "\\u002f"),
+)
+_swap_orjson = _swap_json + (
+    ("\u2028", "\\u2028"),
+    ("\u2029", "\\u2029"),
```
alexcjohnson, line 67: These are apparently JavaScript line terminator characters. The JS parser will see this end-of-line, see that the data structure is incomplete, and throw an error. AFAICT this isn't in itself a security issue, but it's a bug - if you ever wanted these characters in your figure for real it wouldn't work (when inserted in HTML).

The standard library `json` converts at least these unicode characters into escape sequences anyway, so we don't need to worry about them, but `orjson` sends them as unicode characters.

packages/python/plotly/plotly/io/_json.py
```diff
@@ -57,6 +57,25 @@ def coerce_to_strict(const):
...
+)
+
+
+def _safe(json_str, _swap):
+    out = json_str
+    for unsafe_char, safe_char in _swap:
+        if unsafe_char in out:
+            out = out.replace(unsafe_char, safe_char)
```
alexcjohnson, line 75: Testing first whether the character is present in the string is substantially faster than `.replace` when it finds even a single instance of the character - so given that much of the time you won't have some of these characters it's worthwhile including the `if` first. And looping over each character in turn is a lot faster than any solution I could find that only searches the string once, ie a regexp.

packages/python/plotly/test_requirements/requirements_39_optional.txt
```diff
@@ -19,3 +19,4 @@ matplotlib==2.2.3
 scikit-image==0.18.1
 psutil==5.7.0
 kaleido
+orjson==3.8.12
```
alexcjohnson, line 22: I added `orjson` only to py3.7 and py3.9 optional tests, so that we'd run a bunch of the other tests using `json`. Recent `orjson` doesn't support py3.6 anyway.

LiamConnors (approved):
:dancer:

Issue #4193, opened 2023-05-08
fail to build plotly python pkg due to use_2to3 is invalid


hubutui:
I failed to build plotly 5.14.1 with python 3.11, setuptools 67.7.0.
logs:
```text
/usr/lib/python3.11/site-packages/setuptools/config/setupcfg.py:293: _DeprecatedConfig: Deprecated config in `setup.cfg`
!!

        ********************************************************************************
        The license_file parameter is deprecated, use license_files instead.

        By 2023-Oct-30, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://setuptools.pypa.io/en/latest/https://setuptools.pypa.io/en/latest/userguide/declarative_config.html for details.
        ********************************************************************************

!!
  parsed = self.parsers.get(option_name, lambda x: x)(value)
/usr/lib/python3.11/site-packages/setuptools/dist.py:954: DistDeprecationWarning: use_2to3 is ignored.
  ep.load()(self, ep.name, value)
error in plotly setup command: use_2to3 is invalid.
```
quick fix, delete this line, https://github.com/plotly/plotly.py/blob/f006ae830365d6735e5d6fd3c3fb46dca10f539b/packages/python/plotly/setup.py#L498

alexcjohnson:
closed by #4206 - thanks @hubutui and @J815!

J815:
Welcome @alexcjohnson

PR #4206, opened 2023-05-12
fail to build plotly python pkg due to use_2to3 is invalid #4193 

J815:


alexcjohnson:
Thanks @J815 - please add a changelog entry and we'll merge 🎉

J815:
Fixed : removed use_2to3 =failed  in the setup tool which solves the issue of  installing setup tool (line 498)

PR #4210, opened 2023-05-15
Update plotly.js to 2.24.1 and add docs for new features

LiamConnors:
### Documentation PR

- [x] I've [seen the `doc/README.md` file](https://github.com/plotly/plotly.py/blob/master/doc/README.md)
- [x] This change runs in the current version of Plotly on PyPI and targets the `doc-prod` branch OR it targets the `master` branch
- [x] If this PR modifies the first example in a page or adds a new one, it is a `px` example if at all possible
- [x] Every new/modified example has a descriptive title and motivating sentence or paragraph
- [x] Every new/modified example is independently runnable
- [x] Every new/modified example is optimized for short line count	and focuses on the Plotly/visualization-related aspects of the example rather than the computation required to produce the data being visualized
- [x] Meaningful/relatable datasets are used for all new examples instead of randomly-generated data where possible
- [x] The random seed is set if using randomly-generated data in new/modified examples
- [x] New/modified remote datasets are loaded from https://plotly.github.io/datasets and added to https://github.com/plotly/datasets
- [x] Large computations are avoided in the new/modified examples in favour of loading remote datasets that represent the output of such computations
- [x] Imports are `plotly.graph_objects as go` / `plotly.express as px` / `plotly.io as pio`
- [x] Data frames are always called `df`
- [x] `fig = <something>` call is high up in each new/modified example (either `px.<something>` or `make_subplots` or `go.Figure`)
- [x] Liberal use is made of `fig.add_*` and `fig.update_*` rather than `go.Figure(data=..., layout=...)` in every new/modified example
- [x] Specific adders and updaters like `fig.add_shape` and `fig.update_xaxes` are used instead of big `fig.update_layout` calls in every new/modified example
- [x] `fig.show()` is at the end of each new/modified example
- [x] `plotly.plot()` and `plotly.iplot()` are not used in any new/modified example
- [x] Hex codes for colors are not used in any new/modified example in favour of [these nice ones](https://github.com/plotly/plotly.py/issues/2192)

## Code PR

- [x] I have read through the [contributing notes](https://github.com/plotly/plotly.py/blob/master/contributing.md) and understand the structure of the package. In particular, if my PR modifies code of `plotly.graph_objects`, my modifications concern the `codegen` files and not generated files.
- [ ] I have added tests (if submitting a new feature or correcting a bug) or
  modified existing tests.
- [x] For a new feature, I have added documentation examples in an existing or
  new tutorial notebook (please see the doc checklist as well).
[...]

PR #4217, opened 2023-05-19
Jupyterlab 4 support

### LINKED ISSUE #4066: Plotly figures are cropped in JupyterLab 4 (currently in preview)
That rule will not have the proper effect in the upcoming major JupyterLab:

https://github.com/plotly/plotly.py/blob/92ce5bce770dc5390bda0f44a9b6b033d4f2c39e/packages/javascript/jupyterlab-plotly/style/index.css#L26

You should set `height` directly. See https://github.com/jupyterlab/jupyterlab/issues/13227 for more details.

alexcjohnson:
Fixes #4066 - Jupyterlab 4 changes its layout so the `min-height` CSS rule no longer works.
Unfortunately I've been unable to test this locally as something went wrong when I tried to install Jupyterlab 4... will try again next week but I'm pushing this up in case anyone else is interested to push it forward earlier.


alexcjohnson:
Same test failure in the `release_build` workflow as in #4210, which also seems related to jlab4... still needs investigation which would sure be easier if I could get a working install locally 🙄

alexcjohnson:
Went with #4227 instead, which fixes the original issue but does nothing to update the build pipeline. We'll come back to that later when we're ready to update to the new packaging system.

PR #4221, opened 2023-05-30
Updated typo in `entrywidthmode` for legend

tanmaychimurkar:
The current version of the doc shows that the `entrywidthmode` can either be `pixels` (by default) or set to `fraction`. But fraction was written as `'fraction` without the trailing single quote. This merge request fixes that typo by changing it to `'fraction'`.


tanmaychimurkar:
I haven't figured out the reason for the `release build` and `build` to fail since the change in the doc content should not be interfering with the builds.

RyanDoesMath:
My pr just before yours also fixes a small typo and it fails to build. Don't think its an issue.

tanmaychimurkar:
Ah yes, I just saw that. Thanks for the clarification, I was not sure what went wrong from my side.

LiamConnors:
@tanmaychimurkar there was an issue with the tests when you originally created the PR.
That issue is resolved now. Thanks for the PR!

PR #4227, opened 2023-06-02
Jupyterlab v4 take2

alexcjohnson:
Putting #4217 on hold for now since I still have not managed to update any local env I create to Jupyterlab 4 successfully, let's go a different route: take from #4217 the fix that I _think_ will work for jupyterlab 4 https://github.com/plotly/plotly.py/commit/3b4b324b23f3819f665003c3432ad3c6d7fa9136 and the little setup fix https://github.com/plotly/plotly.py/commit/411d66a259aa2f1ba8b8b73529944b571fab9859, but pin our build process to jupyterlab 3 for now https://github.com/plotly/plotly.py/commit/8df887d6e3ceb338a2a65e3b02336bc4e64171e6. Turns out a full upgrade of the process to jupyterlab 4 entails also upgrading the build pipeline from Python 3.7 to 3.8 or later, so a bigger and riskier project than I want right now, and I can't find good information about whether the upgraded version of the extension will work with older versions of jupyterlab anyway.

Fortunately even for jupyterlab 3 users this change has a benefit: we previously set 360px as the minimum output height, but with this change that's just the default, if you set a smaller height that's what you'll get.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -371,7 +371,7 @@ jobs:
           name: Create conda environment
           command: |
             conda create -n env --yes python=3.9 conda-build conda-verify
-            conda install -n env -c conda-forge jupyterlab nodejs=16
+            conda install -n env -c conda-forge jupyterlab=3 nodejs=16
             conda init bash
             mkdir output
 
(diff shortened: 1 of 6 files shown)


WenjieZ:
Hi, has this bugfix been integrated in a release? I encountered the same problem in 5.15.0.

alexcjohnson:
@WenjieZ yes this fix is part of 5.15.0. Jupyter is notorious for having multiple versions installed, you can confirm this is actually being rendered by 5.15 as it will have plotly.js v2.24.1 if you hover on the Plotly logo on a graph. (5.16 released last week has plotly.js v2.25.2)

PR #4231, opened 2023-06-06
Update animations.md

jegrami:
Change two words in paragraph 2 of Animated figures with Plolty Express


LiamConnors:
Thanks for the PR @jegrami!

PR #4244, opened 2023-06-12
Support dataframe interchange protocol

anmyachev:
This pull request continues the work that began in https://github.com/plotly/plotly.py/pull/3387.


packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
```diff
@@ -233,6 +246,24 @@ def test_build_df_with_index():
     assert_frame_equal(tips.reset_index()[out["data_frame"].columns], out["data_frame"])
 
 
+def test_build_df_using_interchange_protocol_mock(add_interchange_module_for_old_pandas):
```
anmyachev, line 249: Using a mock test, "plotly" do not need to have a test for every library that supports this protocol.

nicolaskruchten:
This looks great, thank you! So just so I'm clear on the user-facing impact here: this should make it possible for PX to accept Vaex and Modin dataframes (which don't have `.to_pandas()`) and will use this new (better? faster?) pathway for other non-Pandas systems that happen to also have a `.to_pandas()` like... CuDF and Polars? So hard to keep track of all the flavours 🍨

anmyachev:
> This looks great, thank you! So just so I'm clear on the user-facing impact here: this should make it possible for PX to accept Vaex and Modin dataframes (which don't have `.to_pandas()`) and will use this new (better? faster?) pathway for other non-Pandas systems that happen to also have a `.to_pandas()` like... CuDF and Polars? So hard to keep track of all the flavours 🍨

You're right. The new path will be used even if the library has `to_pandas` method. I believe this is the preferred path, as this protocol aims to become a widely accepted standard, which should be accompanied by good support. From a performance point of view, it seems most likely that this way could be more performant (for example, in the protocol [itself](https://data-apis.org/dataframe-protocol/latest/design_requirements.html#design-requirements) there is the following mention: `Must be zero-copy wherever possible`), at least as good as a simple `to_pandas` call.

anmyachev:
@nicolaskruchten CI is green, may I know what do you think about merging it?

nicolaskruchten:
I'll endorse this PR but I'll defer to @alexcjohnson and @LiamConnors for timing on when to merge it, how to document it etc :)

nicolaskruchten (approved):
💃  This seems good to me in principle although I haven't tested it manually and I don't have a strong sense of what the list is of which non-Pandas dataframes will go through the new codepath and which will be zero-copy, so it's not super-clear to me how to document this.

This PR also needs a changelog entry (which should probably contain some of the answers to the questions above)

anmyachev:
Hi @alexcjohnson! May I have your opinion on this pull request?

MarcoGorelli:
I haven't looked enough into what you do with the DataFrame objects, but I just wanted to note that if `df` is a polars DataFrame, then:
- `pd.api.interchange.from_pandas(df)` will probably make some copy (currently, the result doesn't use pyarrow)
- `df.to_pandas()` should typically be zero-copy

So it may be more user-friendly to first try `to_pandas`, and then fallback to the interchange protocl? (cc @ritchie46 in case I got this wrong)

ritchie46:
> df.to_pandas() should typically be zero-copy

`df.to_pandas(use_pyarrow_extension_array=True)` is zero copy yes.

alexcjohnson:
This looks great, thanks @anmyachev, and thanks @nicolaskruchten for reviewing. I'd love to see a test or two that actually directly use polars and vaex dataframes in px. Compliance with a standard protocol is a great way to achieve this, and in principle if these other packages don't support this it's their problem to solve not ours, but (a) fundamentally what our users care about is that they can use their preferred dataframe manager, and (b) having tests will allow us over time to deepen this support to improve performance.

That and a changelog entry 😎

packages/python/plotly/test_requirements/requirements_39_pandas_2_optional.txt
```diff
@@ -1,6 +1,6 @@
 requests==2.25.1
 tenacity==6.2.0
-pandas==2.0.1
+pandas==2.0.2
```
anmyachev, line 3: Without this change, the protocol cannot be used in tests (for example, vaex test).

anmyachev:
> This looks great, thanks @anmyachev, and thanks @nicolaskruchten for reviewing. I'd love to see a test or two that actually directly use polars and vaex dataframes in px. Compliance with a standard protocol is a great way to achieve this, and in principle if these other packages don't support this it's their problem to solve not ours, but (a) fundamentally what our users care about is that they can use their preferred dataframe manager, and (b) having tests will allow us over time to deepen this support to improve performance.
> 
> That and a changelog entry 😎

@alexcjohnson thanks for the answer! I made the changes you mentioned. Ready for review.

anmyachev:
> Could there be a polars test for this as well? It would be great for us to know that plotly always stays working on both sides. :)

@ritchie46 Sure, I can add. Does polars have a function like `from_pandas` in vaex? (for example, `vaex.from_pandas(iris_pandas` call in the test)

ritchie46:
Yes, there is `pl.from_pandas`. :+1:

anmyachev:
@alexcjohnson friendly ping :)

LiamConnors:
I tried this out by creating a new venv and installing `pandas` and `polars`. When I run an app I get the following traceback asking for `pyarrow>=11.0.0`
```
  File "app.py", line 19, in <module>
    fig = px.scatter(df, x="date", y="float")
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/plotly/express/_chart_types.py", line 66, in scatter
    return make_figure(args=locals(), constructor=go.Scatter)
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/plotly/express/_core.py", line 2012, in make_figure
    args = build_dataframe(args, constructor)
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/plotly/express/_core.py", line 1317, in build_dataframe
    df_pandas = pandas.api.interchange.from_dataframe(df_not_pandas)
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/pandas/core/interchange/from_dataframe.py", line 54, in from_dataframe
    return _from_dataframe(df.__dataframe__(allow_copy=allow_copy))
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/polars/dataframe/frame.py", line 1212, in __dataframe__
    raise ImportError(
ImportError: pyarrow>=11.0.0 is required for converting a Polars dataframe to a dataframe interchange object.
```

nicolaskruchten:
Hmm we definitely don't want a hard dependency on `pyarrow` ... can we check to see if it's present before using this branch if it's routinely-required?

anmyachev:
@LiamConnors thanks for the feedback.

> Hmm we definitely don't want a hard dependency on pyarrow ... can we check to see if it's present before using this branch if it's routinely-required?

I think right now I can rewrite the logic as follows (so that when the version of `plotly` is updated, the code does not stop working for users):
```python
try:
    pandas.api.interchange.from_dataframe(df_not_pandas)
except (ImportError, NotImplemetedError):
    df_not_pandas.to_pandas()
```

> why not swap the branches, and try to_pandas first?

@MarcoGorelli I guess we should try the more standardized and basic way first than specialized. In the case when the dataframe library has both options implemented, then the execution of the code will not reach a potentially better way.

anmyachev:
@nicolaskruchten CI build failed for another reason. Could you manually restart it if possible?

alexcjohnson:
@LiamConnors thanks for testing this out - would you try again, both without `pyarrow` and then with it installed? AFAICT the logic is solid at this point, so if you're happy with how it works I'll be happy to approve and merge. Thanks for the iterations @anmyachev 🙇

packages/python/plotly/test_requirements/requirements_39_pandas_2_optional.txt
```diff
@@ -19,3 +19,5 @@ matplotlib==2.2.3
 scikit-image==0.18.1
 psutil==5.7.0
 kaleido
+vaex
```
YarShev, line 22: Why do you not add modin here?

anmyachev: `test_build_df_using_interchange_protocol_mock` test should be enough for modin. On the other hand, I do not want to expand the list of dependencies, there are already enough of them.

LiamConnors:
I guess this was already there, because I now get this:

```
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/plotly/express/_core.py", line 1318, in build_dataframe
    df_pandas = pandas.api.interchange.from_dataframe(df_not_pandas)
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/pandas/core/interchange/from_dataframe.py", line 54, in from_dataframe
    return _from_dataframe(df.__dataframe__(allow_copy=allow_copy))
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/polars/dataframe/frame.py", line 1212, in __dataframe__
    raise ImportError(
ImportError: pyarrow>=11.0.0 is required for converting a Polars dataframe to a dataframe interchange object.

During handling of the above exception, another exception occurred:

  File "app.py", line 19, in <module>
    fig = px.scatter(df, x="date", y="float")
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/plotly/express/_chart_types.py", line 66, in scatter
    return make_figure(args=locals(), constructor=go.Scatter)
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/plotly/express/_core.py", line 2022, in make_figure
    args = build_dataframe(args, constructor)
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/plotly/express/_core.py", line 1327, in build_dataframe
    df_pandas = df_not_pandas.to_pandas()
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/polars/dataframe/frame.py", line 2076, in to_pandas
    record_batches = self._df.to_pandas()
ModuleNotFoundError: No module named 'pyarrow'
```

anmyachev:
> I guess this was already there, because I now get this:
> 
> ```
> Traceback (most recent call last):
>   File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/plotly/express/_core.py", line 1318, in build_dataframe
>     df_pandas = pandas.api.interchange.from_dataframe(df_not_pandas)
>   File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/pandas/core/interchange/from_dataframe.py", line 54, in from_dataframe
>     return _from_dataframe(df.__dataframe__(allow_copy=allow_copy))
>   File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/polars/dataframe/frame.py", line 1212, in __dataframe__
>     raise ImportError(
> ImportError: pyarrow>=11.0.0 is required for converting a Polars dataframe to a dataframe interchange object.
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "app.py", line 19, in <module>
>     fig = px.scatter(df, x="date", y="float")
>   File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/plotly/express/_chart_types.py", line 66, in scatter
>     return make_figure(args=locals(), constructor=go.Scatter)
>   File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/plotly/express/_core.py", line 2022, in make_figure
>     args = build_dataframe(args, constructor)
>   File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/plotly/express/_core.py", line 1327, in build_dataframe
>     df_pandas = df_not_pandas.to_pandas()
>   File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/polars/dataframe/frame.py", line 2076, in to_pandas
>     record_batches = self._df.to_pandas()
> ModuleNotFoundError: No module named 'pyarrow'
> ```

Do I understand correctly that this should not stop the merge of this pull request?

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1307,7 +1308,25 @@ def build_dataframe(args, constructor):
...
-        if hasattr(args["data_frame"], "to_pandas"):
+        if hasattr(args["data_frame"], "__dataframe__") and version.parse(
+            pd.__version__
+        ) >= version.parse("2.0.2"):
+            import pandas.api.interchange
+
+            df_not_pandas = args["data_frame"]
+            try:
+                df_pandas = pandas.api.interchange.from_dataframe(df_not_pandas)
+            except (ImportError, NotImplementedError) as exc:
```
MarcoGorelli, line 1319: I think you'll want `ModuleNotFoundError`, else you'll get the reported error if `pyarrow` isn't installed

anmyachev: `ModuleNotFoundError` comes from `to_pandas` call. The same behavior should already be on the master branch.

Stack from error above:
```bash
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/plotly/express/_core.py", line 1327, in build_dataframe
    df_pandas = df_not_pandas.to_pandas()
  File "/Users/liamconnors/Desktop/pltest/venv/lib/python3.8/site-packages/polars/dataframe/frame.py", line 2076, in to_pandas
    record_batches = self._df.to_pandas()
ModuleNotFoundError: No module named 'pyarrow'
```

MarcoGorelli: ah you're right - it doesn't make any difference then, thanks

LiamConnors (approved):
Tested with `pyarrow` installed and didn't encounter any other issues.

packages/python/plotly/plotly/express/_core.py
alexcjohnson: I think [we're OK](https://stackoverflow.com/questions/62854761/python3-8-whats-the-difference-between-importerror-and-modulenotfounderror):
```py
>>> issubclass(ModuleNotFoundError, ImportError)
True
```

alexcjohnson (approved):
> Do I understand correctly that this should not stop the merge of this pull request?

Yep, at that point we're out of fallbacks, all we can do is allow that error to propagate, so the user is alerted to install `pyarrow` if they want to use this feature.

💃 Merging, we'll likely release this with the next plotly.js, which I'm guessing will be in a week or so. Thanks again @anmyachev, and everyone else for your comments.

anmyachev:
Thanks for the review!

PR #4282, opened 2023-07-17
Update box-plots.md

patelvyom:
- There is no need for passing dummy variable `y` when stats have already been computed
- `upperfence` values don't match the resulting figure.
- It could be simplified even further by getting rid of the extra call to `add_trace` but I've kept it to follow the documentation styling.
- It might be worth it to separate the example in two cases: (i) with only `q1`, `median`, `q3` provided, (ii) with all values provided.


Coding-with-Adam:
This PR has been mentioned on discourse to test the plugin.
https://community.plotly.com/t/testing-testing-the-new-github-plugin/79843

LiamConnors (approved):
Thanks for the PR @patelvyom. Sorry for the delay in getting this merged.

PR #4286, opened 2023-07-19
only interchange necessary columns

MarcoGorelli:
Trying to address this comment: https://github.com/plotly/plotly.py/pull/3901#issuecomment-1591689292

If it's not a wide plot, then only interchange the columns which are needed

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1396,7 +1396,7 @@ def build_dataframe(args, constructor):
 
     if needs_interchanging:
         try:
-            if wide_mode:
+            if wide_mode or not hasattr(args["data_frame"], "select_columns_by_name"):
                 args["data_frame"] = pd.api.interchange.from_dataframe(
                     args["data_frame"]
                 )
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1018,7 +1018,7 @@ def _get_reserved_col_names(args):
     return reserved_names
 
 
-def _is_col_list(df_input, arg):
+def _is_col_list(columns, arg):
     """Returns True if arg looks like it's a list of columns or references to columns
     in df_input, and False otherwise (in which case it's assumed to be a single column
     or reference to a column).
@@ -1033,7 +1033,7 @@ def _is_col_list(df_input, arg):
         return False  # not iterable
     for c in arg:
         if isinstance(c, str) or isinstance(c, int):
-            if df_input is None or c not in df_input.columns:
+            if columns is None or c not in columns:
                 return False
         else:
             try:
@@ -1059,8 +1059,8 @@ def _isinstance_listlike(x):
         return True
 
 
-def _escape_col_name(df_input, col_name, extra):
-    while df_input is not None and (col_name in df_input.columns or col_name in extra):
+def _escape_col_name(columns, col_name, extra):
+    while columns is not None and (col_name in columns or col_name in extra):
         col_name = "_" + col_name
     return col_name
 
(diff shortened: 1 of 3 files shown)


MarcoGorelli:
🤔 bit confused by the CI failure, it fails on "install chrome driver"?

alexcjohnson:
@LiamConnors would you mind making another "pin chrome" PR in this repo? (Not in this PR but so this and other PRs can update and succeed again!)

nicolaskruchten:
Wow, I'm excited that someone is biting the bullet on this one, thank you!

Would be nice for this to be reused for the jankier `to_pandas()` path as well if possible too, for the shorter term :)

LiamConnors:
> @LiamConnors would you mind making another "pin chrome" PR in this repo? (Not in this PR but so this and other PRs can update and succeed again!)

Opened a PR here to fix it: https://github.com/plotly/plotly.py/pull/4288 @alexcjohnson

MarcoGorelli:
> Would be nice for this to be reused for the jankier to_pandas() path as well if possible too, for the shorter term :)

Sure, but there's no guarantee of what the API to do that would be, right? (I might be missing something though, sorry)

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1307,37 +1307,36 @@ def build_dataframe(args, constructor):
...
-                # should try a different implementation, if available. For example:
-                # def __dataframe__(self, ...):
-                #   if not some_condition:
-                #     self.to_pandas(...)
-                if not hasattr(df_not_pandas, "to_pandas"):
-                    raise exc
-                df_pandas = df_not_pandas.to_pandas()
-            args["data_frame"] = df_pandas
+            args["data_frame"] = df_not_pandas.__dataframe__()
+            columns = args["data_frame"].column_names()
```
alexcjohnson, line 1319: Can we be 100% sure anything returned by `__dataframe__()` will have `column_names` and `select_columns_by_name` methods? If there's any chance an object will come in with either of these missing we should fall back on interchanging the whole thing up front.

MarcoGorelli: yup! you can check the spec here: https://data-apis.org/dataframe-protocol/latest/API.html

alexcjohnson: Oh I know they're in the spec, but I also know not everyone follows a spec to the letter 😉

MarcoGorelli: this can help highlight shortcomings in their implementation then 😉 I tried it out with polars and it works fine there

alexcjohnson: I guess we'll know how to respond when we see:
`AttributeError: 'MyDataFrame' object has no attribute 'select_columns_by_name'`
And in principle you're right that it's not our problem, but we'll be the ones responding to the issue and having to tell our users "don't use this dataframe directly until they fix it." Whereas if we caught this case explicitly we could emit a warning like "This dataframe only partially implements the dataframe interchange protocol. Falling back on a slower full-copy algorithm" so it wouldn't affect usage in px, only performance, and it would be clear where the issue needs to be raised.

MarcoGorelli: thanks for explaining - OK I've added a condition so it'll only use `select_columns_by_name` if that attribute is present

alexcjohnson: Nice. I took a look at also adding a fallback for missing `column_names` and that would be pretty awkward... but if someone has a partial implementation of the protocol presumably `column_names` is an easy piece so would get included early, whereas `select_columns_by_name` could be trickier. So let's leave it as you have it now. Thanks!

alexcjohnson (approved):
💃 Great work @MarcoGorelli, lovely tests!

packages/python/plotly/plotly/express/_core.py
MarcoGorelli: yeah if they don't have `column_names` then `from_dataframe` wouldn't work either, as it uses that internally

https://github.com/pandas-dev/pandas/blob/92792ec063031ae41443dabeb9d12f8daaac3ef1/pandas/core/interchange/from_dataframe.py#L112

Issue #4281, opened 2023-07-14
`to_plotly_json` does not handle `np.array` properly


ChiQiao:
Code to reproduce the issue:
```python
import json
import plotly.graph_objects as go
import numpy as np

fig = go.Figure(go.Scatter(x=np.array([0, 1]), y=np.array([0, 1])))
json.dumps(fig.to_plotly_json())  # TypeError: Object of type ndarray is not JSON serializable
```

itsluketwist:
@ChiQiao - Hey! Whilst the `to_plotly_json` method doesn't provide a `dict` that is immediately `json` compatible, if you use the `PlotlyJSONEncoder` as the `cls` arg when creating the `json` then everything should work fine. 😄 

Fixed code snippet:

```
import json
import plotly.graph_objects as go
import numpy as np

from plotly.utils import PlotlyJSONEncoder  # need this import

fig = go.Figure(go.Scatter(x=np.array([0, 1]), y=np.array([0, 1])))
json.dumps(fig.to_plotly_json(), cls=PlotlyJSONEncoder)  # encoder used as arg here, handles the numpy data  (amongst others)
```

Hopefully this fixes your issue!

ChiQiao:
@itsluketwist Thanks for the information. Do you think this is the expected behavior (i.e., fulfilling the purpose of this method as `Convert figure to a JSON representation as a Python dict`)? If so, please feel free to close this issue.

itsluketwist:
@ChiQiao - agreed, this experience could be better. I've made a PR to hopefully make it so, and discuss the json usage with project maintainers. 😄

itsluketwist:
@ChiQiao - I don't know the reason for the multiple very similar functions here, especially when there is also a `to_dict` method that is the same as `to_plotly_json`. 😅 

You're right that the `to_json` method should also fix your problem! 😄

alexcjohnson:
Closed by #4301 - thanks @itsluketwist 🎉

PR #4301, opened 2023-07-28
Update to_plotly_json docstring, and add new to_json_string helper method

itsluketwist:
This PR aims to resolve some pain points raised in issue [4281](https://github.com/plotly/plotly.py/issues/4281), regarding JSON serialization.

The `to_plotly_json` method simply returns a dictionary, it is not necessarily JSON compatible (it can have invalid data-types included), plotly does provide an encoder - but this isn't immediately obvious.

I've updated the docstring of `to_plotly_json` to include a warning about the above, and point at the available encoder, as well as adding a simple utility method that performs the JSON serialization to `BasePlotlyType`, to keep in line with the `BaseFigure` interface.


packages/python/plotly/plotly/basedatatypes.py
```diff
@@ -5597,12 +5601,24 @@ def to_plotly_json(self):
...
 
+    def to_json(self):
+        """
+        Convert to a JSON string representation.
+
+        Returns
+        -------
+        str
+        """
+        return json.dumps(self.to_plotly_json(), cls=PlotlyJSONEncoder)
```
alexcjohnson, line 5620: There's a better way to do this, using [`plotly.io.to_json`](https://plotly.com/python-api-reference/generated/plotly.io.html#plotly.io.to_json). The main difference is this will use `orjson` if it's installed, which can be much faster for large graphs.

itsluketwist: Ahhh cool, thanks! Updated 😄

packages/python/plotly/plotly/basedatatypes.py
```diff
@@ -3318,6 +3318,8 @@ def to_plotly_json(self):
         """
         Convert figure to a JSON representation as a Python dict
 
+        Note: May include some JSON-invalid data types, use the `PlotlyJSONEncoder` util when encoding.
```
alexcjohnson, line 3321:
```suggestion
        Note: May include some JSON-invalid data types, use the `PlotlyJSONEncoder` util
        or the `to_json` method to encode to a string.
```
(and similar in the other one)

alexcjohnson (approved):
💃 This looks great @itsluketwist - other than my one comment above, just add a changelog entry and we'll be good to merge 🎉

packages/python/plotly/plotly/basedatatypes.py
itsluketwist: Fixed in commit [43d0432](https://github.com/plotly/plotly.py/pull/4301/commits/43d0432805c6b4e406b7e68e688d2f0df3c6c304)

itsluketwist:
@alexcjohnson - all fixed up, should be ready to go! 😄 

Issue https://github.com/plotly/plotly.py/issues/4281 can also be closed too.

Issue #2812, opened 2020-10-07
plotly.express line documentation error in line_shape


abogaard:
Hello, I am trying to pass `line_shape='spline`' as documented (https://plotly.com/python-api-reference/generated/plotly.express.line.html#plotly.express.line) but am getting the following error:

` Invalid value of type 'builtins.str' received for the 'shape' property of scattergl.line
        Received value: 'spline'

    The 'shape' property is an enumeration that may be specified as:
      - One of the following enumeration values:
            ['linear', 'hv', 'vh', 'hvh', 'vhv']`

ronak1009:
I am also getting the error after migrating to Python 3.9 and plotly==5.10.0
Error comes for
```python
fig = plx.line(df,x="x", y="y", color="z", 
 labels=labels, 
 width=224, 
 height=245,
 line_shape= "spline" 
)
```
If I use line_shape=="linear" or remove the parameter from constructor, then the generated svg contains background black:

![image](https://user-images.githubusercontent.com/12443143/205017899-d1de210e-d69a-4dc4-9e57-e5af542620fc.png)

zupdaz:
Getting same error when I load when I load my df from multiple csv.
```
fig = px.line(CAM_table_unpivot, x='Size_class', y='Value', color='Comment 2', line_shape='spline')
fig.update_xaxes(type="log")
```

gvwilson:
Hi - we are trying to tidy up the stale issues and PRs in Plotly's public repositories so that we can focus on things that are still important to our community. Since this one has been sitting for a while, I'm going to close it; if it is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. Alternatively, if it's a request for tech support, please post in our community forum. Thank you - @gvwilson

PR #4305, opened 2023-08-01
px.line: don't use webgl if line_shape="spline" is given

bwest2397:
The documentation for [documentation for plotly.express.line](https://plotly.com/python-api-reference/generated/plotly.express.line.html#plotly.express.line) seems to imply that when calling `px.line`, `line_shape` can be given as `"linear"` or `"spline"` without any other limitations. However, if one calls it with `render_mode="auto"` (or without providing `render_mode`, as `"auto"` is its default value) and with a large enough dataframe (over 1000 total plotted entries), `line_shape="spline"` causes an error. This error is caused by the webgl render mode (`go.Scattergl`) being used by default with large dataframes instead of the svg render mode (`go.Scatter`). This PR works around this issue by not defaulting to the webgl render mode is `line_shape="spline"` is given, as the documentation implies that this _should_ work. It's worth noting that this will likely incur a performance penalty, but I think that anyone rendering that much data is willing to accept the hit :)

This should address https://github.com/plotly/plotly.py/issues/2812

## Code PR

  modified existing tests.


packages/python/plotly/plotly/express/_core.py
```diff
@@ -812,6 +812,7 @@ def make_trace_spec(args, constructor, attrs, trace_patch):
             or (
                 args["render_mode"] == "auto"
                 and len(args["data_frame"]) > 1000
+                and args["line_shape"] != "spline"
```
alexcjohnson, line 815:
```suggestion
                and args.get("line_shape") != "spline"
```
Because not all px functions that have `render_mode` also have `line_shape`.

bwest2397: Oh good catch, thanks!

alexcjohnson:
@bwest2397 thanks for this, looks great and apologies for letting it sit so long. Would you add a line in the changelog under unreleased/fixed? Then we'll merge!

bwest2397:
> @bwest2397 thanks for this, looks great and apologies for letting it sit so long. Would you add a line in the changelog under unreleased/fixed? Then we'll merge!

Done. Assuming you agree with the content of the changelog message, feel free to merge when you're ready.

alexcjohnson (approved):
💃

PR #4324, opened 2023-08-13
Consider necessary columns from complex arguments when interchanging dataframes

guyrosin:
This PR fixes a regression from #4286, where only plain string arguments were considered as necessary columns when interchanging dataframes. This PR considers complex arguments as well (such as `hover_data`).

For example, running the code below resulted in `ValueError: Value of 'hover_data_0' is not the name of a column in 'data_frame'. Expected one of ['petal_length', 'sepal_length'] but received: species`

```python
import numpy as np
import plotly.express as px
import polars as pl

df = px.data.iris()
px.scatter(
    pl.from_pandas(df),
    x="petal_length",
    y="sepal_length",
    hover_data=["species"],  # <-- this was unsupported
)
```

  modified existing tests.

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1419,9 +1419,17 @@ def build_dataframe(args, constructor):
             else:
                 # Save precious resources by only interchanging columns that are
                 # actually going to be plotted.
-                columns = [
+                necessary_columns = [
                     i for i in args.values() if isinstance(i, str) and i in columns
                 ]
+                for field in args:
+                    if field in array_attrables and isinstance(
+                        args[field], (list, dict)
+                    ):
+                        necessary_columns.extend(
+                            [i for i in args[field] if i in columns]
+                        )
+                columns = list(dict.fromkeys(necessary_columns))
                 args["data_frame"] = pd.api.interchange.from_dataframe(
                     args["data_frame"].select_columns_by_name(columns)
                 )
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1423,9 +1423,7 @@ def build_dataframe(args, constructor):
                     i for i in args.values() if isinstance(i, str) and i in columns
                 ]
                 for field in args:
-                    if field in array_attrables and isinstance(
-                        args[field], (list, dict)
-                    ):
+                    if args[field] is not None and field in array_attrables:
                         necessary_columns.extend(
                             [i for i in args[field] if i in columns]
                         )
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1419,15 +1419,13 @@ def build_dataframe(args, constructor):
             else:
                 # Save precious resources by only interchanging columns that are
                 # actually going to be plotted.
-                necessary_columns = [
+                necessary_columns = {
                     i for i in args.values() if isinstance(i, str) and i in columns
-                ]
+                }
                 for field in args:
                     if args[field] is not None and field in array_attrables:
-                        necessary_columns.extend(
-                            [i for i in args[field] if i in columns]
-                        )
-                columns = list(dict.fromkeys(necessary_columns))
+                        necessary_columns.update(i for i in args[field] if i in columns)
+                columns = list(necessary_columns)
                 args["data_frame"] = pd.api.interchange.from_dataframe(
                     args["data_frame"].select_columns_by_name(columns)
                 )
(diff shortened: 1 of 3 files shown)


packages/python/plotly/plotly/express/_core.py
```diff
@@ -1419,9 +1419,17 @@ def build_dataframe(args, constructor):
...
                 # Save precious resources by only interchanging columns that are
                 # actually going to be plotted.
-                columns = [
+                necessary_columns = [
                     i for i in args.values() if isinstance(i, str) and i in columns
                 ]
+                for field in args:
+                    if field in array_attrables and isinstance(
+                        args[field], (list, dict)
+                    ):
```
alexcjohnson, line 1428: `px` is generally much looser with types, and elsewhere seems to assume `array_attrables` are iterables already. So can we remove this `isinstance` check so that other iterables will work too? I don't think we even need to exclude `str`, as a string would already have caused other problems for these args. I'm thinking of tuples at least but I'm sure there are others that will arise. Another example, `reversed([1,2,3])` is not a list...

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1419,9 +1419,17 @@ def build_dataframe(args, constructor):
...
+                        necessary_columns.extend(
+                            [i for i in args[field] if i in columns]
+                        )
+                columns = list(dict.fromkeys(necessary_columns))
```
alexcjohnson, line 1432: Is this just to limit to unique items? Maybe clearer as `list(set(necessary_columns))` then?

packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
```diff
@@ -327,6 +327,32 @@ def test_build_df_from_vaex_and_polars(test_lib):
...
+    args = dict(
+        data_frame=iris_vaex,
+        x="petal_width",
+        y="sepal_length",
+        hover_data=["sepal_width"],
+    )
+    out = build_dataframe(args, go.Scatter)
+    assert_frame_equal(
+        iris_pandas.reset_index()[out["data_frame"].columns], out["data_frame"]
+    )
```
alexcjohnson, line 353: Great test, thanks! Can we also add a flavor that uses a dict, like here: https://plotly.com/python/hover-text-and-formatting/#disabling-or-customizing-hover-of-columns-in-plotly-express

packages/python/plotly/plotly/express/_core.py
guyrosin: You're right!
I used this condition to ignore string arguments, but if you say they are handled elsewhere then sure, why not simplify this code :)

guyrosin: You're right!
I used this condition to ignore string arguments, but if you say they are handled elsewhere then sure, why not simplify this code :)

packages/python/plotly/plotly/express/_core.py
guyrosin: I used it to preserve the order of the columns, but will change to your suggestion as the order doesn't seem to be important here

packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
guyrosin: Thanks, for sure!

packages/python/plotly/plotly/express/_core.py
alexcjohnson: ha ok - the order there doesn't actually matter, so that test could be rewritten to turn the assertion into a set or sorted list comparison `assert sorted(interchange_dataframe.select_columns_by_name.call_args.args[0]) == ["petal_width", "sepal_length"]` or some such. But not a big deal.

alexcjohnson:
Thanks so much @guyrosin ! Would you add a changelog entry, in a new `## UNRELEASED` section at the top? Then I think this is ready to go, and we'll try to get a patch out shortly.

packages/python/plotly/plotly/express/_core.py
guyrosin: Got it! thanks. (btw I tried using `set()` but vaex requires the columns passed to `select_columns_by_name()` to be a `Sequence` for some reason, so I ended up using your suggestion)

guyrosin: Oh, this `call_args.args` property was introduced in python 3.8, so I'm slightly changing it

guyrosin:
That's great to hear! Thanks @alexcjohnson for the feedback

alexcjohnson (approved):
💃 Very nice! Thanks again for identifying AND fixing this regression @guyrosin

PR #4328, opened 2023-08-16
Version changes for v5.16.1

alexcjohnson:
- Update version to 5.16.1
- Finalize changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
-## UNRELEASED
+## [5.16.1] - 2023-08-16
 
 ### Fixed
 - Fixed issue with necessary columns from complex arguments dropped when interchanging dataframes [[#4324](https://github.com/plotly/plotly.py/pull/4324)]
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@
 
 ## Quickstart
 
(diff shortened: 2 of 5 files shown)


alexcjohnson:
Issues encountered in QA'ing the build artifacts here using the [release process](https://github.com/plotly/plotly.py/blob/master/release.md):
- The conda file is not `plotly-5.16.1.tar.bz2` but `plotly-5.16.1-py_0.tar.bz2`. This seems to be OK and does not indicate a dirty build, as it matches the file names in https://anaconda.org/plotly/plotly/files
- When I install from this file I do not get `tenacity` installed automatically like I do with the `pip` build. This also appears to be OK, as (a) the same thing happens if I download 5.16.0 from https://anaconda.org/plotly/plotly/files and install it the same way, (b) it seems like this is deliberate (and endlessly puzzling to the community) https://github.com/conda/conda/issues/1884

Once I have the necessary packages installed both `Figure` and `FigureWidget` seem to work fine in both notebook and lab.

archmoj:
:dancer:

alexcjohnson:
Just a note on the docs update process, so we can use these PR notes to debrief the whole release process when both @LiamConnors and @archmoj are available: eventually I got this to work but it took an hour or so during which time I made several commits that I thought had the right state but did not result in the deployed docs updating. Is there some sequencing, some other job happening behind the scenes you need to wait for before adding the graphing-library-docs commit that triggers the final build?

PR #4357, opened 2023-09-15
remove py2 from build

alexcjohnson:
We haven't supported Py2 for a while, so we should remove the `universal` flag from the `bdist_wheel` command so we don't end up with `py2` in the filename :)


alexcjohnson:
whew, removed a whole bunch more Py2 junk that was still lying around - `__future__` imports, a few `# -*- coding: utf-8 -*-` headers, some manual compatibility shims...

I didn't touch `chart_studio`, which still has lots of `__future__` imports.

archmoj:
LGTM :pray:

LiamConnors (approved):
:dancer:

Issue #3624, opened 2022-03-09
mplexporter not getting Collection offset position due to matplotlib API changes in v3.5


cylammarco:
Using plotly 5.6.0 and matplotlib 3.5.1

[https://matplotlib.org/devdocs/api/prev_api_changes/api_changes_3.5.0.html#classes-methods-and-attributes
](https://matplotlib.org/devdocs/api/prev_api_changes/api_changes_3.5.0.html#classes-methods-and-attributes
)
> [Collection](https://matplotlib.org/devdocs/api/collections_api.html#matplotlib.collections.Collection).set_offset_position and Collection.get_offset_position have been removed; the offset_position of the Collection class is now "screen"

Example code to reproduce the problem:
```
import numpy as np
from matplotlib import pyplot as plt
from plotly import tools

x = np.arange(10)
y = x **2.

fig1 = plt.figure(1)
plt.vlines(x, 0, max(y))

tools.mpl_to_plotly(fig1)
```

gives

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\------\AppData\Local\Programs\Python\Python310\lib\site-packages\plotly\tools.py", line 112, in mpl_to_plotly
    matplotlylib.Exporter(renderer).run(fig)
  File "C:\Users\------\AppData\Local\Programs\Python\Python310\lib\site-packages\plotly\matplotlylib\mplexporter\exporter.py", line 53, in run
    self.crawl_fig(fig)
  File "C:\Users\------\AppData\Local\Programs\Python\Python310\lib\site-packages\plotly\matplotlylib\mplexporter\exporter.py", line 124, in crawl_fig
    self.crawl_ax(ax)
  File "C:\Users\------\AppData\Local\Programs\Python\Python310\lib\site-packages\plotly\matplotlylib\mplexporter\exporter.py", line 146, in crawl_ax
    self.draw_collection(ax, collection)
  File "C:\Users\------\AppData\Local\Programs\Python\Python310\lib\site-packages\plotly\matplotlylib\mplexporter\exporter.py", line 289, in draw_collection
    offset_order = offset_dict[collection.get_offset_position()]
AttributeError: 'LineCollection' object has no attribute 'get_offset_position'. Did you mean: '_offset_position'?
```

alberttorosyan:
facing similar issue with
```
plotly==5.4.0
matplotlib==3.5.0
```

[...]

PR #4372, opened 2023-10-03
Remove use of matplotlib's get_offset_position

NeilGirdhar:


NeilGirdhar:
Fixes https://github.com/plotly/plotly.py/issues/3624 ?

packages/python/plotly/plotly/matplotlylib/mplexporter/exporter.py
```diff
@@ -285,16 +285,13 @@ def draw_collection(
             "zorder": collection.get_zorder(),
         }
 
-        offset_dict = {"data": "before", "screen": "after"}
-        offset_order = offset_dict[collection.get_offset_position()]
```
alexcjohnson, line 289: Thanks @NeilGirdhar - in as far as `"screen"` is now the sole possibility, this is reasonable. I wonder though, for users who still have MPL 3.4 or before, what's the likelihood they have `"data"` and thus this change would break their use case? We could support this with something like:
```py
offset_order = offset_dict[
    collection.get_offset_position()
    if hasattr(collection, "get_offset_position")
    else "screen"
]
```

alexcjohnson:
Does this restore compatibility all the way up to the current MPL v3.8.0? Kind of remarkable that code written 9 years ago, for MPL v1.3, kept working through two major version bumps but broke on a minor 😅 

In addition to my comment above, there are two things we'll want to do before merging:
- Add a changelog entry (make a new `## UNRELEASED` -> `### Fixed` section at the top)
- The [test requirements](https://github.com/plotly/plotly.py/tree/master/packages/python/plotly/test_requirements) files `requirements_*_optional.txt` all include MPL v2.2.3 (only 5 years old 🙈) - perhaps we can leave this version in `requirements_36_optional.txt` to ensure backward compatibility but bump `37`, `38`, `39`, and `39_pandas_2` to the latest supported by each Python version so the [existing matplotlylib tests](https://github.com/plotly/plotly.py/tree/master/packages/python/plotly/plotly/tests/test_optional/test_matplotlylib) run on the versions new users will get by default. Looks like that's:
  - Py3.7: MPL v3.5.3
  - Py3.8: MPL v3.7.3
  - Py3.9: MPL v3.8.0

packages/python/plotly/plotly/matplotlylib/mplexporter/exporter.py
NeilGirdhar: Good point.  Done, as requested.

NeilGirdhar:
@alexcjohnson Done, as requested.

NeilGirdhar:
@alexcjohnson Is there anything else I should do to polish this off?

alexcjohnson (approved):
💃 Nice work @NeilGirdhar! Merging, I expect we'll release this next week with the next plotly.js minor. Thanks!

NeilGirdhar:
Thanks for the quick review process!

NeilGirdhar:
@alexcjohnson Any chance we could get a release soon with this since it breaks WandB?

alexcjohnson:
Sorry for the delay @NeilGirdhar - we’re just finishing up a feature in plotly.js and then will make a release. Should be the next couple of days.

NeilGirdhar:
@alexcjohnson No problem at all!   Looking forward to the release :)

Issue #4285, opened 2023-07-19
python-igraph was renamed to igraph


szhorvat:
The `python-igraph` package was renamed to simply `igraph`. Support for the earlier name is stopping. See https://pypi.org/project/python-igraph/

Please update the references in plotly.py:

https://github.com/search?q=repo%3Aplotly%2Fplotly.py%20python-igraph&type=code

This is basically changing all the `pip install python-igraph` to `pip install igraph`.

LiamConnors:
Sorry for the delay on this. I've created a new PR for this with just the required changes and we'll get it merged as soon as possible https://github.com/plotly/plotly.py/pull/4383

PR #4383, opened 2023-10-14
Update python-igraph references

LiamConnors:
To close https://github.com/plotly/plotly.py/issues/4285

Couldn't find a reason for the ` sudo pip install` references on the original PR, https://github.com/plotly/plotly.py/pull/4289, and it seems to work fine without

### DIFF
--- a/doc/unconverted/python/igraph-networkx-comparison.md
+++ b/doc/unconverted/python/igraph-networkx-comparison.md
@@ -31,7 +31,7 @@ Plotly's Python library is free and open source! [Get started](https://plot.ly/p
 
 
 #### Comparison
-In this tutorial we plot the same network - the coauthorship network of scientists working on network theory and experiment - first as an `igraph.Graph` object, with the Kamada-Kawai layout, and then as a `networkx.Graph`, with the Fruchterman-Reingold layout. Install the Python libraries with `sudo pip install python-igraph` and `sudo pip install networkx`.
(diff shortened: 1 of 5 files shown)


doc/unconverted/python/igraph-networkx-comparison.md
```diff
@@ -31,7 +31,7 @@ Plotly's Python library is free and open source! [Get started](https://plot.ly/p
 
 
 #### Comparison
-In this tutorial we plot the same network - the coauthorship network of scientists working on network theory and experiment - first as an `igraph.Graph` object, with the Kamada-Kawai layout, and then as a `networkx.Graph`, with the Fruchterman-Reingold layout. Install the Python libraries with `sudo pip install python-igraph` and `sudo pip install networkx`.
+In this tutorial we plot the same network - the coauthorship network of scientists working on network theory and experiment - first as an `igraph.Graph` object, with the Kamada-Kawai layout, and then as a `networkx.Graph`, with the Fruchterman-Reingold layout. Install the Python libraries with `pip install igraph` and `sudo pip install networkx`.
```
alexcjohnson, line 34: Same Q

LiamConnors:
@alexcjohnson looking closer at those examples that use sudo, they are actually for Python 2 and Plotly.py v3, so probably best not to update them at all. `pip install igraph` won't work for Python 2. I think we can probably just remove these pages altogether, but I need to look into the correct way to remove v3 pages.

alexcjohnson:
ah yep, makes sense. Not sure if it's worth removing those pages... could be worthwhile in fact adding links from them to dash-cytoscape docs as the more modern and powerful solution for the same kind of needs.

alexcjohnson (approved):
💃 You're still changing the Py2 pages from `python-igraph` to `igraph` - but I think that's OK, really nobody should be using Py2 anymore and recent `igraph` and `python-igraph` are apparently literally the same code.

Issue #1103, opened 2018-08-09
Add type annotation stubs


malmaud:
Hi,
I was wondering if you would be interested in PRs that add type annotation stubs for the plotly.py API. That would enable static type checkers like mypy and Pyre to statically type check Python projects using plotly, and as well as enhancing the IDE experience for editors able to offer richer experiences from type annotations (just as VSCode and PyCharM).

jonmmease:
Hi @malmaud, I would love to do this... but we still support Python 2.7.

That said, in all of our new generated code we use numpydoc docstrings with return types (see, for example, `plotly/graph_objs/_bar.py`). This is enough for PyCharm to figure things out (I haven't really tested other editors very much) and it does provide a nice experience when working with objects in the `graph_objs` hierarchy.

I would be interested in gaining a better understanding of the most broadly compatible way to specify type information for a mixed Python 2/3 project!

malmaud:
Great to hear that. The stubs can live in separate files that sit parallel
to the original Python source files, which preserves Python 2
compatibility.

I recently went through this with the Visdom project. See the
`__init__.pyi` file in
https://github.com/facebookresearch/visdom/tree/17f28cb258ad874d6fc5fcce5a4479abe4377738/py/visdom,
which lives alongside the `__init__.py` that it annotates.

This is the same system as Typescript's `.d.ts` files, if you're familiar.

jonmmease:
That looks really interesting. I don't see any downside to adding these to at least the code generation output, because that would automatically stay up to date (and this covers the majority of the public API).

What has your experience been like keeping these up to date as a codebase evolves (for the human generated code 🙂)? What kind of tooling/testing would you recommend to that help flag inconsistencies?

malmaud:
Well, I won't exaggerate - I became excited about static type checking for Python and started using it for my own research code, but I don't have have direct experience applying it to large multi-developer codebases. I'd look forward to trying and learning, however.

In terms of help from tooling, the only thing I know of is https://github.com/Instagram/MonkeyType. It attempts to produce stubs automatically based on the types it observes variables take during runtime, but I haven't tried it. 

The good news is the worse-case scenario for an incorrect stub is a user gets an incorrect warning (or lack of warning) from their type checker and reports it as an issue. Actual runtime behavior can never be affected by incorrect stubs.

jonmmease:
[...]

PR #4424, opened 2023-11-15
Add py.typed

rhelmot:
This is neither a doc pr nor a code pr, but a packaging PR! Related to #1103, this makes the existing partially-complete type annotations usable for checking projects which import plotly.

### DIFF
--- a/packages/python/plotly/MANIFEST.in
+++ b/packages/python/plotly/MANIFEST.in
@@ -5,3 +5,4 @@ include plotly/_version.py
 include requires-install.txt
 include requires-express.txt
 include plotly/package_data/widgetbundle.js
+include plotly/py.typed
--- a/packages/python/plotly/setup.py
+++ b/packages/python/plotly/setup.py
@@ -554,6 +554,7 @@ setup(
     + validator_packages,
     package_data={
         "plotly": [
+            "py.typed",
             "package_data/*",
             "package_data/templates/*",
             "package_data/datasets/*",


jack-volantautonomy (approved):
Nice

packages/python/plotly/plotly/py.typed
```diff
@@ -0,0 +1 @@
+partial
```
jamesbraza, line 1: `py.typed` is just a marker file, per https://peps.python.org/pep-0561/#packaging-type-information

> This marker applies recursively

It should be blank (e.g. just a file without any contents), so can you remove the text "partial"?

---

Are you interpreting the text `partial` to be like a signifier of partial typing? I don't think anyone actually consumes the content of this file, for example `mypy` just goes based on the file existing: https://github.com/python/mypy/blob/v1.14.1/mypy/modulefinder.py#L344-L345

jamesbraza: Oh I follow what you mean, thanks for sharing. TIL that `partial` means to first check a stub package, then fall back on the runtime package.

Though I still don't think `partial` should be here, as this is not a stub package, this is the actual runtime package. Does that make sense?

In other words, if this was an external stub package filled with `.pyi`, then you'd put `partial`. But that's not the case here, it's the runtime package with `.py` files, so `py.typed` should be blank

rhelmot: I see. Fixed.

PR #4442, opened 2023-11-27
Fix usage of `column_names` which is part of Interchange protocol

anmyachev:
## Code PR

- [ ] I have read through the [contributing notes](https://github.com/plotly/plotly.py/blob/master/contributing.md) and understand the structure of the package. In particular, if my PR modifies code of `plotly.graph_objects`, my modifications concern the `codegen` files and not generated files.
  modified existing tests.

### DIFF
--- a/packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
+++ b/packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
@@ -250,15 +250,24 @@ def test_build_df_with_index():
     assert_frame_equal(tips.reset_index()[out["data_frame"].columns], out["data_frame"])
 
 
+@pytest.mark.parametrize("column_names_as_generator", [False, True])
 def test_build_df_using_interchange_protocol_mock(
-    add_interchange_module_for_old_pandas,
+    add_interchange_module_for_old_pandas, column_names_as_generator
 ):
     class InterchangeDataFrame:
         def __init__(self, columns):
             self._columns = columns
 
-        def column_names(self):
-            return self._columns
+        if column_names_as_generator:
+
+            def column_names(self):
+                for col in self._columns:
+                    yield col
+
+        else:
+
+            def column_names(self):
+                return self._columns
 
     interchange_dataframe = InterchangeDataFrame(
         ["petal_width", "sepal_length", "sepal_width"]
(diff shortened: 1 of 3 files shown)


anmyachev:
Hi @MarcoGorelli! Could you take a look?

MarcoGorelli:
looks fine

anmyachev:
> looks fine

@MarcoGorelli thanks for the feedback!

anmyachev:
Hi @alexcjohnson! I'm not sure if it's necessary to add tests here, do you think it's worth adding? The change looks safe enough. 

For example, I could add a synthetic test where the generator from this function will be returned instead of list.
https://github.com/plotly/plotly.py/blob/63b9ac5d813f26f72035e7c5a91c4ce5af34c028/packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py#L261

alexcjohnson:
I see, so in principle `column_names` could be a non-materialized list - generator, iterator, whatever, but we actually depend on `columns` being materialized, indexable etc... and so far all the dataframes we've encountered have it materialized but in principle that could change any time and still be compliant with the protocol. Do I have that right?

If so, then yes, I think your proposed test would be great. Given that this PR is fixing a case we hadn't previously covered it definitely wants a test one way or another, that would have failed without this change. After that the only other thing we'll want is a changelog entry.

Thanks for the PR!

anmyachev:
> I see, so in principle `column_names` could be a non-materialized list - generator, iterator, whatever, but we actually depend on `columns` being materialized, indexable etc... and so far all the dataframes we've encountered have it materialized but in principle that could change any time and still be compliant with the protocol. Do I have that right?

Yes, you are right. Thanks for the quick response!

> If so, then yes, I think your proposed test would be great. Given that this PR is fixing a case we hadn't previously covered it definitely wants a test one way or another, that would have failed without this change. After that the only other thing we'll want is a changelog entry.

I added a test and a log entry.

packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
```diff
@@ -14,12 +14,12 @@
 @pytest.fixture
 def add_interchange_module_for_old_pandas():
     if not hasattr(pd.api, "interchange"):
-        pd.api.interchange = mock.MagicMock()
```
anmyachev, line 17: This change is necessary so that this attribute is not saved between runs.

alexcjohnson (approved):
💃 Very nicely done @anmyachev - thanks again!

PR #4457, opened 2023-12-14
Update map tiles in docs

LiamConnors:
Update tiles in docs, until we do the next release which allows using stamen tiles again.


alexcjohnson (approved):
💃

PR #4470, opened 2023-12-21
Use plotly.js `base64` API to store and pass typed arrays declared by numpy, pandas, etc. 

archmoj:
This PR make use of [this plotly.js PR](https://github.com/plotly/plotly.js/pull/5230) and changes the default behavior to pass numpy arrays as plotly.js base64 spec.

### Code Changes
* Adds `to_typed_array_spec` function to use for converting numpy arrays to the "typed array spec" type that plotly.js expects when receiving base64 encoded arrays.
* Updates the `validate_coerce` function of the `DataArrayValidator` class to base64 encode numpy arrays and store them in the "typed array spec" format (by calling `to_typed_array_spec`)
* Adds thorough testing to validators
* Adds performance (size and speed) tests


packages/python/plotly/_plotly_utils/basevalidators.py
```diff
@@ -50,6 +50,83 @@ def to_scalar_or_list(v):
...
+int32max = 2147483647
+
+uint8max = 255
+uint16max = 65535
+uint32max = 4294967295
+
+
+def to_typed_array_spec(v):
+    """
+    Convert numpy array to plotly.js typed array sepc
```
LiamConnors, line 78:
```suggestion
    Convert numpy array to plotly.js typed array spec
```

doc/python/b64.md
```diff
@@ -0,0 +1,198 @@
...
+      version: 3
+    file_extension: .py
+    mimetype: text/x-python
+    name: python
+    nbconvert_exporter: python
+    pygments_lexer: ipython3
+    version: 3.9.0
+  plotly:
+    description: How to format axes of 3d plots in Python with Plotly.
+    display_as: b64
```
LiamConnors, line 26:
```suggestion
    display_as: advanced_opt
```

I think we could put this in the Advanced section

doc/python/b64.md
```diff
@@ -0,0 +1,198 @@
...
+    page_type: example_index
+    permalink: python/b64/
+    thumbnail: thumbnail/b64.png
+---
+
+### Simple example showing how arrays of numbers could be passed as base64 typed array objects to plotly.js
+
+```python
+import plotly.graph_objects as go
+
```
LiamConnors, line 40: Once plotly.py automatically does the conversion as shown in the example below, is there any reason why you might want to do it this way?

doc/python/b64.md
```diff
@@ -0,0 +1,198 @@
...
+    z=z
+)])
+
+fig.show()
+```
+
+### Example where base64 is applied to pass values as typed array objects to plotly.js
+
+```python
+import plotly.graph_objects as go
```
LiamConnors, line 75:
```suggestion
### Example where base64 is applied to pass values as typed array objects to plotly.js

```python
```
If we'll do this automatically, can this example go?

doc/python/b64.md
```diff
@@ -0,0 +1,198 @@
...
+    display_as: b64
+    language: python
+    layout: base
+    name: b64
+    order: 1
+    page_type: example_index
+    permalink: python/b64/
+    thumbnail: thumbnail/b64.png
+---
+
```
LiamConnors, line 35: Could you add some context on the benefits of doing this. What does it bring for the user?

doc/python/b64.md
archmoj: Yes. I think it may be useful.

doc/python/b64.md
archmoj: I prefer to keep it for now.

packages/python/plotly/plotly/tests/test_optional/test_utils/test_utils.py
```diff
@@ -372,6 +372,7 @@ def test_invalid_encode_exception(self):
         with self.assertRaises(TypeError):
             _json.dumps({"a": {1}}, cls=utils.PlotlyJSONEncoder)
 
+    @pytest.mark.skip(reason="The encoding is faster now.")
```
marthacryan, line 375: Should we be skipping this?

doc/python/b64.md
```diff
@@ -0,0 +1,198 @@
+---
```
marthacryan, line 1: How is this section used?

packages/python/plotly/_plotly_utils/basevalidators.py
```diff
@@ -50,6 +50,83 @@ def to_scalar_or_list(v):
...
+
+
+def to_typed_array_spec(v):
+    """
+    Convert numpy array to plotly.js typed array spec
+    If not possible return the original value
+    """
+    v = copy_to_readonly_numpy_array(v)
+
+    np = get_module("numpy", should_load=False)
```
marthacryan, line 83: Could you give some context on why this is being used instead of an import at the top?

packages/python/plotly/_plotly_utils/basevalidators.py
```diff
@@ -50,6 +50,83 @@ def to_scalar_or_list(v):
...
+uint16max = 65535
+uint32max = 4294967295
+
+
+def to_typed_array_spec(v):
+    """
+    Convert numpy array to plotly.js typed array spec
+    If not possible return the original value
+    """
+    v = copy_to_readonly_numpy_array(v)
```
marthacryan, line 81: maybe we could use a separate variable to be more clear? And call it something a little more informative than `v`?

marthacryan: (Maybe `np_arr` or something)

packages/python/plotly/plotly/tests/b64.py
```diff
@@ -0,0 +1,82 @@
...
+int16max = 32767
+int32min = -2147483648
+int32max = 2147483647
+
+uint8max = 255
+uint16max = 65535
+uint32max = 4294967295
+
+
+def b64(v):
```
marthacryan, line 27: Maybe this name should be a little more descriptive

marthacryan: Something like `np_arr_to_array_spec`?

packages/python/plotly/plotly/tests/b64.py
```diff
@@ -0,0 +1,82 @@
...
+    If not possible return the original value
+    """
+
+    if not isinstance(v, np.ndarray):
+        return v
+
+    dtype = str(v.dtype)
+
+    # convert default Big Ints until we could support them in plotly.js
+    if dtype == "int64":
```
marthacryan, line 39: Are we documenting the types currently accepted anywhere?

marthacryan: Also might be worth having some kind of warning about this conversion happening

packages/python/plotly/plotly/tests/b64.py
```diff
@@ -0,0 +1,82 @@
...
+    if dtype in plotlyjsShortTypes:
+        arrObj = {
+            "dtype": plotlyjsShortTypes[dtype],
+            "bdata": base64.b64encode(v).decode("ascii"),
+        }
+
+        if v.ndim > 1:
+            arrObj["shape"] = str(v.shape)[1:-1]
+
+        print(arrObj)
```
marthacryan, line 74: extra print statement

packages/python/plotly/plotly/tests/test_optional/test_figure_factory/test_figure_factory.py
```diff
@@ -1012,7 +1016,7 @@ def test_default_dendrogram(self):
         self.assert_fig_equal(dendro["data"][1], expected_dendro["data"][1])
         self.assert_fig_equal(dendro["data"][2], expected_dendro["data"][2])
 
-        self.assert_fig_equal(dendro["layout"], expected_dendro["layout"])
+        # self.assert_fig_equal(dendro["layout"], expected_dendro["layout"])
```
marthacryan, line 1019: Should this be commented out?

packages/python/plotly/_plotly_utils/basevalidators.py
alexcjohnson: Presumably we're doing this because `numpy` isn't a required dependency of `plotly` (nor of `dash` for that matter) - though most people will get it when they install `pandas`. Which begs the question, what _does_ happen if you don't have `numpy` installed? Obviously then nothing is an `np.ndarray` but looks like we need to account for that possibility, as `np` will be `None` then so this would throw an error, unless we change the next line to something like
```
if np is None or not isinstance(v, np.ndarray):
```

doc/python/b64.md
archmoj: This header in this section is copied from other examples.
To run this example you may run
```sh
jupyter notebook doc/python/b64.md
```

packages/python/plotly/_plotly_utils/basevalidators.py
archmoj: Short variable names are not bad and often easier to read.
In this case `v` stands for value and used in the other parts of this code.
So I think we could/should keep it as is.

packages/python/plotly/_plotly_utils/basevalidators.py
archmoj: Thanks @alexcjohnson for the note.
`get_module("numpy", should_load=False)` is used inside other functions in this file.
So that's OK to keep it this way.

packages/python/plotly/_plotly_utils/basevalidators.py
```diff
@@ -207,6 +284,17 @@ def is_homogeneous_array(v):
     return False
 
 
+def is_typed_array_spec(v):
+    """
+    Return whether a value is considered to be a typed array spec for plotly.js
+    """
+    return isinstance(v, dict) and "bdata" in v
```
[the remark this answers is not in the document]
archmoj: Good call. Addressed in [f5df6db](https://github.com/plotly/plotly.py/pull/4470/commits/f5df6dbc2b2b24d4403fd6c18f42024f4117ec07).

packages/python/plotly/_plotly_utils/basevalidators.py
alexcjohnson: We don’t have any tests though that DON’T have numpy available, since it’s in the CI requirements. So likely all these uses are broken for anyone who doesn’t install it, and we should protect them all. Ideally we run a few tests in an env that doesn’t install numpy at all.

packages/python/plotly/plotly/tests/b64.py
archmoj: Removed in [f332922](https://github.com/plotly/plotly.py/pull/4470/commits/f332922883e06f7bbf10e625253f0dc685d07754).

packages/python/plotly/_plotly_utils/basevalidators.py
alexcjohnson: Either that or we make numpy a hard requirement, which isn’t totally out of the question, a large majority of our users surely do install it, but I think it’s better if we can keep it optional.

archmoj: Good point @alexcjohnson.
@marthacryan Could you open a separate PR from master and add a CircleCI job here 
https://github.com/plotly/plotly.py/blob/bd1dd66ed555feda1f8d514b87ddf234970a6a5a/.circleci/config.yml#L587 to test the case of numpy not being installed?
Once merged to the master I would update this branch.

packages/python/plotly/plotly/tests/test_optional/test_figure_factory/test_figure_factory.py
archmoj: Yes we should bring back this test.
Let me investigate if/how `nan` and `null` values are handled between numpy and plotly.js!

packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
```diff
@@ -28,8 +29,9 @@ def add_interchange_module_for_old_pandas():
 
 def test_numpy():
     fig = px.scatter(x=[1, 2, 3], y=[2, 3, 4], color=[1, 3, 9])
-    assert np.all(fig.data[0].x == np.array([1, 2, 3]))
-    assert np.all(fig.data[0].y == np.array([2, 3, 4]))
+
+    assert np.all(fig.data[0].x == b64(np.array([1, 2, 3])))
```
marthacryan, line 33: Do you need to convert this to b64 here?

marthacryan: Or you could use `_b64`

marthacryan:
(Closed and reopened to kick CI)

marthacryan (approved):

archmoj:
@marthacryan Please resolve the conflicts with master; then merge master into this branch. 
Then you may possibly copy the version of tests from master into this branch to see if any tests need any adjustments. 
Thank you!

gvwilson:
thank you everyone

packages/python/plotly/plotly/tests/test_optional/test_utils/test_utils.py
```diff
@@ -372,38 +372,6 @@ def test_invalid_encode_exception(self):
         with self.assertRaises(TypeError):
             _json.dumps({"a": {1}}, cls=utils.PlotlyJSONEncoder)
 
-    def test_fast_track_finite_arrays(self):
```
archmoj, line 375: @marthacryan 
Could we revert this change now or we should drop the test?

marthacryan: I removed this because it was failing. Presumably it's just faster now?

ndrezn (commented):
Psyched for this to land! Killer work team.
ezgif-4-d5a21d0496

test/percy/compare-pandas.py
```diff
@@ -5,25 +5,29 @@
...
 
 for filename in os.listdir("pandas2"):
-    with open(filename, encoding="utf-8") as f1:
-        with open(os.path.join("pandas2", filename)) as f2:
-            fig1 = get_fig(f1.read())
-            fig2 = get_fig(f2.read())
-            if any(l1 != l2 for l1, l2 in zip(fig1, fig2)):
-                print("".join(difflib.unified_diff(fig1, fig2)))
-                raise ValueError(f"Pandas 1/2 difference in {filename}")
+    if filename not in [
```
emilykl, line 18: @marthacryan / @archmoj What is the reason for excluding these particular files?

packages/python/plotly/_plotly_utils/utils.py
```diff
@@ -1,11 +1,116 @@
...
+    return v
+
+
+def is_skipped_key(key):
+    """
+    Return whether any keys in the parent hierarchy are in the list of keys that
+    are skipped for conversion to the typed array spec
+    """
+    skipped_keys = ["geojson", "layer", "range"]
+    return any(skipped_key in key for skipped_key in skipped_keys)
```
emilykl, line 99: @marthacryan Shouldn't the test be `skipped_key == key` rather than `skipped_key in key` ?

Or we really do want to be checking that the skipped key is a substring of `key`?

emilykl: Also I think the docstring comment is no longer accurate

packages/python/plotly/_plotly_utils/utils.py
```diff
@@ -1,11 +1,116 @@
...
+
+def to_typed_array_spec(v):
+    """
+    Convert numpy array to plotly.js typed array spec
+    If not possible return the original value
+    """
+    v = copy_to_readonly_numpy_array(v)
+
+    np = get_module("numpy", should_load=False)
+    if not isinstance(v, np.ndarray):
```
emilykl, line 47: Won't this raise an error if `numpy` is not installed? (Since `get_module` will return `None` in that case.)

Is that ok?

alippai:
Should this work for SVG render mode too? (`render_mode="svg"` in `px.line`)

gvwilson:
cc @marthacryan @emilykl for an answer

Issue #4455, opened 2023-12-08
y axis labels does not appear in scatter plot when using plotly.express with 2 inputs


npcomplete2023:
Hello Plotly,

I experience the following problem,
when trying to generate a figure of type 'line' with multiple traces using the plotly.express
(the traces are defined implicitly, by defining a list of list in the y argument)
the resulting figure indeed includes 2 traces as expected, 
it DOES name correctly the x axis but DOESN'T name correctly the y axis label, which is set to 'value' for some reason.
See attached an example code and a resulting graph (saved as screenshot) 

MOREOVER, you can see the same phenomenon in the plotly tutorial 
https://www.youtube.com/watch?v=GGL6U0k8WYA&t=622s
at minutes 08:50-09:00 of the video. Notice the y-axis label in the video.

Thank you

--------------------Example code--------------
```
import plotly.express as px
pxFig = px.line(x=[0,1,2,3],y=[[10,11,12,13],[12,13,14,15]], labels=dict(x='xLabel', y='yLabel'), title='my Title')
pxFig.show()
```
-------------------------------
Attached image
![plotly_bug_expressMode_multipleYvalues](https://github.com/plotly/plotly.py/assets/153299265/e63adf25-6da0-4787-bbe2-904729fa26c4)

Coding-with-Adam:
hi @npcomplete2023 
Thanks for reporting. This does look like a bug.

Mizba-Anjum:
Hello, I tried tackling this problem in #4474 Let me know if this helps, since it's my first time contributing. The issue seemed like it was a string literal "value" in the express folder

alexcjohnson:
From @nicolaskruchten in #4474:

> In wide mode, the cross dimension is internally treated as if it was called `value` so this already works, in all orientations:
> 
> ```python
> import plotly.express as px
> df = px.data.medals_wide(indexed=True)
> px.bar(df, labels=dict(value="hey there"))
> px.bar(df, labels=dict(value="hey there"), orientation='h')
> ```
> 
> There's probably a discoverability problem but my recommendation would be to add to the docs rather than modify this somewhat complex code.

> There's already some mention of this in the docs here https://plotly.com/python/wide-form/#labeling-axes-legends-and-hover-text but things could always be made clearer :)

[...]

PR #4474, opened 2023-12-29
Update _core.py to fix y label string literal

Mizba-Anjum:
Remove string literal and replace with args y label value. This allowed the y label to be shown when running px.line. Did not feel the need to add a test or documentation but I did run pytest. Related to issue #4455 

## Code PR

  modified existing tests.

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1412,7 +1412,7 @@ def build_dataframe(args, constructor):
                 wide_cross_name = "__x__" if wide_y else "__y__"
 
     if wide_mode:
-        value_name = _escape_col_name(columns, "value", [])
+        value_name = _escape_col_name(columns, args["labels"]["y"], [])
         var_name = _escape_col_name(columns, var_name, [])
 
     if needs_interchanging:
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1412,11 +1412,8 @@ def build_dataframe(args, constructor):
                 wide_cross_name = "__x__" if wide_y else "__y__"
 
     if wide_mode:
-        if "labels" in args:
-            if "y" in args["labels"]:
-                y_value = args["labels"]["y"]
-            else:
-                y_value = "value"
+        if "labels" in args and "y" in args["labels"]:
+            y_value = args["labels"]["y"]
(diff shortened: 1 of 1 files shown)


Coding-with-Adam:
Thank you for the PR, @Mizba-Anjum .
I cloned your repo/fix and can confirm that this code now works. It names correctly the y axis label.
```
import plotly.express as px
pxFig = px.line(x=[0,1,2,3],y=[[10,11,12,13],[12,13,14,15]], labels=dict(x='xLabel', y='yLabel'), title='my Title')
pxFig.show()
```

image

packages/python/plotly/plotly/express/_core.py
```diff
@@ -1412,7 +1412,14 @@ def build_dataframe(args, constructor):
                 wide_cross_name = "__x__" if wide_y else "__y__"
 
     if wide_mode:
-        value_name = _escape_col_name(columns, "value", [])
+        if "labels" in args:
+            if "y" in args["labels"]:
+                y_value = args["labels"]["y"]
+            else:
+                y_value = "value"
```
alexcjohnson, line 1419:
```suggestion
        if "labels" in args and "y" in args["labels"]:
            y_value = args["labels"]["y"]
```
Just a little simplification :)

alexcjohnson:
This mostly looks good to me, but there's a lot of abstraction here, in particular it looks like there are some cases `value_name` should be pulled from `x` instead of `y` in this block:

https://github.com/plotly/plotly.py/blob/95cc3d0dc6b3379d181e28ba90a0b708fad3bc40/packages/python/plotly/plotly/express/_core.py#L1555-L1578

So I'd love if @nicolaskruchten would weigh in here about whether there's a better way to cover this 🙏

nicolaskruchten:
In wide mode, the cross dimension is internally treated as if it was called `value` so this already works, in all orientations:

```python
import plotly.express as px
df = px.data.medals_wide(indexed=True)
px.bar(df, labels=dict(value="hey there"))
px.bar(df, labels=dict(value="hey there"), orientation='h')
```

There's probably a discoverability problem but my recommendation would be to add to the docs rather than modify this somewhat complex code.

nicolaskruchten:
There's already some mention of this in the docs here https://plotly.com/python/wide-form/#labeling-axes-legends-and-hover-text but things could always be made clearer :)

PR #4478, opened 2024-01-09
Fix rendering of example in interactive-html-export.md

Arcader717:
Should fix the rendering of the python example. The html example still isn't rendering correctly. Basically just a typo fix

### Documentation PR

- [x] I've [seen the `doc/README.md` file](https://github.com/plotly/plotly.py/blob/master/doc/README.md)

### DIFF
--- a/doc/python/interactive-html-export.md
+++ b/doc/python/interactive-html-export.md
@@ -72,7 +72,7 @@ You can insert Plotly output and text related to your data into HTML templates u
 
 Then use the following Python to replace `{{ fig }}` in the template with HTML that will display the Plotly figure "fig":
 
-```
+```python
 import plotly.express as px
 from jinja2 import Template
 


Coding-with-Adam:
Thank you for this PR, @Arcader717.
Could you please be more specific.  Which rendering problem are you referring to?

Arcader717:
> Thank you for this PR, @Arcader717.
> Could you please be more specific.  Which rendering problem are you referring to?

Sure. The python example is for using Django, but instead of rendering as a codeblock, the example renders as plain text. The PR should fix this problem

The html example mentioned in the sentence after refers to lines 62-71 in the same file. The issue is that the example renders as part of the html of the page

I hope this clarifies what I was referring to

LiamConnors:
Hi @Arcader717, thanks for the PR! I merged a fix last week to fix some rendering issues with that page
https://github.com/plotly/plotly.py/pull/4479 shortly after the original content was added. So your PR is to fix the same issue, I think. Does it render okay for you now on the page?

Arcader717:
> Does it render okay for you now on the page?

Yes, thank you @LiamConnors 

it renders correctly now

PR #4497, opened 2024-01-26
_bullet has an issue building measure and range bars if the data set …

ksheehy:
…has negative values. The issue is with the method of sorting the data from highest to lowest. The data needs to be sorted so both, the largest positive and negative values are plotted first.

Here is the current implementation - you can see since the sort function puts the -5000 tick mark last and the bar on top of all less negative values.
Screenshot 2024-01-26 130212

Here is a fix.
Screenshot 2024-01-26 130307

I did a little search on this function and it looks like it is being depreciated. I appologize if I have butchered the PR process.

### DIFF
--- a/packages/python/plotly/plotly/figure_factory/_bullet.py
+++ b/packages/python/plotly/plotly/figure_factory/_bullet.py
@@ -88,19 +88,19 @@ def _bullet(
                 range_colors[0], range_colors[1], len(df.iloc[row]["ranges"]), "rgb"
             )
             x = (
-                [sorted(df.iloc[row]["ranges"])[-1 - idx]]
+                [sorted(df.iloc[row]["ranges"], key=abs, reverse=True)[idx]]
                 if orientation == "h"
                 else [0]
             )
             y = (
                 [0]
                 if orientation == "h"
-                else [sorted(df.iloc[row]["ranges"])[-1 - idx]]
+                else [sorted(df.iloc[row]["ranges"], key=abs, reverse=True)[idx]]
             )
             bar = go.Bar(
                 x=x,
                 y=y,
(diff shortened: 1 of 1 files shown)


Coding-with-Adam:
Thank you for reporting this, @ksheehy

Coding-with-Adam:
Hi @archmoj 
Here's the code pen for this issue:
https://codepen.io/charming-data/pen/ExJwxRY

archmoj:
Thanks very much for the PR.
This looks good.
Please add a test in `packages/python/plotly/plotly/tests/test_optional/test_figure_factory/test_figure_factory.py` to lock this bug.

packages/python/plotly/plotly/figure_factory/_bullet.py
```diff
@@ -88,19 +88,19 @@ def _bullet(
...
                 [0]
                 if orientation == "h"
-                else [sorted(df.iloc[row]["ranges"])[-1 - idx]]
+                else [sorted(df.iloc[row]["ranges"], key=abs, reverse=True)[idx]]
             )
             bar = go.Bar(
                 x=x,
                 y=y,
-                marker=dict(color=inter_colors[-1 - idx]),
+                marker=dict(color=inter_colors[sorted(df.iloc[row]["ranges"]).index(sorted(df.iloc[row]["ranges"], key=abs, reverse=True)[idx])]),
```
archmoj, line 103: Formatting
```suggestion
                marker=dict(
                    color=inter_colors[
                        sorted(df.iloc[row]["ranges"]).index(
                            sorted(df.iloc[row]["ranges"], key=abs, reverse=True)[idx]
                        )
                    ]
                ),
```

PR #4524, opened 2024-02-15
Pin sphinx dependencies

LiamConnors:
We are encountering this https://github.com/sphinx-doc/sphinx/issues/11890

```
Running Sphinx v3.5.4

Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
make: *** [Makefile:34: html] Error 2
```

when trying to generate some of the reference docs. 
Setting these to the last versions that worked successfully for the docs build

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -21,6 +21,12 @@ squarify
 scikit-image==0.18.1
 scikit-learn
 sphinx==3.5.4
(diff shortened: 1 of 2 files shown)


archmoj (approved):
Looking good to me.
:dancer:

PR #4562, opened 2024-03-29
Fix `orjson` loading error

### LINKED ISSUE #3567: With newer versions of orjson, users need to specify the json engine explicitly (bug?)
Hey

I found out I get an
`AttributeError: partially initialized module 'orjson' has no attribute 'OPT_NON_STR_KEYS'`
if I don't specify this
`plotly.io.json.config.default_engine = 'orjson'`
when using orjson v3.6.6 (latest as of 25jan2022)

Also, additional note for whoever might have this issue: you don't need to uninstall orjson if you don't want to use it. just set the engine to 'json' explicitly. 

I'm using orjson because of the performance claims, although I ran some tests switching between the 2 engines and they seem to yield the same results: using go.Candlestick with 10000 candlesticks and some 4-5 indicators, getting ~0.8sec in each case for creating the plot. My purpose is to improve the dash server performace, but it seems it makes no difference (the web page still renders slower than the ticker even with 600 candles)

emilykl:
Closes #3567 
Also addresses [plotly/dash#2797](https://github.com/plotly/dash/issues/2797)

## Summary

Address the error described in #3567 by adjusting the module loading logic in `get_module`.

The error can be reproduced by following [these steps](https://github.com/plotly/plotly.py/issues/3567#issuecomment-2025505791).

As best I can tell, the error is caused by a race condition between the multiple threads in the Flask hot reloader, and the fact that Python adds a module to `sys.modules` before the module has been executed [[1]](https://github.com/python/cpython/blob/main/Lib/importlib/_bootstrap.py#L928C9-L935C48) [[2]](https://tenthousandmeters.com/blog/python-behind-the-scenes-11-how-the-python-import-system-works/). If one thread attempts to load `orjson` after it has been added to `sys.modules` but before it has been executed, we'll get the error.

## Fix

- Change logic in `get_module` so that when `should_load=True`, we just call `import_module` right away instead of checking `sys.modules` first. 
- This resolves the issue because `import_module` contains logic to verify that the module is fully initialized
- Logic for when `should_load=False` remains the same: If the module is in `sys.modules` we return it, otherwise we return `None`

## Code PR

  modified existing tests.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -37,5 +37,5 @@ orjson
 dash-bio
 jinja2<3.1
 parmed<=3.4.4; python_version<"3.8"
-polars
 dask==2022.2.0
+polars
--- a/packages/python/plotly/_plotly_utils/optional_imports.py
+++ b/packages/python/plotly/_plotly_utils/optional_imports.py
@@ -23,15 +23,14 @@ def get_module(name, should_load=True):
     if not should_load:
         return sys.modules.get(name, None)
 
-    else:
-        if name not in _not_importable:
-            try:
-                return import_module(name)
-            except ImportError:
-                _not_importable.add(name)
-            except Exception:
-                _not_importable.add(name)
-                msg = f"Error importing optional module {name}"
-                logger.exception(msg)
+    if name not in _not_importable:
+        try:
+            return import_module(name)
+        except ImportError:
+            _not_importable.add(name)
+        except Exception:
+            _not_importable.add(name)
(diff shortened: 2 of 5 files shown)


T4rk1n (approved):
:woman_dancing:

alexcjohnson (approved):
💃 very nice diagnosis & fix!

archmoj:
Some tests are failing on CI.
Over to @LiamConnors :pray:

PR #4594, opened 2024-05-01
Update Plotly.py version for docs

LiamConnors:
This PR updates the docs to use the latest, just released, plotly.py - 5.22
Release step here: https://github.com/plotly/plotly.py/blob/master/release.md#update-documentation-site

Note: Docs build failed for three pages because of an incompatibility between the latest version of geoparse and the library versions we are using in examples.
Pinning it here to the previous version to get the docs for the released merged, and then I can look at updating the requirements for the docs to later versions.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,4 +1,4 @@
-plotly==5.21.0
+plotly==5.22.0
 jupytext==1.1.1
 ipywidgets==7.7.2
 jupyter-client<7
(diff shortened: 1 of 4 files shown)


doc/requirements.txt
```diff
@@ -45,3 +45,4 @@ jinja2<3.1
 parmed<=3.4.4; python_version<"3.8"
 dask==2022.2.0
 polars
+geoparse<=2.0.3
```
archmoj, line 48: Do you know which `geoparse` version used before?

LiamConnors: The last successful build used 2.0.3

gvwilson (approved):

Issue #4491, opened 2024-01-22
Update all docs examples to run on Python 3.10 with later libraries


LiamConnors:
https://github.com/plotly/plotly.py/pull/4490

PR #4190, opened 2023-05-05
Px pandas2

alexcjohnson:
Fixes #4179 
Fixes #4065 

Pandas v2.0 removed `df.append`, need to change to `pd.concat`. Note that `pd.concat` has been available since at least [v0.13](https://pandas.pydata.org/pandas-docs/version/0.13/generated/pandas.concat.html) so there should be no issue with older versions _requiring_ that we use `df.append`.

Tested by having the `px` Percy tests generate the same HTML files with both `pandas==1.1.5` (the oldest version with a Py3.9 wheel) and `pandas==2.0.1` (latest) and check that the HTML is identical before shipping it off to Percy.

- [x] I have read through the [contributing notes](https://github.com/plotly/plotly.py/blob/master/contributing.md) and understand the structure of the package. In particular, if my PR modifies code of `plotly.graph_objects`, my modifications concern the `codegen` files and not generated files.
- [x] I have added tests (if submitting a new feature or correcting a bug) or
  modified existing tests.
- [x] I have added a CHANGELOG entry if fixing/changing/adding anything substantial.

PR #4596, opened 2024-05-01
Update docs dependencies and Python version

LiamConnors:
This PR:

- Updates the Python version used to build the docs to Python 3.10
- Updates most libraries to the latest versions
- Fixes examples that no longer work with later versions. - e.g. removal of `np.int` and `df.append`. https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#removal-of-prior-version-deprecations-changes
- Removes pinned `geoparse` from https://github.com/plotly/plotly.py/pull/4594 - no longer required with latest version of `pandas`
- Specifies xarray version as I encountered the following when updating: https://github.com/plotly/plotly.py/issues/4598

To close https://github.com/plotly/plotly.py/issues/4491


doc/python/icicle-charts.md
```diff
@@ -305,7 +305,7 @@ def build_hierarchical_dataframe(df, levels, value_column, color_columns=None):
     Levels are given starting from the bottom to the top of the hierarchy,
     ie the last level corresponds to the root.
     """
-    df_all_trees = pd.DataFrame(columns=['id', 'parent', 'value', 'color'])
+    df_trees = []
```
gvwilson, line 308: I don't understand this change but I'm willing to take it on faith :-)

gvwilson (approved):
There are three places where it appears that a single call to Pandas has been replaced by a loop. I don't have enough context to understand the change but am willing to take it on faith; would be happy to discuss.

doc/python/icicle-charts.md
emilykl: @LiamConnors My only suggestion here (and at the other spot) is that the `df_trees` variable be renamed to clarify that it's a _list_ rather than a df.

emilykl:
@LiamConnors Are the Pandas example code changes still backwards-compatible with Pandas 1.x?

doc/python/icicle-charts.md
LiamConnors: Yeah, sorry, I should have given more context. 

The original example uses `df.append`, to append each new dataframe in the loop to the first dataframe, but that method was removed. 

So here I instead use a list, and add each to a list and use `pd.concat` on the list instead.

We had to make this update also in the `plotly.express` library itself at one stage and that's the approach we took: https://github.com/plotly/plotly.py/pull/4190

@emilykl good suggestion. Will update the variable name
`pd.concat` has existed at least since https://pandas.pydata.org/pandas-docs/version/0.13/generated/pandas.concat.html

PR #4600, opened 2024-05-03
Update radar-chart.md

hailorrr:


Coding-with-Adam (commented):
hi @hailorrr 
Thank you for this PR. Can you please share more about your suggested changes. Why do you suggest using those new theta labels?

PR #4612, opened 2024-05-17
Handle `extras` option in `IntegerValidator`

archmoj:
`extras` are added to plotly.js integer validator in https://github.com/plotly/plotly.js/pull/6990/commits/37dee4852e335067508add73de7ee7a59e8db045
i.e. to handle cases like *bold* and *normal* in `font.weight` which could also be a number between 1-1000.

This commit would be required in updating to upcoming plotly.js v2.33.0.

@LiamConnors 
cc: @emilykl @gvwilson

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Updated
 - Specify Python version 3.8-3.11 for development virtual environments and pin `pytest` at version 8.1.1 to match.
+- Update `IntegerValidator` to handle `extras` option to allow supporting additional keyword values. For example, 'bold' and 'normal' as well as integers as used in font weights [#4612].
 
 ## [5.22.0] - 2024-05-01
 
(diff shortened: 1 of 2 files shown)


LiamConnors (approved):

emilykl:
Logic looks good to me. Thanks @archmoj !

Do we have tests for the validators? Can we add a test for this change?

archmoj:
> Logic looks good to me. Thanks @archmoj !
> 
> Do we have tests for the validators? Can we add a test for this change?

Good call. 
It would be nice to add tests for this PR.
@LiamConnors you may add a test in [packages/python/plotly/_plotly_utils/tests/validators/test_dash_validator.py](https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/_plotly_utils/tests/validators/test_integer_validator.py)

Thank you!

archmoj:
@LiamConnors The tests used to pass before [your commit](https://github.com/plotly/plotly.py/pull/4612/commits/c0a775703427f80c2f3a0cb2b342620f6376bc4b) which simply added a line to the changelog. But now [it fails](https://app.circleci.com/pipelines/github/plotly/plotly.py/3242/workflows/73d0a877-caa9-4ea9-839a-68c070cee807/jobs/85466/parallel-runs/0/steps/0-108).
Do you have any idea why this is happening?

LiamConnors:
@archmoj if you update from master to bring in this change https://github.com/plotly/plotly.py/pull/4625
it should fix it

archmoj:
I'll add tests in a separate PR later.

PR #4644, opened 2024-06-19
Add plotly[express] extra for easily installing Plotly Express dependencies

### LINKED ISSUE #2279: Requires pandas, but not documented?
I'm new to plotly. I installed plotly with `pip install plotly==4.5.4` and when I tried to generate a graph, I got 'ImportError: Plotly express requires pandas to be installed.'

The getting-started page says "This package contains everything you need to write figures to standalone HTML files." I looked through the documentation, but couldn't find anywhere that said pandas was required, or how to install pandas.

To reproduce:
```
>>> import plotly.express as px
  File "<stdin>", line 1, in <module>
  File "/Users/ken/blog/corona/env/lib/python3.7/site-packages/plotly/express/__init__.py", line 12, in <module>
    Plotly express requires pandas to be installed."""
ImportError: Plotly express requires pandas to be installed.
```
Maybe this is intentional (see #2125), but I think the documentation should be clearer. Automatically installing pandas would be even better.

ndrezn:
Closes #2279

This PR:
- Adds the install shortcut `plotly[express]`, explained in more detail below
- Renames the following files, to match conventions used in other Plotly repos:
  - `requirements.txt` to `requires-install.txt`
  - `optional-requirements.txt` to `requires-optional.txt`
- Adds an additional file `requires-express.txt`, which contains the dependencies required by Plotly Express which are not required by the base Plotly package (currently, this is just `numpy`)
- Adds additional detail to the error message displayed when a user tries to import Plotly Express without having numpy installed
- Does a very light round of cleanup on `requires-optional.txt`, removing outdated dependencies which are definitely not needed anymore

------

This PR adds the ability to install Plotly with:
```
pip install plotly[express]
```
which bundles `numpy` in the install, which is required by `plotly.express`,

This also revises the error message raised when using `plotly.express` if `numpy` is not installed, providing further instructions on how to install.

Open to ideas on better wording.

```
(plotly.py) ➜  plotly.py git:(better-pandas-warning) python
Python 3.12.3 (main, Apr  9 2024, 08:09:14) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import plotly.express
  File "<stdin>", line 1, in <module>
  File "/Users/nathandrezner/Plotly/plotly.py/packages/python/plotly/plotly/express/__init__.py", line 10, in <module>
    raise ImportError(
ImportError: Plotly Express requires numpy to be installed. You can install numpy using pip with:

$ pip install numpy

Or install Plotly Express and its dependencies directly with:

$ pip install "plotly[express]"

You can also use Plotly Graph Objects to create a large number of charts without installing
numpy. See examples here: https://plotly.com/python/graph-objects/
```


ndrezn:
We can plausibly drop the note about graph objects but basically I want some sort of explanation of why `pandas` isn't a dependency.

emilykl:
We shouldn't recommend `pip install plotly.express` , I believe it only exists for backwards compatibility and it's confusing to have multiple install paths.

If we want to have a one-liner "install everything" command that should be a change made inside Plotly.py.

ndrezn:
I was thinking of adding:
```
pip install plotly[express]
```
as an option which goes with usual pip patterns for optional dependencies. If we went that route I would expect that we drop support for `pip install plotly.express` entirely.

.gitignore
```diff
@@ -2,7 +2,7 @@
 0.html
 iframe_figures/
 packages/python/plotly/plotly/tests/test_orca/images/linux/failed/
-
+*.DS_Store
```
gvwilson, line 5: is this related to the pandas warning? if not, should it be in a separate PR?

.gitignore
```diff
@@ -62,3 +62,7 @@ packages/python/plotly/jupyterlab_plotly/nbextension/index.js*
 
 test/percy/*.html
 test/percy/pandas2/*.html
+
+# Build artifacts
+packages/python/plotly/plotly/package_data/datasets/amphibians.csv.gz
```
gvwilson, line 67: same as above - are these files related to the pandas warning? if not, should they be in a separate single-purpose PR?

contributing.md
```diff
@@ -137,17 +137,17 @@ conda activate plotly-dev
 
 ### Install requirements - (Non-Windows)
 ```bash
-(plotly_dev) $ pip install -r packages/python/plotly/requirements.txt
-(plotly_dev) $ pip install -r packages/python/plotly/optional-requirements.txt
+(plotly_dev) $ pip install -r packages/python/plotly/requires-install.txt
```
gvwilson, line 140: are these related to the pandas warning change?

contributing.md
```diff
@@ -170,7 +170,7 @@ documentation on _development mode_.
 This repo uses the [Black](https://black.readthedocs.io/en/stable/) code formatter,
 and the [pre-commit](https://pre-commit.com/) library to manage a git commit hook to
 run Black prior to each commit.  Both pre-commit and black are included in the
-`packages/python/plotly/optional-requirements.txt` file, so you should have them
+`packages/python/plotly/requires-dev.txt` file, so you should have them
```
gvwilson, line 173: is this change required by the pandas warning change?

contributing.md
```diff
@@ -258,7 +258,7 @@ We take advantage of two tools to run tests:
 
 ### Running Tests with `pytest`
 
-Since our tests cover *all* the functionality, to prevent tons of errors from showing up and having to parse through a messy output, you'll need to install `optional-requirements.txt` as explained above.
+Since our tests cover *all* the functionality, to prevent tons of errors from showing up and having to parse through a messy output, you'll need to install `requires-dev.txt` as explained above.
```
gvwilson, line 261: again, related to pandas warning?

gvwilson (approved):
I assumed from the PR title that this was going to improve the install-time or runtime error messages for "you need Pandas" but there seem to be other changes as well? I trust you...

ndrezn:
@gvwilson there are other changes following @emilykl 's comment. It might make sense to break this into two PRs to be honest...

gvwilson:
:+1: on breaking the PR - thank you

.gitignore
ndrezn: I'll remove this commit from this PR.

contributing.md
ndrezn: No, this is related to the updates to requirements files.

emilykl:
IMO the naming should be `plotly[pandas]` rather than `plotly[express]` -- much more clear what it's actually doing.

I'm in favor of this change, but let's not merge it right away. This PR changes a number of lines in `setup.py` so it's not just an addition; it could feasibly break the install process for some users. Need to do really solid QA to make sure it runs smoothly.

ndrezn:
@gvwilson on further thought I think it makes sense to keep these changes in one PR. The text in the warning message is directly related to the installation recommendations that are also added in this PR -- I've updated the title & description to make the primary change more clear.

gvwilson:
:+1:

packages/python/plotly/requires-pandas.txt
```diff
@@ -0,0 +1,2 @@
+pandas; python_version <= '3.8' # Loose requirement for older Python as older numpy is not supported
+pandas>=2.2.0; python_version > '3.0' # Minimum pin to ensure compatibility with all versions of numpy
```
alexcjohnson, line 2:
```suggestion
pandas>=2.2.0; python_version > '3.8' # Minimum pin to ensure compatibility with all versions of numpy
```
Otherwise this will combine with the other one and this will win for all Py3.x, right?

ndrezn: Yep... typo... good catch

ndrezn: Actually I realize this is just wrong. Should be <3.9; >=3.9 rather than <=3.9; >3.8. AFK but I'll fix when I'm back.

ndrezn:
@emilykl is this PR good to go with those changes?

ndrezn:
@marthacryan could you take a look at this?

gvwilson:
See also #4790

marthacryan:
@ndrezn So is `pip install plotly[pandas]` exactly the same as `pip install pandas`?

emilykl:
@marthacryan `pip install plotly[pandas]` would be roughly equivalent to `pip install plotly && pip install pandas` I believe.

marthacryan:
Oh interesting. What is the benefit here over just using `pip install plotly pandas`?

ndrezn:
The benefit is we recommend a specific version number for compatibility. I don't think we have issues with different versions of Pandas but this makes it easy to pin a version of Pandas for use with Plotly if we ever encounter issues.

ndrezn:
@marthacryan / @emilykl bump

gvwilson:
conversation with @ndrezn he believes this should go in before #4790 (Narwhals support) so moving it to the 3.0 release cycle.

marthacryan:
> The benefit is we recommend a specific version number for compatibility. I don't think we have issues with different versions of Pandas but this makes it easy to pin a version of Pandas for use with Plotly if we ever encounter issues.
> 
> In fact we're doing that here, where we install the latest pandas for new versions of Python but force older pandas for older Python. People in the original issue were having funky errors caused by this incompatibility: [#2279 (comment)](https://github.com/plotly/plotly.py/issues/2279#issuecomment-721024313)

Oh ok that makes sense! Maybe for documenting the usage of that we could make that more clear?

packages/python/plotly/requires-express.txt
```diff
@@ -0,0 +1,8 @@
+### Required dependencies for Plotly Express    ###
+###                                             ###
+### To install, run:                            ###
+###      $ pip install -r requires-express.txt  ###
+###                                             ###
+###################################################
+
+numpy
```
ndrezn, line 8: We should add a version pin here, we want to recommend a specific version of Numpy

packages/python/plotly/plotly/express/__init__.py
```diff
@@ -2,13 +2,24 @@
...
     raise ImportError(
         """\
-Plotly express requires numpy to be installed."""
+Plotly Express requires numpy to be installed. You can install numpy using pip with:
+
+$ pip install numpy
+
+Or install Plotly Express and its dependencies directly with:
+
+$ pip install "plotly[express]"
```
ndrezn, line 18: Previously we settled on `plotly[pandas]` to install Plotly with our recommended version of pandas; do we want to stick with that (`plotly[numpy]`) or make it "the dependencies for Plotly Express"? If we have other dependencies then `plotly[express]` could e.g. require `plotly[numpy]` & whatever else.

I don't have a strong opinion but something to make an active decision on.

packages/python/plotly/requires-install.txt
```diff
@@ -1,9 +1,10 @@
...
 ###                                             ###
 ### To install, run:                            ###
-###      $ pip install -r requirements.txt      ###
+###      $ pip install -r requires-install.txt  ###
 ###                                             ###
 ###################################################
 
 ## dataframe agnostic layer ##
 narwhals>=1.13.3
+packaging
```
ndrezn, line 10: Do we want a version pin?

packages/python/plotly/plotly/express/__init__.py
emilykl: If I remember correctly you originally proposed `plotly[express]` and I'm the one who argued for `plotly[pandas]` instead 😄 mea cupla

On reflection, and especially since the Plotly Express dependencies have changed following the Narwhals integration, I think it makes more sense to have `plotly[express]` as shorthand for "Plotly, plus whatever dependencies happen to be required by the current version of Plotly Express". That way if the dependencies change in the future we don't have to change the install name.

packages/python/plotly/requires-express.txt
emilykl: I believe we shouldn't pin specific versions unless absolutely necessary — this is why:

https://iscinumpy.dev/post/bound-version-constraints/

packages/python/plotly/plotly/express/__init__.py
ndrezn: 🫡 lol, sounds good to me!

packages/python/plotly/requires-install.txt
emilykl: No

ndrezn: 🆗

contributing.md
```diff
@@ -138,17 +138,17 @@ We will support Python 3.12 and higher versions soon.
 
 ### Install requirements - (Non-Windows)
 ```bash
-(plotly_dev) $ pip install -r packages/python/plotly/requirements.txt
-(plotly_dev) $ pip install -r packages/python/plotly/optional-requirements.txt
+(plotly_dev) $ pip install -r packages/python/plotly/requires-install.txt
```
LiamConnors, line 141: Is this step required in contributing.md? I think it's covered by the editable install of `plotly` further down.

gvwilson: @ndrezn @emilykl if you can spare time today, it would be great to get this one in - thanks

LiamConnors:
Looks good to me. Could you add a changelog entry for it too.

contributing.md
emilykl: good catch @LiamConnors , I'll remove this

(I verified that the setup instructions still work without this step, at least on Mac)

gvwilson (approved):

PR #4663, opened 2024-07-14
Switch to use anywidget

marthacryan:
Co-authored by @manzt! This updates the logic for the `FigureWidget` to rely on [anywidget](https://anywidget.dev/).
A quote about anywidget from the documentation for context:
> anywidget is both a [specification](https://anywidget.dev/en/afm) and toolset for authoring reusable web-based widgets for interactive computing environments. As a specification, it defines a standard for widget front-end code. As a toolkit, it provides tools for authoring widget front-end code according to the specification, as well as a Python library that simplifies creating and publishing custom [Jupyter Widgets](https://ipywidgets.readthedocs.io/en/latest/).

**This will not affect the `go.Figure` class**. High level explanation of the difference between using `go.Figure` and `go.FigureWidget`:
* `go.Figure`: The chart generated by this object will be interactive on the **client side**. For many purposes, this is all a user needs to do with a chart. The interactions are things like zooming in, or any of the various operations in the toolbar.
* `go.FigureWidget`: The difference here is that the figure is interactive on the **server side** as well.
  * Interactions with the plotly chart can be sent to the python kernel and used in other widgets.
  * Interactions with other widgets can be used to control the chart


.circleci/config.yml
```diff
@@ -41,6 +41,7 @@ commands:
       - run:
           name: Install dependencies
           command: |
+            cd packages/javascript/jupyterlab-plotly && npm run bundle && cd ../../..
```
manzt, line 44: I _think_ we don't need this anymore after updating `setup.py`

packages/python/plotly/js/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "scripts": {
+    "build": "esbuild --bundle --alias:plotly.js=plotly.js/dist/plotly --format=esm --minify --outfile=../plotly/bundle.js widget.ts",
+    "watch": "npm run build -- --watch --sourcemap=inline",
```
manzt, line 4: sourcemaps will make it easier to debug when developing the widget.

packages/python/plotly/js/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "scripts": {
+    "build": "esbuild --bundle --alias:plotly.js=plotly.js/dist/plotly --format=esm --minify --outfile=../plotly/bundle.js widget.ts",
```
manzt, line 3: I added the `--minify` and `--alias` flags. `minify` makes the final bundle smaller. 

The alias here is to make the import nicer for typescript type checking types (tsc), while esbuild needs to point to the actual files (./dist).

PR #4679, opened 2024-07-18
Specify which maps require token

LiamConnors:


emilykl (approved):
This is great, thanks @LiamConnors 👍 

This entire property may change in the future as part of https://github.com/plotly/plotly.js/pull/7015, but in the meantime it's great to have this info directly in the docstring.

PR #4690, opened 2024-07-23
Revert some docs dependencies updates

LiamConnors:
The version of `sphinx` that we use to build the reference docs isn't compatible with Python 3.10.

So I'm reverting some of the changes in https://github.com/plotly/plotly.py/pull/4596 to get the reference docs building again so we can get the new ones for 5.23 out.

Will need to take more time to update and test `sphinx`


gvwilson (approved):
:+1: go ahead and merge

PR #4695, opened 2024-07-26
Add tests for b64 PR

marthacryan:
- [x] Add 2 speed performance tests for scatter plot: float32 and float64
- [x] Add speed performance test for scatter3d
- [x] Add 2 performance size tests for scatter plot: uint8 and uint32
- [x] Checking all the arrayOK attributes validate & render properly

In another PR:
- [ ] Test every place in validation that we added b64 (see: marker support for numpy number types)

### DIFF
--- a/packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
+++ b/packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
@@ -20,57 +20,67 @@ def test_performance_b64_scatter3d():
     z_list = z.tolist()
     c_list = c.tolist()
     list_start = time.time()
-    fig = go.Figure(
-        data=[
-            go.Scatter3d(
-                x=x_list,
-                y=y_list,
-                z=z_list,
-                marker=dict(color=c_list),
-                mode="markers",
-                opacity=0.2,
-            )
-        ]
-    )
+    fig = go.Figure(data=[go.Scatter3d(
+        x=x_list,
+        y=y_list,
+        z=z_list,
+        marker=dict(color=c_list),
+        mode="markers",
+        opacity=0.2,
+    )])
+    fig.show()
     list_time_elapsed = time.time() - list_start
 
     # Test the performance with base64 arrays
     np_start = time.time()
-    fig = go.Scatter3d(
+    fig = go.Figure(data=[go.Scatter3d(
         x=x,
         y=y,
         z=z,
         marker=dict(color=c),
         mode="markers",
         opacity=0.2,
-    )
+    )])
+    fig.show()
     np_time_elapsed = time.time() - np_start
 
     # np should be faster than lists
-    assert (np_time_elapsed / list_time_elapsed) < 0.005
+    assert (np_time_elapsed / list_time_elapsed) < 0.5
 
-
-def test_performance_b64_float64():
-    np_arr_1 = np.random.random(10000)
-    np_arr_2 = np.random.random(10000)
+FLOAT_TEST_CASES = [
+    (
+        "float32", # dtype
+        0.45 # difference threshold
+    ),
+    (
+        'float64',
+        0.55
+    )
+]
+@pytest.mark.parametrize('dtype, expected_size_difference', FLOAT_TEST_CASES)
+def test_performance_b64_float(dtype, expected_size_difference):
+    np_arr_1 = np.random.random(10000).astype(dtype)
+    np_arr_2 = np.random.random(10000).astype(dtype)
     list_1 = np_arr_1.tolist()
     list_2 = np_arr_2.tolist()
 
     # Test the performance of the base64 arrays
     np_start = time.time()
-    fig = go.Scatter(x=np_arr_1, y=np_arr_2)
+    fig = go.Figure(data=[go.Scatter(x=np_arr_1, y=np_arr_2)])
+    fig.show()
     np_time_elapsed = time.time() - np_start
 
     # Test the performance of the normal lists
     list_start = time.time()
-    fig = go.Scatter(x=list_1, y=list_2)
+    fig = go.Figure(data=[go.Scatter(x=list_1, y=list_2)])
+    fig.show()
     list_time_elapsed = time.time() - list_start
 
     # np should be faster than lists
-    assert (np_time_elapsed / list_time_elapsed) < 0.3
+    assert (np_time_elapsed / list_time_elapsed) < expected_size_difference
 
 
-DTYPE_TEST_CASES = [
+INT_SIZE_PERFORMANCE_TEST_CASES = [
     (
         "uint8", # dtype
         256, # max_val
@@ -82,8 +92,8 @@ DTYPE_TEST_CASES = [
         900000
     )
 ]
-@pytest.mark.parametrize('dtype, max_val, expected_size_difference', DTYPE_TEST_CASES)
-def test_size_performance_b64_uint8(dtype, max_val, expected_size_difference):
+@pytest.mark.parametrize('dtype, max_val, expected_size_difference', INT_SIZE_PERFORMANCE_TEST_CASES)
+def test_size_performance_b64_int(dtype, max_val, expected_size_difference):
     np_arr_1 = (np.random.random(100000) * max_val).astype(dtype)
     np_arr_2 = (np.random.random(100000) * max_val).astype(dtype)
 
--- a/packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
+++ b/packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
@@ -1,6 +1,7 @@
 import time
 import numpy as np
 import plotly.graph_objs as go
+import pytest
 
 np.random.seed(1)
 
@@ -14,14 +15,18 @@ def test_performance_b64_scatter3d():
     c = np.random.randint(size=N, low=-10, high=10, dtype="int8")
 
     # Test the performance with lists
+    x_list = x.tolist()
+    y_list = y.tolist()
+    z_list = z.tolist()
+    c_list = c.tolist()
     list_start = time.time()
     fig = go.Figure(
         data=[
             go.Scatter3d(
-                x=x.tolist(),
-                y=y.tolist(),
-                z=z.tolist(),
-                marker=dict(color=c.tolist()),
+                x=x_list,
+                y=y_list,
+                z=z_list,
+                marker=dict(color=c_list),
                 mode="markers",
                 opacity=0.2,
             )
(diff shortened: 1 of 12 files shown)


packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
```diff
@@ -138,6 +139,45 @@ def test_name_heuristics():
...
+    b64_time_elapsed = time.time() - b64_start
+
+    # Test the performance of the raw arrays
+    raw_start = time.time()
+    df_raw = pd.DataFrame(dict(x=rand_arr_1, y=rand_arr_2))
+    fig = px.scatter(df_raw, x="x", y="y")
+    raw_time_elapsed = time.time() - raw_start
+
+    # b64 should be faster than raw
+    assert b64_time_elapsed < raw_time_elapsed
```
archmoj, line 161: Let's test that this would run e.g. 15% faster. 
Something like: 
```puthon
   assert (b64_time_elapsed / raw_time_elapsed) < 0.85
```

packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
```diff
@@ -138,6 +139,45 @@ def test_name_heuristics():
     assert fig.data[0].hovertemplate == "y=%{marker.size}<br>x=%{y}<extra></extra>"
 
 
+def test_performance_b64():
+    rand_arr_1 = np.array(np.random.random(100000))
```
archmoj, line 143: Please check this part of checklist: 
```
- [ ] The random seed is set if using randomly-generated data in new/modified examples
```

packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
```diff
@@ -138,6 +139,45 @@ def test_name_heuristics():
...
+    rand_arr_2 = np.array(np.random.random(100000))
+    b64_arr_1 = b64(rand_arr_1)
```
archmoj, line 145: The `rand_arr_1` is already a numpy array, right?
I think we should cast the numpy array to a list using `.tolist()` and then compare the results of numpy(b64) vs normal arrays.

marthacryan: You're totally right - no need to convert it to a numpy array!
But as for the casting `.tolist()`, this works for the speed performance test but fails on size performance tests (in those cases it performs the same). I'll leave the speed performance with the `.tolist()` and leave it out for the size performance.

packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
```diff
@@ -138,6 +140,64 @@ def test_name_heuristics():
...
 
+def test_performance_b64():
+    rand_arr_1 = np.random.random(100000)
+    rand_arr_2 = np.random.random(100000)
+    b64_arr_1 = b64(rand_arr_1)
+    b64_arr_2 = b64(rand_arr_2)
+
+    # Test the performance of the base64 arrays
+    b64_start = time.time()
+    df_b64 = pd.DataFrame(dict(x=b64_arr_1, y=b64_arr_2))
```
archmoj, line 151: Let's not use `pandas.DataFrame` here that could be slow.

packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
```diff
@@ -138,6 +140,64 @@ def test_name_heuristics():
...
+    fig = px.scatter(df_b64, x="x", y="y", width=800, height=800)
```
archmoj, line 152: In this case it might be better to use `go.Figure(data=go.Scatter(...))` instead of `px.scatter`.

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
```diff
@@ -0,0 +1,47 @@
...
+    fig = go.Scatter(x=b64_arr_1, y=b64_arr_2)
+    b64_time_elapsed = time.time() - b64_start
+
+    # Test the performance of the raw arrays
+    raw_start = time.time()
+    fig = go.Scatter(x=raw_arr_1, y=raw_arr_2)
+    raw_time_elapsed = time.time() - raw_start
+
+    # b64 should be faster than raw
+    assert (b64_time_elapsed / raw_time_elapsed) < 0.85
```
archmoj, line 30: Could you print out the actual ratio to see if we could further reduce `0.85`?

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def test_size_performance_b64_uint8():
+    rand_arr_1 = np.random.random(100000).astype("uint8")
+    rand_arr_2 = np.random.random(100000).astype("uint8")
```
archmoj, line 35: We need to multiply `[0, 1)` random numbers by 256 before casting to `uint8`; otherwise what would get just zeros.
```suggestion
    rand_arr_1 = (np.random.random(100000) * 256).astype("uint8")
    rand_arr_2 = (np.random.random(100000) * 256).astype("uint8")
```

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
```diff
@@ -0,0 +1,47 @@
...
+from plotly.tests.b64 import b64, _b64
+
+
+def test_performance_b64():
+    rand_arr_1 = np.random.random(100000)
+    rand_arr_2 = np.random.random(100000)
+    raw_arr_1 = rand_arr_1.tolist()
+    raw_arr_2 = rand_arr_2.tolist()
+    b64_arr_1 = b64(rand_arr_1)
+    b64_arr_2 = b64(rand_arr_2)
```
archmoj, line 17: Do we need `b64` conversion here?
Can't we simply pass `rand_arr_1` `numpy` input data?

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
```diff
@@ -0,0 +1,47 @@
...
+import numpy as np
+import plotly.graph_objs as go
+import plotly.io as pio
+from plotly.tests.b64 import b64, _b64
+
+
+def test_performance_b64():
+    rand_arr_1 = np.random.random(100000)
+    rand_arr_2 = np.random.random(100000)
+    raw_arr_1 = rand_arr_1.tolist()
```
archmoj, line 14: Using `raw` for this variable is rather confusing; whereas raw data here should point to raw user data in numpy format.
Therefore I'd suggest we use `normal_arr` instead of `raw_arr`.

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
```diff
@@ -0,0 +1,47 @@
...
+from unittest import TestCase
+import pytest
+import numpy as np
+import plotly.graph_objs as go
+import plotly.io as pio
+from plotly.tests.b64 import b64, _b64
+
+
+def test_performance_b64():
+    rand_arr_1 = np.random.random(100000)
```
archmoj, line 12: These are `float64` numbers by current numpy default.
How about clarifying the test name to be `test_performance_b64_float64` similar to `unit8` one.

archmoj:
For `scatter3d` one you may consider tweaking this example: 
```py
import numpy as np

np.random.seed(1)

N = 10000

x = np.random.randn(N)
y = np.random.randn(N).astype('float32')
z = np.random.randint(size=N, low=0, high=256, dtype='uint8')
c = np.random.randint(size=N, low=-10, high=10, dtype='int8')

fig = go.Figure(data=[go.Scatter3d(
    x=x,
    y=y,
    z=z,
    marker=dict(color=c),
    mode='markers',
    opacity=0.2
)])

fig.show()

print(fig.to_json())
```

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
archmoj: Alternative you could use `randint` similar to https://github.com/plotly/plotly.py/pull/4695#issuecomment-2263263589.

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
archmoj: Since the `to_json` return full figure it contains lots of other setup inside `layout.template`.
I think it would be better to compare the difference between those two sizes e.g. 
```b64_time_elapsed -normal_time_elapsed < 12345```

Comment (marthacryan) [reply packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py:30]: I think the ratio might make more sense here because different environments will have different elapsed times, but hopefully they'll be proportionally slower if they're in a slower environment.

File: packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py (hunk 8 above)
Comment (marthacryan) [reply packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py:17]: I'm confused - that variable is a numpy array, so it wouldn't be in the b64 format?

File: packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py (hunk 9 above)
Comment (marthacryan) [reply packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py]: I changed it to `list_1` and `list_2`, does that work?

File: packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py (hunk 6 above)
Comment (archmoj) [reply packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py]: Yes I wanted this for sizes not performance :)

File: packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py (hunk 8 above)
Comment (archmoj) [reply packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py]: Exactly. #4470 is about passing numpy arrays directly. So yes we shouldn't need to cast to b64 here.

File: packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py (hunk 9 above)
Comment (archmoj) [reply packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py]: Perfect. Less is more.

File: packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py (hunk 6 above)
Comment (marthacryan) [reply packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py]: Ah ok will do

File: packages/python/plotly/test_requirements/requirements_312_core.txt [hunk 11]
```diff
@@ -1,3 +1,4 @@
 requests==2.25.1
 tenacity==6.2.0
 pytest==7.4.4
+numpy==1.23.2
```
archmoj, line 4: Why we have to pin down this version here?

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
```diff
@@ -0,0 +1,79 @@
...
+        y=y,
+        z=z,
+        marker=dict(color=c),
+        mode="markers",
+        opacity=0.2,
+    )
+    np_time_elapsed = time.time() - np_start
+
+    # np should be faster than raw
+    assert (np_time_elapsed / list_time_elapsed) < 0.75
```
archmoj, line 45: Could we further reduce the 0.75 ratio here?

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
```diff
@@ -0,0 +1,79 @@
...
+    fig = go.Scatter(x=np_arr_1, y=np_arr_2)
+    np_time_elapsed = time.time() - np_start
+
+    # Test the performance of the raw arrays
+    list_start = time.time()
+    fig = go.Scatter(x=np_arr_1.tolist(), y=np_arr_2.tolist())
+    list_time_elapsed = time.time() - list_start
+
+    # np should be faster than raw
+    assert (np_time_elapsed / list_time_elapsed) < 0.75
```
archmoj, line 63: Could we further reduce the 0.75 ratio here?

packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
```diff
@@ -138,6 +140,52 @@ def test_name_heuristics():
...
+    np_arr_time_elapsed = time.time() - np_arr_start
+
+    # Test the performance of the raw arrays
+    list_start = time.time()
+    fig = px.scatter(x=np_arr_1.tolist(), y=np_arr_2.tolist(), width=800, height=800)
+    fig.update_layout(margin=dict(l=0, r=0, t=0, b=0))
+    list_time_elapsed = time.time() - list_start
+
+    # b64 should be faster than raw
+    assert (np_arr_time_elapsed / list_time_elapsed) < 0.7
```
archmoj, line 160: Let's add `test_performance_b64_float32` as well as for size which could demonstrate similar/better results.
Thank you!

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def test_size_performance_b64_uint8():
```
archmoj, line 66: Also please parameterize/add tests for `uint32`.
Thank you!

packages/python/plotly/test_requirements/requirements_312_core.txt
marthacryan: Oh that totally fixed the CI. Thanks!

packages/python/plotly/plotly/tests/test_core/test_graph_objs/test_performance.py
marthacryan: I did some tests and it's hovering around .7 so I think that's as low as it'll go

packages/python/plotly/test_requirements/requirements_310_core.txt
```diff
@@ -1,3 +1,4 @@
 requests==2.25.1
 tenacity==6.2.0
 pytest==7.4.4
+numpy==1.23.2
```
archmoj, line 4: I am wondering why we need to adjust these requirements?
Don't we have other tests that load `numpy`?
If so, can't we move these tests to run with those together?

marthacryan: Oh yeah the tests that use numpy normally are in `optional`. Do you want me to move it there?

archmoj (approved):
Please revise the PR description a bit to better reflect the kind of tests added.
:dancer:
Nicely completed.
Thanks @marthacryan :trophy: :star2: :medal_sports:

PR #4707, opened 2024-08-05
Base64 Validator tests

marthacryan:
Adds validator tests for #4470 

## Code PR

  modified existing tests.
~- [ ] For a new feature, I have added documentation examples in an existing or
  new tutorial notebook (please see the doc checklist as well).~
~- [ ] I have added a CHANGELOG entry if fixing/changing/adding anything substantial.~
~- [ ] For a new feature or a change in behaviour, I have updated the relevant docstrings in the code to describe the feature or behaviour (please see the doc checklist as well).~

### DIFF
--- a/packages/python/plotly/_plotly_utils/tests/validators/test_angle_validator.py
+++ b/packages/python/plotly/_plotly_utils/tests/validators/test_angle_validator.py
@@ -3,6 +3,7 @@ import pytest
 # from ..basevalidators import AngleValidator
 from _plotly_utils.basevalidators import AngleValidator
 import numpy as np
+from plotly.tests.b64 import b64
 
 
 # Fixtures
@@ -50,6 +51,13 @@ def test_aok_acceptance(val, validator_aok):
     assert validator_aok.validate_coerce(tuple(val)) == val
     assert np.array_equal(validator_aok.validate_coerce(np.array(val)), np.array(val))
 
+
+# Test base64 array
+def test_aok_base64_array(validator_aok):
+    val = b64(np.array([1, 2, 3], dtype="int64"))
+    coerce_val = validator_aok.validate_coerce(val)
+    assert coerce_val["bdata"] == "AQID"
+    assert coerce_val["dtype"] == "i1"
 
 # ### Test coercion above 180 ###
 @pytest.mark.parametrize(
--- a/packages/python/plotly/_plotly_utils/tests/validators/test_angle_validator.py
+++ b/packages/python/plotly/_plotly_utils/tests/validators/test_angle_validator.py
@@ -59,6 +59,7 @@ def test_aok_base64_array(validator_aok):
     assert coerce_val["bdata"] == "AQID"
     assert coerce_val["dtype"] == "i1"
 
+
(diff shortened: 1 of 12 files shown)


packages/python/plotly/_plotly_utils/tests/validators/test_angle_validator.py
```diff
@@ -51,6 +52,14 @@ def test_aok_acceptance(val, validator_aok):
     assert np.array_equal(validator_aok.validate_coerce(np.array(val)), np.array(val))
 
 
+# Test base64 array
+def test_aok_base64_array(validator_aok):
+    val = b64(np.array([1, 2, 3], dtype="int64"))
```
archmoj, line 57: Great that you tested `int64` i.e. [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) in JavaScript.
Since we decided not to handle BigInts on the plotly.js side (for now) `int64` and `uint64` are passed as other types in regards with the minimum and maximum values used in the array. 
See here: https://github.com/plotly/plotly.py/blob/644992f0fe4c3f869ba45e5627e65bf1f6537f7c/packages/python/plotly/_plotly_utils/basevalidators.py#L89-L112

It would be great if you add tests for each of those cases.
Thank you!

archmoj: The `angle` valType is restricted to small integers between -360 and 360. So let's add those tests to `integer` valType instead.

packages/python/plotly/_plotly_utils/tests/validators/test_angle_validator.py
```diff
@@ -3,6 +3,7 @@
 # from ..basevalidators import AngleValidator
 from _plotly_utils.basevalidators import AngleValidator
 import numpy as np
+from plotly.tests.b64 import b64
```
archmoj, line 6: I don't think you need to load this function.
Can't you simply pass numpy arrays in the tests below?

marthacryan: Hm this fails without adding b64. Why do you think that is?

marthacryan: I get `coerce_val = [1 2 3]`

marthacryan:
Ran into issues with testing string array validators and realized that typed string arrays isn't currently supported. We might want to include a warning / update the behavior to reflect this.

packages/python/plotly/_plotly_utils/tests/validators/test_compoundarray_validator.py
```diff
@@ -1,7 +1,8 @@
 import pytest
 from _plotly_utils.basevalidators import CompoundArrayValidator
 from plotly.graph_objs.layout import Image
-
+import numpy as np
+from plotly.tests.b64 import b64
```
archmoj, line 5: Are these used in this test?

archmoj (approved):
Nicely done.
:dancer:

PR #4727, opened 2024-08-23
Add tests for the keys that should be skipped in base64 conversion

marthacryan:
Adds tests for #4470 to test that certain keys are not being converted to the { base64, dtype } object. These keys were found [here](https://github.com/plotly/plotly.js/blob/056799dfc705a4533935ae3169cc94d35bc44830/test/image/convert_b64.py#L22). Geojson shouldn't be converted because it would then be an invalid geojson object.


packages/python/plotly/plotly/tests/test_optional/test_graph_objs/test_skipped_b64_keys.py
```diff
@@ -0,0 +1,64 @@
...
+            'showscale': False,
+            'type': 'choropleth',
+            'z': {
+                'bdata': 'Cg==',
+                'dtype': 'i1'
+            }
+        }
+        fig.show()
+
+        self.assert_fig_equal(fig.data[0], exp_data)
```
archmoj, line 64: How about rewriting the to test to make it clear which attribute we are testing?
E.g.
```py
class TestSkippedBase64Keys(NumpyTestUtilsMixin, TestCase):
    def test_np_geojson(self):

        normal_coordinates = [[
            [-87, 35],
            [-87, 30],
            [-85, 30],
            [-85, 35],
        ]]

        numpy_coordinates = np.array(normal_coordinates)

        data = [{
            "type": "choropleth",
            "locations": ["AL"],
            "featureidkey": "properties.id",
            "z": np.array([10]),
            "geojson": {
                "type": "Feature",
                "properties": {
                    "id": "AL"
                },
                "geometry": {
                    "type": "Polygon",
                    "coordinates": numpy_coordinates
                }
            }
        }]

        fig = go.Figure(data=data)
        fig.show()

        self.assert_fig_equal(fig["data"][0]["geojson"]["geometry"]["coordinates"], normal_coordinates)
```

packages/python/plotly/plotly/tests/test_optional/test_graph_objs/test_skipped_b64_keys.py
```diff
@@ -0,0 +1,87 @@
...
+                "center": {"lon": 0.5, "lat": 51},
+            },
+        }
+        data = [{"type": "scattermapbox"}]
+
+        fig = go.Figure(data=data, layout=layout)
+
+        # TODO: This is failing because the actual value of the "dash" field
+        # is converted to the { b64, dtype } object.
+        # assert fig.layout['mapbox']['layers'][0]['line']['dash'] == (2.5, 1)
```
archmoj, line 73: Thanks very much for the PR.
As discussed let's enable this test and actually try to fix this bug on the plotly.py side.

packages/python/plotly/plotly/tests/test_optional/test_graph_objs/test_skipped_b64_keys.py
```diff
@@ -0,0 +1,87 @@
+from unittest import TestCase
+import numpy as np
+from plotly.tests.test_optional.optional_utils import NumpyTestUtilsMixin
+import plotly.graph_objs as go
+
+
+class TestSkippedBase64Keys(NumpyTestUtilsMixin, TestCase):
```
archmoj, line 7:
```suggestion
class TestShouldNotUseBase64InUnsupportedKeys(NumpyTestUtilsMixin, TestCase):
```

packages/python/plotly/_plotly_utils/basevalidators.py
```diff
@@ -488,9 +489,10 @@ def description(self):
         )
 
     def validate_coerce(self, v):
-
         if is_none_or_typed_array_spec(v):
             pass
+        elif 'layer' in self.parent_name or 'range' in self.parent_name or 'geojson' in self.parent_name:
```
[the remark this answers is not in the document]
marthacryan: > Any change to refactor it? 

Just pushed a change! Let me know if you have naming or other refactoring suggestions :)

> Are there any other data types that we need to put this condition in specially for ranges?

So, this is the only place that calls `to_typed_array_spec`, so I think that should cover any conversions that are happening. Does that answer your question?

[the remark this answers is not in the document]
archmoj: Exactly.

packages/python/plotly/plotly/tests/test_optional/test_graph_objs/test_skipped_b64_keys.py
```diff
@@ -0,0 +1,87 @@
...
+                    },
+                ],
+                "center": {"lon": 0.5, "lat": 51},
+            },
+        }
+        data = [{"type": "scattermapbox"}]
+
+        fig = go.Figure(data=data, layout=layout)
+
+        # TODO: This is failing because the actual value of the "dash" field
```
archmoj, line 71: Please remove this to do if it is done.

archmoj (approved):
Excellent PR and bug fix.
:dancer:

PR #4728, opened 2024-08-26
Fix CI failure

marthacryan:
Fixes the failing CI based on suggestions in [this comment](https://github.com/conda/conda-libmamba-solver/issues/283#issuecomment-1949560105).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -400,6 +400,8 @@ jobs:
       - run:
           name: Create conda environment
           command: |
+            conda config --remove channels defaults
+            conda config --add channels conda-forge
             conda create -n env --yes python=3.9 conda-build=3.28.4 conda-verify
             conda install -n env -c conda-forge jupyterlab=3 nodejs=16
             conda init bash


BRONSOLO:
I suggest @jonmmease and/or @LiamConnors should review : ) Removing my assignment (though code changes look OK to me in general)

archmoj (approved):
Thanks very much for the fix.
Looking good to me.
@LiamConnors Wondering this might be something we should also add to the release process?

:dancer:

LiamConnors:
> Thanks very much for the fix. Looking good to me. @LiamConnors Wondering this might be something we should also add to the release process?
> 
> 💃

@archmoj which part of the release process might benefit from adding something based on this? I thought this PR fixes it

archmoj:
> > Thanks very much for the fix. Looking good to me. @LiamConnors Wondering this might be something we should also add to the release process?
> > 💃
> 
> @archmoj which part of the release process might benefit from adding something based on this? I thought this PR fixes it

Just in case you noticed a similar bug during the release.

PR #4743, opened 2024-08-30
Update percy tests to use maps

LiamConnors:


archmoj:
Do we set unused plot config on any of these?

LiamConnors:
> Do we set unused plot config on any of these?

I don't understand. Do you have an example of what this looks like?

LiamConnors:
> > > Do we set unused plot config on any of these?
> > 
> > 
> > I don't understand. Do you have an example of what this looks like? And what's the impact of setting/not setting it for these tests?
> 
> I mean if you search inside these tests any of them using set mapbox token?

None of the tests are using that.

archmoj:
Thanks. Wondering why some tests are failing?

gvwilson (approved):

PR #4762, opened 2024-09-11
Remove references to plotly.js CDN and requirejs

### LINKED ISSUE #4336: Latex not working in jupyter notebook v7.
I recently upgraded to jupyter notebook v7 and I noticed that some plotly plots that previously rendered flawlessly now fail to render. I get the following output:

Screenshot from 2023-08-22 12-27-21

After some exploration I found out that it comes from an error using MathJax in `jupyterlab-plotly`. It tries to access `MathJax.config.startup.output`, but `MathJax.config.startup` is `undefined`.

It is very easy to reproduce this error:
- Create a new python environment (I tried with python 3.8).
- `pip install jupyter plotly`
- Open a jupyter notebook with `jupyter notebook`.
- Run a cell with the following code:

```python
from plotly.graph_objects import Figure

Figure().update_layout(title="$Title$")
```

marthacryan:
This removes references in the base renderers to the CDN and requirejs. 
* Latest plotly.js used to be on a CDN, but it isn't anymore (we decided it was too much maintenance years ago)
* Requirejs is essentially unmaintained and newer versions of jupyter lab and jupyter notebook don't support it anymore, so this fixes some of the notebook 7 and lab 4 issues we've been seeing. 
* Removes setting `window._Plotly` because that appears to be unused

This should fix #4336.

### DIFF
--- a/packages/python/plotly/plotly/io/_html.py
+++ b/packages/python/plotly/plotly/io/_html.py
@@ -3,7 +3,7 @@ from pathlib import Path
 import webbrowser
 
 from _plotly_utils.optional_imports import get_module
-from plotly.io._utils import validate_coerce_fig_to_dict, plotly_cdn_url
+from plotly.io._utils import validate_coerce_fig_to_dict
 from plotly.offline.offline import _get_jconfig, get_plotlyjs
 
 _json = get_module("json")
@@ -270,9 +270,8 @@ def to_html(
     elif include_plotlyjs == "cdn":
         load_plotlyjs = """\
         {win_config}
-        <script charset="utf-8" src="{cdn_url}"></script>\
     """.format(
-            win_config=_window_plotly_config, cdn_url=plotly_cdn_url()
+            win_config=_window_plotly_config
         )
 
     elif include_plotlyjs == "directory":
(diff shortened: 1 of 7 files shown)


packages/python/plotly/plotly/io/_html.py
```diff
@@ -57,22 +57,10 @@ def to_html(
         is included in the output.  HTML files generated with this option are
         fully self-contained and can be used offline.
 
-        If 'cdn', a script tag that references the plotly.js CDN is included
-        in the output. The url used is versioned to match the bundled plotly.js.
-        HTML files generated with this option are about 3MB smaller than those
```
archmoj, line 62: I think we should keep `cdn` option as it could help reduce smaller HTML files.
But for loading that we shouldn't use `requirejs`.

marthacryan: Oh ok - but we don't have a cdn currently right?

archmoj: Yes we publish every version to CDN.
It's only the `-latest` version is no longer updated (https://github.com/plotly/plotly.js/pull/5697) since plotly.js v2 and stayed as latest v1 version i.e. `1.58.5`.
So we should be able to continue loading from CDN by specifying the exact version.
To achieve that you could add a separate `script` tag instead of `requirejs` to load e.g.
```html
<script src="https://cdn.plot.ly/plotly-2.35.1.min.js" charset="utf-8"></script>
```
Or alternatively for async loading
```html
<script type="module">
    import "https://cdn.plot.ly/plotly-2.35.2.min.js"
    Plotly.newPlot("gd", [{ y: [1, 2, 3] }])
</script>
```

PR #4763, opened 2024-09-11
Use modern `native ES6 import` to load plotly.js bundle instead of `requirejs` which is no longer under active development

### LINKED ISSUE #4336: Latex not working in jupyter notebook v7.

marthacryan:
**Note: this PR breaks support for notebook < 7.** #4805 is already working on removing that code and adding a warning. 

This removes references in the base renderers to requirejs. Requirejs is essentially unmaintained and newer versions of jupyter lab and jupyter notebook don't support it anymore, so this fixes some of the notebook 7 and lab 4 issues we've been seeing. Replaces requirejs support with native javascript modules: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules

Alternative to #4762 that doesn't remove the CDN. 

Fixes #4336.


archmoj:
Great PR.
Please add a changelog also please mention if it is fixing #4336?

marthacryan:
Note: the orca tests are also failing on master

packages/python/plotly/plotly/io/_base_renderers.py
```diff
@@ -275,27 +273,13 @@ def activate(self):
...
-            paths: {{
-                'plotly': ['{plotly_cdn}']
-            }}
-        }});
-        require(['plotly'], function(Plotly) {{
-            window._Plotly = Plotly;
-        }});
-        }}
         </script>
+        <script src=\"{plotly_cdn}\" charset="utf-8"></script>
```
archmoj, line 282: How about using a `script` tag with *module* type i.e. [native ES6 import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) to have async loading.

Something like: 
```html
<script type="module">import \"{plotly_cdn}\"</script>
```

packages/python/plotly/plotly/io/_base_renderers.py
```diff
@@ -309,15 +293,7 @@ def activate(self):
...
-        if (typeof require !== 'undefined') {{
-        require.undef("plotly");
-        define('plotly', function(require, exports, module) {{
-            {script}
-        }});
-        require(['plotly'], function(Plotly) {{
-            window._Plotly = Plotly;
-        }});
-        }}
+        {script}
```
archmoj, line 296: Also here similar to my comment https://github.com/plotly/plotly.py/pull/4763/files#r1757352938

You may add two scripts: one for configs; and the other to `import` plotly.js.

packages/python/plotly/plotly/io/_base_renderers.py
marthacryan: Sure

packages/python/plotly/plotly/io/_base_renderers.py
marthacryan: I think this is actually embedding the entire plotly.js library so we wouldn't be including an `import` for plotly.js

alexcjohnson (approved):
💃 I'm on board - not aware of any good use cases for requirejs here. Maybe there were in the past, but mostly those use cases would have applied to direct plotly.js use rather than plotly.py anyway.

archmoj:
Please update the description of the PR and include links & info about `requirejs`.

packages/python/plotly/plotly/io/_base_renderers.py
```diff
@@ -275,27 +273,13 @@ def activate(self):
...
-            paths: {{
-                'plotly': ['{plotly_cdn}']
-            }}
-        }});
-        require(['plotly'], function(Plotly) {{
-            window._Plotly = Plotly;
-        }});
-        }}
         </script>
+        <script type="module">import \"{plotly_cdn}\"</script>
```
archmoj, line 282: Wondering if we need to set `charset="utf-8"` here?

marthacryan: What would be the advantage of that?

archmoj: I think we may skip it for now since out minified bundle does not have non ASCII characters.
But the advantage of setting `charset` is that if one wants to switch to the unminified bundle (e.f. for testing purposes) ; then they don't need to figure out these details as out unminified bundle may still have (or would possibly have) such characters. 
Here is an example of PRs that fixed unicode variable names in our for of d3 module: https://github.com/plotly/d3/pull/6/files

CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fixed a bug in integer validation of arrays that threw an error when an array contained a mix of strings and integers.
 
+- Fixed a bug in JupyterLab >= 4 and Jupyter Notebook >= 7 that caused latex to not render in plotly charts. 
+
+- Removed require.js from the generated javascript in plotly charts.
```
archmoj, line 26:
```suggestion
- Use modern [native ES6 import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) to load plotly.js bundle instead of requirejs which is [no longer under active development](https://github.com/requirejs/r.js/compare/2.3.6...2.3.7)
```

packages/python/plotly/plotly/io/_base_renderers.py
archmoj: Yes. But I think if we could use ES6 import here there would be the benefit of having async loading.
Could you please give that a try?
To do that you need to add a separate `script` tag` after closing this script.
Thank you!

marthacryan: If you scroll up in the code, this case is stated as follows:
```
# If not connected then we embed a copy of the plotly.js
# library in the notebook
```
So we should probably keep that in, right?

archmoj (approved):
Great PR :trophy: 
Nicely completed.
:dancer:

archmoj:
@LiamConnors do you have permission to merge this?

LiamConnors:
@marthacryan and @archmoj, when building Plotly.py using this PR and the master branch of Plotly.js, figures don't load in a classic notebook and the following error is in the console. 

image

gvwilson:
@archmoj @marthacryan please sync with @LiamConnors about the issue he flagged (previous comment in this thread)

LiamConnors:
Regarding the error saw in my previous comment. It only is an issue with notebook<7

archmoj:
Is there any blocker for merging this PR?

marthacryan:
> Is there any blocker for merging this PR?

@archmoj @LiamConnors @gvwilson I believe there isn't anything blocking merging since Liam was able to use this with Notebook 7.

archmoj:
:dancer:

gvwilson:
merge it!

PR #4781, opened 2024-10-05
fixing a bug that causes the hovertext to show %{text} rather than th…

rl-utility-man:
There is a bug here that causes the hovertext to show one line as %{text} rather than showing the continent name.  I have fixed it the way the comments imply that the example intended.  either we should add a text parameter as I've done here, or the example could remove the %{text} from the hovertext and from the comments.

This is a one line bug fix bringing the example in congruence with its comments, so I'm going to skip the checklist.  Happy to fill it out later if it would be helpful.


LiamConnors:
Thanks for the PR @rl-utility-man! Looks great

PR #4784, opened 2024-10-07
Convert base64 in BaseFigure.to_dict instead of validate_coerce

marthacryan:
Converting to base64 in BaseFigure.to_dict() offers the same performance improvements without changing the output of the data field of the Figure object. This will only be called through `show()` or in dash

### DIFF
--- a/packages/python/plotly/plotly/io/_utils.py
+++ b/packages/python/plotly/plotly/io/_utils.py
@@ -1,7 +1,112 @@
-from _plotly_utils.basevalidators import is_homogeneous_array, to_typed_array_spec
+import base64
+from _plotly_utils.basevalidators import (
+    copy_to_readonly_numpy_array,
+    is_homogeneous_array,
+    to_typed_array_spec,
+)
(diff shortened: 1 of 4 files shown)


packages/python/plotly/plotly/io/_utils.py
```diff
@@ -24,6 +25,14 @@ def validate_coerce_fig_to_dict(fig, validate):
                 typ=type(fig), v=fig
             )
         )
+
+    # Add base64 conversion before sending to the front-end
+    for trace in fig_dict["data"]:
+        for key, value in trace.items():
+            if is_homogeneous_array(value):
+                print("to typed array: key:", key, "value:", value)
```
gvwilson, line 33: is this a stray debugging statement?

archmoj:
Nice discovery! Thanks to @marthacryan and @emilykl 
PR of the year! :trophy: :medal_sports: :star2: 
This would simplify the changes made to the tests and validation steps on the #4470 PR.
:dancer:

Issue #4820, opened 2024-10-22
`build-doc` failing with AttributeError: module 'plotly.validators.layout.hoverlabel' has no attribute 'GrouptitlefontValidator'


LiamConnors:
`build-doc` is failing with AttributeError: module 'plotly.validators.layout.hoverlabel' has no attribute 'GrouptitlefontValidator'

```
~/Desktop/plotly.py/packages/python/plotly/plotly/io/_kaleido.py:338: UserWarning:

full_figure_for_development is not recommended or necessary for production use in most circumstances. 
To suppress this warning, set warn=False

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[10], line 1
----> 1 full_fig = fig.full_figure_for_development()
      2 print(full_fig)

File ~/Desktop/plotly.py/packages/python/plotly/plotly/basedatatypes.py:3476, in BaseFigure.full_figure_for_development(self, warn, as_dict)
   3450 """
   3451 Compute default values for all attributes not specified in the input figure and
   3452 returns the output as a "full" figure. This function calls Plotly.js via Kaleido
   (...)
   3472     The full figure
   3473 """
   3474 import plotly.io as pio
-> 3476 return pio.full_figure_for_development(self, warn, as_dict)

File ~/Desktop/plotly.py/packages/python/plotly/plotly/io/_kaleido.py:350, in full_figure_for_development(fig, warn, as_dict)
    347 else:
    348     import plotly.graph_objects as go
--> 350     return go.Figure(fig, skip_invalid=True)

File ~/Desktop/plotly.py/packages/python/plotly/plotly/graph_objs/_figure.py:641, in Figure.__init__(self, data, layout, frames, skip_invalid, **kwargs)
      5 def __init__(
      6     self, data=None, layout=None, frames=None, skip_invalid=False, **kwargs
      7 ):
      8     """
      9     Create a new :class:Figure instance
     10 
   (...)
    639         is invalid AND skip_invalid is False
    640     """
--> 641     super(Figure, self).__init__(data, layout, frames, skip_invalid, **kwargs)

File ~/Desktop/plotly.py/packages/python/plotly/plotly/basedatatypes.py:556, in BaseFigure.__init__(self, data, layout_plotly, frames, skip_invalid, **kwargs)
    553 self._layout_validator = LayoutValidator()
    555 # ### Import Layout ###
--> 556 self._layout_obj = self._layout_validator.validate_coerce(
[...]

PR #4821, opened 2024-10-22
Regenerate plotly.py from plotly.js

LiamConnors:
This PR regenerates Plotly.py based on Plotly.js

https://github.com/plotly/plotly.py/issues/4820


archmoj (approved):
Good catch!
Thank you @LiamConnors

PR #4823, opened 2024-10-22
Switch to use anywidget

marthacryan:
Co-authored by @manzt! This updates the logic for the `FigureWidget` to rely on [anywidget](https://anywidget.dev/).
A quote about anywidget from the documentation for context:
> anywidget is both a [specification](https://anywidget.dev/en/afm) and toolset for authoring reusable web-based widgets for interactive computing environments. As a specification, it defines a standard for widget front-end code. As a toolkit, it provides tools for authoring widget front-end code according to the specification, as well as a Python library that simplifies creating and publishing custom [Jupyter Widgets](https://ipywidgets.readthedocs.io/en/latest/).

## How to test
Create a python environment. 
```
cd packages/python/plotly/js
npm i && npm run build
cd ..
pip install -e .
pip install jupyter anywidget
jupyter lab
```
Docs on how to use `FigureWidget`: https://plotly.com/python/figurewidget/

## More context
**This will not affect the `go.Figure` class**. High level explanation of the difference between using `go.Figure` and `go.FigureWidget`:
* `go.Figure`: The chart generated by this object will be interactive on the **client side**. For many purposes, this is all a user needs to do with a chart. The interactions are things like zooming in, or any of the various operations in the toolbar.
* `go.FigureWidget`: The difference here is that the figure is interactive on the **server side** as well.
  * Interactions with the plotly chart can be sent to the python kernel and used in other widgets.
  * Interactions with other widgets can be used to control the chart

### Some info about the code changes:
* This changes our requirement for using jupyter widgets from `ipywidgets` to `anywidget`. The majority of the files changed in this PR are only changing that. 
* This adds a folder into `packages/python/plotly` called `js`, which contains:
  * `package.json`: A very small amount of dependencies and scripts to build `bundle.js`
  * `widget.ts`: The code used by jupyter to render and interact with the widget. This is transpiled into `packages/python/plotly/plotly/bundle.js` and referenced from `basewidget.py`. 
  * `tsconfig.json`: Configurations for transpiling `widget.ts`.


ndrezn:
Shot in the dark, but any chance these AnyWidget changes will help with https://github.com/plotly/plotly.py/issues/4827?

packages/python/chart-studio/test_requirements/requirements_37.txt
```diff
@@ -6,6 +6,6 @@ retrying==1.3.3
 pytest==3.5.1
 pandas==0.23.2
 numpy==1.14.3
-ipywidgets==7.2.0
+anywidget
```
ndrezn, line 9: Should we have a version pin here?

packages/python/plotly/requirements.txt
```diff
@@ -7,3 +7,4 @@
 
 ## retrying requests ##
 tenacity>=6.2.0
+anywidget>=0.9.0
```
ndrezn, line 10: Lots of proposed new requirements coming into Plotly.py... How does requiring AnyWidget affect the bundle size? Does it need to be a mandatory dependency, or is it safe to error out and ask the user to install it manually? 

Should we add e.g. `pip install plotly[notebook]` so we can _recommend_ rather than enforce AnyWidget install?

See: https://github.com/plotly/plotly.py/pull/4644

marthacryan: Oh yeah actually it isn't bundled with `plotly.py` - I added an [error message](https://github.com/plotly/plotly.py/blob/4ab2a0cc6b70a5bd7f6b0348f8a7f5aa3834318a/packages/python/plotly/plotly/missing_anywidget.py) in this PR that tells users to download `anywidget` if they try to create a `FigureWidget` without `anywidget` installed. This is identical to the behavior with `ipywidgets` currently, except replaces `ipywidgets` with `anywidget`. `ipywidgets` and `anywidget` aren't significantly different sizes.

ndrezn: I'll do testing locally to get a feel for the error cases

manzt: > How does requiring AnyWidget affect the bundle size?

anywidget is a pure python and primarily wraps ipywidgets. 

```sh
anywidget v0.9.13
├── ipywidgets v8.1.5
│   ├── comm v0.2.2
│   │   └── traitlets v5.14.3
│   ├── ipython v8.12.3
│   │   ├── appnope v0.1.4
│   │   ├── backcall v0.2.0
│   │   ├── decorator v5.1.1
│   │   ├── jedi v0.19.1
│   │   │   └── parso v0.8.4
│   │   ├── matplotlib-inline v0.1.7
│   │   │   └── traitlets v5.14.3
│   │   ├── pexpect v4.9.0
│   │   │   └── ptyprocess v0.7.0
│   │   ├── pickleshare v0.7.5
│   │   ├── prompt-toolkit v3.0.47
│   │   │   └── wcwidth v0.2.13
│   │   ├── pygments v2.18.0
│   │   ├── stack-data v0.6.3
│   │   │   ├── asttokens v2.4.1
│   │   │   │   └── six v1.16.0
│   │   │   ├── executing v2.0.1
│   │   │   └── pure-eval v0.2.3
│   │   └── traitlets v5.14.3
│   ├── jupyterlab-widgets v3.0.13
│   ├── traitlets v5.14.3
│   └── widgetsnbextension v4.0.13
└── psygnal v0.11.1
```

NB: we have [a PR](https://github.com/manzt/anywidget/pull/579) experimenting with dropping ipywidgets all together for a smaller module graph.

ndrezn: @manzt very cool!!

packages/python/chart-studio/test_requirements/requirements_37.txt
manzt: Min version I think should be 0.9.0. If you want to pin an exact version, I'd use the latest: v0.9.13.

packages/python/plotly/requirements.txt
ndrezn: @marthacryan coming back to my original point, we don't specify `ipywidgets` here, so I don't think we need to add `anywidget` (I _do_ think that our system for managing dev/required dependencies is quite odd, but that's out of scope of this PR and for now I'd say just leave this `reqs` unchanged).

packages/python/plotly/js/widget.ts
```diff
@@ -1,7 +1,7 @@
 // Adapted from ../../../javascript/jupyterlab-plotly/src/Figure.ts for anywidget.
-import type { DOMWidgetModel } from "@jupyter-widgets/base";
-import _ from "lodash";
+import _ from "https://esm.sh/lodash-es@4.17.21";
```
manzt, line 2: Adding an esm URL like this will break the build in enviroments that don't have an internet connection (which you probably don't want). I'd maybe first try `npm install lodash-es` and importing from `lodash-es` here and see if it works.

packages/python/plotly/js/widget.ts
```diff
@@ -1,7 +1,7 @@
...
 import Plotly from "plotly.js";
+/*global define */
```
manzt, line 4: Shouldn't be a problem, but what is the purpose of define here?

packages/python/plotly/js/widget.ts
marthacryan: It does work! Thank you, I'll push that

packages/python/plotly/js/widget.ts
```diff
@@ -1679,7 +1678,7 @@ function py2js_deserializer(v: any, widgetManager?: any) {
     for (var i = 0; i < v.length; i++) {
       res[i] = py2js_deserializer(v[i]);
     }
-  } else if (_.isPlainObject(v)) {
+  } else if (_.isObject(v)) {
```
manzt, line 1681: Personally, I'd just make my own helper for this. Not sure how much lodash is used in the rest of the file, but a nice typescript helper i often copy into projects:

```ts
function isObject(input: unknown): input is Record<string, unknown> {
  return typeof input === 'object' && input !== null && !Array.isArray(input);
};
```

packages/python/plotly/js/widget.ts
marthacryan: Just removed this - it was added to address an error I was seeing related to requirejs. I think this wasn't actually the required fix but I kept it in because I hadn't tested removing it yet. Just tested it and it looks like we're good!

packages/python/chart-studio/test_requirements/requirements_37.txt
marthacryan: Pinned!

marthacryan:
> Shot in the dark, but any chance these AnyWidget changes will help with #4827?

Just tested and it does!

packages/python/plotly/js/widget.ts
archmoj: @marthacryan typedArrays should also be excluded.
```js
var ab = ArrayBuffer;
var dv = DataView;

function isTypedArray(a) {
    return ab.isView(a) && !(a instanceof dv);
}
```

ndrezn:
Hey @mscolnick -- looks like this AnyWidget support gets Plotly charts playing nicely in Marimo:
Screenshot 2024-11-01 at 11 41 21 AM

We're seeing some issues making it reactive but we'll be taking a closer look!

ndrezn:
@marthacryan could you validate the behaviour of various events accessible in `FigureWidget`? By my understanding only `on_click` is documented: https://plotly.com/python/click-events/.

Here's a snippet showing these events:
```python
from random import randint

fig = px.scatter(x=range(0,10), y=[randint(0,10) for i in range(0,10)])
fig.update_layout(dragmode='select')

figurewidget = go.FigureWidget(fig)

trace = figurewidget.data[0]

trace.on_change
trace.on_click
trace.on_deselect
trace.on_hover
trace.on_selection
trace.on_unhover

figurewidget
```

marthacryan:
@ndrezn Which CI artifact are you using? And could you give me more info on the steps you took? I'm not able to reproduce

packages/python/plotly/js/widget.ts
```diff
@@ -945,13 +962,13 @@ export class FigureView extends DOMWidgetView {
     }
   }
 
-  processPhosphorMessage(msg: any) {
-    this._processLuminoMessage(msg, super["processPhosphorMessage"]);
-  }
+  // processPhosphorMessage(msg: any) {
+  //   this._processLuminoMessage(msg, super["processPhosphorMessage"]);
+  // }
```
marthacryan, line 967: delete

packages/python/plotly/plotly/basewidget.py
```diff
@@ -1,30 +1,21 @@
+import pathlib
 import ipywidgets as widgets
```
marthacryan, line 2: knife emoji

packages/python/plotly/plotly/io/_renderers.py
```diff
@@ -395,6 +398,19 @@ def show(fig, renderer=None, validate=True, **kwargs):
                 "Mime type rendering requires nbformat>=4.2.0 but it is not installed"
             )
 
+        parent_process = psutil.Process().parent().cmdline()[-1]
+
+        if "jupyter-notebook" in parent_process and jupyter_notebook.__version__ < "7":
```
marthacryan, line 403: Maybe put this in `FigureWidget` in addition

emilykl:
@marthacryan I followed your steps for testing locally and the `FigureWidget` seems to be working properly!
Screen Shot 2024-11-07 at 4 35 27 PM

I also tested that the warnings for `jupyterlab<3` and `notebook<7` are showing up:
Screen Shot 2024-11-07 at 4 37 17 PM
Screen Shot 2024-11-07 at 4 38 53 PM

emilykl (approved):
🚀 

At least one more reviewer would be ideal since I'm not a frequent Jupyter Notebook user, but looks great to me.

packages/python/plotly/plotly/io/_utils.py
```diff
@@ -1,6 +1,29 @@
...
+import warnings
+import psutil
+
+jupyter_notebook = optional_imports.get_module("notebook")
+jupyter_lab = optional_imports.get_module("jupyterlab")
+
+
+def display_jupyter_version_warnings():
+
+    parent_process = psutil.Process().parent().cmdline()[-1]
```
alexcjohnson, line 15: IDK how this would happen in practice, but in principle `.parent()` can [return None](https://psutil.readthedocs.io/en/latest/#psutil.Process.parent) which would make this fail. In that case I guess we aren't in a jupyter context so we can short-circuit the rest of this.

packages/python/plotly/plotly/io/_utils.py
```diff
@@ -1,6 +1,29 @@
...
+
+    if "jupyter-notebook" in parent_process and jupyter_notebook.__version__ < "7":
```
alexcjohnson, line 17: Can we move the `optional_imports.get_module` calls inside these conditionals? Something like:
```py
if "jupyter-notebook" in parent_process:
    jupyter_notebook = optional_imports.get_module("notebook")
    if jupyter_notebook.__version__ < "7":
```
and similar for `jupyterlab`. Those imports can be slow (particularly `jupyterlab`, it's nearly a second on my computer), no need to increase everyone's startup time (for example Dash apps). And calling `optional_imports.get_module` on an already-loaded module or a nonexistent module you've previously asked for is very fast.

I wonder about the existing optional import calls in `_renderers.py` as well - looks like `IPython` and `IPython.display` are automatically imported when you're in an IPython or jupyter environment, so I bet we could just add `should_load=False` to those calls. `nbformat` is not automatically loaded, so it would need to move into the `show` method where it's called, like I'm suggesting here.

marthacryan: Good point! I moved them into the if statements. I'm not sure about the other optional calls in that file but I can make an issue to investigate that more

marthacryan: Just made this issue: https://github.com/plotly/plotly.py/issues/4856

packages/python/plotly/plotly/io/_utils.py
marthacryan: I added some code to address this! Let me know if you think there's a better approach than a try/except

.circleci/config.yml
```diff
@@ -46,6 +46,9 @@ commands:
             . venv/bin/activate
             pip install --upgrade pip wheel
             pip install -r ./test_requirements/requirements_<<parameters.py>>_optional.txt
+            cd js
+            npm install
```
archmoj, line 50: Please consider using `npm ci` instead of `npm i` everywhere to have more stable packages as well as being faster on CircleCI.

manzt:
Just had a chance to read the [blog post](https://plotly.com/blog/plotly-dash-major-release/), congrats on the pre-release! I have a quick question about this part:

> "The other big change in Plotly.py 6.0 is that we are dropping support for Jupyter Notebook versions before 7.0, which allows us to support the AnyWidget standard for notebook extensions."

This language suggests this change is related to anywidget. anywidget should be compatible with at least Jupyter Notebook v6, so just wanted to clarify why error/warning. Is there some API that is missing that the `FigureWidget` required?

```sh
uvx juv init example.ipynb
uvx juv add example.ipynb anywidget==0.9.13 plotly==6.0.0rc0
uvx juv run --jupyter=notebook@6 example.ipynb
```

```py

f = go.FigureWidget()
f
# Warning: Plotly version >= 6 requires Jupyter Notebook >= 7 but you have 6.0.0 installed.
``` 

```py
import anywidget
import traitlets

class CounterWidget(anywidget.AnyWidget):
    _esm = """
    function render({ model, el }) {
      let count = () => model.get("value");
      let btn = document.createElement("button");
      btn.innerHTML = `count is ${count()}`;
      btn.addEventListener("click", () => {
        model.set("value", count() + 1);
        model.save_changes();
      });
      model.on("change:value", () => {
        btn.innerHTML = `count is ${count()}`;
      });
      el.appendChild(btn);
    }
    export default { render };
    """
    value = traitlets.Int(0).tag(sync=True)
    
CounterWidget() # shows button
```

image

ndrezn:
Hey @manzt, that's a mistake in the blog post. We're dropping support for <7.0 primarily because it simplifies our build and release process. I'll make sure that gets corrected -- thanks for taking the time to deep dive on this!

Emil4221:
plotly_620_error
I got plotly in my requirements.txt file, but it does not seem to install anywidget automatically.

manzt:
I believe that anywidget is an optional dependency in plotly, so you'll also need to add "anywidget" to your requirements.txt to use the FigureWidget.

PR #4831, opened 2024-10-24
Remove tenacity dependency

ndrezn:
The `tenacity` dependency in Plotly.py (added in https://github.com/plotly/plotly.py/pull/2911) is used today in one place for Orca to provide a pretty naive retry decorator. It's easy enough to write our own with identical functionality.

This means it's possible to remove the sole dependency of Plotly.py. Yay!

The meaningful change is in cf4bfd1


alexcjohnson (approved):
💃 Love it!

archmoj:
Please mention this PR in the change log. 
Thank you!

PR #4840, opened 2024-10-28
Docs updates for Plotly.py version 6

LiamConnors:
- Migration guide from v5 to v6
- Add new performance page with info on 
- Add new dataframe support information
- Remove pages for removed features


doc/python/v6-changes.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Removed Attributes
+
+The following attributes have been removed in Plotly.py 6.
+
+### `titlefont`,`titleposition`, `titleside`, and `titleoffset`
+
+The `titlefont`,`titleposition`, `titleside`, and `titleoffset` attributes have been removed. Replace them with `title.font`, `title.position`, `title.side`, and `title.offset`.
+
+Here's an example using `layout.tilefont`:
```
LiamConnors, line 70:
```suggestion
Here's an example using `layout.titlefont`:
```

doc/python/performance.md
```diff
@@ -0,0 +1,298 @@
...
+* High-performance multidimensional trace types: `splom`, or `parcoords`
+* 3-d trace types `scatter3d`, `surface`, `mesh3d`, `cone`, `streamtube`
+* Mapbox Gl JS-powered trace types: `scattermapbox`, `choroplethmapbox`, `densitymapbox`
+
+### WebGL Limitations and Tradeoffs
+
+WebGL is a powerful technology for accelerating computation but comes with some strict limitations:
+
+1. GPU requirement: WebGL is a GPU (graphics card) technology and therefore requires specific hardware which is available in most but not all cases and is supported by most but not all browsers.
+2. Rasterization: WebGL-rendered data is drawn as a grid of pixels rather than as individual shapes, so can appear pixelated or fuzz in certain cases, and when exported to static file formats will appear pixelated on zoom. In addition: text rendering will differ between SVG and WebGL-powered traces.
```
gvwilson, line 58: comma after "In addition"

doc/python/performance.md
```diff
@@ -0,0 +1,298 @@
...
+3. Context limits: browsers impose a strict limit on the number of WebGL "contexts" that any given web document can access. WebGL-powered traces in `plotly` can use multiple contexts in some cases but as a general rule, **it may not be possible to render more than 8 WebGL-involving figures on the same page at the same time.** See the following section, Multiple WebGL Contexts, for more details.
+4. Size limits: browsers impose hardware-dependent limits on the height and width of figures using WebGL which users may encounter with extremely large plots (e.g. tens of thousands of pixels of height).
```
gvwilson, line 60: why "of height" rather than "in height or width"? is it really just height that matters?

doc/python/performance.md
```diff
@@ -0,0 +1,298 @@
...
+
+In addition to the above limitations, the WebGL-powered version of certain SVG-powered trace types (`scattergl`, `scatterpolargl`) are not complete drop-in replacements for their SVG counterparts yet
+* Available symbols will differ.
+* Area fills are not yet supported in WebGL.
+* Range breaks on time-series axes are not yet supported.
+* Axis range heuristics may differ.
+
+#### Multiple WebGL Contexts
+
+*New in 5.19*
```
gvwilson, line 70: is this relevant for 6.0 ?

doc/python/performance.md
```diff
@@ -0,0 +1,298 @@
...
+
+Most browsers have a limit of between 8 and 16 WebGL contexts per page. A Plotly WebGL-based figure may use multiple WebGL contexts, but generally you'll be able to render between 4 and 8 figures on one page.
```
gvwilson, line 72: ?? if the browser limit is 8-16, why is our limit 4-8?

doc/python/performance.md
```diff
@@ -0,0 +1,298 @@
...
+
+If you exceed the browser limit on WebGL contexts, some figures won't render and you'll see an error. In the console in Chrome, for example, you'll see the error: "Too many active WebGL contexts. Oldest context will be lost".
+
+If you encounter WebGL context limits when using WebGL-based figures, you can use [Virtual WebGL](https://github.com/greggman/virtual-webgl), which virtualizes a single WebGL context into multiple contexts.
+
+To use it, in the environment where your Plotly figures are being rendered, load the Virtual WebGL script, "https://unpkg.com/virtual-webgl@1.0.6/src/virtual-webgl.js", for example, using a `<script>` tag.
```
gvwilson, line 78: add a note to the effect that this will be slower than "pure" GL (otherwise, people will ask why we don't do it all the time - at least, I would)

doc/python/performance.md
```diff
@@ -0,0 +1,298 @@
...
+
+fig.update_traces(marker_line=dict(width=1, color='DarkSlateGray'))
+
+fig.show()
+```
+
+
+#### WebGL with 1,000,000 points with Graph Objects
+
+If Plotly Express does not provide a good starting point, it is also possible to use [the more generic `go.Scattergl` class from `plotly.graph_objects`](/python/graph-objects/).
```
gvwilson, line 126: good starting point for what? (I know from context, but maybe worth repeating here?)

doc/python/performance.md
```diff
@@ -0,0 +1,298 @@
...
+fig.show()
+```
+
+See https://plotly.com/python/reference/scattergl/ for more information and chart attribute options!
+
+## NumPy and NumPy Convertible Arrays for Improved Performance
+
+*New in Plotly.py version 6*
+
+Improve the performance of generating Plotly figures that use a large number of data points by using NumPy arrays and other objects that Plotly can convert to NumPy arrays, such as Pandas and Polars Series.
```
gvwilson, line 159: Maybe "You can improve..." ?

doc/python/performance.md
```diff
@@ -0,0 +1,298 @@
...
+- float32
+- float64
+- int8
+- uint8
+- int16
+- uint16
+- int32
+- uint32
+
+*If the array dtype is **int64** and **uint64**, often the default dtype for arrays in NumPy when no dtype is specified, those dtypes will be changed to other types internally by Plotly.py where possible. When working with NumPY directly, you can [specify the `dtype`](https://numpy.org/doc/stable/user/basics.types.html#array-types-and-conversions-between-types) when creating `ndarray` objects.
```
gvwilson, line 182: "int16 or uint64" rather than "and"

doc/python/performance.md
```diff
@@ -0,0 +1,298 @@
...
+
+### Unsupported Attributes
+
+Arrays passed to attributes with the following names are not supported:
```
gvwilson, line 186: does "not supported" mean "they won't work"? or just "they won't get performance benefits"?

doc/python/performance.md
```diff
@@ -0,0 +1,298 @@
...
+fig.show()
+```
+
+## Datashader
+
+Use [Datashader](https://datashader.org/) to reduce the size of a dataset passed to the browser for rendering by creating a rasterized representation of the dataset. This makes it ideal for working with datasets of tens to hundreds of millions of points. 
+
+### Passing Datashader Rasters as a Tile Map Image Layer
+
+We visualize here the spatial distribution of taxi rides in New York City. A higher density
```
gvwilson, line 228: "The example below shows the spatial distribution of taxi rides in New York City, which are concentrated on major avenues." instead of first two sentences

doc/python/px-arguments.md
```diff
@@ -33,19 +33,33 @@ jupyter:
...
+The examples on this page use datasets available in the `data` package in `px`. `px.data` contains functions that when called return a dataset as a DataFrame. Some of the datasets included in `px.data` are:
 
-[`px.imshow` operates on matrix-like data](/python/imshow/) you might store in a `numpy` or `xarray` array and functions like [`px.choropleth` and `px.choropleth_mapbox` can operate on geographic data](/python/maps/) of the kind you might store in a GeoPandas `GeoDataFrame`. This page details how to provide column-oriented data to most Plotly Express functions.
+- `carshare` - Each row represents the availability of car-sharing services near the centroid of a zone in Montreal over a month-long period.
+- `election` - Each row represents voting results for an electoral district in the 2013 Montreal mayoral election.
+- `iris` - Each row represents a flower.
 
+To access the `iris` dataset, we call it's function and assign it to a variable:
```
gvwilson, line 46: "its" not "it's"

doc/python/px-arguments.md
```diff
@@ -33,19 +33,33 @@ jupyter:
...
 
+```python
+import plotly.express as px
+
+df = px.data.iris()
+df.head()
+```
+
+By default `px.data` functions return Pandas DataFrames, but you can specify an alternative DataFrame type using `return_type`. `pandas`, `polars`, and `pyarrow` are supported return types.
```
gvwilson, line 55: "dataframe type" not "DataFrame type"

doc/python/px-arguments.md
```diff
@@ -33,19 +33,33 @@ jupyter:
...
 
-### Plotly Express works with  Long-, Wide-, and Mixed-Form Data
+```python
+df = px.data.iris(return_type='polars')
+df.head()
+```
 
-*Until version 4.8, Plotly Express only operated on long-form (previously called "tidy") data, but [now accepts wide-form and mixed-form data](/python/wide-form/) as well.*
+## Long, Wide-, and Mixed-Form Data
```
gvwilson, line 62: puzzled: why no dash after Long when there is one after Wide? (I wouldn't have a dash after either.)

doc/python/px-arguments.md
```diff
@@ -101,33 +115,55 @@ fig.show()
...
-### Input Data as Pandas `DataFrame`s
+## DataFrame Input
+
+The first argument of every `px` function is `data_frame`. If you provide a DataFrame as a `px` function's first argument, you can then specify column names as strings from the DataFrame as other arguments.
+
+### Supported DataFrames
+
+`px` functions natively support the following types of DataFrames:
+
+- pandas
```
gvwilson, line 126: "Pandas" rather than "pandas" (?) not sure what the capitalization should be

doc/python/px-arguments.md
```diff
@@ -101,33 +115,55 @@ fig.show()
...
+- Modin
 
-As shown above, `px` functions supports natively pandas DataFrame. Arguments can either be passed as dataframe columns, or as column names if the `data_frame` argument is provided.
+> To use Polars, PyArrow, or Modin with Plotly Express, you'll need to have NumPy installed. You can install it with: `pip install numpy`.
+To use [trendlines](/python/linear-fits/), you'll also need to have pandas installed. 
 
-#### Passing columns as arguments
+`px` uses [Narwhals](https://narwhals-dev.github.io/narwhals/) to natively support these DataFrames. Other types of DataFrames that are currently supported by Narwhals may also work with `px`, for example cuDF.
+
+You can also pass DataFrames that are not natively supported, but support the dataframe interchange protocol, or have a `to_pandas()` method.
```
gvwilson, line 136: again, do we use "DataFrame" for the generic object or "dataframe"?

doc/python/px-arguments.md
```diff
@@ -137,18 +173,17 @@ fig = px.scatter(df, x=df.sepal_length, y=df.sepal_width, size=df.petal_length,
 fig.show()
 ```
 
-### Columns not in the `data_frame` argument
-
-In the addition to columns from the `data_frame` argument, one may also pass columns from a different DataFrame, _as long as all columns have the same length_. It is also possible to pass columns without passing the `data_frame` argument.
+### Using Columns from Multiple DataFrames
 
-However, column names are used only if they correspond to columns in the `data_frame` argument, in other cases, the name of the keyword argument is used. As explained below, the `labels` argument can be used to set names.
+You can also use columns from multiple dataframes in the one `px` function, as long as all the DataFrame columns you use have the same length. In this example, we pass `df1` as the `data_frame` argument to `px.bar` and then us a column from `df2` for the `y` argument.
```
gvwilson, line 178: "in one" as opposed to "in the one"

doc/python/px-arguments.md
```diff
@@ -239,7 +233,7 @@ fig.show()
 
 ### Passing dictionaries or array-likes as the data_frame argument
 
-The column-based argument `data_frame` can also be passed with a `dict` or `array`. Using a dictionary can be a convenient way to pass column names used in axis titles, legend entries and hovers without creating a pandas DataFrame.
+The column-based argument `data_frame` can also be passed with a `dict` or `array`. Using a dictionary can be a convenient way to pass column names used in axis titles, legend entries and hovers without creating a DataFrame.
```
gvwilson, line 236: why "data_frame" in this case?

gvwilson (approved):

doc/python/px-arguments.md
LiamConnors: In this case it's name of the px parameter.

doc/python/performance.md
LiamConnors: Yeah, just keeping the 5.19 in so users that are using an older version will know if this is available to them

doc/python/performance.md
LiamConnors: The amount of contexts used can vary by figure and I believe this was the range we saw in testing

doc/python/px-arguments.md
LiamConnors: I say "pandas" as that's how it is in their docs: https://pandas.pydata.org/docs/user_guide/index.html

doc/python/performance.md
```diff
@@ -0,0 +1,298 @@
...
+    name: High Performance Visualization
+    order: 14
+    permalink: python/performance/
+    redirect_from:
+    - python/webgl-vs-svg/
+    - python/datashader/
+    thumbnail: thumbnail/webgl.jpg
+---
+
+## WebGL
```
ndrezn, line 39: I'd recommend:
1. Dataframe types
2. Numpy Arrays
3. WebGL
4. Datashader

for section order

doc/python/performance.md
LiamConnors: just not supported for the performance benefits. updating

doc/python/performance.md
LiamConnors: @archmoj do you know if the virtual webgl would make things slower?  or what's the reason not to use it all the time?

doc/python/performance.md
LiamConnors: @archmoj do you happen to know?

LiamConnors:
Added an example for grouped stacked bar charts @archmoj  for https://github.com/plotly/plotly.js/pull/7009

doc/python/performance.md
archmoj: Yes it is slower.

doc/python/px-arguments.md
```diff
@@ -101,33 +115,60 @@ fig.show()
...
+
+`px` functions natively support the following types of dataframes:
+
+- pandas
+- Polars
+- PyArrow
+- Modin
 
-As shown above, `px` functions supports natively pandas DataFrame. Arguments can either be passed as dataframe columns, or as column names if the `data_frame` argument is provided.
+`px` uses [Narwhals](https://narwhals-dev.github.io/narwhals/) to natively support these dataframes. Other types of dataframes that are currently supported by Narwhals may also work with `px`.
```
gvwilson, line 131: Move this up above the list to replace the "px functions natively support..." sentence? That feels redundant given this more detailed explanation.

doc/python/v6-changes.md
```diff
@@ -0,0 +1,152 @@
...
+This page outlines the changes in Plotly.py version 6 and cases where you may need to update your charts or tools that you use for working with Plotly.py.
+
+<!-- #region -->
+## Jupyter Notebook Support
+
+Versions of Jupyter Notebook earlier than version 7 are no longer supported. To upgrade to the latest Jupyter Notebook:
+
+```
+pip install notebook --upgrade
+```
```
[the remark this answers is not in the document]
LiamConnors: I think `plotly-dev` should be working now with this merged https://github.com/plotly/plotly.py/pull/4866

doc/python/px-arguments.md
LiamConnors: Reordered the content

doc/python/performance.md
```diff
@@ -0,0 +1,303 @@
...
+    name: High Performance Visualization
+    order: 14
+    permalink: python/performance/
+    redirect_from:
+    - python/webgl-vs-svg/
+    - python/datashader/
+    thumbnail: thumbnail/webgl.jpg
+---
+
+## DataFrames Types
```
emilykl, line 39:
```suggestion
## DataFrame Types
```

doc/python/performance.md
```diff
@@ -0,0 +1,303 @@
...
+
+*New in Plotly.py version 6*
+
+Plotly Express natively supports [pandas, Polars, PyArrow, and Modin dataframes](/python/px-arguments). When building figures with Plotly Express, changing your dataframe library may help improve performance.
```
emilykl, line 43:
```suggestion
Plotly Express natively supports [Pandas, Polars, PyArrow, and Modin dataframes](/python/px-arguments). When building figures with Plotly Express, changing your dataframe library may help improve performance.
```

emilykl, line 43: Maybe worth clarifying here specifically what's changed in v6.

Polars and PyArrow were previously supported via the dataframe interchange protocol, but performance is _significantly_ faster in v6. Modin I believe was not previously supported at all (I might be misremembering though).

doc/python/performance.md
```diff
@@ -0,0 +1,303 @@
...
+
+## NumPy and NumPy Convertible Arrays for Improved Performance
+
+*New in Plotly.py version 6*
+
+You can improve the performance of generating Plotly figures that use a large number of data points by using NumPy arrays and other objects that Plotly can convert to NumPy arrays, such as Pandas and Polars Series.
```
emilykl, line 49:
```suggestion
You can improve the performance of generating Plotly figures that use a large number of data points by passing data as NumPy arrays, or in a format that Plotly can convert easily to NumPy arrays, such as Pandas and Polars Series or DataFrames. These formats will usually show better performance than passing data as a Python list.
```

doc/python/performance.md
```diff
@@ -0,0 +1,303 @@
...
+
+Plotly.py uses Plotly.js for rendering, which supports typed arrays. In Plotly.py, NumPy array and NumPy-convertible arrays are base64 encoded before being passed to Plotly.js for rendering.
```
emilykl, line 51:
```suggestion
Plotly.py uses Plotly.js for rendering, which supports typed arrays. In Plotly.py, NumPy arrays and NumPy-convertible arrays are base64 encoded before being passed to Plotly.js for rendering.
```

doc/python/performance.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Arrays and Data Types Supported
+
+The following types of objects in Python are supported for improved performance:
+
+- Numpy `numpy.ndarray` objects.
+- Pandas Index, Pandas Series, Polars Series, and PyArrow Chunked Array objects.
+- Array objects that can be converted to `numpy.ndarray` objects. i.e., they implement `"__array__"` or `"__array_interface__"` and return a `numpy.ndarray`.
```
emilykl, line 59: I think it would be clearer here to refer to DataFrame objects rather than Series objects -- in the vast majority of cases, at least in `px`, I think users are passing a DataFrame object as the first argument and then column names for the `x`, `y` etc. arguments (rather than passing in individual series to `x` and `y`).

emilykl: Also "supported for improved performance" is vague -- should be more specific ("the following object types benefit from base64 encoding").

doc/python/performance.md
```diff
@@ -0,0 +1,303 @@
...
+- float32
+- float64
+- int8
+- uint8
+- int16
+- uint16
+- int32
+- uint32
+
+*If the array dtype is **int64** or **uint64**, often the default dtype for arrays in NumPy when no dtype is specified, those dtypes will be changed to other types internally by Plotly.py where possible. When working with NumPY directly, you can [specify the `dtype`](https://numpy.org/doc/stable/user/basics.types.html#array-types-and-conversions-between-types) when creating `ndarray` objects.
```
emilykl, line 72: What happens if you pass an array with a dtype which is not on this list?

doc/python/performance.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Unsupported Attributes
+
+Arrays passed to attributes with the following names are not supported for the performance benefits:
```
emilykl, line 76: Again I think we should specify which performance features we are talking about rather than saying "the performance benefits".

doc/python/performance.md
```diff
@@ -0,0 +1,303 @@
...
+`plotly` figures are rendered by web browsers, which broadly speaking have two families of capabilities for rendering graphics:
+
+- The SVG API, which supports vector rendering.
+- The Canvas API, which supports raster rendering, and can exploit GPU hardware acceleration via a browser technology known as WebGL.
+
+Each `plotly` trace type is rendered with either SVG or WebGL. The following trace types use WebGL for rendering:
+
+* Accelerated versions of SVG trace types: `scattergl`, `scatterpolargl`,
+* High-performance multidimensional trace types: `splom`, or `parcoords`
+* 3-d trace types `scatter3d`, `surface`, `mesh3d`, `cone`, `streamtube`
```
emilykl, line 123: I'm more used to seeing `3D` rather than `3-d` (but if we use `3-d` elsewhere in the docs we should keep it for consistency)

doc/python/performance.md
```diff
@@ -0,0 +1,303 @@
...
+* Mapbox Gl JS-powered trace types: `scattermapbox`, `choroplethmapbox`, `densitymapbox`
+
+### WebGL Limitations and Tradeoffs
+
+WebGL is a powerful technology for accelerating computation but comes with some strict limitations:
```
emilykl, line 128:
```suggestion
WebGL is a powerful technology for accelerating rendering but comes with some strict limitations:
```

doc/python/px-arguments.md
```diff
@@ -101,33 +115,55 @@ fig.show()
 Many more examples of wide-form and messy data input can be found in our [detailed wide-form support documentation](/python/wide-form/).
 
 
-### Input Data as Pandas `DataFrame`s
+## Dataframe Input
+
+The first argument of every `px` function is `data_frame`. If you provide a dataframe as a `px` function's first argument, you can then specify column names as strings from the dataframe as other arguments.
+
+### Supported Dataframes
```
emilykl, line 122:
```suggestion
### Supported DataFrame Types
```

doc/python/px-arguments.md
```diff
@@ -101,33 +115,55 @@ fig.show()
...
+
+`px` functions natively support pandas, Polars, and PyArrow dataframes. `px` uses [Narwhals](https://narwhals-dev.github.io/narwhals/) to provide this native dataframe support. Other types of dataframes that are currently supported by Narwhals, for example cuDF and Modin, may also work with `px`.
```
emilykl, line 124: Note: Need to make sure we are consistent with messaging about whether Modin is supported (it's included in the list of supported DataFrame types in `doc/python/performance.md`)

doc/python/px-arguments.md
```diff
@@ -101,33 +115,55 @@ fig.show()
...
 
-As shown above, `px` functions supports natively pandas DataFrame. Arguments can either be passed as dataframe columns, or as column names if the `data_frame` argument is provided.
+PySpark dataframes are also supported and are converted to pandas dataframes internally by Plotly Express.
 
-#### Passing columns as arguments
+> There is a known issue where `px.icicle`, `px.treemap`, and `px.sunburst` used with a `path` parameter won't work with cuDF dataframes.
+
+#### Additional Dependencies Required
+
+- To use Polars, PyArrow, or other dataframes supported natively with Plotly Express, you'll need to have NumPy installed. You can install it with: `pip install numpy`.
```
emilykl, line 134: A little misleading, since NumPy is required for Pandas dataframes as well, it's just that NumPy is a dependency of Pandas so it's likely to be already installed if you're instantiating a Pandas dataframe.

doc/python/px-arguments.md
```diff
@@ -239,7 +233,7 @@ fig.show()
 
 ### Passing dictionaries or array-likes as the data_frame argument
 
-The column-based argument `data_frame` can also be passed with a `dict` or `array`. Using a dictionary can be a convenient way to pass column names used in axis titles, legend entries and hovers without creating a pandas DataFrame.
+The column-based argument `data_frame` can also be passed with a `dict` or `array`. Using a dictionary can be a convenient way to pass column names used in axis titles, legend entries and hovers without creating a dataframe.
```
emilykl, line 236:
```suggestion
The `data_frame` argument can also accept a `dict` or `array` in addition to DataFrame objects. Using a dictionary can be a convenient way to pass column names used in axis titles, legend entries and hovers without creating a dataframe.
```

doc/python/px-arguments.md
LiamConnors: Good catch. I'd changed it in one but not the other. Thanks!

doc/python/performance.md
LiamConnors: Yeah, I'd 3D is better. This is mostly existing content that has been moved to this page. We seem to have a mix, but 3D is used more, so I'll update it.

doc/python/performance.md
LiamConnors: Good point. Added. Thanks

doc/python/px-arguments.md
LiamConnors: That's a fair point. Updated

doc/python/performance.md
LiamConnors: It won't be converted to the typed array spec.

doc/python/performance.md
LiamConnors: Great points. Updated text for both. Let me know your thoughts. Thanks!

LiamConnors:
@marthacryan could you take a look at the typed arrays info and let me know if it looks good to you

doc/python/v6-changes.md
```diff
@@ -0,0 +1,157 @@
...
+
+The following attributes have been removed in Plotly.py 6.
+
+### `titlefont`,`titleposition`, `titleside`, and `titleoffset`
+
+The `titlefont`,`titleposition`, `titleside`, and `titleoffset` attributes have been removed. Replace them with `title.font`, `title.position`, `title.side`, and `title.offset`.
+
+Here's an example using `layout.titlefont`:
+
+```python
```
emilykl, line 74: @LiamConnors I don't love the idea of including code in the docs which specifically _won't_ work.

What if you combine the two code snippets into one, and include the old layout definition as commented-out code, followed by the new layout definition?

Also, maybe it would be clearer to define the layout using a `dict(title=dict(...))` rather than `go.Layout()` since the former is much more common in my experience — I'm not sure I've ever seen anyone use `go.Layout()`.

emilykl:
@LiamConnors I've been through the whole set of changes -- looks good!

Some high-level comments:
- Consistency about what dataframes are supported (already discussed, but I wonder if there's a way to avoid having to update the docs in multiple places if/when we add additional supported dataframes -- is it possible to link to a single source of truth for every one of those mentions?
- Usage of "DataFrame" vs. "dataframe" — It seems to me like the docs are _mostly_ consistent about using "dataframe" when referring to the general concept of a dataframe, and "DataFrame" specifically when referring to the specific Python DataFrame objects, _but_ there are a few cases that feel borderline to me so maybe worth a look over. (Can point out specific cases if helpful.)

doc/python/v6-changes.md
LiamConnors: Yeah, that's a fair point. The complete example probably isn't needed anyway. Reworked to add just a comment with the old method and to use dict rather than `go.Layout`

LiamConnors:
> Consistency about what dataframes are supported (already discussed, but I wonder if there's a way to avoid having to update the docs in multiple places if/when we add additional supported dataframes -- is it possible to link to a single source of truth for every one of those mentions?

That's a great point. I think the px-arguments page can be the source of truth. So I've update the performance page to link that for full details on supported dataframe libraries

doc/python/px-arguments.md
```diff
@@ -101,33 +115,55 @@ fig.show()
...
 
-#### Passing columns as arguments
+> There is a known issue where `px.icicle`, `px.treemap`, and `px.sunburst` used with a `path` parameter won't work with cuDF dataframes.
+
+#### Additional Dependencies Required
+
+- Plotly Express requires NumPy. You can install it with `pip install numpy` if it's not installed by the dataframe library you are using.
+- To use [trendlines](/python/linear-fits/), you'll also need to have pandas installed.
+- To use PySpark dataframes or dataframes that support the dataframe interchange protocol, you'll need to have pandas installed.
+
```
LiamConnors, line 137:
```suggestion
- To use PySpark dataframes, you'll need to have pandas installed. To use dataframes that support the dataframe interchange protocol, you'll need to have PyArrow installed.
```

CHANGELOG.md
```diff
@@ -8,12 +8,29 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Removed
 - Drop deprecated `pointcloud` and `heatmapgl` traces from the API
 - Drop `tenacity` dependency [#4831](https://github.com/plotly/plotly.js/pull/4831)
+- Drop support for Jupyter Notebook 6 and earlier [#4822](https://github.com/plotly/plotly.py/pull/4822)
```
LiamConnors, line 11:
```suggestion
- Drop support for Jupyter Notebook version 6 and earlier [#4822](https://github.com/plotly/plotly.py/pull/4822)
```

CHANGELOG.md
```diff
@@ -34,7 +51,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fixed a bug in integer validation of arrays that threw an error when an array contained a mix of strings and integers.
 
-- Fixed a bug in JupyterLab >= 4 and Jupyter Notebook >= 7 that caused latex to not render in plotly charts. 
+- Fixed a bug in JupyterLab >= 4 and Jupyter Notebook >= 7 that caused latex to not render in plotly charts.
```
gvwilson, line 54: LaTeX not latex

doc/python/bar-charts.md
```diff
@@ -304,6 +304,83 @@ fig.update_layout(barmode='stack')
 fig.show()
 ```
 
+### Bar Chart with Relative Barmode
+
+With "relative" barmode, the bars are stacked on top of one another, with negative values
+below the axis, positive values above.
```
gvwilson, line 310: below the axis _and_ positive values above.

gvwilson (approved):

CHANGELOG.md
LiamConnors, line 54:
```suggestion
- Fixed a bug in JupyterLab >= 4 and Jupyter Notebook >= 7 that caused LaTeX to not render in plotly charts.
```

LiamConnors: oh, it looks like that and some others were added to a past release in error, and not the unreleased version, so hadn't seen. will update and move

doc/python/v4-migration.md
emilykl, line 1: @LiamConnors Do we need to get rid of this document entirely? Is there any issue caused by keeping it around? I could imagine someone out there trying to migrate from a very old version of Plotly might be glad to have it. 

But definitely not opposed to removing it if there's a good reason.

doc/python/v6-changes.md
```diff
@@ -0,0 +1,137 @@
...
+<!-- #region -->
+## Jupyter Notebook Support
+
+Versions of Jupyter Notebook earlier than version 7 are no longer supported. To upgrade to the latest Jupyter Notebook:
+
+```
+pip install notebook --upgrade
+```
+
+## Change to anywidget for go.FigureWidget
```
emilykl, line 48: Is it worth mentioning here that `ipywidgets` is no longer a dependency? (maybe a question for @marthacryan ?)

doc/python/v6-changes.md
```diff
@@ -0,0 +1,137 @@
...
+pip install anywidget
+```
+
+## Processing NumPy and NumPy-Convertible Arrays
+
+Plotly.py now takes advantage of recent changes in how Plotly.js handles typed arrays for improved performance. See the [performance page](python/performance/) for more details.
+
+## Dataframe Support
+
+Plotly Express now uses [Narwhals](https://narwhals-dev.github.io/narwhals/) to natively support pandas, Polars, and PyArrow.
```
emilykl, line 62: Probably worth mentioning at the top level here that performance of Polars and PyArrow is significantly improved (and maybe link to the performance page as well).

doc/python/v6-changes.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Removed Attributes
+
+The following attributes have been removed in Plotly.py 6.
+
+### `titlefont`,`titleposition`, `titleside`, and `titleoffset`
+
+The `titlefont`,`titleposition`, `titleside`, and `titleoffset` attributes have been removed. Replace them with `title.font`, `title.position`, `title.side`, and `title.offset`.
```
emilykl, line 70:
```suggestion
The layout attributes `titlefont`,`titleposition`, `titleside`, and `titleoffset` attributes have been removed. Replace them with `title.font`, `title.position`, `title.side`, and `title.offset`.
```

doc/python/v6-changes.md
```diff
@@ -0,0 +1,137 @@
...
+fig.show()
+```
+
+## Removed Traces
+
+The following traces have been removed.
+
+### `heatmapgl`
+
+The `heatmapgl` trace has been removed. Use [`heatmap`](/python/heatmaps/) instead.
```
emilykl, line 105: How different are the APIs for `heatmapgl` vs. `heatmap`, and `pointcloud` vs. `scattergl`? Is it worth providing a more detailed migration guide explaining how to adapt your code? Or is it fairly straightforward?

doc/python/v6-changes.md
emilykl, line 1: @LiamConnors Just realized there is nothing in here about the Mapbox -> Map migration!

I realize the new Map traces were already added in the last release, but since the deprecation warnings are new, lots of people will probably realize for the first time. Should probably add a header about this and link to the [Maplibre migration doc](https://plotly.com/python/mapbox-to-maplibre/).

doc/python/v6-changes.md
LiamConnors: it's a dependency of `anywidget` so I'd say not worth mentioning.

doc/python/getting-started.md
```diff
@@ -66,17 +66,15 @@ or `conda`:
...
-> Note: **No internet connection, account, or payment is required to use plotly.py.** Prior to version 4, this library could operate in either an "online" or "offline" mode. The documentation tended to emphasize the online mode, where graphs get published to the Chart Studio web service. In version 4, all "online" functionality was removed from the `plotly` package and is now available as the separate, optional, `chart-studio` package (See below). **plotly.py version 4 is "offline" only, and does not include any functionality for uploading figures or data to cloud services.**
+If you want to use Plotly Express, install its required dependencies with:
 
-
-```python
-import plotly.express as px
-fig = px.bar(x=["a", "b", "c"], y=[1, 3, 2])
-fig.write_html('first_figure.html', auto_open=True)
+```
+pip install plotly[express]==5.24.1
```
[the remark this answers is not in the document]
LiamConnors: It's a good question. I suppose it guarantees a user gets the version the docs are about when they install, but it would be nice to have it auto-update.

doc/python/v4-migration.md
LiamConnors: There's no major issue in keeping it around. Mostly removed it because v4 was released over five years ago, and I felt it's probably not very useful anymore, but I see your point. It may be useful to someone, so I'll look at how to keep it around.

CHANGELOG.md
```diff
@@ -3,17 +3,36 @@ All notable changes to this project will be documented in this file.
...
-- Drop `tenacity` dependency [#4831](https://github.com/plotly/plotly.js/pull/4831)
-
-### Updated
-
-- Updated plotly.py to use base64 encoding of arrays in plotly JSON to improve performance.
-- Add `subtitle` attribute to all Plotly Express traces
-- Make plotly-express dataframe agnostic via Narwhals [#4790](https://github.com/plotly/plotly.py/pull/4790)
+- Drop deprecated `pointcloud` and `heatmapgl` traces from the API [[#4815](https://github.com/plotly/plotly.py/pull/4815)]
+- Drop `tenacity` dependency [[#4831](https://github.com/plotly/plotly.py/pull/4831)]
+- Drop support for Jupyter Notebook version 6 and earlier [[#4822](https://github.com/plotly/plotly.py/pull/4822)]
```
archmoj, line 12: Which lower versions are supported?

CHANGELOG.md
```diff
@@ -3,17 +3,36 @@ All notable changes to this project will be documented in this file.
...
+
+### Updated
+- Update plotly.py to use base64 encoding of arrays in plotly JSON to improve performance [[#4470](https://github.com/plotly/plotly.py/pull/4470)].
```
archmoj, line 15:
```suggestion
- Update plotly.py to use base64 encoding of typed arrays e.g. numpy in plotly JSON to keep precision intact and improve performance [[#4470](https://github.com/plotly/plotly.py/pull/4470)].
```

CHANGELOG.md
```diff
@@ -3,17 +3,36 @@ All notable changes to this project will be documented in this file.
...
+- Make plotly-express dataframe agnostic via Narwhals [[#4790](https://github.com/plotly/plotly.py/pull/4790)].
+- Update `go.FigureWidget` to use `anywidget` [[#4823](https://github.com/plotly/plotly.py/pull/4823)]
+- Use modern [native ES6 import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) to load plotly.js bundle instead of requirejs which is [no longer under active development](https://github.com/requirejs/r.js/compare/2.3.6...2.3.7) [[#4736](https://github.com/plotly/plotly.py/pull/4763)]
+- Update Plotly.js from version 2.34.2 to version 3.0.0-rc0 See the [plotly.js CHANGELOG](https://github.com/plotly/plotly.js/blob/master/CHANGELOG.md#300-rc0----2024-11-11) for more information. These changes are reflected in the auto-generated `plotly.graph_objects` module. Notable changes include:
```
archmoj, line 19:
```suggestion
- Update Plotly.js from version 2.34.2 to version 3.0.0-rc.0 See the [plotly.js CHANGELOG](https://github.com/plotly/plotly.js/blob/master/CHANGELOG.md#300-rc0----2024-11-11) for more information. These changes are reflected in the auto-generated `plotly.graph_objects` module. Notable changes include:
```

CHANGELOG.md
```diff
@@ -3,17 +3,36 @@ All notable changes to this project will be documented in this file.
...
+  - Drop support for deprecated attributes `titlefont`, `titleposition`, `titleside`, and `titleoffset` [[#7212](https://github.com/plotly/plotly.js/pull/7212)].
```
archmoj, line 20: Could one still use a sting in place of titles? If no then `title` should also be mentioned for this item.

CHANGELOG.md
```diff
@@ -3,17 +3,36 @@ All notable changes to this project will be documented in this file.
...
+  - Drop support for deprecated `annotation.ref` attribute (use `annotation.xref` and `annotation.yref` instead) [[#7215](https://github.com/plotly/plotly.js/pull/7215)]
+  - Drop support for deprecated error bar `opacity` attribute (use alpha channel of error bar `color` attribute instead) [[#7214](https://github.com/plotly/plotly.js/pull/7216)]
+  - Drop support for deprecated attribute `gl3d.cameraposition` (use `gl3d.camera` instead) [[#7217](https://github.com/plotly/plotly.js/pull/7217)]
+  - Drop deprecated `plot3dPixelRatio` from config [[#7231](https://github.com/plotly/plotly.js/pull/7231)]
+  - Drop deprecated `zauto`, `zmin` and `zmax` from the surface trace [[#7234](https://github.com/plotly/plotly.js/pull/7234)]
+  - Drop deprecated `autotick` attributes from cartesian axes [[#7236](https://github.com/plotly/plotly.js/pull/7236)]
+  - Drop `transforms` from the API [[#7240](https://github.com/plotly/plotly.js/pull/7240), [#7254](https://github.com/plotly/plotly.js/pull/7254)]
+  - Make offsetgroup work with barmode "stacked" and "relative" for bar traces [[#7009](https://github.com/plotly/plotly.js/pull/7009)]
```
archmoj, line 30: /How about moving this item to the top of the list or even presenting it in the `Added` section?

CHANGELOG.md
```diff
@@ -3,17 +3,36 @@ All notable changes to this project will be documented in this file.
...
+  - Deprecate mapbox traces and provide links to migration docs for plotly.js and plotly.py users [[#7260](https://github.com/plotly/plotly.js/pull/7260)]
```
archmoj, line 31: Also perhaps we could move this item up and mention the users are encouraged to switch to new map options.

CHANGELOG.md
```diff
@@ -34,10 +53,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fixed a bug in integer validation of arrays that threw an error when an array contained a mix of strings and integers.
 
-- Fixed a bug in JupyterLab >= 4 and Jupyter Notebook >= 7 that caused latex to not render in plotly charts.
```
archmoj, line 37: Are these corrections to the `5.24.1` changelog?

doc/python/LaTeX.md
```diff
@@ -35,7 +35,7 @@ jupyter:
...
+Figure titles, axis labels and annotations all accept LaTeX directives for rendering mathematical formulas and notation, when the entire label is surrounded by dollar signs `$...$`. This rendering is handled by the [MathJax library](https://www.mathjax.org/) (version 2.7.5), which must be loaded in the environment where figures are being rendered. MathJax is included by default in Jupyter-like environments. When embedding Plotly figures in other contexts it may be required to ensure that MathJax is separately loaded, for example via a `<script>` tag pointing to a content-delivery network (CDN).
```
archmoj, line 38: 1. The latest version of MathJax is v3 which is supported by plotly.js ? Is MathJax v3 supported here too? 
2. Instead of/In addition to https://www.mathjax.org/ we may provide a link to the related version(s) here: https://www.npmjs.com/package/mathjax?activeTab=versions

doc/python/getting-started.md
```diff
@@ -138,20 +136,19 @@ See [_Displaying Figures in Python_](/python/renderers/) for more information on
...
 
-
 #### Jupyter Notebook Support
 
 For use in the classic [Jupyter Notebook](https://jupyter.org/), install the `notebook` and `ipywidgets`
 packages using `pip`:
 
 ```
-$ pip install "notebook>=5.3" "ipywidgets>=7.5"
+pip install "notebook>=7.0" "anywidget>=0.9.13"
```
archmoj, line 145: Should we guard against notebook v8 and higher to avoid potential breakages in feature?

doc/python/performance.md
```diff
@@ -0,0 +1,311 @@
...
+
+- The SVG API, which supports vector rendering.
+- The Canvas API, which supports raster rendering, and can exploit GPU hardware acceleration via a browser technology known as WebGL.
+
+Each `plotly` trace type is rendered with either SVG or WebGL. The following trace types use WebGL for rendering:
+
+* Accelerated versions of SVG trace types: `scattergl`, `scatterpolargl`,
+* High-performance multidimensional trace types: `splom`, or `parcoords`
+* 3D trace types `scatter3d`, `surface`, `mesh3d`, `cone`, `streamtube`
+* Mapbox Gl JS-powered trace types: `scattermapbox`, `choroplethmapbox`, `densitymapbox`
```
archmoj, line 132: Do we want to keep mapbox traces here?
If so then we may add a deprecation note?

doc/python/performance.md
```diff
@@ -0,0 +1,311 @@
...
+`plotly` figures are rendered by web browsers, which broadly speaking have two families of capabilities for rendering graphics:
+
+- The SVG API, which supports vector rendering.
+- The Canvas API, which supports raster rendering, and can exploit GPU hardware acceleration via a browser technology known as WebGL.
+
+Each `plotly` trace type is rendered with either SVG or WebGL. The following trace types use WebGL for rendering:
+
+* Accelerated versions of SVG trace types: `scattergl`, `scatterpolargl`,
+* High-performance multidimensional trace types: `splom`, or `parcoords`
+* 3D trace types `scatter3d`, `surface`, `mesh3d`, `cone`, `streamtube`
```
archmoj, line 131:
```suggestion
* 3D trace types `scatter3d`, `surface`, `mesh3d`, `cone`, `streamtube`, `isosurface`, `volume`
```

CHANGELOG.md
LiamConnors: They were added in error to the changelog file here https://github.com/plotly/plotly.py/pull/4763/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed after the release and don't appear in the release notes on the release tab

doc/python/performance.md
LiamConnors: Good point. Swapped out for the new traces

doc/python/getting-started.md
LiamConnors: I don't think so. We can always update this if there is an issue in a future version.

doc/python/LaTeX.md
LiamConnors: Yes, it should be the same as for Plotly.js (https://github.com/plotly/plotly.js/pull/6073/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R90) so I've updated it.

doc/python/v6-changes.md
LiamConnors: My understanding from the original PR comments https://github.com/plotly/plotly.js/pull/7213 and from testing it with a `heatmapgl` vs `heatmap` that it can be swapped out. Curious if you have any more information on that @archmoj. For `pointcloud`, yeah, maybe, an example could be good. But it looks like we have nothing documented in examples for this at all. @alexcjohnson, would you know what a minimal example would look like for `pointcloud` that could then be converted to `scattergl`?

doc/python/getting-started.md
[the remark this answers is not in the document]
emilykl: In any case, if we keep the version pin, it should be `6.0.0` for this docs update right?

CHANGELOG.md
archmoj: Yes for For Jupyter notebooks. 
Should you revise the following note?
```
Drop support for Jupyter Notebook version 6 and earlier
```

CHANGELOG.md
archmoj: Thank you! Resolved.

CHANGELOG.md
archmoj: Resolved.

doc/python/getting-started.md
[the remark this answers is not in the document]
archmoj: Actually I may argue that we should keep the pin for a while.
Namely if we want to use the RC version, without the pip the latest v5 would be installed not the `6.0.0-rc.0` version.

doc/python/getting-started.md
archmoj: Debatable : )

doc/python/getting-started.md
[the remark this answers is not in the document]
emilykl: True — it depends on whether we are releasing these docs at the time the RC is released, or waiting until the full `6.0.0` release.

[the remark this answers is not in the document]
LiamConnors: We won't release the docs with the RC as we don't have proper versioning to manage it.

CHANGELOG.md
LiamConnors: You mean to specify the versions that do work as well as those that are dropped?

emilykl (approved):
Thanks @LiamConnors 🚀 🚀 🚀

doc/python/performance.md
```diff
@@ -0,0 +1,311 @@
...
+- float32
+- float64
+- int8
+- uint8
+- int16
+- uint16
+- int32
+- uint32
+
+*If the array dtype is **int64** or **uint64**, often the default dtype for arrays in NumPy when no dtype is specified, those dtypes will be changed to other types internally by Plotly.py where possible. When working with NumPY directly, you can [specify the `dtype`](https://numpy.org/doc/stable/user/basics.types.html#array-types-and-conversions-between-types) when creating `ndarray` objects.
```
archmoj, line 77: Slight improvement could be: 
When working with integers it is preferable to specify the `dtype` i.e. to avoid an extra step of finding the minimum and maximum of the list which by default uses Big-Integers (int64).

LiamConnors, line 77:
```suggestion
*If the array dtype is **int64** or **uint64**, often the default dtype for arrays in NumPy when no dtype is specified, those dtypes will be changed to supported types internally by Plotly.py where possible. When working with NumPy directly, you can [also specify the `dtype`](https://numpy.org/doc/stable/user/basics.types.html#array-types-and-conversions-between-types) when creating `ndarray` objects, and Plotly.py won't need to make the conversion internally.
```

LiamConnors: Thanks @archmoj! is your comment because you feel like it's not clear on the benefits of specifying the data type. Do you think this is better?

doc/python/v6-changes.md
```diff
@@ -0,0 +1,141 @@
...
+pip install anywidget
+```
+
+## Processing NumPy and NumPy-Convertible Arrays
+
+Plotly.py now takes advantage of recent changes in how Plotly.js handles typed arrays for improved performance. See the [performance page](python/performance/) for more details.
+
+## Dataframe Support
+
+Plotly Express now uses [Narwhals](https://narwhals-dev.github.io/narwhals/) to natively support pandas, Polars, and PyArrow.
```
LiamConnors, line 62:
```suggestion
Plotly Express now uses [Narwhals](https://narwhals-dev.github.io/narwhals/) to natively support pandas, Polars, and PyArrow. With this change, the [performance](https://plotly.com/python/performance/) of using Polars or PyArrow with Plotly Express is significantly improved. 
```

CHANGELOG.md
LiamConnors: Updated

doc/python/performance.md
LiamConnors: added

doc/python/performance.md
archmoj: Yes. Your suggestion is great. Thank you!

doc/python/v6-changes.md
archmoj: Please make an issue for future and resolve it here.

doc/python/px-arguments.md
```diff
@@ -101,33 +115,52 @@ fig.show()
...
 
-#### Passing columns as arguments
+PySpark dataframes are also supported and are converted to pandas dataframes internally by Plotly Express.
+
+#### Additional Dependencies Required
+
+- Plotly Express requires NumPy. You can install it with `pip install numpy` if it's not installed by the dataframe library you are using.
+- To use [trendlines](/python/linear-fits/), you'll also need to have pandas installed.
+- To use PySpark dataframes, you'll need to have pandas installed. To use dataframes that support the dataframe interchange protocol, you'll need to have PyArrow installed.
+### Example: Using a Pandas DataFrame with `px.bar`
```
LiamConnors, line 135:
```suggestion

### Example: Using a Pandas DataFrame with `px.bar`
```

PR #4790, opened 2024-10-09
feat: make plotly-express dataframe agnostic via narwhals

FBruzzesi:
# Description

This PR migrates plotly-express module logic from pandas to narwhals. In this way, pandas is not a required dependency for plotly-express (or at least for its entirety - e.g. trendlines will still require pandas for now) and users coming with polars, 
 pyarrow or other eager dataframes supported in narwhals do not need to depend on pandas in the first place.
 
Closes #4749 

## Code PR

- [x] I have read through the [contributing notes](https://github.com/plotly/plotly.py/blob/master/contributing.md) and understand the structure of the package. In particular, if my PR modifies code of `plotly.graph_objects`, my modifications concern the `codegen` files and not generated files.
- [x] I have added tests (if submitting a new feature or correcting a bug) or
  modified existing tests.
- [ ] For a new feature, I have added documentation examples in an existing or
  new tutorial notebook (please see the doc checklist as well).
- [x] I have added a CHANGELOG entry if fixing/changing/adding anything substantial.
- [ ] For a new feature or a change in behaviour, I have updated the relevant docstrings in the code to describe the feature or behaviour (please see the doc checklist as well).

## Out of scope for the PR

- No documentation change has been done so far
- ~Adapt plotly data accordingly~ included in this PR
- timeseries/trendlines

cc: @MarcoGorelli @LiamConnors

PR #4843, opened 2024-11-01
feat: allow to load plotly data with other backends

FBruzzesi:
# Description

Allow to load plotly data directly via pandas, polars and pyarrow, without depending directly on any.

It would indirectly address the following [comment](https://github.com/plotly/plotly.py/pull/4790#discussion_r1825172072) in #4790 

## Comments

I found that the docstrings were out of sync with other formats used in the library and the rendering is not what's expected: [plolty.data api reference page](https://plotly.com/python-api-reference/generated/plotly.data.html#module-plotly.data).

I tried to change it to consistently use numpy style.

## Code PR

  modified existing tests.

### DIFF
--- a/packages/python/plotly/plotly/data/__init__.py
+++ b/packages/python/plotly/plotly/data/__init__.py
@@ -1,7 +1,12 @@
 """
 Built-in datasets for demonstration, educational and test purposes.
 """
+import os
+from importlib import import_module
+
 import narwhals.stable.v1 as nw
+
+AVAILABLE_BACKENDS = {"pandas", "polars", "pyarrow"}
 
 
 def gapminder(
@@ -372,9 +377,10 @@ def _get_dataset(d, return_type):
     """
     Loads the dataset using the specified backend.
 
-    Notice that the available backends are 'pandas', 'polars', 'pyarrow' and they all
-    have a `read_csv` function. Therefore we can dynamically load the library via
-    `importlib.import_module` and then call `backend.read_csv(filepath)`.
+    Notice that the available backends are 'pandas', 'polars', 'pyarrow' and they all have
+    a `read_csv` function (pyarrow has it via pyarrow.csv). Therefore we can dynamically
+    load the library using `importlib.import_module` and then call
+    `backend.read_csv(filepath)`.
 
     Parameters
     ----------
@@ -388,23 +394,20 @@ def _get_dataset(d, return_type):
     -------
     Dataframe of `return_type` type
     """
-    import os
-    from importlib import import_module
-
-    AVAILABLE_BACKENDS = {"pandas", "polars", "pyarrow"}
-
     filepath = os.path.join(
         os.path.dirname(os.path.dirname(__file__)),
         "package_data",
         "datasets",
         d + ".csv.gz",
     )
+
     if return_type not in AVAILABLE_BACKENDS:
         msg = f"Unsupported return_type. Found {return_type}, expected one of {AVAILABLE_BACKENDS}"
         raise NotImplementedError(msg)
 
     try:
-        backend = import_module(return_type)
+        module_to_load = "pyarrow.csv" if return_type == "pyarrow" else return_type
+        backend = import_module(module_to_load)
     except ModuleNotFoundError:
         msg = f"return_type={return_type}, but {return_type} is not installed"
         raise ModuleNotFoundError(msg)
--- a/packages/python/plotly/plotly/data/__init__.py
+++ b/packages/python/plotly/plotly/data/__init__.py
@@ -51,9 +51,11 @@ def gapminder(
         df = df.filter(nw.col("year") == year)
     if datetimes:
         df = df.with_columns(
-            nw.concat_str([nw.col("year").cast(nw.String()), nw.lit("-01-01")]).cast(
-                nw.Datetime(time_unit="ns")
-            )
+            # Concatenate the year value with the literal "-01-01" so that it can be
+            # casted to datetime from "%Y-%m-%d" format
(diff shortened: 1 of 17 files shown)


packages/python/plotly/plotly/data/__init__.py
```diff
@@ -1,32 +1,65 @@
...
+    df = nw.from_native(
+        _get_dataset("gapminder", return_type=return_type), eager_only=True
+    )
     if year:
-        df = df[df["year"] == year]
+        df = df.filter(nw.col("year") == year)
     if datetimes:
-        df["year"] = (df["year"].astype(str) + "-01-01").astype("datetime64[ns]")
+        df = df.with_columns(
+            nw.concat_str([nw.col("year").cast(nw.String()), nw.lit("-01-01")]).cast(
```
archmoj, line 54: Would you please explain and possibly add a comment about these operations?
Is it working in every time stamp scenario? Or are there any limitations?
Thank you!

packages/python/plotly/plotly/data/__init__.py
```diff
@@ -128,95 +194,216 @@ def election_geojson():
...
+        "datasets",
+        d + ".csv.gz",
     )
+    if return_type not in AVAILABLE_BACKENDS:
+        msg = f"Unsupported return_type. Found {return_type}, expected one of {AVAILABLE_BACKENDS}"
+        raise NotImplementedError(msg)
+
+    try:
+        backend = import_module(return_type)
+        return backend.read_csv(filepath)
```
archmoj, line 406: Could we also handle the case of `backend` not having a `read_csv` function please?

FBruzzesi: This should never happen, as pandas, pyarrow and polars all implement `read_csv`. Yet I split the loading and reading csv into two separate try-except blocks and added a custom message

packages/python/plotly/plotly/data/__init__.py
FBruzzesi: Changed method used and added in line comment

emilykl:
This looks great, @FBruzzesi . I'll add a few small style comments.

packages/python/plotly/plotly/data/__init__.py
```diff
@@ -128,95 +196,221 @@ def election_geojson():
...
-    import pandas
+    return_type: {'pandas', 'polars', 'pyarrow'}
+        Type of the resulting dataframe
+
+    Returns
+    -------
+    Dataframe of `return_type` type
+    """
     import os
+    from importlib import import_module
```
emilykl, line 392: @archmoj @LiamConnors Is there some performance / other reason you can think of why the imports in this file are all nested inside the individual functions rather than being at the top? If not, I'd prefer to move them to the top.

@FBruzzesi I'd prefer `AVAILABLE_BACKENDS` to be defined at the top as well.

packages/python/plotly/plotly/data/__init__.py
LiamConnors: It looks like this returns `no attribute 'read_csv'` for `pyarrow` (`pyarrow==18.0.0`):

```
Exception: Unable to read 'carshare' dataset due to: module 'pyarrow' has no attribute 'read_csv'
```

Does it need to be accessed with
```
from pyarrow import csv
```
 https://arrow.apache.org/docs/python/generated/pyarrow.csv.read_csv.html

FBruzzesi: Thanks for the catch, just fixed with the last commit

packages/python/plotly/plotly/data/__init__.py
FBruzzesi: Moved imports and `AVAILABLE_BACKENDS` to top of the file with the last commit

packages/python/plotly/plotly/data/__init__.py
LiamConnors: Works well now. Thanks!

archmoj:
Please add a changelog entry.

packages/python/plotly/plotly/data/__init__.py
```diff
@@ -1,32 +1,72 @@
 """
 Built-in datasets for demonstration, educational and test purposes.
 """
+import os
+from importlib import import_module
 
+import narwhals.stable.v1 as nw
 
-def gapminder(datetimes=False, centroids=False, year=None, pretty_names=False):
+AVAILABLE_BACKENDS = {"pandas", "polars", "pyarrow"}
```
FBruzzesi, line 9: If that's of interest, adding modin and cudf is also quite straightforward and all the operations are supported via narwhals

emilykl:
@FBruzzesi Thanks very much for this PR. We have reviewed on the Plotly side and we're happy with these changes. 🚀 

We think the easiest next step is to merge this branch into the existing Narwhals branch -- could you go ahead and do that?

FBruzzesi:
> @FBruzzesi Thanks very much for this PR. We have reviewed on the Plotly side and we're happy with these changes. 🚀
> 
> We think the easiest next step is to merge this branch into the existing Narwhals branch -- could you go ahead and do that?

Hey @emilykl thanks for the feedback! That's great to hear! I kept it separate to lower the amount of changes, but definitly happy to merge it in there since you already reviewed it

emilykl:
@FBruzzesi Yes I think keeping it separate is definitely the right instinct and it made the review much easier. Now that we've reviewed it, I think it will make the merge process smoother to merge the two branches before merging into master. That will also allow updating the tests to use the new return types for px datasets.

FBruzzesi:
Closing because merge in #4790

PR #4866, opened 2024-11-12
Unpin jupyterlab version in optional requirements

marthacryan:
The JupyterLab version was pinned in the optional requirements because JupyterLab 3 allowed for both npm _and_ python packaged extensions, while JupyterLab 4 dropped support for npm extensions. Now that #4779 is merged, we no longer need to support npm extensions, so we can unpin this dependency.

### DIFF
--- a/packages/python/plotly/optional-requirements.txt
+++ b/packages/python/plotly/optional-requirements.txt
@@ -45,7 +45,7 @@ scipy
 
 ## jupyter ##
 jupyter
-jupyterlab ~=3.0  # To package js extension as federated module
+jupyterlab
 ipykernel
 
 ## deps for _county_choropleth.py figure factory


packages/python/plotly/optional-requirements.txt
```diff
@@ -45,7 +45,7 @@ scipy
 
 ## jupyter ##
 jupyter
-jupyterlab ~=3.0  # To package js extension as federated module
+jupyterlab
```
archmoj, line 48: Thanks @marthacryan for the quick fix.

archmoj (approved):

filbranden:
Thanks for the fix! Can we please get a new release of plotly in PyPI that includes this fix? Interested in integrating it with notebook 7 and I found this is what's blocking it. Thanks!

marthacryan:
@filbranden We're about to do a major release so this should all be on PyPi soon! But I believe JupyterLab 4 should work with Plotly? Where are you seeing issues with integrating?

filbranden:
Thank you @marthacryan!

This is what I'm seeing:

```
$ jupyter labextension list
...
        jupyterlab-plotly v5.24.1 enabled  X
...
   The following extensions may be outdated or specify dependencies that are incompatible with the current version of jupyterlab:
        jupyterlab-plotly

   If you are a user, check if an update is available for these packages.
   If you are a developer, re-run with `--verbose` flag for more details.
```

And the `--verbose` output is:

```
"jupyterlab-plotly@5.24.1" is not compatible with the current JupyterLab
Conflicting Dependencies:
JupyterLab        Extension              Package
>=2.0.0 <3.0.0    >=1.2.3 <2.0.0         @lumino/messaging
>=2.3.1-alpha.0 <3.0.0>=1.8.1 <2.0.0         @lumino/widgets
```

filbranden:
Installing `dist/plotly-5.24.1+352.gd31c72885-py3-none-any.whl` built from trunk fixes this warning

Granted, not sure if this warning amounts to any issues in production, but from that warning I tracked down this PR so that's how I got here 😃 

Knowing a new release is forthcoming is much appreciated!

marthacryan:
@filbranden Thanks for the context! Is this affecting the behavior of plotly in your actual notebook editor? The `jupyter labextension list` output is a remnant of older builds of JupyterLab / notebook, but the actual behavior I've seen in Notebook 7 has been working. I'm interested if that's broken for you! Glad to hear that the wheel is working though :D

filbranden:
No in that case nevermind, I just assumed there was something wrong with it based on that warning from this command, if everything is supposed to work correctly then that should be the case. Thanks for following up!

marthacryan:
No problem! Thanks for reaching out

PR #4890, opened 2024-11-19
chore: Address pandas deprecation warning in test

MarcoGorelli:


emilykl (approved):
`H` has been deprecated since pandas version 2.2.0.

Here is the documentation page with the deprecation notice: https://pandas.pydata.org/docs/user_guide/timeseries.html#timeseries-offset-aliases

Thank you @MarcoGorelli for this change!

PR #4900, opened 2024-11-22
Fix: Display deprecated warnings for mapbox traces

### LINKED ISSUE #4730: Add deprecation warning for Mapbox traces
Following #4726, add a deprecation warning to the Python console for Mapbox traces. Something like:
> `px.scatter_mapbox` is deprecated and will be removed in a future release. Please use `px.scatter_map` with MapLibre instead. For more information on migrating to MapLibre, refer to the migration guide [LINK].

archmoj:
Closes #4730.

Adds deprecation warnings which are printed to the console when using one of the following objects / functions:

graph objects:
- `go.Choroplethmapbox`
- `go.Densitymapbox`
- `go.Scattermapbox`

express:
- `px.choropleth_mapbox`
- `px.density_mapbox`
- `px.scatter_mapbox`
- `px.line_mapbox`

Since the graph objects files are generated by codegen, this PR also modifies the codegen process to automatically add the deprecation warnings to the generated files.


emilykl:
@LiamConnors When you have a chance, could you sanity check whether the warnings are working as expected for you on this branch? I.e. that we see a warning when using the deprecated traces, but no warnings otherwise.

LiamConnors:
It's working as expected. Looks good to me.

emilykl:
Thanks Liam. @archmoj @gvwilson This is ready for review.

gvwilson (approved):
thank you

packages/python/plotly/codegen/datatypes.py
```diff
@@ -400,6 +410,19 @@ def __init__(self"""
...
 
+    if node.name_property in deprecated_mapbox_traces:
+        buffer.write(
+            f"""
+        warn(
+            "*{node.name_property}* is deprecated!"
+            + " Use *{node.name_property.replace("mapbox", "map")}* instead."
+            + " Learn more at: https://plotly.com/python/mapbox-to-maplibre/",
+            stacklevel=2,
+            category=DeprecationWarning,
```
archmoj, line 421: @emilykl Good improvement. And I can approve this PR.
As follow ups to this PR:
Could you possibly 
1. double check that `stacklevel=2` is the best option we could use?
2. should we use similar `stacklevel` and `category` options for other warnings in plotly.py?
Thank you!

emilykl: I believe `stacklevel=2` is the correct option in this case -- it causes the warning to appear to be emitted at the place in the user's code where they have called the deprecated function, for example:

```
Making a go.Choroplethmapbox figure...
/Users/ekl/code/plotly.py/my_script.py:32: DeprecationWarning:

*choroplethmapbox* is deprecated! Use *choroplethmap* instead. Learn more at: https://plotly.com/python/mapbox-to-maplibre/
```

(Also, for `DeprecationWarning` specifically, using any other `stacklevel` for these deprecations causes the warning to not appear at all.)

It's just coincidence that `stacklevel=2` is the correct option in this case for both the `go` and `px` deprecations... the correct stacklevel will differ in every case depending on exactly where the warning is being emitted. 

I don't think it's worth the effort to go through the codebase and clean up existing warnings, but I do think it makes sense to use `stacklevel` and `category` consistently going forward. It seems at least some of the warnings already in place in Plotly.py are using them correctly.

PR #4902, opened 2024-11-25
test: Pass native dataframe in `test_pass_df_columns` and `test_category_order_with_category_as_x`

MarcoGorelli:
It looks like in this test a Narwhals DataFrame is being passed to `px.histogram` - however, a user would pass in their native frame, not a Narwhals one, so I think the test should be testing when `tips.to_native()` is passed, not when `tips` is passed. At least, that's what all the other tests do - I presume this was an accident? I might be missing something though

cc @FBruzzesi


FBruzzesi:
Most likely my bad, nice catch @MarcoGorelli

marthacryan (approved):
LGTM

PR #4914, opened 2024-11-28
fix: px.timeline was raising when x_start and/or x_end were already datetime for Polars / PyArrow

### LINKED ISSUE #4913: Plotly express process_dataframe_timeline expects datetime as strings
If the user supplies a dataframe with correct dtypes (datetime), it will err as it's trying to cast them from string.

https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/express/_core.py#L2125-L2129

It would be better if the casting is conditional, because right now I have to cast my datetime back to string, just for the function to cast it back to datetime.

plotly version 6.0.0rc1

MarcoGorelli:
closes #4913


emilykl (approved):
Looks good to me. Thank you @MarcoGorelli for the fix and thank you for adding the test! 🚀

PR #4916, opened 2024-11-28
fix: add narwhals to conda recipe

LiamConnors:

### DIFF
--- a/packages/python/plotly/recipe/meta.yaml
+++ b/packages/python/plotly/recipe/meta.yaml
@@ -25,6 +25,7 @@ requirements:
     - setuptools
   run:
     - python
+    - narwhals >=1.13.3
 
 test:
   imports:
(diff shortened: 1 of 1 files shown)


emilykl:
@LiamConnors Under what circumstances is this `meta.yaml` file used? Is it just used by CI/development or by end users of Plotly.py?

emilykl:
> @LiamConnors Under what circumstances is this `meta.yaml` file used? Is it just used by CI/development or by end users of Plotly.py?

Answered offline, the `meta.yaml` specifies dependencies to install when using `conda install`. 

Ideally IMO we should have just a single source of truth for dependencies so we don't need to update them in two places but in the meantime yes let's add `narwhals` in here to fix the conda installs.

packages/python/plotly/recipe/meta.yaml
```diff
@@ -25,6 +25,7 @@ requirements:
     - setuptools
   run:
     - python
+    - narwhals >=1.13.3
```
emilykl, line 28:
```suggestion
    - narwhals >=1.15.1
```

Update `narwhals` version following merge of #4921

PR #4921, opened 2024-12-01
fix: pandas DataFrame with non-default Index was resulting in scrambled data in scatter_matrix

### LINKED ISSUE #3576: parallel_coordinates is incorrect when using `color` and non-sorted index
This is somewhat of a weird bug, but I don't think it would be that uncommon of a use-case.

First let's see it working as expected.

If you run:

```
df = pd.DataFrame({
        'A': [1, 2, 3, 4],
        'B': [1, 2, 3, 4],
        'C': [1, 2, 3, 4],
    },
    index=[0, 1, 2, 3]
    #index=[3, 2, 1, 0]
)
df

px.parallel_coordinates(df)
```

you'll see

Screen Shot 2022-01-30 at 11 43 31 AM

Then run:

```
px.parallel_coordinates(df, color='B')
```

and you'll see

Screen Shot 2022-01-30 at 11 43 58 AM

Everything looks good so far.

But then change the index of the DataFrame from  `index=[0, 1, 2, 3]` to `index=[3, 2, 1, 0]`

```
df = pd.DataFrame({
        'A': [1, 2, 3, 4],
        'B': [1, 2, 3, 4],
        'C': [1, 2, 3, 4],
    },
    #index=[0, 1, 2, 3]
    index=[3, 2, 1, 0]
)
df

px.parallel_coordinates(df)
```

This will return exactly the same graph as before. But when you use the `color` param, you get this

```
px.parallel_coordinates(df, color='B')
```

Screen Shot 2022-01-30 at 11 45 58 AM

Which is incorrect. For example, instead of 1-1-1 (which is still a single row, despite the change in index), the line goes from 1-4-1.

### LINKED ISSUE #3577: scatter_matrix is incorrect when using color and non-sorted index
related to https://github.com/plotly/plotly.py/issues/3576

```
df = pd.DataFrame({
        'A': [1, 2, 3, 4],
        'B': [1, 2, 3, 4],
        'C': [1, 2, 3, 4],
    },
    index=[3, 2, 1, 0]
)
px.scatter_matrix(df)
px.scatter_matrix(df, color='C')
```

`px.scatter_matrix(df)` produces:

Screen Shot 2022-01-31 at 9 51 05 AM

`px.scatter_matrix(df, color='C')` produces:

Screen Shot 2022-01-31 at 9 49 30 AM

### LINKED ISSUE #4788: Using a `DataFrame` with a scrambled index and `color` in `px.scatter_matrix()` yields mismatched/null colors
When using a `DataFrame` with scrambled (and optionally sparse) index values (like you'd get from `sklearn.model_selection.train_test_split()`) and the `color` field in a `px` function, the `DataFrame` is split into a `dict` of `Series` in [`process_args_into_dataframe()`](https://github.com/plotly/plotly.py/blob/7181c4dbdd86400b2b984434e25128507aa8ea69/packages/python/plotly/plotly/express/_core.py#L1100). The column corresponding to the `color` argument has its index reset via `to_unindexed_series()`, but the rest of the columns do not get reset. For sparse shuffled indices, the reset indices of the color column are re-joined with the original, sparse indices of the other columns, so indices that match are joined but the ones that do not match end up as `NaN`s. For non-sparse indices, the result is arguably worse: the indices all get re-joined upon re-creation of the `DataFrame`, and since the indices are not sparse, they will all match, but to incorrect values, so you don't get any indication that something went wrong other than the colors mismatching. While this doesn't seem to affect `px.scatter()` in a simple test (it seems that doesn't end up returning any `df_output` at all), I'm sure other `px` functions are affected by this.

I'm not entirely sure what the purpose of these calls to `to_unindexed_series` are for, considering that the indices will be re-joined properly if they are not reset, and commenting out the code at https://github.com/plotly/plotly.py/blob/7181c4dbdd86400b2b984434e25128507aa8ea69/packages/python/plotly/plotly/express/_core.py#L1249-L1251 fixes the issue for my use-case, but I have no idea what other errors this may introduce. I can try removing all calls to the function and running tests to see if this causes any regressions.

Re-joining the `DataFrame` using the original indices and just never resetting them seems like the best option to me, but I'm not sure if other parts of the code depend on certain columns having ordered/non-sparse indices. Maybe a simple `reset_index` on the entire `DataFrame` up-front would be best, and then we don't have to worry about some columns getting reset while others don't?

Here's a minimal example:

```python
import numpy as np
import pandas as pd
import plotly.express as px

n_rows_total = 100
n_rows_to_sample = 50

df = pd.DataFrame(
    {
        "v1": np.random.randn(n_rows_total),
        "v2": np.random.randn(n_rows_total),
        "v3": np.random.randn(n_rows_total),
        "v_color": np.random.randint(low=0, high=3, size=n_rows_total).astype(float),
    }
)

# Correct result
px.scatter_matrix(df, color="v_color").show()

# Colors (and color values) will match with incorrect data points
px.scatter_matrix(df.sample(n_rows_total), color="v_color").show()

# "null" color values
px.scatter_matrix(df.sample(n_rows_to_sample), color="v_color").show()
```

### LINKED ISSUE #4917: plotly.express.scattermatrix displays no color if supplied a pandas.DataFrame with a DataTime index
The color does not get diplayed when using a DataFrame with a DateTime index.

Have a look at the following example:
```
import pandas as pd
import plotly.express as px
df = pd.DataFrame({
        "x":[1,2,3,4],
        "y":[10,20,10,20],
        "z":[-1,-2,-3,-4],
        "color":[1,2,3,4]
    })
df.index = pd.DatetimeIndex(["1/1/2020 10:00:00+00:00", "2/1/2020 11:00:00+00:00", "3/1/2020 10:00:00+00:00", "4/1/2020 11:00:00+00:00"])
px.scatter_matrix(df, color = "color")
```

Image

There are no problems when you comment out the line `df.index = ...`.

Maybe this issue is related to https://github.com/plotly/plotly.py/issues/4788

MarcoGorelli:
These issue get closed "for free" with the latest Narwhals release, as a result of https://github.com/narwhals-dev/narwhals/issues/1474 having been addressed:

- closes #4917
- closes #4788
- closes #3577
- closes #3576

Furthermore, it allows for the Plotly codebase to be simplified a bit, as `maybe_reset_index` isn't necessary any more. So, I've:
- made that simplification
- bumped the minimum Narwhals 
- added a test to check this stays fixed

Demo:

Latest stable plotly release:

image

This branch:

image

### DIFF
--- a/packages/python/plotly/plotly/express/_core.py
+++ b/packages/python/plotly/plotly/express/_core.py
@@ -1168,16 +1168,13 @@ def _escape_col_name(columns, col_name, extra):
     return col_name
 
 
-def to_unindexed_series(x, name=None, native_namespace=None):
-    """Assuming x is list-like or even an existing Series, returns a new Series (with
-    its index reset if pandas-like). Stripping the index from existing pd.Series is
-    required to get things to match up right in the new DataFrame we're building.
-    """
+def to_named_series(x, name=None, native_namespace=None):
+    """Assuming x is list-like or even an existing Series, returns a new Series named `name`."""
     # With `pass_through=True`, the original object will be returned if unable to convert
     # to a Narwhals Series.
     x = nw.from_native(x, series_only=True, pass_through=True)
     if isinstance(x, nw.Series):
-        return nw.maybe_reset_index(x).rename(name)
+        return x.rename(name)
     elif native_namespace is not None:
         return nw.new_series(name=name, values=x, native_namespace=native_namespace)
     else:
@@ -1306,7 +1303,7 @@ def process_args_into_dataframe(
                                 length,
                             )
                         )
-                    df_output[col_name] = to_unindexed_series(
+                    df_output[col_name] = to_named_series(
                         real_argument, col_name, native_namespace
                     )
                 elif not df_provided:
@@ -1343,7 +1340,7 @@ def process_args_into_dataframe(
                     )
                 else:
                     col_name = str(argument)
-                    df_output[col_name] = to_unindexed_series(
+                    df_output[col_name] = to_named_series(
                         df_input.get_column(argument), col_name
                     )
             # ----------------- argument is likely a column / array / list.... -------
@@ -1362,7 +1359,7 @@ def process_args_into_dataframe(
                             argument.name is not None
                             and argument.name in df_input.columns
                             and (
-                                to_unindexed_series(
+                                to_named_series(
                                     argument, argument.name, native_namespace
                                 )
                                 == df_input.get_column(argument.name)
@@ -1380,7 +1377,7 @@ def process_args_into_dataframe(
                         % (field, len_arg, str(list(df_output.keys())), length)
                     )
 
-                df_output[str(col_name)] = to_unindexed_series(
+                df_output[str(col_name)] = to_named_series(
                     x=argument,
                     name=str(col_name),
                     native_namespace=native_namespace,
--- a/packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
+++ b/packages/python/plotly/plotly/tests/test_optional/test_px/test_px_input.py
@@ -435,6 +435,29 @@ def test_splom_case(backend):
     assert np.all(fig.data[0].dimensions[0].values == ar[:, 0])
 
 
+def test_scatter_matrix_indexed_pandas():
+    # https://github.com/plotly/plotly.py/issues/4917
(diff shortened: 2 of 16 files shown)


emilykl (approved):
Thanks for the fix, @MarcoGorelli. Sorry for the delayed review, I wanted to take the time to make sure I understood the differences between the Pandas and Narwhals index behavior. Looks good! 🚀

PR #4922, opened 2024-12-02
fix: Skip base64 conversion for empty arrays (fixes #4919)

### LINKED ISSUE #4919: Exception raised with empty dataframe in v6.0.0rc0
Our App Studio test suite caught an issue with this code in the latest RC of plotly.py:
```
df = pd.DataFrame({
  'x': [],
  'y': [],
})
fig = px.bar(df, x="x", y="y")
```

In v5.24.1, this produces an empty graph as expected
In v6.0.0rc0, this raises an Exception: `ValueError: zero-size array to reduction operation maximum which has no identity`

This is a simple case, but a more realistic use case would be when a dataframe is filtered down to nothing due to user input.

emilykl:
Closes #4919 

The error was caused by trying to take the min/max of an empty array. 

The error only shows up when the empty array is typed, and is specifically a 'big int' type (`int64` or `uint64`), because the min/max call only happens in those cases.

This PR:
- Modifies `to_typed_array_spec()` function to just skip b64 encoding entirely in the case of an empty array
- Adds tests for the scenario where an empty array of type `int64` is passed into a figure (`go` and `px`). These tests fail on `master` but pass on this branch.

### DIFF
--- a/packages/python/plotly/_plotly_utils/utils.py
+++ b/packages/python/plotly/_plotly_utils/utils.py
@@ -51,8 +51,8 @@ def to_typed_array_spec(v):
 
     # convert default Big Ints until we could support them in plotly.js
     if dtype == "int64":
-        max = v.max() if v else 0
-        min = v.min() if v else 0
+        max = v.max() if v.size > 0 else 0
+        min = v.min() if v.size > 0 else 0
         if max <= int8max and min >= int8min:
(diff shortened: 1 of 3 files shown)


marthacryan:
I think it might be a better solution to just not convert empty arrays to base64.

marthacryan (approved):

PR #4926, opened 2024-12-03
patch: deepcopy figure fix

### LINKED ISSUE #4925: Invalid value of type 'builtins.dict'  error when copying figure in 6.0.0rc0
The example below throws an error in 6.0.0rc0 but not in 5.24.1

```
    raise ValueError(
ValueError: 
    Invalid value of type 'builtins.dict' received for the 'x' property of scatter
        Received value: {'dtype': 'i2', 'bdata': 'oAelB6oHrwe0B7kHvgfDB8gHzQfSB9cH'}

    The 'x' property is an array that may be specified as a tuple,
    list, numpy array, or pandas Series
```

```
from dash import Dash, dcc, html
import plotly.express as px

import copy

app = Dash(__name__)

gapminder = px.data.gapminder().query("(country=='Canada') | (country=='United States') | (country=='Mexico')")

fig = px.line(gapminder, x="year", y="gdpPercap", color='country')
fig.update_yaxes(title_text='Year')
fig.update_xaxes(title_text='Adoption Rate')

# Default dimensions
fig.update_layout(
    margin=dict(l=60, r=0, t=0, b=60),
)

fig_custom_height = copy.deepcopy(fig)
fig_custom_height.update_layout(height=800)

fig_custom_short_height = copy.deepcopy(fig)
fig_custom_short_height.update_layout(height=200)

app.layout = [
                dcc.Graph(
                figure = fig_custom_height,
                ),
                dcc.Graph(
                figure = fig_custom_short_height,
            ),

]

if __name__ == '__main__':
    app.run_server(debug=True)

```

FBruzzesi:
## Code PR

Closes #4925 by reverting the changes in  https://github.com/plotly/plotly.py/commit/960adb9b9a89387d05343497de1df5d3df592698 to check for `is_none_or_typed_array_spec`.

Looking for indication if worth adding a test. I see that in few tests a deepcopy of figure is called, thus... I am a bit confused 🙃

### DIFF
--- a/packages/python/plotly/_plotly_utils/basevalidators.py
+++ b/packages/python/plotly/_plotly_utils/basevalidators.py
@@ -223,6 +223,17 @@ def type_str(v):
     return "'{module}.{name}'".format(module=v.__module__, name=v.__name__)
 
 
+def is_typed_array_spec(v):
+    """
+    Return whether a value is considered to be a typed array spec for plotly.js
+    """
+    return isinstance(v, dict) and "bdata" in v and "dtype" in v
+
+
+def is_none_or_typed_array_spec(v):
+    return v is None or is_typed_array_spec(v)
+
+
 # Validators
 # ----------
 class BaseValidator(object):
@@ -393,8 +404,7 @@ class DataArrayValidator(BaseValidator):
 
     def validate_coerce(self, v):
 
-        if v is None:
-            # Pass None through
+        if is_none_or_typed_array_spec(v):
             pass
         elif is_homogeneous_array(v):
             v = copy_to_readonly_numpy_array(v)
@@ -591,8 +601,7 @@ class EnumeratedValidator(BaseValidator):
         return False
 
     def validate_coerce(self, v):
-        if v is None:
-            # Pass None through
+        if is_none_or_typed_array_spec(v):
             pass
         elif self.array_ok and is_array(v):
             v_replaced = [self.perform_replacemenet(v_el) for v_el in v]
@@ -636,8 +645,7 @@ class BooleanValidator(BaseValidator):
         )
 
     def validate_coerce(self, v):
-        if v is None:
-            # Pass None through
+        if is_none_or_typed_array_spec(v):
             pass
         elif not isinstance(v, bool):
             self.raise_invalid_val(v)
@@ -661,8 +669,7 @@ class SrcValidator(BaseValidator):
         )
 
     def validate_coerce(self, v):
-        if v is None:
-            # Pass None through
+        if is_none_or_typed_array_spec(v):
             pass
         elif isinstance(v, str):
             pass
@@ -752,8 +759,7 @@ class NumberValidator(BaseValidator):
         return desc
 
     def validate_coerce(self, v):
-        if v is None:
-            # Pass None through
+        if is_none_or_typed_array_spec(v):
             pass
         elif self.array_ok and is_homogeneous_array(v):
             np = get_module("numpy")
@@ -899,8 +905,7 @@ class IntegerValidator(BaseValidator):
         return desc
 
     def validate_coerce(self, v):
-        if v is None:
-            # Pass None through
+        if is_none_or_typed_array_spec(v):
             pass
         elif v in self.extras:
             return v
@@ -1063,8 +1068,7 @@ class StringValidator(BaseValidator):
         return desc
 
     def validate_coerce(self, v):
-        if v is None:
-            # Pass None through
+        if is_none_or_typed_array_spec(v):
             pass
         elif self.array_ok and is_array(v):
 
@@ -1365,8 +1369,7 @@ class ColorValidator(BaseValidator):
         return valid_color_description
 
     def validate_coerce(self, v, should_raise=True):
-        if v is None:
-            # Pass None through
+        if is_none_or_typed_array_spec(v):
             pass
         elif self.array_ok and is_homogeneous_array(v):
             v = copy_to_readonly_numpy_array(v)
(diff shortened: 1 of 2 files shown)


packages/python/plotly/_plotly_utils/basevalidators.py
```diff
@@ -1708,16 +1710,17 @@ def description(self):
...
         """.format(
             plotly_name=self.plotly_name,
-            array_ok=", or a list, numpy array or other iterable thereof"
-            if self.array_ok
-            else "",
+            array_ok=(
+                ", or a list, numpy array or other iterable thereof"
+                if self.array_ok
+                else ""
+            ),
```
FBruzzesi, line 1717: 🤔 different black version?

marthacryan:
I'd love to see a test that _fails_ before these changes! I'm a little concerned that the existing ones didn't catch this

FBruzzesi:
Hey @marthacryan , thanks for getting back to me. Should the example in the issue enough as a test?

stefanv:
Here's a reproducing example that came out of the skimage CI:

```python
import numpy as np

import plotly
import plotly.express as px

plotly.io.renderers.default = "sphinx_gallery_png"

img = np.zeros((50, 50))
fig = px.imshow(img, binary_string=True)

fig.add_trace(
    go.Scatter(
        x=np.array([[1,2,3],[4,5,6]]),
        y=np.array([[1,2,3],[4,5,6]]),
        name="1"
    )
)

plotly.io.show(fig)
```

Which results in:

```
ValueError:
    Invalid value of type 'builtins.dict' received for the 'x' property of scatter
        Received value: {'dtype': 'i1', 'bdata': 'AQIDBAUG', 'shape': '2, 3'}

    The 'x' property is an array that may be specified as a tuple,
    list, numpy array, or pandas Series
```

If you toggle `binary_string` to False, it brings up another unexpected error:

```
ValueError:
    Invalid value of type 'builtins.dict' received for the 'z' property of heatmap
        Received value: {'dtype': 'f8', 'bdata': 'AAAAAAAAAAAAAAAAAA...AAA=', 'shape': '50, 50'}

    The 'z' property is an array that may be specified as a tuple,
    list, numpy array, or pandas Series
```

This is plotly 6.0.0rc0 and kaleido 1.0.0rc0.

FBruzzesi:
I am not particularly confident with this part of the code. Minimal example failing in master:

```py
import copy
import plotly.express as px

gapminder = px.data.gapminder()
fig = px.line(gapminder, x="year", y="gdpPercap", color="country")
fig_copy = copy.deepcopy(fig)
```

The entire error trace gives some more insights, here what I could figure out (maybe it is trivial, yet it wasn't for me)

- `deepcopy` calls `__reduce__`, which calls  `to_dict`, which calls `convert_to_base64`.
- At this point every value is converted in the following form:
    ```py
    'x' = {'dtype': 'i2', 'bdata': 'oAelB6oHrwe0B7kHvgfDB8gHzQfSB9cH'},
    'y' = {'dtype': 'f8', 'bdata': 'ESBU8...7M'}
    ```
- Then `deepcopy` will call `_reconstruct`, which tries to run `Figure(*converted_values)`, hence `BaseFigure`, and finally this will call `validate_coerce` which will fail.

Hope it helps

packages/python/plotly/_plotly_utils/basevalidators.py
marthacryan: Oh interesting! Yeah not sure why that's happening but it looks fine?

packages/python/plotly/_plotly_utils/tests/validators/test_fig_deepcopy.py
```diff
@@ -0,0 +1,11 @@
...
+import plotly.express as px
+
+
+def test_deepcopy():
+    gapminder = px.data.gapminder()
+    fig = px.line(gapminder, x="year", y="gdpPercap", color="country")
+
+    fig_copy = copy.deepcopy(fig)
+
+    assert fig_copy is not None
```
marthacryan, line 11: This test looks good thank you for doing that! Do you think you could add a simple check that the data did copy correctly? Also, where are the other deepcopy tests you mentioned? Does it make more sense to include this test with those?

FBruzzesi: Thanks @marthacryan , what's the best way to assert that the values are the same? is it `fig_copy.to_dict() == fig.to_dict()`?

Here are few other figure deepcopy's in the test suite:
- [test_update_subplots](https://github.com/plotly/plotly.py/blob/725290068d5ea3f7f344eec7d7b84c58b85cfb38/packages/python/plotly/plotly/tests/test_core/test_update_objects/test_update_subplots.py#L364)
- [test_update_traces](https://github.com/plotly/plotly.py/blob/725290068d5ea3f7f344eec7d7b84c58b85cfb38/packages/python/plotly/plotly/tests/test_core/test_update_objects/test_update_traces.py#L295)
- [test_deepcopy_pickle.py](https://github.com/plotly/plotly.py/blob/725290068d5ea3f7f344eec7d7b84c58b85cfb38/packages/python/plotly/plotly/tests/test_io/test_deepcopy_pickle.py#L38)

packages/python/plotly/_plotly_utils/basevalidators.py
FBruzzesi: Yeah I would not worry too much about this, I just wonder why that's the case 😇

marthacryan (approved):
Looks great, thanks @FBruzzesi!

packages/python/plotly/_plotly_utils/tests/validators/test_fig_deepcopy.py
marthacryan: I think using `fig.to_dict()` is fine here. There are other ways to compare them but I just wanted a baseline check that the data is still there

PR #4968, opened 2025-01-14
Remove mention of fig.show

LiamConnors:


marthacryan (approved):
LGTM! I think we can do a larger review of this part later but for now this keeps it accurate

PR #4969, opened 2025-01-14
Remove specific version from README

marthacryan:

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@
 
 ## Quickstart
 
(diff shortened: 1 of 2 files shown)


LiamConnors:
There's a reference to updating that file here: https://github.com/plotly/plotly.py/blob/master/release.md?plain=1#L28-L33

marthacryan:
@LiamConnors Good point! Just removed it.

LiamConnors (approved):

PR #4951, opened 2024-12-21
Acepace remove future

acepace:
This removes some dead code, starting with the low hanging fruit mentioned in #4817

Playing around with the zip imports but it's not PR ready yet.

PR #4978, opened 2025-01-22
feat: modifying code generation to reduce bundle size

gvwilson:
feat: modify code generation to reduce bundle size

| branch   | format  | bytes    | %age |
| -------- | ------- | -------- | ---- |
| master   | .whl    | 14803768 |      |
| codegen2 | .whl    | 12215667 | -18% |
| master   | .tar.gz |  8100014 |      |
| codegen2 | .tar.gz |  6114458 | -24% |

1.  Assign an empty string to the `data_docs` field of generated
    validators.  (This has a major impact because those docs are
    duplicated many times.)

1.  Alias name of base validator during import in
    `codegen/validators.py`.

1.  Remove the long list of CSS colors from help strings for color
    properties.

1.  Replace `super(Parent, self)` with `super()` in generated code.

1.  Modify `commands.py` to run code generation.

1.  Remove comments from generated code.

1.  Replaced named arguments in constructors with positional arguments.

1.  Drop use of sys.version_info and TYPE_CHECKING.  Removed the check
    for Python < 3.7 using `sys.version_info` and as a backup checking
    `typing.TYPE_CHECKING`; this saves a little space and also cleans
    up the code.

1.  Introduce `_init_provided()` for `BaseFigure` and `BasePlotlyType`
    that calls a helper function `_initialize_provided()` to replace
    repetitions of:

```
_v = arg.pop("something", None)
_v = something if something is not None else _v
if _v is not None:
    self["something"] = _v
```

Used in #5008.


gvwilson:
see also #4951

emilykl:
@gvwilson High-level question -- what is the practical impact of some of these changes on the docs / docstrings, for example, removing the list of CSS colors or the list of dict properties?

Completely agree that having those large chunks of text repeated many times over is a very poor use of bundle bytes. But are there cases where it makes the docstrings less useful, and if so are there any changes we need to make in the docs to mitigate the impact?

codegen/__init__.py
```diff
@@ -26,6 +26,10 @@
     get_data_validator_instance,
 )
 
+# Target Python version for code formatting with Black.
+# Must be one of the values listed in pyproject.toml.
+BLACK_TARGET_VERSION = "py311"
```
emilykl, line 31: [Black docs](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#t-target-version) say to pass ALL Python versions supported by the code as a `--target-version` — should we follow that recommendation?

I guess the configuration in the `pyproject.toml` doesn't play nice with the codegen, which is why we need to redefine `target_version` here?

emilykl:
@gvwilson Could we add a single line at the top of all codegen files saying `# THIS FILE IS AUTO-GENERATED`?

I realize that runs _exactly_ opposite to the goal of reducing bundle size, but it would make development a lot easier to have all code-generated files identified in the file itself.

(Could be a separate PR.)

codegen/resources/plot-schema.json
emilykl, line 1: Why is this file showing up in the diff? Bad merge? Should probably just defer to whatever version of this file is on `main`.

codegen/utils.py
```diff
@@ -456,9 +452,7 @@ def get_validator_params(self):
 
         if self.is_compound:
             params["data_class_str"] = repr(self.name_datatype_class)
-            params["data_docs"] = (
-                '"""' + self.get_constructor_params_docstring() + '\n"""'
-            )
+            params["data_docs"] = '"""\n"""'
```
emilykl, line 455: Do we know if removing this docstring has any impact on the docs, or otherwise impacts the dev process? I am assuming not, but I do wonder why it was there in the first place.

commands.py
```diff
@@ -299,9 +299,18 @@ def update_plotlyjs_dev():
     run_codegen()
 
 
-if __name__ == "__main__":
-    if "updateplotlyjsdev" in sys.argv:
+def main():
+    if len(sys.argv) != 2:
+        print(USAGE, file=sys.stderr)
```
emilykl, line 304: 👍

plotly/basedatatypes.py
```diff
@@ -834,6 +847,14 @@ def _ipython_display_(self):
         else:
             print(repr(self))
 
+    def _init_provided(self, name, arg, provided):
```
emilykl, line 850: Could this function be called something like `_set_property(self, prop_name, value, defaults)`?

I find the existing name hard to parse.

emilykl:
@gvwilson Looks good, see my comments for some questions I have, but none of them are blockers to merging — except possibly the removal of the `TYPE_CHECKING` if clause, which I don't fully understand.

codegen/__init__.py
```diff
@@ -267,36 +267,24 @@ def perform_codegen():
         root_datatype_imports.append(f"._deprecations.{dep_clas}")
 
     optional_figure_widget_import = f"""
-if sys.version_info < (3, 7) or TYPE_CHECKING:
```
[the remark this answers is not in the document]
gvwilson: `typing.TYPE_CHECKING` is false at runtime but treated as true by type-checking tools, so it's a way to conditionalize behavior. I think we should rely on comment annotations to tell pyright and other tools what we want them to do rather than trying to work around them like this, but I'll defer to people who know more than I do - @T4rk1n do you have a preference?

gvwilson:
> @gvwilson High-level question -- what is the practical impact of some of these changes on the docs / docstrings, for example, removing the list of CSS colors or the list of dict properties?
> 
> Completely agree that having those large chunks of text repeated many times over is a very poor use of bundle bytes. But are there cases where it makes the docstrings less useful, and if so are there any changes we need to make in the docs to mitigate the impact?

@emilykl agreed - we will need to provide a canonical list of colors and fonts. @LiamConnors would you like to figure out where you'd like to put this and then I can add the URL into plotly.js (to be copied into plotly.py by codegen) that can stay in the docstring?

codegen/__init__.py
gvwilson: didn't realize Black wanted all versions - I'll update that. Am i correct that we currently support 3.9 - 3.12 inclusive ?

gvwilson:
> @gvwilson Could we add a single line at the top of all codegen files saying `# THIS FILE IS AUTO-GENERATED`?

Excellent idea - it will only add a few bytes per file but will save a lot of grief.

codegen/resources/plot-schema.json
gvwilson: agreed.

codegen/utils.py
gvwilson: didn't seem to have an impact…

plotly/basedatatypes.py
gvwilson: sure - I'll incorporate that in the next update to the PR. thanks for the suggestion.

codegen/__init__.py
gvwilson: I've modified `BLACK_TARGET_VERSION` to match `pyproject.toml`.

LiamConnors (approved):

PR #4979, opened 2025-01-23
Create outliers

eldercamposds:
creates a multi-column plot for outlier analysis

### DIFF
--- a/packages/python/plotly/plotly/outliers
+++ b/packages/python/plotly/plotly/outliers
@@ -1,0 +1,15 @@
+df outliers(df : pandas.DataFrame, num_col : int ):
+
+  colunas = df.select_dtypes(include=['float64', 'int64']).columns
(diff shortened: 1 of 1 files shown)


gvwilson:
Hi - can you please provide some more detail on this? In particular, it looks like the first line of the PR isn't valid Python (the first `df` should be `def`).

eldercamposds (commented):
I wrote "def " wrong, sorry

gvwilson:
Hi @eldercamposds - I'm going to close this one again as there are no tests or documentation, the file doesn't appear to have a `.py` extension, and it (probably) doesn't belong in the `plotly` directory. To get this on track, please open an issue in this repository to explain what you're trying to accomplish (i.e., what feature you want to add and why) so we can discuss a plan there. thanks - @gvwilson

PR #5002, opened 2025-01-30
Move `packages/python/plotly` to top level

emilykl:
- Moves the contents of `packages/python/plotly` to the top level of the repository
- Updates paths across the repository to correspond to the new location

The full diff is pretty unwieldy, Github does not perform well with large diffs.

Link to the diff which excludes the commit where files were moved (more manageable): https://github.com/plotly/plotly.py/pull/5002/files/32adcc3de1cf3f6d15b35a4ed85e981bc9a8b24b..54d076b7c734d99e8efe9a96491cbe6b09adbc3c


gvwilson (approved):

marthacryan:
So sad that github diff doesn't work with this PR 😭 I'll just leave comments on what I find here because I basically can't leave comments in the diff. 
* README.md could include a link to the plotly-geo repo since that has its own section there. Just in case people want to make issues or even just to track down where it is. 
* Looks like the `.gitignore` needs to be updated - specifically:
```
packages/javascript/jupyterlab-plotly/lib/
plotly/jupyterlab_plotly/labextension/
plotly/jupyterlab_plotly/nbextension/index.js*
```
* I feel like we should remove the whole section on tox in the `CONTRIBUTING.md` but that's for a different PR

Overall, I think this is ready to merge even if there will be some finishing touches afterward.

marthacryan (approved):

gvwilson:
@emilykl would you like to push the big green button or shall I?

emilykl:
@gvwilson I'll make the changes Martha suggested and then merge!

LiamConnors:
The make file for the API reference docs needs an update for the new structure:
https://github.com/plotly/plotly.py/blob/simplify-dir-structure/doc/apidoc/Makefile
It still references `packages/python...`

emilykl:
Thanks @LiamConnors , good catch.

I've updated the paths -- how would you recommend I verify that the Makefile is correct now?

LiamConnors:
> Thanks @LiamConnors , good catch.
> 
> I've updated the paths -- how would you recommend I verify that the Makefile is correct now?

If you install the `requirements.txt` in `doc` and then go to `doc/apidoc`, running `make html` should generate the output.
I've tested it with these changes and the API docs build correctly.

LiamConnors (approved):

PR #5004, opened 2025-01-31
Add note on DDK compatibility 

LiamConnors:


ndrezn (approved):

ndrezn:
@T4rk1n do you agree that documentation is the right solution here? I don't think we can fix this in the package as it's a missing pin in old versions.

PR #5019, opened 2025-02-07
Unpin test requirements

marthacryan:
* Replaces the separated files which pinned different versions of packages for each version of python with two files that list the unpinned requirements for core / optional tests
* Updates the CI config to be more parameterized 
* Removes the no numpy tests - it turns out that it wasn't actually testing without numpy, and since we added narwhals our tests do in fact fail without numpy
* Removes orca tests from config
* Adds uv into a couple of more spots to speed up the dependency downloads
* Removes unused commit step from plotly.js dev build job
* Removes unused import of ipython and sage salvus in `plotly/tools.py`


tests/test_optional/test_px/test_px_input.py
```diff
@@ -14,8 +14,7 @@
 import warnings
 
 
-# FIXME: don't test with vaex if vaex isn't installed
-if (optional_imports.get_module("vaex") is None) and (sys.version_info >= (3, 12)):
+if optional_imports.get_module("vaex") is None:
```
emilykl, line 17: @marthacryan I would keep the upper bound here but just change it to complement what's in the `requirements.txt` so that if vaex randomly doesn't get installed when it should, the test will fail.

```suggestion
if optional_imports.get_module("vaex") is None and sys.version_info > (3, 9):
```

.circleci/config.yml
```diff
@@ -516,18 +395,28 @@ workflows:
...
+            parameters:
+              python_version:
+                - "3.8"
+                - "3.9"
+                - "3.10"
+                - "3.11"
+                - "3.12"
+            name: "Test optional with python << parameters.python_version >>"
+      - test_optional_py:
+          name: "Test optional with pandas 1.2.4"
```
emilykl, line 418:
```suggestion
          name: "Test optional with python << parameters.python_version >> and pandas << parameters.pandas_version >>"
```

(does this work?)

marthacryan: Unfortunately it doesn't look like the other one is working so I don't think we can actually use this here :(

test_requirements/requirements_core.txt
```diff
@@ -0,0 +1,3 @@
+requests
+pytest
+narwhals
```
LiamConnors, line 3: Is `plotly` installed by the time these requirements are installed? If so, might it be a good idea to remove `narwhals` here to ensure `narwhals` was correctly installed as a dependency of `plotly`?

marthacryan: Ah yeah that's a good idea. @emilykl Do you agree?

marthacryan: Ok - I think this change requires merging #5044 because narwhals isn't being installed by default. I'll merge with main once that's in and remove narwhals here!

marthacryan: Ok - I think this change requires merging #5044 because narwhals isn't being installed by default. I'll merge with main once that's in and remove narwhals here!

marthacryan: Done!

emilykl, line 3: @marthacryan remove `narwhals` here too!

marthacryan: Oh good catch! Didn't realize it was in both

.circleci/config.yml
```diff
@@ -43,15 +48,24 @@ commands:
...
+            not:
+              equal:
+                - <<parameters.pandas_version>>
+                - ""
+          steps:
+            - run:
+                name: Install pandas
+                command: |
+                  source .venv/bin/activate
+                  pip install pandas==<<parameters.pandas_version>>
```
emilykl, line 67: @marthacryan I think this should be 
```suggestion
                  uv pip install pandas==<<parameters.pandas_version>>
```

marthacryan: That actually caused issues for some reason.. I had that originally and it threw a weird error.

.circleci/config.yml
```diff
@@ -516,18 +399,28 @@ workflows:
...
+      - test_optional_py:
+          matrix:
+            parameters:
+              python_version:
+                - "3.8"
+                - "3.9"
+                - "3.10"
+                - "3.11"
+                - "3.12"
+            name: "Test optional with python << parameters.python_version >>"
```
emilykl, line 420: Get rid of this since it's not working?
```suggestion
```

.circleci/config.yml
emilykl: Otherwise `uv pip install pip && python -m pip install` should work

.circleci/config.yml
```diff
@@ -142,110 +136,35 @@ jobs:
...
+      python_version: <<parameters.python_version>>
     steps:
       - test_optional:
-          py: "312_np2"
-
+          pandas_version: <<parameters.pandas_version>>
+          numpy_version: <<parameters.numpy_version>>
 
   # Percy
   python_39_percy:
```
emilykl, line 170: @marthacryan Line 207 in this job should be `uv pip install "pandas==1.1.5"`

emilykl (approved):
@marthacryan LGTM once my final comments are addressed! 🚀 

This is a huge win for maintainability.

A few big-picture thoughts that could be addressed in another PR:

- IMO we should bring back the no-numpy job (to prevent `numpy` dependency from accidentally slipping into core Plotly.py)

- If we're testing `pandas<2` with Python 3.9, should we also test `pandas<2` for other Python versions? Probably worth a discussion.

PR #5023, opened 2025-02-10
Merge test and tests folder and clean up files

marthacryan:
* Removes `test/vv`
* Moves `test/percy` to `tests/percy` to consolidate test directories
  * Updates `.gitignore` and circle CI config to point to new location of `percy` directory


emilykl:
@marthacryan IMO -- move the existing tests which were inside `tests/` into `tests/unit/` so that we have

```
tests/
  - percy/
  - unit/
    - test_core/ 
    - etc.    
```

Not strictly necessary because it seems like `pytest tests/` "just works" and skips the `percy/` dir anyway, but I think it would be ideal to have a directory which contains _only_ the Pytest tests in case we need it.

emilykl (approved):
🚀

PR #5029, opened 2025-02-11
Use uv instead of pip for CI dependencies

marthacryan:


.circleci/config.yml
```diff
@@ -39,51 +39,52 @@ commands:
...
+            source .venv/bin/activate
             pytest tests/test_plotly_utils/
           no_output_timeout: 20m
       - run:
           name: Test io
           command: |
-            . venv/bin/activate
-            pytest tests/test_io
+            source .venv/bin/activate
+            python -m pytest tests/test_io
```
emilykl, line 77: should be consistent about `python -m pytest` vs. `pytest` -- I believe since we are doing an install step first (`uv pip install .`), just `pytest` is fine

.circleci/config.yml
```diff
@@ -39,51 +39,52 @@ commands:
...
           no_output_timeout: 20m
       - run:
-          name: Test dependencdies not imported
+          name: Test dependencies not imported
```
emilykl, line 80: ❤️

.circleci/config.yml
```diff
@@ -426,17 +423,11 @@ jobs:
             sudo apt-get install -y poppler-utils libxtst6 xvfb libgtk2.0-0 libgconf-2-4 libnss3 libasound2 rename
             echo 'export PATH="/home/circleci/project/node_modules/.bin:$PATH"' >> $BASH_ENV
 
-      - save_cache:
```
emilykl, line 429: @marthacryan Are you thinking the install is fast enough now that we can skip this caching step?

emilykl: Oh wow I just noticed the cache keys for saving and loading are different 🤦‍♀️ (`v1` vs. `v2`). So effectively we have not been caching since https://github.com/plotly/plotly.py/pull/4937.

.circleci/config.yml
marthacryan: Ah yeah I added that because for whatever reason test_io was failing because it wasn't picking up `anywidget` in the environment. Adding this fixed it, presumably because it uses the dependencies that are in the python environment? Not sure why the other ones are working but I could make them all use this? I think using the global pytest cli can cause dependency issues in general

.circleci/config.yml
marthacryan: Yeah

.circleci/config.yml
emilykl: Oh interesting. Let's change them all to `python -m pytest` then?

marthacryan: Done!

emilykl (approved):
🚀

PR #5044, opened 2025-02-18
Add back in express extra dependencies and required dependencies

### LINKED ISSUE #5041: Release builds in CI not installing dependencies and not recognising [express] extra
If you download and install a release build from CI, it no longer picks up that there is an `express` extra:

```
Processing .../plotly-6.0.0-py3-none-any.whl (from plotly==6.0.0)
WARNING: plotly 6.0.0 does not provide the extra 'express'
````

Additionally, required dependencies like `narwhals` are not installed. 

@marthacryan

### PR DESCRIPTION
**Pull request by marthacryan**
Fixes #5041 - @LiamConnors I haven't tested this through the CI artifacts yet but locally these changes are working for me!

Edit: tested the CI artifact and it's also working. Also, check out what it looks like in test.pypi.org! https://test.pypi.org/project/plotly/

### CODE REVIEW

**Comment by emilykl** [comment]
Are we OK with getting rid of standalone `requires-install` and `requires-express` files? I can't think of any reason why it should be an issue, but not sure why we had them in the first place.

@marthacryan Make sure to remove them from `MANIFEST.in`.

**Comment by marthacryan** [comment]
@emilykl Those files were used as arguments for the `setup.py`, but they had so few requirements in them that I felt like it made more sense to just put them inline. It shouldn't cause any issues for anyone.

Also, I'll remove them from the `MANIFEST.in`~

File: pyproject.toml [hunk 1]
```diff
@@ -38,9 +38,14 @@ classifiers = [
 requires-python = ">=3.8"
 license = {file="LICENSE.txt"}
 version = "6.0.0"
-dynamic = ["dependencies"]
+dependencies = [
+    "narwhals>=1.13.3",
```
emilykl, line 42: @marthacryan bump to `1.15.1`

emilykl, line 42:
```suggestion
    "narwhals>=1.15.1",
```

emilykl (approved):
LGTM! 🚀

PR #5052, opened 2025-02-21
Add initial test pypi release workflow

marthacryan:
Adds a github actions workflow that runs on push to the `main` branch that will release most recent changes to test.pypi.org for easier testing.


.github/workflows/test-release.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Install npm dependencies
+      run: |
+        cd js
+        npm ci --verbose
+        npm run build --verbose
+
+    - name: Install pypa/build
+      run: >-
+        python3 -m
+        pip install
```
ndrezn, line 33: Is it worth using `uv` here? no external dependencies here so not a lot to reconcile but nonetheless.

marthacryan: Yeah I don't know that it is worth it - we'd just be installing one package

ndrezn:
We now have admin access on the `plotly` Test PyPi namespace.

.github/workflows/test-release.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    - name: Install npm dependencies
+      run: |
+        cd js
+        npm ci --verbose
+        npm run build --verbose
+
+    - name: Install pypa/build
+      run: >-
+        python3 -m
```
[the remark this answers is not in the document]
marthacryan: I think this was just what was in the sample in the python packing documentation for releasing with GitHub actions

emilykl (approved):

PR #5062, opened 2025-03-04
Support Kaleido v1 in Plotly.py

emilykl:
Support both the Kaleido v0 and v1 APIs in Plotly.py. 

- Update `fig.write_image()`, `fig.to_image()`, `pio.write_image()`, and `pio.to_image()` to work with both Kaleido versions
- Add deprecation warnings for:
  - the `engine` argument to `to_image()` and `write_image()`
  - Use of Orca
  - Use of Kaleido v0
  - Setting defaults via `kaleido.scope.*`
- Add a new way of setting image export defaults: `plotly.io.defaults.*`. This can be used in place of `kaleido.scope.*` and supports all the same settings: `default_format`, `default_height`, `default_width`, `default_scale`, `mathjax`, and `topojson`. `pio.defaults` is backwards-compatible with Kaleido v0.
- Add `pio.write_images()` function for generating multiple images at a time
  - Required due to the new Kaleido architecture; _much_ faster than calling `write_image()` in a loop
  - Add `pio.get_chrome()` function which wraps `kaleido.get_chrome_sync()`, and add a CLI entry point `plotly_get_chrome` for installing Chrome from the command line
- Update `tests/test_optional/test_kaleido/test_kaleido.py` tests to work with either Kaleido v0 or v1. They now also actually check that an image of the right format and dimensions has been generated, as opposed to using mocking to check that the right arguments are being passed to Kaleido.
- Add test for `pio.write_images()`
- Update CI to run all tests with Kaleido v1, and add an additional job which runs the Kaleido tests with Kaleido v0, to protect against regressions

------

TODO

- [x] Update CI to run all tests with Kaleido v1
- [x] Add CI job to run Kaleido tests with Kaleido v0
- [x] Finalize UX for what happens if Chrome is not found
- [x] Finalize UX / API for EPS
- [x] Figure out what to do about 'renderers' Kaleido test
- [x] Add deprecation warning for `engine` argument to `write_image` / `to_image`
- [x] Re-add removed Orca support but with DeprecationWarning
- [x] Fix DeprecationWarnings not showing up
- [x] Expose API for setting default `mathjax` and `topojson`


plotly/io/_kaleido.py
```diff
@@ -35,100 +47,52 @@ def to_image(
...
-          - 'webp'
-          - 'svg'
-          - 'pdf'
-          - 'eps' (Requires the poppler library to be installed and on the PATH)
+            - 'png'
+            - 'jpg' or 'jpeg'
+            - 'webp'
+            - 'svg'
+            - 'pdf'
+            - 'eps' (Requires the poppler library to be installed and on the PATH)
```
gvwilson, line 55: do we still handle EPS?

plotly/io/_kaleido.py
```diff
@@ -35,100 +47,52 @@ def to_image(
...
     validate: bool
         True if the figure should be validated before being converted to
         an image, False otherwise.
 
-    engine: str
-        Image export engine to use:
-         - "kaleido": Use Kaleido for image export
-         - "orca": Use Orca for image export
-         - "auto" (default): Use Kaleido if installed, otherwise use orca
+    engine (deprecated): str
```
gvwilson, line 85: do we print a deprecation warning if this argument is used?

plotly/io/_kaleido.py
```diff
@@ -137,12 +101,32 @@ def to_image(
...
+    # Convert figure to dict (and validate if requested)
     fig_dict = validate_coerce_fig_to_dict(fig, validate)
-    img_bytes = scope.transform(
-        fig_dict, format=format, width=width, height=height, scale=scale
-    )
+
+    # Request image bytes
+    if kaleido_major > 0:
+        # Kaleido v1
+        opts = {
```
gvwilson, line 110: this took me a couple of moments to parse :-)

plotly/io/_kaleido.py
```diff
@@ -341,7 +316,24 @@ def full_figure_for_development(fig, warn=True, as_dict=False):
             "To suppress this warning, set warn=False"
         )
 
-    fig = json.loads(scope.transform(fig, format="json").decode("utf-8"))
+    if kaleido_major > 0:
+        # Kaleido v1
+        try:
+            json_file = Path(tempfile.mkstemp(suffix=".json")[1])
```
gvwilson, line 322: why write to a temporary file rather than write to an in-memory buffer?

tests/test_optional/test_kaleido/test_kaleido.py
```diff
@@ -1,66 +1,31 @@
...
-    original_scope = pio._kaleido.scope
-    pio._kaleido.scope = scope_mock
-    try:
-        yield scope_mock
-    finally:
-        pio._kaleido.scope = original_scope
+def check_image(path_or_buffer, size=(700, 500), format="PNG"):
+    if format == "PDF":
+        img = PdfReader(path_or_buffer)
+        # TODO: There is a conversion factor needed here
```
gvwilson, line 15: why?

gvwilson (commented):
Left a few comments and a question, happy to read through again as soon as you want.

plotly/io/_kaleido.py
emilykl: Not yet, assuming we are in agreement about removing this argument, I'll add one

plotly/io/_kaleido.py
emilykl: Actually I think this can be removed now that https://github.com/plotly/Kaleido/issues/276 is closed.

plotly/io/_kaleido.py
emilykl: Not yet supported by Kaleido v1 -- here is the issue: https://github.com/plotly/Kaleido/issues/280

tests/test_optional/test_kaleido/test_kaleido.py
emilykl: Part of the TODO is for me to educate myself on how Plotly currently determines PDF size when writing to PDF. :) 

But the `size` argument is measured in pixels (at least for raster file types) and PDFs have no concept of pixels so I wouldn't expect the PDF to report the same "size" as passed in the argument necessarily.

ayjayt:
@ayjayt just mentioning myself to add this to my github pendings

plotly/io/_kaleido.py
emilykl: @gvwilson Following up on this — Kaleido v1 does not support EPS yet. So either we drop support for EPS entirely, or document that EPS is only available with Kaleido v0 and add an informative error message.

gvwilson: please do the latter - thanks

tests/test_optional/test_kaleido/test_kaleido.py
ayjayt: I will also take a looksy on that issue, lets say for this and the above tomorrow. I am flying today.

plotly/io/_kaleido.py
ayjayt: I can take a look at all this, but does `calc_bytes_sync` not fulfill this use case? I would imagine it would return an UTF-8 stream in the case of format being json

emilykl: Good point @ayjayt, I suppose this logic could be implemented by _only_ calling `calc_fig_sync` in Plotly.py (not using `write_fig_sync` at all).

I can't remember if there's a reason I rejected that approach initially but I'll give it a shot -- that would resolve the buffer issue, you're right.

ayjayt: I'm still going to implement the buffer FYI

marthacryan:
Do we need to maintain backwards compatibility with older Kaleido versions here?

plotly/io/_kaleido.py
marthacryan: I'm in agreement that we should remove this argument

tests/test_optional/test_kaleido/test_kaleido.py
marthacryan, line 1: Are there percy tests that use kaleido to make sure that the actual images generated are correct?

plotly/io/_kaleido.py
```diff
@@ -35,100 +47,52 @@ def to_image(
...
+            - 'jpg' or 'jpeg'
+            - 'webp'
+            - 'svg'
+            - 'pdf'
+            - 'eps' (Requires the poppler library to be installed and on the PATH)
 
-        If not specified, will default to:
-             - `plotly.io.kaleido.scope.default_format` if engine is "kaleido"
-             - `plotly.io.orca.config.default_format` if engine is "orca"
+        If not specified, will default to `plotly.io.kaleido.scope.default_format`
```
LiamConnors, line 57: Do these references to `plotly.io.kaleido.scope..` need to be something else if that's no longer imported here
https://github.com/plotly/plotly.py/pull/5062/files#diff-40813ac13aafeaa7135da57a86bec52cae638135b0acf9324677c9c4eee2ba2bR1

emilykl: Yes, those docstrings should be changed -- I need to research where those defaults are defined in Kaleido v1. (@ayjayt do you know off the top of your head?)

ayjayt: https://github.com/plotly/Kaleido/blob/d493b9dc0c04940721dc2d67e8b342d31fc6bb18/src/py/kaleido/_fig_tools.py#L9-L13

The order is *png default* overridden by *user-specified output path w/ extension* overriden by *user-specified format in opts*

.circleci/config.yml
```diff
@@ -117,6 +117,35 @@ commands:
...
+          name: Install dependencies
+          command: |
+            curl -LsSf https://astral.sh/uv/install.sh | sh
+            uv venv
+            source .venv/bin/activate
+            uv pip install .
+            uv pip install -r ./test_requirements/requirements_optional.txt
+            # Install Kaleido v1 instead of the default version
+            uv pip uninstall -y kaleido
+            uv pip install 'git+https://github.com/plotly/Kaleido.git@v1.0.0rc10#subdirectory=src/py'
```
ayjayt, line 135: `uv pip install kaleido>=1.0.0 || uv pip install 'git+https://github.com/plotly/Kaleido.git@latest-tag#subdirectory=src/py'` is my recommendation

emilykl: 👍 I'll just update the CI manually once Kaleido 1.0.0 full version is on PyPI.

plotly/io/_kaleido.py
```diff
@@ -35,65 +49,61 @@ def to_image(
...
+
     # Handle engine
     # -------------
+    if engine is not None:
+        warnings.warn(
+            f"DeprecationWarning: The 'engine' argument is deprecated. Kaleido will be the only supported engine after {ENGINE_SUPPORT_TIMELINE}.",
+            DeprecationWarning,
+        )
+        engine = "auto"
+
```
LiamConnors, line 104: Should `engine = "auto"` here be in an `else` block so we'll default to using Kaleido when nothing is specified for engine?

emilykl: Good q -- my feeling is that we should NOT change the default engine behavior until the next major release, because that could be a breaking change. I.e. currently if no engine is specified, we decide which engine to use based on what is installed -- if we change it to default to Kaleido, that will potentially cause errors for users who don't specify `engine` and only have Orca installed.

LiamConnors: I agree. I think the way engine is set here does change the existing behaviour. If you call `to_image` with `engine="orca"`, this sets it to "auto", which means kaleido will still be used if kaleido is also installed. And the user won't see the orca deprecation argument further down.

emilykl: OH -- you're absolutely right, thank you for catching @LiamConnors !

plotly/io/_kaleido.py
```diff
@@ -7,9 +7,32 @@
...
+kaleido_scope_default_getwarning = (
+    lambda x: f"""
+Accessing plotly.io.kaleido.scope.{x} is deprecated and will be removed after {ENGINE_SUPPORT_TIMELINE}.
+Please use plotly.io.defaults.{x} instead.
+"""
+)
+
+kaleido_scope_default_setwarning = (
+    lambda x: f"""
+Setting plotly.io.kaleido.scope.{x} is deprecated and will be removed after {ENGINE_SUPPORT_TIMELINE}. "
```
LiamConnors, line 23:
```suggestion
Setting plotly.io.kaleido.scope.{x} is deprecated and will be removed after {ENGINE_SUPPORT_TIMELINE}. 
```

emilykl: fixed, thanks!

plotly/io/_defaults.py
```diff
@@ -0,0 +1,16 @@
...
+    """
+    Class to store default settings for image generation.
+    """
+
+    def __init__(self):
+        self.default_format = "png"
+        self.default_width = 700
+        self.default_height = 500
+        self.default_scale = 1
+
```
LiamConnors, line 14: Will there be a way to set `mathjax`, `topojson`, and `mapbox_access_token`, the other export settings available in v0?

emilykl: `mapbox_access_token` isn't relevant anymore since Mapbox traces are no longer supported; the other two I guess we should expose. I'll add a TODO.

emilykl:
Quick comment to note that commit [54985b8](https://github.com/plotly/plotly.py/pull/5062/commits/54985b81f0134093a2ff564539d7f67d7c5d030e) on this branch has been released as beta version [v6.1.0b0](https://github.com/plotly/plotly.py/releases/tag/v6.0.1b0).

plotly/io/_kaleido.py
emilykl: Resolved -- these defaults are now defined in the Plotly codebase in [`plotly/io/_defaults.py`](https://github.com/plotly/plotly.py/blob/c99073627719ddc818fe01c2d223f451dfa02f53/plotly/io/_defaults.py) and I've updated the docstrings.

plotly/io/_defaults.py
emilykl: `topojson` has been added but I have not tested it yet; `mathjax` is added in `defaults` but not functional quite yet

emilykl: `pio.defaults.topojson` and `pio.defaults.mathjax` are both confirmed functional now (just need a new Kaleido release to fix a small bug with `mathjax`)

gvwilson (approved):

PR #4822, opened 2024-10-22
Drop notebook 6

marthacryan:
**Note: this PR is on top of the changes from #4823 because the plotly.js version used in plotly.py is based on the jupyterlab extension code, and we needed a new place to store the plotly.js version. The widget seemed like a good place because all of the other javascript is removed in this PR.**

Fixes #4779. 

It appears that the entire `packages/javascript/jupyterlab_plotly` directory contains code that is either used to support very old versions of jupyter lab/notebook/voila, or is not used at all. This PR removes the entire directory in favor of having a very small `packages/python/plotly/js` directory that contains only the widget code (see #4663 for adding the widget code into that directory).

The last time that the jupyter extension was significantly changed was 3 years ago in #3142. @fcollonval described the files in the `packages/javascript/jupyterlab_plotly` directory in [this comment](https://github.com/plotly/plotly.py/pull/3142#issuecomment-819637791). He did add support for federated extensions (new jupyter lab/notebook build), but I wasn't able to find evidence that having an extension installed in jupyterlab >=3 or jupyter notebook >= 7 is even necessary for plotly. @fcollonval if you know of any reason that that isn't true, please let me know! 

To test:
* Checkout branch
* Create a new python env (conda, mamba, pipenv, etc)
* `git clean -f -x -d`
* `cd packages/python/plotly`
* `pip install -e .`
* `pip install jupyter`
* Try it out in jupyter lab: `jupyter lab`
* Try it out in jupyter notebook: `jupyter notebook`

PR #5096, opened 2025-03-17
Add back the mime renderer JL extension

### LINKED ISSUE #5056: plotly on jupyter creates HUGE notebooks file sizes
Hi, we are seeing an issue with plotly v6.0.0 where running in jupyter creates HUGE notebook files, resulting in crashed kernel and inability to save.

Sample code to reproduce the issue (full example can be found [here](https://gist.github.com/urirosenberg/c1b3c1f3eaa48472ec0b5368ceb25be9)):
```python
import plotly.express as px
import pandas as pd
import numpy as np

# Create sample data
dates = pd.date_range(start='2024-01-01', end='2024-12-31', freq='M')
values = np.random.randn(12).cumsum()
categories = ['A', 'B', 'C', 'D']
values_bar = [23, 45, 56, 78]

# Line Chart
fig1 = px.line(x=dates, y=values, 
               title='Monthly Trend 2024',
               labels={'x': 'Date', 'y': 'Value'})
fig1.show()

# Scatter Plot with custom styling
fig2 = go.Figure()
fig2.add_trace(go.Scatter(
    x=np.random.randn(100),
    y=np.random.randn(100),
    mode='markers',
    marker=dict(
        size=10,
        color=np.random.randn(100),
        colorscale='Viridis',
        showscale=True
    )
))
fig2.update_layout(title='Scatter Plot with Color Scale')
fig2.show()

# Bar Chart
fig3 = px.bar(x=categories, y=values_bar,
              title='Category Distribution',
              labels={'x': 'Category', 'y': 'Value'},
              color=values_bar,
              color_continuous_scale='Reds')
fig3.show()
```
Using plotly version 5.24.1, this generates a >1M file.
Using plotly version 6.0.0, this generates a 19M file.

Looking at the v6.0.0 notebook source json file, I see massive amounts of JS code.

marthacryan:
Fixes #5056 by adding back a mime renderer extension removed in #4822. We didn't realize at the time that this part of the code was serving an important purpose in keeping filesizes down for offline users.


marthacryan:
Pinning the version of shapely here for tests to pass, but we should address that in another PR.

gvwilson (approved):

.circleci/config.yml
```diff
@@ -166,6 +163,61 @@ jobs:
           pandas_version: <<parameters.pandas_version>>
           numpy_version: <<parameters.numpy_version>>
 
+  check-js-build:
```
emilykl, line 166: I would name this something more specific, e.g. `check-jupyterlab-extension-js-build`

.circleci/config.yml
```diff
@@ -166,6 +163,65 @@ jobs:
...
+      - run:
+          name: Skip if no JS changes
+          command: |
+            git diff --name-only origin/main...HEAD | grep '^js/' || exit 0
+      - run:
+          name: Copy current files to a temporary directory
+          command: |
+            cp -R plotly/labextension/ plotly/labextension-tmp/
+      - run:
+          name: Install dependencies
```
emilykl, line 184:
```suggestion
          name: Install dependencies and build Jupyterlab extension
```

.circleci/config.yml
```diff
@@ -166,6 +163,65 @@ jobs:
...
+            source .venv/bin/activate
+            uv pip install jupyter
+            cd js
+            npm ci
+            npm run build
+      - run:
+          name: Check JupyterLab build artifacts
+          command: |
+            # 1. Compare all files except 'Only in' cases
+            diff -qr plotly/labextension-tmp/static plotly/labextension/static | grep -v '^Only in' > build_diff.txt || true
```
emilykl, line 197: I'm guessing the `Only in` cases are cases where just the filename is different?

.circleci/config.yml
```diff
@@ -166,6 +163,65 @@ jobs:
...
+            cat build_diff.txt
+
+            # Remove the "load" line from both package.json files before comparing
+            grep -v '"load": "static/' plotly/labextension/package.json > pkg1.json
+            grep -v '"load": "static/' plotly/labextension-tmp/package.json > pkg2.json
+
+            # Compare stripped versions
+            diff pkg1.json pkg2.json > package_json_diff.txt || true
```
emilykl, line 205: What's the purpose of the `|| true`?

.circleci/config.yml
```diff
@@ -166,6 +163,65 @@ jobs:
...
+            if [ -s build_diff.txt ] || [ -s package_json_diff.txt ]; then
+              echo "❌ Build artifacts differ:"
+              echo "--- Unexpected diffs ---"
+              cat build_diff.txt
+              echo "--- Unexpected package.json diffs ---"
+              cat package_json_diff.txt
+              echo "Please replace the `plotly/labextension` directory with the artifacts of this CI run."
+              exit 1
+            else
+              echo "✅ Build artifacts match expected output (ignoring known 'load' hash in package.json)."
```
emilykl, line 217:
```suggestion
              echo "✅ Build artifacts in `plotly/labextension` match expected output (ignoring known 'load' hash in package.json)."
```

.circleci/config.yml
```diff
@@ -269,22 +325,13 @@ jobs:
 
     steps:
       - checkout
-
-      - run:
-          name: initial NPM Build
```
emilykl, line 274: I sort of think we still want this step when doing the release build... although I suppose we don't have a separate "release build" step at the moment that's not also run on every push to a PR.

So adding a separate "release build" step could come in another PR.

js/package.json
```diff
@@ -1,16 +1,36 @@
...
-    "lodash-es": "^4.17.21",
-    "plotly.js": "3.0.1"
-  },
-  "devDependencies": {
-    "@types/plotly.js": "^2.33.4",
-    "esbuild": "^0.23.1",
-    "typescript": "^5.6.2"
+    "name": "jupyterlab-plotly",
+    "main": "lib/mimeExtension.js",
+    "version": "6.0.0",
```
emilykl, line 4: this should be 6.0.1, right?

js/package.json
```diff
@@ -1,16 +1,36 @@
...
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/plotly/plotly.py"
+    },
+    "description": "The plotly Jupyter extension",
```
emilykl, line 9:
```suggestion
    "description": "The Plotly JupyterLab extension",
```

emilykl:
@marthacryan Looks good! Can you update `CONTRIBUTING.md` to explain how the Jupyterlab extension build process works (and also explain the new CI step)?

js/package.json
marthacryan: Oh right, I didn't want to add another place that people need to update the version so I wasn't sure what to add here. I'm not even really sure what the purpose of the version is here but it didn't work without a version added.

.circleci/config.yml
marthacryan: Ended up moving away from this

.github/workflows/check-js-build.yml
```diff
@@ -0,0 +1,77 @@
+on: push
+
+jobs:
+  check-js-build:
+    name: Check JS build artifacts
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        persist-credentials: false
```
emilykl, line 10: is this needed?

.github/workflows/check-js-build.yml
```diff
@@ -0,0 +1,77 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.x"
+        
+    - name: Install Node
+      uses: actions/setup-node@v2
+      with:
+        node-version: '18'
```
emilykl, line 19: node 18 is EOL, use 22 or 24

.github/workflows/check-js-build.yml
```diff
@@ -0,0 +1,77 @@
...
+    - name: Copy current files to a temporary directory
+      run: |
+        cp -R plotly/labextension/ plotly/labextension-tmp/
+
+    - name: Check if JS or labextension files changed
+      run: |
+        git fetch origin ${{ github.base_ref }}
+        CHANGED=$(git diff --name-only origin/${{ github.base_ref }}...HEAD | grep -E '^js/|^plotly/labextension/' || true)
+        if [ -z "$CHANGED" ]; then
+          exit 0
```
emilykl, line 30: This won't skip the rest of the steps, need to find out how to exit the workflow in GH actions

.github/workflows/check-js-build.yml
```diff
@@ -0,0 +1,77 @@
...
+        npm run build
+    - name: Check JupyterLab build artifacts
+      run: |
+        # 1. Hash contents of all static files, sort by content hash
+        find plotly/labextension/static -type f -exec sha256sum {} \; | awk '{print $1}' | sort > new_hashes.txt
+        find plotly/labextension-tmp/static -type f -exec sha256sum {} \; | awk '{print $1}' | sort > old_hashes.txt
+
+        # 2. Compare the sorted content hashes
+        diff old_hashes.txt new_hashes.txt > content_diff.txt
+        cat content_diff.txt
```
emilykl, line 50:
```suggestion
```

.github/workflows/check-js-build.yml
marthacryan: oh lol I got that from the other github actions file. I'll update that one too

.github/workflows/check-js-build.yml
marthacryan: good point. Also got that from the other script

emilykl (approved):
🚀

PR #5112, opened 2025-03-25
Fix issue of plotly.js being embedded twice in notebook

marthacryan:
Fixes part of #5056. This bug was introduced by #4763.


emilykl (approved):
🔥

PR #5122, opened 2025-04-01
fix a typo in the documentation of show

williamstein:
The docs for the height parameter said that it sets the width.


LiamConnors (approved):
Thanks for the PR!

Issue #5058, opened 2025-02-27
Update the subplot x and y axis title using add_trace


Folarin14:
The snippet below gotten from [plotly documentation](https://plotly.com/python/subplots/) is unneccesrily verbose. Users should be able to update x and y axis title for respective subplots by passing an argument to the add_trace() method where the row and col are also already specified. I can't imagine having more than a few subplots to make with this approach. Having a master x and y axis title is only great if all subplots are displaying the same type of data which is not always the case in science

```
from plotly.subplots import make_subplots
import plotly.graph_objects as go

# Initialize figure with subplots
fig = make_subplots(
    rows=2, cols=2, subplot_titles=("Plot 1", "Plot 2", "Plot 3", "Plot 4")
)

# Add traces
fig.add_trace(go.Scatter(x=[1, 2, 3], y=[4, 5, 6]), row=1, col=1)
fig.add_trace(go.Scatter(x=[20, 30, 40], y=[50, 60, 70]), row=1, col=2)
fig.add_trace(go.Scatter(x=[300, 400, 500], y=[600, 700, 800]), row=2, col=1)
fig.add_trace(go.Scatter(x=[4000, 5000, 6000], y=[7000, 8000, 9000]), row=2, col=2)

# Update xaxis properties
fig.update_xaxes(title_text="xaxis 1 title", row=1, col=1)
fig.update_xaxes(title_text="xaxis 2 title", range=[10, 50], row=1, col=2)
fig.update_xaxes(title_text="xaxis 3 title", showgrid=False, row=2, col=1)
fig.update_xaxes(title_text="xaxis 4 title", type="log", row=2, col=2)

# Update yaxis properties
fig.update_yaxes(title_text="yaxis 1 title", row=1, col=1)
fig.update_yaxes(title_text="yaxis 2 title", range=[40, 80], row=1, col=2)
fig.update_yaxes(title_text="yaxis 3 title", showgrid=False, row=2, col=1)
fig.update_yaxes(title_text="yaxis 4 title", row=2, col=2)
```
Something along the lines of the below snipppet would be better. Even better is if you can loop through the number of traces you created defined by the rows and cols argument of the subplot. See [issue](https://github.com/plotly/plotly.py/issues/5053)
`fig.add_trace(go.Scatter(x=[1, 2, 3], y=[4, 5, 6]), row=1, col=1, xaxes_title='Date', yaxes_title='rainfall')`

Thanks

gvwilson:
[...]

PR #5141, opened 2025-04-12
Added x and y axis titles to add_trace function #5058

mpatrikios:
Added ability to add xaxis_title and yaxis_title parameters in add_trace() per #5058


emilykl:
Hi @mpatrikios , thank you for the contribution!

IMO it doesn't make sense to pass axes titles as arguments to `add_trace()` since axes titles are properties of subplots, not traces, and a subplot may have multiple traces. What should happen if multiple traces are added to the same subplot but different values are passed for the axis titles?

robertclaus:
Given the age and last comment on the PR I am closing this for now. Please re-open if you disagree!

PR #5154, opened 2025-04-25
Fix script tag for notebook renderers

### LINKED ISSUE #4953: Plots not shown in HTML export for "vscode+notebook" renderer
### Environment data

Version: 1.96.2 (user setup)
OS: Windows 11 Pro
Jupyter extension version:  2025.1.2024121801
Python Expension Version: 2024.22.1

### Steps to reproduce

```
import plotly.express as px
fig = px.scatter(x=[0, 1, 2, 3, 4], y=[0, 1, 4, 9, 16])
fig.show(renderer='vscode+notebook')

!jupyter nbconvert --to html filename.ipynb
```

### Expected behaviour

Plots are visible when exporting to HTML

### Actual Behaviour

Plots are shown in the notebook but they are not shown in the HTLM file

Side note: I am aware of https://github.com/microsoft/vscode-jupyter/issues/6999 indeed i am using the solution found in that issue

marthacryan:
This updates the `notebook` renderer to use a plain script tag when embedding all of plotly.js rather than using `<script type="module">`. This fixes #4953. For more context on JS modules, see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules. 

Some context:
* `<script type="module">` was added in #4763. 
* Before that PR, it was `<script type="text/javascript">`: https://github.com/plotly/plotly.py/blame/26c4443b1a2efc83a5af3efd179c68f5e13cdfad/packages/python/plotly/plotly/io/_base_renderers.py#L292. It seems like using `type="text/javascript"` is no longer necessary since HTML 5: https://stackoverflow.com/questions/25737619/what-is-the-difference-between-script-type-text-javascript-and-script.

### DIFF
--- a/plotly/io/_base_renderers.py
+++ b/plotly/io/_base_renderers.py
@@ -295,7 +295,7 @@ class HtmlRenderer(MimetypeRenderer):
         {win_config}
         {mathjax_config}
         </script>
-        <script type="module">{script}</script>
+        <script>{script}</script>
         """.format(
                     script=get_plotlyjs(),
                     win_config=_window_plotly_config,


emilykl:
Not my area but I have no objection if this fixes the issue and doesn't break anything else

marthacryan:
Also this has been tested by someone who was encountering this issue and it works! https://github.com/plotly/plotly.py/issues/4953#issuecomment-2831486117

alexcjohnson (approved):
💃 Makes sense to me.

PR #5164, opened 2025-05-02
Kaleido doc string updates

LiamConnors:

### DIFF
--- a/plotly/io/_kaleido.py
+++ b/plotly/io/_kaleido.py
@@ -14,7 +14,7 @@ ENGINE_SUPPORT_TIMELINE = "September 2025"
 
 PLOTLY_GET_CHROME_ERROR_MSG = """
 
-Kaleido requires Google Chrome to be installed. 
+Kaleido requires Google Chrome to be installed.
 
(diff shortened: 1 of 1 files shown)


plotly/io/_kaleido.py
```diff
@@ -549,47 +549,67 @@ def write_images(
...
           - 'jpg' or 'jpeg'
           - 'webp'
           - 'svg'
           - 'pdf'
 
-        If not specified, this will default to `plotly.io.defaults.default_format`.
+        Use a list to specify formats for each figure or dict in the list
+        provided to the `fig` argument.
+        Specify format as a `str` to apply the same format to all exported images.
+        If not specified, will default to `plotly.io.defaults.default_format`.
```
emilykl, line 572:
```suggestion
        If not specified, and the corresponding `file` argument has a file extension, then `format` will default to the
        file extension. Otherwise, will default to `plotly.io.defaults.default_format`.
```

emilykl (approved):
Looks great, thank you @LiamConnors ! 🚀

Issue #1208, opened 2018-09-28
Long install time from windows.


T4rk1n:
When `pip install plotly` from a windows machine, it takes anywhere from a minute to 5 minutes. From linux it took  no more than 10 seconds.

DarkGhost7:
taking me forever to install on windows 10 with pip install dash, seems like plotly is taking up most of the time?

Shaban0818:
Same in my case. Windows 10 and plotly taking forever to install.

AP2008:
I've been trying to install for about 30 mins. It still hasn't finished. Should I cancel ?
I am on Windows 10 Python 3.8.
![image](https://user-images.githubusercontent.com/44343120/95179115-d5bc8200-07dd-11eb-94ba-2bc0c0591321.png)

nicolaskruchten:
That does seem like a long time... did it eventually succeed?

nitrocalcite:
I also experienced this on Windows 10, Python 3.6.10, and after ~5 minutes I canceled it.  However, I didn't get any error messages & plotly at least appears to be working fine; perhaps it's getting hung somewhere?

parmanchaddha:
Same thing here. Using poetry on my mac, running `poetry add plotly` on an _empty_ environment took about 30 seconds. Executing the same command on my windows 10 pro machine took about 5 minutes...

AP2008:
It was probably an issue with my python installation. I reinstalled and now it takes about 5 mins.

Prajval-1608:
Same here, windows 10 and python 3.7.9. Waiting for plotly to install since more than 5 minutes.

beingandbrian:
I am also having this issue. Is there a solve for this?

florian6973:
Same issue on Windows 11

adridem:
Same issue, stuck on Windows 11

eroller:
plotly install/update is pretty slow on WSL too. Slower than all the other packages I use. Is there a good justification for it?

ahmedwaqas92:
Installation was getting stuck on Windows 11 for me as well, I aborted the installation and tried the following:

``` pip install plotly -vvv ```

installed pretty much within 30 seconds.

nick-st:
`pip install plotly` took ages, but 
`pip install plotly==5.14.1`

okoye-dev:
pip install dash is taking almost 30minutes here and it's still now done.
Is this just because it's a windows machine? Windows 11 btw.

[...]

PR #5173, opened 2025-05-09
Refactor validators

bmaranville:
## Code PR

- [x] I have read through the [contributing notes](https://github.com/plotly/plotly.py/blob/main/CONTRIBUTING.md) and understand the structure of the package. In particular, if my PR modifies code of `plotly.graph_objects`, my modifications concern the `codegen` files and not generated files.
  modified existing tests.

# Refactor of validators
This refactor is based on `release-6.0.1rc0`

Helps address #1208 

## Motivation
The large number of files in the `plotly/validators` folder slows down package installation, esp. on systems where files are inspected for viruses during install (e.g. Windows)

The majority of the code is duplicated across these files as well, as there are e.g. >650 instances of nearly-identical classes inheriting `_plotly_utils.basevalidators.BooleanValidator`

## Mitigation
Things to note, in the current validator arrangement:

1. Nearly all the validator classes in `plotly.validators` inherit directly from classes in `_plotly_utils.basevalidators`
1. None of them override any methods at all, they just provide custom kwargs to `__init__`
1. Almost all access to the validators in the codebase is through the `plotly.validator_cache.ValidatorCache.get_validator(parent_name, prop_name)`
1. All these validator python source modules are _generated_ by the plotly build system, by running `python commands.py codegen`

In this PR, the following changes are made:

1. During `codegen`, instead of generating and writing thousands of python files to `plotly/validators/...`, we create a single JSON file store at `plotly/validators/_validators.json` and for each `(parent_name, prop_name)` combination we add an entry to the store with 
   - "superclass": the name of the validator class to instantiate
   - "params": the custom kwargs to use when instantiating the validator class
1. The `ValidatorCache.get_validator` method is modified so that rather than calculating an import path and then importing a validator class from the file tree, it looks up the custom kwargs and validator class name from the JSON store, then creates (and caches) a new instance of the validator class from
`basevalidators`, with the custom init kwargs, finally returning the instance
1. Usages of direct import of validator classes from `plotly.validators` in the codebase are refactored to use the `ValidatorCache` instead (there were only a handful of these cases)

The net result is that the number of files to be installed is reduced from 14981 to 1652.


gvwilson:
this is looking good - thank you very much @bmaranville. We won't try to get it into the 6.1 release, but it will definitely be part of 6.2.

gvwilson:
Note: plotly.py 6.1.0rc0 is 7549710 bytes as a .tar.gz; these changes get that down to 6601687 bytes, which is a 12.5% reduction on top of the reductions in #4978. cc @alexcjohnson

bmaranville:
> this is looking good - thank you very much @bmaranville. We won't try to get it into the 6.1 release, but it will definitely be part of 6.2.

That's great!  Thanks very much for walking me through the code paths related to this issue.  I look forward to seeing it in the 6.2 release.

gvwilson:
closed in favor of #5214 (which rebases these changes on top of `main`).

PR #5175, opened 2025-05-09
replace rgb codes with CSS colors including plotly recommended colors

rl-utility-man:
this PR is entirely trying to apply the following expectation.  I often chose CSS colors closer to those in the original graphics even if they were not on the plotly list
....
- [X] Hex codes for colors are not used in any new/modified example in favour of [these nice ones](https://github.com/plotly/plotly.py/issues/2192)
...

### DIFF
--- a/doc/python/horizontal-bar-charts.md
+++ b/doc/python/horizontal-bar-charts.md
@@ -91,8 +91,8 @@ fig.add_trace(go.Bar(
     name='SF Zoo',
     orientation='h',
     marker=dict(
-        color='rgba(246, 78, 139, 0.6)',
-        line=dict(color='rgba(246, 78, 139, 1.0)', width=3)
+        color='hotpink',
+        line=dict(color='deeppink', width=3)
     )
 ))
 fig.add_trace(go.Bar(
@@ -101,8 +101,8 @@ fig.add_trace(go.Bar(
     name='LA Zoo',
     orientation='h',
     marker=dict(
-        color='rgba(58, 71, 80, 0.6)',
-        line=dict(color='rgba(58, 71, 80, 1.0)', width=3)
+        color='dimgray',
+        line=dict(color='black', width=3)
     )
 ))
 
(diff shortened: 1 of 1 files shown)


gvwilson (approved):

gvwilson:
thanks @rl-utility-man - I'll merge this in as part of 6.2.

doc/python/horizontal-bar-charts.md
```diff
@@ -200,15 +198,15 @@ for yd, xd in zip(y_data, x_data):
                                     x=space + (xd[i]/2), y=yd,
                                     text=str(xd[i]) + '%',
                                     font=dict(family='Arial', size=14,
-                                              color='rgb(248, 248, 255)'),
+                                              color=f"{'white'*(i<2)}{'black'*(i>=2)}"),
```
LiamConnors: Why this f-string for the color? Could it not just be a named color that is close to `rgb(248, 248, 255)`?

doc/python/horizontal-bar-charts.md
```diff
@@ -118,10 +118,8 @@ import plotly.graph_objects as go
 top_labels = ['Strongly<br>agree', 'Agree', 'Neutral', 'Disagree',
               'Strongly<br>disagree']
 
-colors = ['rgba(38, 24, 74, 0.8)', 'rgba(71, 58, 131, 0.8)',
-          'rgba(122, 120, 168, 0.8)', 'rgba(164, 163, 204, 0.85)',
-          'rgba(190, 192, 213, 1)']
-
+colors = ['DarkBlue', 'MediumBlue', 'cyan', 'mediumpurple', 'thistle']
+#, 'silver'
```
LiamConnors: Can this comment go?

doc/python/horizontal-bar-charts.md
rl-utility-man: This is color of the text that goes on the bars, which is currently CSS GhostWhite and is already has less than ideal contrast with lightest bar.  When I went to CSS colors, I ended up with bigger differences between the lightest and darkest bars and struggled to find one color that had enough contrast for all of them.  I make the text on the first two bars white and the text on the last three bars black to provide adequate contrast.

Original:
image

Proposed:
image

Here is the proposed bar graph with all white text -- the middle bar in particular is nearly illegible.  Yellow and gray perform similarly.
image

Here's a potential alternative fix --swapping Cyan for DarkSlateBlue and then using all white text
image

I welcome thoughts.  I am happy to comment this to explain, perhaps:  "make the text on the first two bars white and the text on the last three bars black to provide adequate contrast."  I use this style of conditional F-string a fair bit in my production code and this is a fairly rich, polished example where it might be helpful to show that technique.

LiamConnors: The alternative fix may be better, as I find the hover text when hovering on the cyan bar quite hard to read
image

LiamConnors:
thanks @rl-utility-man

PR #5180, opened 2025-05-12
Allow `shared_yaxes` to work with secondary axes

gmjw:
- [x] I have read through the [contributing notes](https://github.com/plotly/plotly.py/blob/main/CONTRIBUTING.md) and understand the structure of the package. In particular, if my PR modifies code of plotly.graph_objects, my modifications concern the codegen files and not generated files.
- [x]  I have added tests (if submitting a new feature or correcting a bug) or modified existing tests.
- [x]  For a new feature, I have added documentation examples in an existing or new tutorial notebook (please see the doc checklist as well).
- [x]  I have added a CHANGELOG entry if fixing/changing/adding anything substantial.
- [x]  For a new feature or a change in behaviour, I have updated the relevant docstrings in the code to describe the feature or behaviour (please see the doc checklist as well).

The `shared_yaxes` parameter to make_subplots did not previously work on _secondary_ y-axes. It had no effect.

This PR fixes this, without affecting how the parameter is applied to normal plots (without secondary y-axes). I have added a simple test to show how the new code "matches" secondary y-axes. The sharing for secondary y-axes is applied right-to-left, in the same way that the sharing for primary y-axes is applied left-to-right (I believe this results in the clearest separation of tick marks when viewing the final plot).

I believe this is a bugfix/improvement. I'm not sure if it's a new feature worthy of any documentation.

I will add some images showing a before/after comparison in comments.


gmjw:
This is the output of the test adjusted in the PR, with `shared_yaxes=True`, before and after the changes. You can see that after the changes the secondary y-axes are shared across rows, whereas previously only primary y-axes were shared.

fig_test_before
fig_test_after

gmjw:
This is my first PR to this repo, I've done my best but please let me know if there's anything I've forgotten or that could be done better. 

In particular I'm not sure if/how I should add to the CHANGELOG file.

tests/test_core/test_subplots/test_make_subplots.py
```diff
@@ -1828,115 +1828,127 @@ def test_secondary_y_traces(self):
...
     def test_secondary_y_subplots(self):
-        fig = subplots.make_subplots(
-            rows=2,
-            cols=2,
-            specs=[
-                [{"secondary_y": True}, {"secondary_y": True}],
-                [{"secondary_y": True}, {"secondary_y": True}],
-            ],
-        )
+        for shared_y_axes in [False, True]:
```
gmjw, line 1831: Ideally I would have passed in these options using `mark.parametrize`, but I don't think that is possible with unittest TestCase methods.

gmjw:
Thanks @gvwilson - just checking in to see if this could be merged soon - thanks!

plotly/_subplots.py
```diff
@@ -746,8 +746,12 @@ def _check_hv_spacing(dimsize, spacing, name, dimvarname, dimname):
...
             grid_ref[r][c] = subplot_refs
 
-    _configure_shared_axes(layout, grid_ref, specs, "x", shared_xaxes, row_dir)
-    _configure_shared_axes(layout, grid_ref, specs, "y", shared_yaxes, row_dir)
+    _configure_shared_axes(layout, grid_ref, specs, "x", shared_xaxes, row_dir, False)
+    _configure_shared_axes(layout, grid_ref, specs, "y", shared_yaxes, row_dir, False)
+    if secondary_y:
+        _configure_shared_axes(
+            layout, grid_ref, specs, "y", shared_yaxes, row_dir, True
+        )
```
emilykl: @gmjw I think there's a logic issue here -- `secondary_y` is set in the loop above, which loops through all the `specs` for each subplot. So after the above loop terminates, `secondary_y` corresponds only to the value provided for `secondary_y` in the spec of the final subplot.

You can see how this causes issues if you set `'secondary_y': False` for the final subplot only; it breaks shared y-axes for every subplot in the grid.

The fix might be as simple as something like this:

```suggestion
    any_secondary_y = any([spec.get("secondary_y", False) for spec_row in specs for spec in spec_row])
    if any_secondary_y:
        _configure_shared_axes(
            layout, grid_ref, specs, "y", shared_yaxes, row_dir, True
        )
```

although I haven't fully thought through all the edge cases.

gmjw: Hi @emilykl - thanks for spotting this! I should have read through the function more carefully.

I've adjusted almost as you suggest - based on the loop above, it looks like I can rely on "secondary_y" being in `spec` provided `spec` is not None.

gmjw:
Hi @gvwilson - just to say that I hope this is ready to merge now, if you have time to review. Thanks!

gvwilson:
cc @emilykl one more review and merge?

gmjw:
Hi @emilykl - no rush but just wanted to bump this PR for another look if you have time, many thanks!

emilykl (approved):
Yes, this can be merged — sorry for the delay @gmjw and thank you for your patience!

PR #5201, opened 2025-05-25
switched from hex codes to CSS colors

rl-utility-man:
this PR is entirely trying to apply the following expectation. the Crimson is very close to the original; the DarkBlue is on plotly's list and contrasts well with the background and the Crimson and works for people who are red/green color blind.
....

 Hex codes for colors are not used in any new/modified example in favour of these nices ones: 
 https://github.com/plotly/plotly.py/issues/2192
...


LiamConnors:
Thanks @rl-utility-man!

PR #5207, opened 2025-05-27
Support setting plotly.js path for Kaleido v1 image export

### LINKED ISSUE #5206: Manually setting Plotly.js path is not possible with Kaleido v1
Previously, for image export, it was possible to pass a local path and/or custom URL for Plotly.js by setting `pio.kaleido.scope.plotlyjs`. This is not [documented](https://plotly.com/python/static-image-export/#image-export-settings-kaleido) but it's nonetheless [used by our own image test pipeline in Plotly.js](https://github.com/plotly/plotly.js/blob/b5603685a8e8ea4ada6d97d4af3fc60ec470b3cb/test/image/make_baseline.py#L55). 

The following works in Plotly.py v6.0.1, as well as v6.1.2 with Kaleido v0:
```python
import plotly.io as pio

pio.kaleido.scope.plotlyjs = "path/to/local/plotly.js"
```

But with Plotly.py v6.1.2 and Kaleido v1 it gives the following error:
```
AttributeError: 
Attribute plotly.io.defaults.plotlyjs is not valid.
```

IMO there's no reason to stop supporting this option and it should be relatively straightforward to pass through into Kaleido v1.

emilykl:
Closes #5206 

- Adds a `pio.defaults.plotlyjs` entry for setting the Plotly.js path
- Makes sure the value of `pio.defaults.plotlyjs` is passed through to Kaleido when using Kaleido v1

### DIFF
--- a/plotly/io/_defaults.py
+++ b/plotly/io/_defaults.py
@@ -13,6 +13,7 @@ class _Defaults(object):
         self.default_scale = 1
         self.mathjax = None
         self.topojson = None
+        self.plotlyjs = None
 
 
 defaults = _Defaults()
--- a/plotly/io/_kaleido.py
+++ b/plotly/io/_kaleido.py
@@ -369,6 +369,12 @@ To downgrade to Kaleido v0, run:
         from kaleido.errors import ChromeNotFoundError
 
         try:
+            kopts = {}
+            if defaults.plotlyjs:
+                kopts["plotlyjs"] = defaults.plotlyjs
+            if defaults.mathjax:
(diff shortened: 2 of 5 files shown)


gvwilson (approved):

emilykl:
FYI @gvwilson @ayjayt this is ready to merge except that I need to update the Kaleido RC version referenced in `pyproject.toml` since `rc13` is no longer compatible — once a newer RC is available on PyPI I will update this branch and then merge.

gvwilson:
thank you

PR #5214, opened 2025-06-04
refactor validators to remove duplicate files

gvwilson:
This PR replaces #5173 by rebasing @bmaranville's changes on top of `main`. Please see #5173 for detailed notes about the motivation and changes. The important changes are in:

- codegen/__init__.py
- codegen/datatypes.py
- codegen/validators.py
- doc/python/marker-style.md
- plotly/_subplots.py
- plotly/basedatatypes.py
- plotly/figure_factory/_annotated_heatmap.py
- plotly/io/_templates.py
- plotly/validator_cache.py

- [X] I have read through the [contributing notes](https://github.com/plotly/plotly.py/blob/main/CONTRIBUTING.md) and understand the structure of the package. In particular, if my PR modifies code of `plotly.graph_objects`, my modifications concern the `codegen` files and not generated files.
- [X] I have added tests (if submitting a new feature or correcting a bug) or
  modified existing tests.
- [X] For a new feature, I have added documentation examples in an existing or
- [X] I have added a CHANGELOG entry if fixing/changing/adding anything substantial.
- [X] For a new feature or a change in behaviour, I have updated the relevant docstrings in the code to describe the feature or behaviour (please see the doc checklist as well).


gvwilson:
@alexcjohnson if you have time to look this over we'd be grateful for your feedback

codegen/__init__.py
```diff
@@ -21,14 +21,15 @@
...
+    get_data_validator_params,
+    get_validator_params,
+    write_validator_json,
     get_data_validator_instance,
 )
 
 # Target Python version for code formatting with Black.
 # Must be one of the values listed in pyproject.toml.
-BLACK_TARGET_VERSIONS = "py38 py39 py310 py311 py312"
+BLACK_TARGET_VERSIONS = "py39 py310 py311 py312"
```
emilykl: revert this line -- needs to remain the same as what's in `pyproject.toml`
```suggestion
BLACK_TARGET_VERSIONS = "py38 py39 py310 py311 py312"
```

codegen/validators.py
```diff
@@ -1,118 +1,91 @@
...
-    # filepath = opath.join(outdir, "validators", *node.parent_path_parts, "__init__.py")
-    filepath = opath.join(
-        outdir, "validators", *node.parent_path_parts, "_" + node.name_property + ".py"
-    )
-
-    write_source_py(validator_source, filepath, leading_newlines=2)
+    filepath = opath.join(outdir, "validators", "_validators.json")
+    with open(filepath, "w") as f:
+        f.write(json.dumps(params, indent=4))
+        # f.write(str(params))
```
emilykl: 🔪

plotly/basedatatypes.py
```diff
@@ -468,7 +468,12 @@ class is a subclass of both BaseFigure and widgets.DOMWidget.
             if a property in the specification of data, layout, or frames
             is invalid AND skip_invalid is False
         """
-        from .validators import DataValidator, LayoutValidator, FramesValidator
+        from .validator_cache import ValidatorCache
+
+        DataValidator = ValidatorCache.get_validator("", "data")
+        FramesValidator = ValidatorCache.get_validator("", "frames")
+        LayoutValidator = ValidatorCache.get_validator("", "layout")
+        # from .validators import DataValidator, LayoutValidator, FramesValidator
```
emilykl: 🔪

codegen/validators.py
bmaranville: Does the knife symbol mean you want to remove this code?

emilykl: Ha, yes, sorry @bmaranville -- by 🔪 I mean delete this line 😄  (the commented-out one)

emilykl:
@bmaranville The file `plotly/validators/_validators.json` needs to be flagged in the `pyproject.toml` so that it's included in the built package.

Looks like the correct way is to add `validators/_validators.json` to the list under `[tool.setuptools.package-data]`.

emilykl:
Thanks for the PR @bmaranville ! Other than my comments above, looks good to me. The tests are passing and validation seems to be working as usual in the examples I've tried. (@alexcjohnson I'm curious whether you can see any possible pitfalls or edge cases here -- happy to do more testing of specific cases if you have concerns.)

alexcjohnson:
@bmaranville nice work, this is a clever solution! Two small thoughts about it:
- We should be able to reduce the whitespace significantly, to make the file smaller. Could even collapse it to a single line, I'm not sure there's much value keeping it human-readable, but maybe @emilykl has opinions about this. If we really want to optimize file size we could also change its structure a bit - like if every entry is a dict `{params, superclass}` this could be converted to a length-2 list. But the whitespace is the biggest piece of this.
- `json.load` is pretty fast, ~41ms on my computer. But it's even faster if we just make this a Python file. In my quick test I just added:
```py
true=True
false=False
null=None
v = 
```
to the beginning of the file to convert the JSON to Python, and then `from validators._validators import v` took only ~25ms. (If we do this for real we should be able to tweak the `json.dump` to output Python in the first place so we don't need to alias `true`, `false`, and `null`)

@emilykl the only thing I'm not 100% confident of is whether this has any type checking implications. I don't think so, I think the only type checking that matters is on `graph_objs`, not `validators`, but we do currently have `if TYPE_CHECKING` blocks in all the `__init__.py` files in validators, ensuring that type checkers will load them all. If you haven't done so already it's probably worth playing around on this branch a bit to convince ourselves that we don't lose anything important with this.

codegen/__init__.py
gvwilson: added

plotly/basedatatypes.py
gvwilson: cut

gvwilson:
@alexcjohnson :+1: on reducing whitespace, but I'd like to merge this one first, then combine it with #5218, and then tidy up the whitespace in the result - #5218 switches code formatting and checking from `black` to `ruff`, so if we're going to figure out how to configure a tool to stay silent about things like long lines, I'd like to do that once. I've created #5225 and will assign it to myself.

codegen/validators.py
gvwilson: removed

plotly/basedatatypes.py
```diff
@@ -468,7 +468,11 @@ class is a subclass of both BaseFigure and widgets.DOMWidget.
             if a property in the specification of data, layout, or frames
             is invalid AND skip_invalid is False
         """
-        from .validators import DataValidator, LayoutValidator, FramesValidator
+        from .validator_cache import ValidatorCache
+
+        DataValidator = ValidatorCache.get_validator("", "data")
+        FramesValidator = ValidatorCache.get_validator("", "frames")
+        LayoutValidator = ValidatorCache.get_validator("", "layout")
```
alexcjohnson: Minor style point: `DataValidator` et al are instances, not classes as they were previously, so should be camel_case.

gvwilson: good catch - fixed in the latest push

pyproject.toml
```diff
@@ -63,7 +63,7 @@ plotly = [
     "package_data/*",
     "package_data/templates/*",
     "package_data/datasets/*",
-    "plotly/validators/_validators.json"
```
gvwilson: oops - thank you

emilykl: no prob, it's not obvious, I originally assumed it should have the leading `plotly/` but apparently not

emilykl:
> * We should be able to reduce the whitespace significantly, to make the file smaller. Could even collapse it to a single line, I'm not sure there's much value keeping it human-readable, but maybe @emilykl has opinions about this.

@alexcjohnson I'm in favor of this. I don't think there's any particular reason to keep it human-readable, although it's also not a HUGE difference with compression. Looks like about 194 KB compressed with whitespace vs. 163 KB compressed without whitespace, from a quick test I did locally.

> * `json.load` is pretty fast, ~41ms on my computer. But it's even faster if we just make this a Python file. ... convert the JSON to Python, and then `from validators._validators import v` took only ~25ms. (If we do this for real we should be able to tweak the `json.dump` to output Python in the first place so we don't need to alias `true`, `false`, and `null`)

@alexcjohnson @gvwilson I'm in favor of this -- that's a big performance improvement. Something for a future PR? (as a bonus, then we wouldn't have to worry about including this file in the `pyproject.toml`, since Python files are included automatically.)

> @emilykl the only thing I'm not 100% confident of is whether this has any type checking implications. I don't think so, I think the only type checking that matters is on `graph_objs`, not `validators`, but we do currently have `if TYPE_CHECKING` blocks in all the `__init__.py` files in validators, ensuring that type checkers will load them all. If you haven't done so already it's probably worth playing around on this branch a bit to convince ourselves that we don't lose anything important with this.

I'm not a _heavy_ user of type checking so it's possible I'll miss something here. Syntax highlighting is working as expected with these changes. I turned on stricter Pylance type checking and haven't noticed any issues yet but I'll keep an eye out. Hopefully we can put these changes out in an RC to catch any issues before full release.

PR #5223, opened 2025-06-07
added Python decorator to the markdown

rl-utility-man:
The graph on this new example is not displaying.  The new code had line numbers when the rest of the examples did not.  I noticed that there was no python keyword/decorator right at the start of the code block, so added it in this PR.  Is that sufficient to  solve the problem?


LiamConnors (approved):

LiamConnors:
Thanks @rl-utility-man. This should be good to solve the problem. I checked the build output in CI and it renders correctly there

PR #5249, opened 2025-06-26
Add `scipy-stubs` as dev dependency

jorenham:
This adds [`scipy-stubs`](https://github.com/scipy/scipy-stubs) as an optional development dependency. It can help quite a bit with IDE supports, such as autocompletion and introspection. It requires no configuration or mypy plugins, and there's zero runtime impact.  So even if you don't care about typing annotations, you can view this as a free DX improvement :)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,7 @@ dev_optional = [
     "pytz",
     "scikit-image",
     "scipy",
+    "scipy-stubs;python_version>='3.10'",
     "shapely",
     "statsmodels",
     "vaex;python_version<='3.9'",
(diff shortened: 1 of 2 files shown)


emilykl:
@jorenham Thanks for the contribution!

We try to keep our dev dependencies pretty minimal and only include packages which are strictly required for the dev process. If you want to make a case for how specifically to use `scipy-stubs` in the dev process, we are open to it; that info should be added to `CONTRIBUTING.md`.

jorenham:
> If you want to make a case for how specifically to use `scipy-stubs` in the dev process, we are open to it; that info should be added to `CONTRIBUTING.md`.

It's a stub-package, so there's no need to *use* it; once installed, the benefits I listed above will automatically come into effect. So no config or code changes required.
So it's not something that contributors even need to know about for them to be able to contribute.

Issue #5136, opened 2025-04-09
add_hline not working for make_subplots if not populated


NoniosTheMad:
Whenever you add a hline/vline to a plot it should display it at the given position. However, when the figure is not yet populated with data, figures made with make_subplots never show the line.

```python
import plotly.graph_objects as go
from plotly.subplots import make_subplots

for name, fig in {
    "go.Figure": go.Figure(),
    "make_subplots": make_subplots(rows=1, cols=1)
}.items():
    fig.add_hline(y=0.5)
    fig.add_trace(go.Scatter(x=[1, 2], y=[0, 1]))
    fig.update_layout(title=name)
    fig.show()
```

![Image](https://github.com/user-attachments/assets/175a5227-f90a-47ec-8eed-e0002d2d0aa6)
![Image](https://github.com/user-attachments/assets/9fc719f4-1053-4f2c-ad0a-45f65a988181)

`pip list`

Package                   Version
------------------------- -----------
asttokens                 3.0.0
attrs                     25.3.0
colorama                  0.4.6
comm                      0.2.2
debugpy                   1.8.13
decorator                 5.2.1
executing                 2.2.0
fastjsonschema            2.21.1
ipykernel                 6.29.5
ipython                   9.0.2
ipython_pygments_lexers   1.1.1
jedi                      0.19.2
jsonschema                4.23.0
jsonschema-specifications 2024.10.1
jupyter_client            8.6.3
jupyter_core              5.7.2
matplotlib-inline         0.1.7
narwhals                  1.31.0
nbformat                  5.10.4
nest-asyncio              1.6.0
packaging                 24.2
parso                     0.8.4
pip                       24.3.1
platformdirs              4.3.7
**plotly**                    6.0.1
prompt_toolkit            3.0.50
psutil                    7.0.0
pure_eval                 0.2.3
Pygments                  2.19.1
python-dateutil           2.9.0.post0
pywin32                   310
pyzmq                     26.3.0
referencing               0.36.2
rpds-py                   0.23.1
six                       1.17.0
stack-data                0.6.3
tornado                   6.4.2
traitlets                 5.14.3
typing_extensions         4.12.2
wcwidth                   0.2.13

gvwilson:
thanks for the bug report @NoniosTheMad

leavor:
plotly.py verson 6.4.0,  has same issue

PR #5254, opened 2025-06-27
Fix add hline not working for make_subplot if not populated

Erikp15:
This is a proposal fix for issue #5136
I first reproduced the issue and created two unit tests, `test_add_hline_empty_subplots.py` and `test_hline_subplots_bug.py`, that check if the bug is resolved. Then I implemented the fix (outlined below), which passed both of the new unit tests, and in the process, no old tests were broken.

## Root cause
`add_shape()` is called with `exclude_empty_subplots=True`. For a fresh `make_subplots` figure, every subplot is flagged empty, so the h/v line shape is filtered out and never appears, even after traces are later added.

## Fix overview
When `add_hline`/`add_vline` are invoked and the figure contains no traces (`len(self.data)==0`), the code now disables the exclude empty subplots logic before delegating to `add_shape`, allowing the shape to be created and later displayed once data is added.

## Code PR

- [X] I have read through the [contributing notes](https://github.com/plotly/plotly.py/blob/main/CONTRIBUTING.md) and understand the structure of the package. In particular, if my PR modifies code of `plotly.graph_objects`, my modifications concern the code generator and *not* the generated files.
- [X] I have added tests or modified existing tests.


Erikp15:
Hi @gvwilson , just checking in to see if there's anything else needed from my end on this PR. Happy to make any changes if needed.

emilykl:
Hi @Erikp15, it seems to me that the `exclude_empty_subplots` argument is working as intended (i.e., it indeed does not draw the hline if the subplot is empty), so I'm not sure we want to change that behavior.

What I would recommend, if you don't want to exclude empty subplots, is to pass `exclude_empty_subplots=False` to `add_hline()` — would that work for you?

robertclaus:
Given that this PR is nearing a year old and ended with a potential workaround, I'm going to close it for now. Please re-open if I misunderstood it's status!

PR #5258, opened 2025-06-30
Fix issue with default renderer when `ipython` is installed

### LINKED ISSUE #5257: Default renderer is incorrect when `ipython` is installed in environment
- If the `ipython` package is installed in the environment, the default renderer (`pio.renderers.default`) is set to `"plotly_mimetype+notebook"` _even when running in a normal Python script_.
- This causes undesirable behavior for `fig.show()`: Calling `fig.show()` causes HTML/Javascript output to be printed to the terminal, rather than launching a browser window to show the plot. 
- If `ipython` is NOT installed, then `pio.renderers.default` is set to `"browser"`, as it should be, and `fig.show()` behaves normally.

emilykl:
Closes #5257 

Reverts change to `plotly/io/_renderers.py` added in #5096 to keep the original logic flow of the if statement.

Adds a test to ensure that the default renderer is `"browser"` while running the tests; since the `test_optional` workflow installs `jupyter` which installs `ipython`, this should prevent a regression.

## Steps to test

1. Verify incorrect behavior on `main`: 
    - `pip install jupyter`
    - Run `python -c "import plotly.express as px; fig = px.scatter(x=[1], y=[1]); fig.show()"`. This SHOULD show a figure in the browser, but instead it prints the HTML/JS to the terminal output.
2. Verify correct behavior on this branch:
    - Run `python -c "import plotly.express as px; fig = px.scatter(x=[1], y=[1]); fig.show()"` and make sure the figure is shown in the browser
3.  Make sure figure still renders correctly in Jupyter notebook:
    - This is a little tricky because Jupyter rendering seems to not play nicely with the local install (at least on my machine) and I'm not sure why yet. But it's not a problem introduced by this PR; the same is true on `main`.
    -  For that reason I recommend downloading [the build artifacts from the CI](https://output.circle-artifacts.com/output/job/cb116553-3b67-4d85-9153-f3cb8776fe85/artifacts/0/output.tgz) and installing the built wheel.
    - Run `jupyter notebook` to launch the notebook editor, and create a new notebook
    - Make sure the following code correctly renders a figure in the notebook:
  ```python
import plotly.express as px
fig = px.scatter(x=[1], y=[1])
fig.show()
```


marthacryan (approved):
LGTM!

camdecoster:
I followed the test steps above and everything worked correctly (including the Jupyter notebook with a local package).

PR #5262, opened 2025-07-03
Add support for overriding x tick with non arithmetic progression values

robertoffmoura:
Hi,
This change adds support for specifying tickvals and ticktext when the values don't follow an arithmetic progression, i.e., they're not evenly spaced.

## Code PR

- [x] For a new feature, I have added documentation examples (please see the doc checklist as well).
- [x] For a new feature or a change in behavior, I have updated the relevant docstrings in the code.

### DIFF
--- a/plotly/matplotlylib/mpltools.py
+++ b/plotly/matplotlylib/mpltools.py
@@ -436,9 +436,9 @@ def prep_ticks(ax, index, ax_type, props):
             tick0 = tickvalues[0]
             dticks = [
                 round(tickvalues[i] - tickvalues[i - 1], 12)
-                for i in range(1, len(tickvalues) - 1)
+                for i in range(1, len(tickvalues))
             ]
-            if all([dticks[i] == dticks[i - 1] for i in range(1, len(dticks) - 1)]):
+            if all([dticks[i] == dticks[i - 1] for i in range(1, len(dticks))]):
                 dtick = tickvalues[1] - tickvalues[0]
             else:
                 warnings.warn(
@@ -448,6 +448,8 @@ def prep_ticks(ax, index, ax_type, props):
                 raise TypeError
         except (IndexError, TypeError):
             axis_dict["nticks"] = props["axes"][index]["nticks"]
+            if props["axes"][index]["tickvalues"] is not None:
+                axis_dict["tickvals"] = props["axes"][index]["tickvalues"]
         else:
             axis_dict["tick0"] = tick0
             axis_dict["dtick"] = dtick
@@ -496,6 +498,15 @@ def prep_ticks(ax, index, ax_type, props):
 
     if formatter == "LogFormatterMathtext":
         axis_dict["exponentformat"] = "e"
+    elif (
+        formatter == "FuncFormatter" and props["axes"][index]["tickformat"] is not None
+    ):
+        to_remove = ["dticktickmode"]
+        for key in to_remove:
+            if key in axis_dict:
+                axis_dict.pop(key)
+        axis_dict["ticktext"] = props["axes"][index]["tickformat"]
+        axis_dict["tickvals"] = props["axes"][index]["tickvalues"]
     return axis_dict
 
 
--- a/plotly/matplotlylib/mpltools.py
+++ b/plotly/matplotlylib/mpltools.py
@@ -487,14 +487,15 @@ def prep_ticks(ax, index, ax_type, props):
     formatter = axis.get_major_formatter().__class__.__name__
     if ax_type == "x" and "DateFormatter" in formatter:
         axis_dict["type"] = "date"
-        try:
-            axis_dict["tick0"] = mpl_dates_to_datestrings(axis_dict["tick0"], formatter)
-        except KeyError:
-            pass
-        finally:
+        tickvalues = props["axes"][index]["tickvalues"]
+        if tickvalues is not None:
+            # custom ticks: export the exact locations and drop the
+            # arithmetic tick0/dtick spec, which plotly would use instead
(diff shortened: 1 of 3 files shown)


robertoffmoura:
Hi @emilykl,
This adds support for x ticks that don't follow an arithmetic progression when converting matplotlib figures to plotly.

Currently, the conversion ignores custom x tick values specified in the matplotlib figure.

See the snippet below, which prints `None` before this change:
```
import warnings
import plotly.tools as tls

xticks = [0.01, 0.53, 0.75]  # non-arithmetic progression

fig = plt.figure()
plt.plot([0, 1], [0, 1])
plt.xticks(xticks)
fig.savefig("matplotlib.png")

p = tls.mpl_to_plotly(plt.gcf())
print(p.layout.xaxis.tickvals) # (0.01, 0.53, 0.75)
p.write_image("plotly.png")
```

Here's a screenshot of the matplotlib figure with custom x tick values:
matplotlib

Before this change, the plotly conversion ignores the 0.75 x tick value.
plotly_before

After this change, the converted plot correctly shows the 0.75 x tick value.
plotly_after

plotly/matplotlylib/mpltools.py
```diff
@@ -496,6 +498,15 @@ def prep_ticks(ax, index, ax_type, props):
 
     if formatter == "LogFormatterMathtext":
         axis_dict["exponentformat"] = "e"
+    elif (
+        formatter == "FuncFormatter" and props["axes"][index]["tickformat"] is not None
+    ):
+        to_remove = ["dticktickmode"]
+        for key in to_remove:
+            if key in axis_dict:
+                axis_dict.pop(key)
```
camdecoster: I think `"dticktickmode"` is supposed to be `"dtick", "tickmode"`. If that's the case, could you follow the pattern from a few lines up?
```suggestion
        axis_dict.pop("dtick", None)
        axis_dict.pop("tickmode", None)
```

plotly/matplotlylib/mpltools.py
camdecoster: If a user has set custom tick values on a date axis, you'll need to convert those as well. What do you think of this?
```suggestion
axis_dict.pop("dtick", None)
axis_dict.pop("tickmode", None)
if axis_dict.get("tickvals") is not None:
        axis_dict["tickvals"] = mpl_dates_to_datestrings(
            axis_dict["tickvals"], formatter
        )
axis_dict["range"] = mpl_dates_to_datestrings(props["xlim"], formatter)
```

camdecoster, line 447: Given that `tickvals` now gets set in the `except` block, this message isn't accurate anymore. Could you update it?

plotly/matplotlylib/mpltools.py
```diff
@@ -496,6 +498,15 @@ def prep_ticks(ax, index, ax_type, props):
 
     if formatter == "LogFormatterMathtext":
         axis_dict["exponentformat"] = "e"
+    elif (
+        formatter == "FuncFormatter" and props["axes"][index]["tickformat"] is not None
```
camdecoster: Should this also handle when `FixedFormatter` is being used?

plotly/matplotlylib/tests/test_renderer.py
camdecoster, line 1: Could you please add a test that checks the `yticks`? Those should also be fixed by your changes, but it would be good to check it here.

plotly/matplotlylib/tests/test_renderer.py
```diff
@@ -199,3 +199,27 @@ def test_filled_path_collection_date_xaxis():
     filled = [t for t in plotly_fig.data if t.fill == "toself"]
     assert len(filled) >= 1
     assert all(isinstance(x, str) for x in filled[0].x)
+
+
+def test_non_arithmetic_progression_xtickvals():
```
camdecoster, line 249: Any reason why you didn't use `fix, ax = plt.subplots()` like the other tests? It would be nice to be consistent.

camdecoster (changes requested):
Thanks for the PR! Could you please review my comments and add a CHANGELOG entry?

plotly/matplotlylib/mpltools.py
robertoffmoura: Done.

plotly/matplotlylib/mpltools.py
robertoffmoura: Updated to say "exporting explicit tick values instead of dtick".

plotly/matplotlylib/tests/test_renderer.py
robertoffmoura: No reason. Updated to use `fix, ax = plt.subplots()` for consistency.

plotly/matplotlylib/tests/test_renderer.py
robertoffmoura: Done

plotly/matplotlylib/mpltools.py
robertoffmoura: Done.

plotly/matplotlylib/mpltools.py
robertoffmoura: Thanks! I keyed off props["axes"][index]["tickvalues"] instead of axis_dict["tickvals"] because tickvals is only set for unevenly spaced ticks.

camdecoster (commented):
Looks good. Before I approve, could you please address the merge conflict?

robertoffmoura:
Thanks for the review! I rebased and addressed the merge conflict.

camdecoster (approved):
Thanks for the update!

PR #5271, opened 2025-07-08
corrected an incorrect description entry

rl-utility-man:
This fixes a couple unclear aspects of the page.  The distinction between this page and the other choropleth page is that this one shows tile maps.  It shows some examples that are not US counties.  I clarified this.

Also, I noted that mapbox traces are deprecated on the link to the mapbox reference.


LiamConnors (approved):
Thanks for fixing this @rl-utility-man

PR #5274, opened 2025-07-10
Update CONTRIBUTING.md to include jupyter lab local development instr…

marthacryan:


CONTRIBUTING.md
```diff
@@ -120,6 +120,12 @@ you can install all packages with:
 pip install -e '.[dev]'
 ```
 
+If you're testing local changes in Jupyter Lab or Jupyter Notebook, you'll want to run:
```
emilykl: Could you explain when to run this command -- does it need to be run only when setting up the environment, or after every change? Also briefly what does it do / what happens if you don't run it?

marthacryan: Just tried to update it! How does it look?

PR #5282, opened 2025-07-16
Expose `plotly.io.get_chrome()`

### LINKED ISSUE #5272: Bug Report: `plotly.io.install_chrome()` documented but doesn't exist
The official Plotly documentation shows `plotly.io.install_chrome()` as a method for installing Chrome for static image export, but this method doesn't actually exist in the `plotly.io` module, causing `AttributeError` when users try to use it.

### **Documentation vs Reality**

**📖 What the docs show:**
```python
import plotly.io as pio
pio.install_chrome()
```
*Source: https://plotly.com/python/static-image-export/#chrome*

**💥 What actually happens:**
```python
AttributeError: module 'plotly.io' has no attribute 'install_chrome'
```

**✅ What actually works:**
```bash
plotly_get_chrome
```

### **Steps to Reproduce**
1. Follow the official static image export documentation
2. Try to run `plotly.io.install_chrome()`
3. Observe the `AttributeError`

### **Expected Behavior**
Either:
- The `plotly.io.install_chrome()` method should exist and work as documented, OR
- The documentation should be corrected to only show the working `plotly_get_chrome` command

### **Current Workaround**
Users must use the command-line tool instead:
```python
import subprocess
subprocess.run(["plotly_get_chrome", "-y"], check=True)
```

### **Environment**
- **Plotly version:** 5.x+ (affects current versions)
- Kaleido>=1.0
- **Python:** 3.9+
- **OS:** All platforms

### **Impact**
- **User confusion:** Official docs lead users to non-working code
- **Development friction:** Users waste time debugging non-existent methods
- **Documentation credibility:** Undermines trust in other documented APIs

### **Suggested Fix**
**Option 1 (Preferred):** Implement the missing method
```python
# Add to plotly/io/__init__.py
def install_chrome():
    """Install Chrome for static image export using Kaleido."""
    import subprocess
    subprocess.run(["plotly_get_chrome", "-y"], check=True)
```

**Option 2:** Update documentation to remove the non-existent method and only show the working command-line approach.

### **Related Files**
- **Documentation:** https://plotly.com/python/static-image-export/
- **API Reference:** Missing from plotly.io module
- **Working CLI tool:** `plotly_get_chrome` command

---

*This issue affects users trying to automate Chrome installation for static image export in CI/CD pipelines and development environments.*

emilykl:
Closes #5272 

Expose `plotly.io.get_chrome()` as a function which can be called from within a Python script.

`get_chrome()` accepts one optional argument, `path`, which is the location to download Chrome to; if not supplied, the default location defined by `choreographer` will be used.

This PR also slightly refactors `get_chrome()` into two functions:
- `get_chrome()`, which is intended to be called directly in code, and
- `plotly_get_chrome()`, which is a CLI wrapper for `get_chrome()` and intended to be called via the `plotly_get_chrome` command from the command line.

On this branch, running the following code should work:

```python
import plotly.io as pio

pio.get_chrome()
```

### DIFF
--- a/plotly/io/_kaleido.py
+++ b/plotly/io/_kaleido.py
@@ -775,11 +775,13 @@ which can be installed using pip:
         return go.Figure(fig, skip_invalid=True)
 
 
-def get_chrome() -> None:
+def plotly_get_chrome() -> None:
     """
     Install Google Chrome for Kaleido (Required for Plotly image export).
-    This function can be run from the command line using the command `plotly_get_chrome`
-    defined in pyproject.toml
+    This function is a command-line wrapper for `plotly.io.get_chrome()`.
+
+    When running from the command line, use the command `plotly_get_chrome`;
+    when calling from Python code, use `plotly.io.get_chrome()`.
     """
 
     usage = """
@@ -813,7 +815,6 @@ Install it using `pip install 'kaleido>=1.0.0'` or `pip install 'plotly[kaleido]
 
     # Handle "--path" flag
     chrome_install_path = None
-    user_specified_path = False
     if "--path" in cli_args:
         path_index = cli_args.index("--path") + 1
         if path_index < len(cli_args):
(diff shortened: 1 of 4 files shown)


plotly/io/_kaleido.py
```diff
@@ -813,16 +815,60 @@ def get_chrome() -> None:
...
+    Get the path to the Chrome executable for Kaleido.
+    This function is used by the `plotly_get_chrome` command line utility.
+
+    Parameters
+    ----------
+    path: str or Path or None
+        The path to the directory where Chrome should be installed.
+        If None, the default download path will be used.
+    """
+    if not kaleido_available() or kaleido_major() < 1:
```
gvwilson, line 858: why a confirmation? (not arguing, just think it's very unlikely someone would call the function by accident)

gvwilson (approved):

LiamConnors (approved):

plotly/io/_kaleido.py
emilykl: Do you mean the one on line 837? I agree it's unlikely to be called by accident, but I mainly put it there so users could confirm that the install path makes sense.

In any case, the confirmation only appears when using the command-line `plotly_get_chrome` command, not when calling the `plotly.io.get_chrome()` function within a script.

Issue #5266, opened 2025-07-06
suggest updating the figurewidget doc: only for Jupyter, with anywidget


denis-bz:
suggest updating the https://plotly.com/python/figurewidget/ doc, along the lines
> only for Jupyter widgets, with `pip install anywidget` 

is that correct ?

(For my understanding, why is that so -- how is rendering to Jupyter different from other renderers ?

Movies of data packets of different sizes and speeds trundling through plotly might be fun.)

LiamConnors:
Thanks @denis-bz for the suggestion! Just want to make sure I understand correctly - is the issue that the figurewidget page examples require `anywidget` but that's not mentioned on any of the pages in this section?

denis-bz:
@LiamConnors, yes, please update the doc; \
is there a way to update a *single* canvas in a browser, not jupyter ?

Thanks eh

PR #5284, opened 2025-07-17
Update figurewidget.md

LiamConnors:
Add info on installing anywidget:

https://github.com/plotly/plotly.py/issues/5266

### DIFF
--- a/doc/python/figurewidget.md
+++ b/doc/python/figurewidget.md
@@ -22,7 +22,7 @@ jupyter:
     pygments_lexer: ipython3
     version: 3.6.5
   plotly:
-    description: Introduction to the new Plotly FigureWidget
+    description: Introduction to the Plotly FigureWidget
     display_as: chart_events
     language: python
     layout: base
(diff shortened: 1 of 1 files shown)


doc/python/figurewidget.md
```diff
@@ -22,7 +22,7 @@ jupyter:
     pygments_lexer: ipython3
     version: 3.6.5
   plotly:
-    description: Introduction to the new Plotly FigureWidget
```
marthacryan, line 25: I'm not sure if we need to add "new" here. When would we remove that word? And also the `FigureWidget` already existed, we just changed the dependency from `ipywidgets` to `anywidget`

marthacryan:
Other than the one comment this LGTM!

doc/python/figurewidget.md
LiamConnors: I'm actually removing the word new, not adding it.

marthacryan: Oh ok then I very much support that!

marthacryan (approved):

Issue #5286, opened 2025-07-21
Deprecation warnings related to `engine` parameter raised when not specifying `engine` parameter


jdbeel:
When running notebooks with `plotly[kaleido]==6.2.0` there are instances when deprecation warnings are emitted given the same parameterization of a plot function. I had hoped to include a very sensible reproduction case for this issue, but unfortunately it's so much stranger than that.

The issue only appears to happen when instantiating a class from another library I am working with. This library is opening a handle to some binaries on my system, which is my only idea as to why there would be any difference from Jupyter or plotly's perspective. I've attached a notebook with a requirements file, but I'll go over the issue here as well.

In short,

```
import os
os.environ["PLOTLY_RENDERER"] = "svg"

import pmlb
import plotly.express as px

from howso.engine import Trainee

df = pmlb.fetch_data("iris")

t = Trainee()

fig = px.scatter(df, x="sepal-width", y="sepal-length", color="target")
fig.show()
```

produces the following warning:

```
file:///C:/Users/JacobBeel/workspace/plotly.py/plotly/io/_base_renderers.py:123: DeprecationWarning:


Support for the 'engine' argument is deprecated and will be removed after September 2025.
Kaleido will be the only supported engine at that time.
```

Whereas 

```
import os
os.environ["PLOTLY_RENDERER"] = "svg"

import pmlb
import plotly.express as px

from howso.engine import Trainee

df = pmlb.fetch_data("iris")

fig = px.scatter(df, x="sepal-width", y="sepal-length", color="target")
fig.show()
```

does not. Note that all I've done is not instantiate the `Trainee` class.

[plotly_warning.zip](https://github.com/user-attachments/files/21351684/plotly_warning.zip)

I have managed a fix which I'm going to PR, but it doesn't really address the strangeness of the reproduction.

gvwilson:
@emilykl Kaleido-related (please move to that repo if more appropriate)

emilykl:
Closed by https://github.com/plotly/plotly.py/pull/5287.

PR #5287, opened 2025-07-21
fix: Resolve issue causing extraneous `engine` deprecation warnings

jdbeel:
## Code PR

- [ ] I have added tests or modified existing tests.


jdbeel:
Fixes the issue in #5286

I don't believe any docstrings need updating as the ones for the classes/methods I changed do not reference the default for the engine argument. The default was also changed to `None` further up the API. I believe this should basically be a NOP for most users/flows. If this warrants a change in the CHANGELOG, please let me know. I figured for the above reasons it did not.

emilykl (approved):
Thanks for running this down @jdbeel, this is great. 

I agree, this should be a noop since inside the `to_image` function, `engine` is set to `"auto"` if `None` is passed.

Once the `engine` argument is removed from the `to_image` and `write_image` functions, we should remove it from these functions as well. Until then this is a good solution.

🚀

PR #5296, opened 2025-07-23
Upgrade to plotly.js to 3.0.3

marthacryan:


plotly/graph_objs/bar/_insidetextfont.py
```diff
@@ -395,11 +402,18 @@ def _prop_descriptions(self):
             `color`.
         family
             HTML font family - the typeface that will be applied by
-            the web browser. The web browser can only apply a font
```
gvwilson: uh oh - looks like something's gone wrong with the `plot-schema.json` file. The text that's being deleted is much shorter than the text that's being added - the latter is the old text that I replaced with the shorter version to reduce bundle size.

gvwilson (approved):

gvwilson:
thanks @marthacryan

PR #5299, opened 2025-07-29
fix: add missing arguments to commands.py

gvwilson:
cc @LiamConnors


gvwilson:
@LiamConnors if this worked for you, let's merge it in

PR #5315, opened 2025-08-11
fix a typo in `_ternary_contour.py`

Lexachoc:
It should be `ilr` instead of `llr`.


emilykl (approved):
Good catch, thank you @Lexachoc ! 🚀

PR #5333, opened 2025-08-23
added a cross reference to creating a custom color scale in graph objects

rl-utility-man:
There are px examples that do not work in GO.  This link clarifies where to find the equivalent GO example.

### DIFF
--- a/doc/python/colorscales.md
+++ b/doc/python/colorscales.md
@@ -157,7 +157,7 @@ fig.show()
 
 ### Explicitly Constructing a Color scale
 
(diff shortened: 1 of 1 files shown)


gvwilson (approved):

rl-utility-man:
@camdecoster I'm just getting back to Plotly doc contributions and noticed that this very terse one appears to be approved but was never merged.  Let me know if there's anything I can do to support its further consideration.

camdecoster (approved):

PR #5366, opened 2025-10-07
Add references to cloud

LiamConnors:


red-patience:
Are the jupyter version bumps intentional?

LiamConnors:
> Are the jupyter version bumps intentional?

Yeah, it doesn't mean very much. Just when you open a page in Jupyter notebook it updates to the Python version you are using. That info is not actually used anywhere in the docs build.

Issue #3065, opened 2021-02-05
Error on using add_vline with text annotation for data with date-time x axis


MR0205:
Hi, I am trying to use the library for a simple visualisation for the first time, and I stumbled upon supposedly a bug trying to draw a vertical line with text annotation on a graph with a date-time x axis.

```
import plotly.express as px
df = px.data.stocks(indexed=True)
fig = px.line(df)
fig.add_vline(x="2018-09-24", annotation_text="test" )
fig.show()
```

I get the following error message:

`TypeError: unsupported operand type(s) for +: 'int' and 'str'`

Could anyone please help me confirm that it is indeed a bug ? Thank you.

nicolaskruchten:
Yep, definitely a bug! Thanks for the clear and reproducible report :)

I'm not immediately clear why we need to compute the midpoint of anything here https://github.com/plotly/plotly.py/blob/03979d105c65dda3df3a155322eaff18f203b03f/packages/python/plotly/plotly/shapeannotation.py#L58 but this is the problem. Ideally we would need to rework this to use the `(x|y)anchor` attribute instead of trying to do math on the coordinates in the Python layer.

awrobel1:
Hi @MR0205, 

If you're looking for a temporary workaround, you can get the annotation to work by converting your `x` value into milliseconds since epoch as follows: 
```
import plotly.express as px
import datetime
df = px.data.stocks(indexed=True)
fig = px.line(df)
fig.add_vline(x=datetime.datetime.strptime("2018-09-24", "%Y-%m-%d").timestamp() * 1000, annotation_text="test" )
fig.show()
```

MR0205:
@awrobel1 Thanks for your idea, I thought on using a rectangle as a replacement, where you would make both edges start at the limit of a single day. Not sure though that will completely imitate the vertical line, due to possible problem when zooming.

YukunYangNPF:
Following this too!

tomshaffner:
The above suggestion from @awrobel1 was great for me. In my case I simplified a tad with `x=datetime.datetime(2021,5,27).timestamp() * 1000`. No need to convert from a string first in this case.

nicholas-esterer:
[...]

PR #5368, opened 2025-10-08
Add datetime support to _mean()

m-ad:
This PR extends the  `_mean()` function in `shapeannotation.py` to allow for sequences of `datetime` objects as input. This fixes issue #3065, specifically it enables annotations on `hline`/`vline` objects that use `datetime` axes.

## Code PR

- [X] I have read through the [contributing notes](https://github.com/plotly/plotly.py/blob/main/CONTRIBUTING.md) and understand the structure of the package. In particular, if my PR modifies code of `plotly.graph_objects`, my modifications concern the code generator and *not* the generated files.
- [X] I have added tests or modified existing tests.

### DIFF
--- a/plotly/shapeannotation.py
+++ b/plotly/shapeannotation.py
@@ -1,10 +1,32 @@
 # some functions defined here to avoid numpy import
 
+from datetime import datetime
+from numbers import Number
+
 
 def _mean(x):
     if len(x) == 0:
         raise ValueError("x must have positive length")
-    return float(sum(x)) / len(x)
+
+    # Numeric sequence: default behaviour
+    if all(isinstance(v, Number) for v in x):
+        return float(sum(x)) / len(x)
+
+    # Datetime sequence: delegate to _mean_datetime
+    if all(isinstance(v, datetime) for v in x):
+        return _mean_datetime(x)
+
+    # Fallback for non-numeric and non-datetime types: return first element
+    return x[0]
+
+
+def _mean_datetime(x: list[datetime]) -> datetime:
+    """Return midpoint of a sequence of datetime objects (assumes homogenous)."""
+
+    timestamps = [v.timestamp() for v in x]  # convert to POSIX timestamps
+    avg = sum(timestamps) / len(timestamps)
+    tzinfo = x[0].tzinfo  # extract timezone info from first element
+    return datetime.fromtimestamp(avg, tz=tzinfo)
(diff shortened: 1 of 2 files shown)


gvwilson:
@emilykl please review

emilykl:
@m-ad Thank you for the PR!

I'm open to something along these lines as a "patch" fix to enable `add_vline()` and `add_hline()` to work with datetime axes, although as @nicolaskruchten [mentioned](https://github.com/plotly/plotly.py/issues/3065#issuecomment-774181759), the "correct" fix would be to refactor the code such that we're not trying to do math with data values.

Ultimately, now that [shape labels are natively supported in Plotly.js](https://github.com/plotly/plotly.js/pull/6454), I think the longer-term fix here is to refactor `add_hline()` and `add_vline()` to add `annotation_text` as a shape label rather than as a separate annotation; that will allow us to skip the step of computing averages of the data values. I mention it here just in case that's a task you'd be interested in -- I've opened issue https://github.com/plotly/plotly.py/issues/5373 to track.

Otherwise, something like what you've proposed here will work as a temporary fix.

There's a few issues with the logic as written; for example, falling back to returning `x[0]` is incorrect when `x[0]` and `x[1]` are not equal. We should simply raise an error in that case because we have no way of calculating an average. 

That means cases like the reproducible example in https://github.com/plotly/plotly.py/issues/3065, where the date value is passed as a string, will require a different approach. I'm not sure exactly what the best approach is; trying to parse every value as a datetime seems a bit clunky, but technically speaking it will produce the correct result. Or maybe we just document this limitation and specify that if you are trying to use `add_hline()` or `add_vline()` for a datetime axis you must pass the value as a datetime.

robertclaus:
@m-ad since we haven't seen any activity on this PR in a while, we may close it in the near future. Let us know if you are still working on it!

m-ad:
> @m-ad since we haven't seen any activity on this PR in a while, we may close it in the near future. Let us know if you are still working on it!

@robertclaus I am not currently working on it. This is a is a band-aid-style fix and as @emilykl said, the nicer and more "correct" version would be to implement the changes to `add_hline()` and `add_vline()` proposed in #5373. Someone appears to be working on that, which would ideally make this PR unnecessary. That said, if you still like to have my band-aid fix around until #5373 is closed, I'd be happy to finalize this PR.

Issue #4512, opened 2024-02-08
ValueError: Mime type rendering requires nbformat>=4.2.0 but it is not installed


NotShrirang:
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File [d:\Files\Training\Python\Visualization](file:///D:/Files/Training/Python/Visualization) Assignment\Startup-Data-Visualizer\venv\Lib\site-packages\IPython\core\formatters.py:922, in IPythonDisplayFormatter.__call__(self, obj)
    [920](file:///D:/Files/Training/Python/Visualization%20Assignment/Startup-Data-Visualizer/venv/Lib/site-packages/IPython/core/formatters.py:920) method = get_real_method(obj, self.print_method)
    [921](file:///D:/Files/Training/Python/Visualization%20Assignment/Startup-Data-Visualizer/venv/Lib/site-packages/IPython/core/formatters.py:921) if method is not None:
--> [922](file:///D:/Files/Training/Python/Visualization%20Assignment/Startup-Data-Visualizer/venv/Lib/site-packages/IPython/core/formatters.py:922)     method()
    [923](file:///D:/Files/Training/Python/Visualization%20Assignment/Startup-Data-Visualizer/venv/Lib/site-packages/IPython/core/formatters.py:923)     return True

File [d:\Files\Training\Python\Visualization](file:///D:/Files/Training/Python/Visualization) Assignment\Startup-Data-Visualizer\venv\Lib\site-packages\plotly\basedatatypes.py:832, in BaseFigure._ipython_display_(self)
    [829](file:///D:/Files/Training/Python/Visualization%20Assignment/Startup-Data-Visualizer/venv/Lib/site-packages/plotly/basedatatypes.py:829) import plotly.io as pio
    [831](file:///D:/Files/Training/Python/Visualization%20Assignment/Startup-Data-Visualizer/venv/Lib/site-packages/plotly/basedatatypes.py:831) if pio.renderers.render_on_display and pio.renderers.default:
--> [832](file:///D:/Files/Training/Python/Visualization%20Assignment/Startup-Data-Visualizer/venv/Lib/site-packages/plotly/basedatatypes.py:832)     pio.show(self)
    [833](file:///D:/Files/Training/Python/Visualization%20Assignment/Startup-Data-Visualizer/venv/Lib/site-packages/plotly/basedatatypes.py:833) else:
[...]

PR #5380, opened 2025-10-16
include nbformat as a plotly[express] dependency

afeld:
Required for MimeType rendering. Closes https://github.com/plotly/plotly.py/issues/4512.

## Code PR

- [ ] ~~I have added tests or modified existing tests.~~
- [ ] ~~For a new feature, I have added documentation examples (please see the doc checklist as well).~~
- [ ] ~~For a new feature or a change in behavior, I have updated the relevant docstrings in the code.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Updated
+
+- Include `nbformat` as a dependency for `plotly[express]`. [[#5380](https://github.com/plotly/plotly.py/pull/5380)]
+
 ## [6.3.1] - 2025-10-02
 
 ### Updated
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,10 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-express = ["numpy"]
+express = [
+    "numpy",
+    "nbformat>=4.2.0"
+]
 kaleido = ["kaleido>=1.0.0"]
 dev_core = [
     "pytest",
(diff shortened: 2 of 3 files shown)


afeld:
Hmm, [the remaining build error](https://app.circleci.com/pipelines/github/plotly/plotly.py/5281/workflows/2875b85d-4dd3-420e-9a67-1fbb9e300918/jobs/122213) doesn't sense to me… [Doesn't seem to be happening on `main`](https://app.circleci.com/pipelines/github/plotly/plotly.py?branch=main), but at a glance, I don't get why this change would have caused it. Any suggestions?

emilykl:
@afeld AFAIK `nbformat` is only required when working with Jupyter notebooks, is there a scenario you've encountered where you require `nbformat` when working outside Jupyter?

In general we try to keep Plotly Express dependencies pretty minimal and only require packages which are needed in all scenarios.

Would be open to maybe adding a `[jupyter]` extra or something along those lines though.

afeld:
> `nbformat` is only required when working with Jupyter notebooks … 

Sure. Do you have a sense of how many Plotly Express users are doing so in Jupyter vs. not?

> is there a scenario you've encountered where you require nbformat when working outside Jupyter?

No, I exclusively(?) use Plotly with Jupyter, with build tools like [nbconvert](https://nbconvert.readthedocs.io/) and [Jupyter Book](https://jupyterbook.org/).

> In general we try to keep Plotly Express dependencies pretty minimal and only require packages which are needed in all scenarios.

The impression I got was that `plotly` contains the bare minimum, while `plotly[express]` was the "batteries included" / "give me what I need to make charts quickly" option. (Re-reading the [Getting Started](https://plotly.com/python/getting-started/#overview), realizing that is not what it says.)

> Would be open to maybe adding a [jupyter] extra or something along those lines though.

That makes sense to me!

```sh
pip install plotly[express,jupyter]
```

would still be easy for users. I'll update.

PR #5384, opened 2025-10-28
Documentation conversion into Mkdocs

daexs:
Converted the plotly.py documentation build into Mkdocs
- Create a Mkdocs build for the plotly.py docs and reference pages, including styling.
- Modifies the code generator to create output that is compatible with Markdown rendering. For example, `\\\\[i\\\\]\\\\[{i}\\\\]` ensures that Mkdocs will correctly render this portion of the docstring into `[i][i]`.
- Modifies the examples for any code block errors, jupyter special commands, and formatting issues. For example, lists do not render correctly unless a new line was added.
- Modifies link references within example and reference pages for readability and converted url links into internal markdown links for Mkdocs to handle.
- Home/Landing page is `plotly/plotly.py/README.md` and index pages are placed in each individual category of examples under `Overview`.

### DIFF
--- a/docs/pages/css/plotly-styles.css
+++ b/docs/pages/css/plotly-styles.css
@@ -342,9 +342,8 @@ article {
     color: #a0aaba !important;
 }
 
-
-[data-md-color-scheme="slate"] .--footer-column ul a {
-    color: #a0aaba;
+[data-md-color-scheme="slate"] .--footer-column ul a:not(.subscribe-button) {
+    color: #a0aaba !important;
 }
 
 [data-md-color-scheme="slate"] .--footer-meta a {


doc/requirements.txt
```diff
@@ -1,48 +0,0 @@
-plotly==6.3.0
-anywidget
-cufflinks==0.17.3
-dash-bio
```
LiamConnors, line 4: One of the tests is failing because this is missing. I think we can restore it for now and completely remove it once we take down the old docs

CONTRIBUTING.md
```diff
@@ -38,16 +38,17 @@ it is important to understand the structure of the code and the repository.
     These are organized in subdirectories according to what they test:
     see the "Setup" section below for more details.
 
--   Documentation is found in `doc/`, and its structure is described in [its README file](doc/README.md).
+-   Documentation is found in `doc/`, and its structure is described in its README file.
```
LiamConnors, line 41: Could we update the readme to have the new doc build process. 
And add a github action that runs all the commands required to build the docs and save them as an artifact.

bin/codegen/__init__.py
```diff
@@ -29,6 +28,11 @@
 )
 
 
+PROJECT_ROOT = Path(__file__).parent.parent.parent
+PLOT_SCHEMA_RELATIVE = Path("resources") / "plot-schema.json"
+PLOT_SCHEMA = PROJECT_ROOT / PLOT_SCHEMA_RELATIVE
+
+
```
LiamConnors, line 35: Why is this change needed in codegen for docs?

emilykl:
@LiamConnors @daexs The `perform_codegen()` function is missing the reformat step.

See lines 326-330 of `commands.py`: [Link](https://github.com/plotly/plotly.py/blob/11d50b056752033df792fc56efccae5d3835cf5d/codegen/__init__.py#L326C1-L330C30)

These lines should be added at the end of `perform_codegen()` in `bin/codegen/__init__.py`.

There is also a missing import statement. `import argparse` should be added at the top of `bin/utils.py`.

emilykl:
Actually, hmm. It looks like the entire `reformat_code()` function is missing from this branch entirely.

To be honest I'm not very enthusiastic about mixing codegen changes in with this PR. I realize there are some changes to the codegen which are required in order for Mkdocs to work properly, but I would prefer to keep the codegen changes to _only_ what is strictly necessary for the docs to work, and keep any other codegen changes for another PR.

emilykl:
In any case, the `reformat_code()` function is pretty minimal. It's just:


```python
import subprocess

...

def reformat_code(outdir):
    """Reformat Python code using settings in pyproject.toml."""
    subprocess.call(["ruff", "format", *make_paths(outdir)])
```

So the immediate issue can be fixed by adding that function to `bin/utils.py`.

emilykl:
@daexs There are two changes I had to make in order to build the docs successfully on my machine:

- Add `import argparse` at the top of `bin/utils.py`

- Add `--extra docs` to the `reference` command in the Makefile, so that the final command is: `${RUN} --extra docs bin/generate_graph_objects_docs.py` (This tells `uv` to run the command in an environment where the `docs` requirements are installed)

doc/requirements.txt
daexs: Sounds good, I've added the file back in

daexs:
> Actually, hmm. It looks like the entire `reformat_code()` function is missing from this branch entirely.
> 
> To be honest I'm not very enthusiastic about mixing codegen changes in with this PR. I realize there are some changes to the codegen which are required in order for Mkdocs to work properly, but I would prefer to keep the codegen changes to _only_ what is strictly necessary for the docs to work, and keep any other codegen changes for another PR.

@emilykl I tried adding the lines of code (functions, import statements) you mentioned were missing, but it ended up breaking the process of regenerating the code. I took a look at older commits and it seems like there was a lot of refactoring that was done to make the current codegen work. This is also causing the `dev_build` test to fail because the test expects a `commands.py` in the top level but it was refactored into `bin/utils.py`. I have left them as is in this PR for now because I tried to undo some things, but it created a larger mess. I will have to spend more time checking why each change was made.

If I understand correctly, `reformat_code` might have been deleted because the make rule `make generate` runs the reformatting process with ruff right after the code is generated.

emilykl:
@daexs Yeah after looking at the history I realized all the codegen refactor was done by Greg, so don't worry about it!

PR #5388, opened 2025-10-31
Update plotly.js to v3.2.0 and associated files

camdecoster:
### Description

Update plotly.js to v3.2.0 and associated files.

### DIFF
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -1,5045 +1,5045 @@
 {
-  "name": "jupyterlab-plotly",
-  "version": "6.0.1",
-  "lockfileVersion": 3,
-  "requires": true,
-  "packages": {
-    "": {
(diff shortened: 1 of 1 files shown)


js/package-lock.json
emilykl, line 1: @camdecoster can we keep the previous indentation here? Doesn't matter THAT much but it's like 60% of the diff for this PR and has a good chance of getting reverted the next time we bump the Plotly.js version.

(I guess I spoke too soon in saying we don't need a JSON/JS formatter for Plotly.py...)

emilykl (approved):

js/package-lock.json
camdecoster: Sorry about that. ESLint took over with a default of 2 spaces per level since no formatter is set for this repo.

PR #5415, opened 2025-11-10
Fix bug where numpy datetime contained in Python list gets converted to integer

### LINKED ISSUE #3957: Incorrect handling of `np.datetime64[ns]` date values
Dates specified using numpy `datetime64` with `[ns]` precision are handled incorrectly in plotting. Dates with microsec `[us]` precision or lower work as expected.
```
t0 = np.datetime64("2022-10-20T00:00:00").astype("datetime64[ns]")
t1 = np.datetime64("2022-10-21T00:00:00").astype("datetime64[ns]")

fig = go.Figure()
fig.add_trace(go.Scatter(x=[t0, t1], y=[3, 4], mode="lines"))
fig.show()
```
newplot

### LINKED ISSUE #5355: plotly 6.3.0 does not work with numpy datetime
I upgraded my requirements from `plotly==5.9.0` to `plotly==6.3.0` and observed a regression. Markers on the plot disappeared without any error.

Simply put, my code takes a figure as input and plots a dot at the end of each line. Simple reproducible example:
```python
import numpy as np
import pandas as pd
import plotly.graph_objs as pgo
import plotly.subplots

np.random.seed(42)

# random data
dates = pd.date_range(start="2025-01-01", periods=3, freq="D")
data = np.random.randn(3, 1)
df = pd.DataFrame(data, index=dates, columns=["value"])

# this is plotted elsewhere
fig = plotly.subplots.make_subplots()

for column in df.columns:
    trace = pgo.Scatter(
        x=df.index,
        y=df[column],
        mode="lines",
    )

    fig.add_trace(trace)

# take the figure and
# render last point of each line using fig.data
for i, trace in enumerate(fig.data):
    input_x = trace.x[-1]
    input_y = trace.y[-1]
    print("Input x:", input_x, type(input_x))
    print("Input y:", input_y, type(input_y))

    fig.add_scatter(
        x=(input_x,),
        y=(input_y,),
        mode="markers+text",
    )
    print(f"Plotted x value:", fig.data[-1].x)

fig.show()
```

If I run this with different plotly versions I get different type for the timestamp:

Image

With plotly<6, the figure looks like this

Image
With plotly>=6, it looks like this 

Image

I am aware that I can convert numpy datetime back to datetime.datetime to fix my problem but why does plotly treat numpy datetime as a unix timestamp? It's not being converted to an actual date and there is no error either.
Simply put I would expect x axis to be a date and not an impossibly large number:
```python
import numpy as np
import plotly.subplots

fig = plotly.subplots.make_subplots()

x = np.datetime64('2025-09-26T01:00:00.000000000')
y = 1.23

fig.add_scatter(
    x=(x,),
    y=(y,),
    mode="markers+text",
)
print(f"Plotted x value:", fig.data[-1].x)

fig.show()
```

Image

emilykl:
Closes #3957, closes #5355

Fixes an issue where numpy datetimes with nanosecond precision, contained in a Python list and passed to a `go` trace constructor, were treated as integers rather than dates, and displayed on a numerical axis rather than a date axis.

This is more severe than it seems at first glance, since pandas `Timestamp` objects (which are represented by the numpy `datetime64` datatype under the hood) use nanosecond precision by default, even when not required, so it is very easy to create a nanosecond precision datetime without intending to.
```python
import pandas as pd
x = pd.to_datetime('2025-09-26')
print(x.unit)
# output: ns
```

## The issue

Before this fix, the following code produces the incorrect result below. The x-axis should be a date axis, but instead it shows up as a numeric axis with very large numbers.

```python
import pandas as pd

x_vals = list(pd.to_datetime(['2025-01-01', '2025-01-02', '2025-01-03']).to_numpy())
y_vals = [10, 20, 15]
fig = go.Figure([go.Scatter(x=x_vals, y=y_vals, mode='lines+markers')])
fig.show()
```
Screenshot 2025-11-11 at 11 22 55 AM

## The fix
These nanosecond-precision datetimes are converted to ints during the coercion step; specifically by a call to the numpy [`.item()`](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.item.html) function which converts a single value from a numpy type to a built-in Python type. For `datetime64` types with microsecond precision or less, a call to `.item()` returns a Pandas `datetime` type. But since the Pandas datetime does not support any precision greater than microseconds, calling `.item()` on a `datetime64` with nanosecond precision or greater simply returns a large integer.

This PR resolves the issue by adding an additional check to verify whether the value is a `datetime64` with nanosecond precision, and if so, casts it to microsecond precision before calling `.item()`.

## To test

Run the code snippet given under "The issue" above. On this branch, it should produce a chart with the x-axis being a date axis:
Screenshot 2025-11-11 at 11 21 58 AM

### DIFF
--- a/_plotly_utils/basevalidators.py
+++ b/_plotly_utils/basevalidators.py
@@ -29,7 +29,7 @@ def to_non_numpy_type(np, v):
     Python datetimes only support microsecond precision. So we cast
     datetime64[ns] to datetime64[us] to ensure it remains a datetime.
 
-    Should only be used in contexts where we already know `np` is defined
+    Should only be used in contexts where we already know `np` is defined.
     """
     if hasattr(v, "dtype") and v.dtype == np.dtype("datetime64[ns]"):
         return v.astype("datetime64[us]").item()


_plotly_utils/basevalidators.py
```diff
@@ -22,6 +22,20 @@ def fullmatch(regex, string, flags=0):
...
 
 
+def to_non_numpy_type(np, v):
+    """
+    Convert a numpy scalar value to a native Python type.
+    Calling .item() on a datetime64[ns] value returns an integer, since
+    Python datetimes only support microsecond precision. So we cast
+    datetime64[ns] to datetime64[us] to ensure it remains a datetime.
+
+    Should only be used in contexts where we already know `np` is defined
```
camdecoster:
```suggestion
    Should only be used in contexts where we already know `np` is defined.
```

camdecoster (approved):
Looks good! I suggested one very small change.

PR #4782, opened 2024-10-07
Update geopandas example

sarthak-12:
Please uncomment this block and take a look at this checklist if your PR is making substantial changes to **documentation**/impacts files in the `doc` directory. Check all that apply to your PR, and leave the rest unchecked to discuss with your reviewer! Not all boxes must be checked for every PR :)

If your PR modifies code of the `plotly` package, we have a different checklist
below :-).

### Documentation PR

- [x] I've [seen the `doc/README.md` file](https://github.com/plotly/plotly.py/blob/master/doc/README.md)
- [x] This change runs in the current version of Plotly on PyPI and targets the `doc-prod` branch OR it targets the `master` branch
- [x] If this PR modifies the first example in a page or adds a new one, it is a `px` example if at all possible
- [x] Every new/modified example has a descriptive title and motivating sentence or paragraph
- [x] Every new/modified example is independently runnable
- [x] Every new/modified example is optimized for short line count	and focuses on the Plotly/visualization-related aspects of the example rather than the computation required to produce the data being visualized
- [x] Meaningful/relatable datasets are used for all new examples instead of randomly-generated data where possible
- [ ] The random seed is set if using randomly-generated data in new/modified examples
- [ ] New/modified remote datasets are loaded from https://plotly.github.io/datasets and added to https://github.com/plotly/datasets
- [x] Large computations are avoided in the new/modified examples in favour of loading remote datasets that represent the output of such computations
- [x] Imports are `plotly.graph_objects as go` / `plotly.express as px` / `plotly.io as pio`
- [x] Data frames are always called `df`
- [x] `fig = <something>` call is high up in each new/modified example (either `px.<something>` or `make_subplots` or `go.Figure`)
- [x] Liberal use is made of `fig.add_*` and `fig.update_*` rather than `go.Figure(data=..., layout=...)` in every new/modified example
- [x] Specific adders and updaters like `fig.add_shape` and `fig.update_xaxes` are used instead of big `fig.update_layout` calls in every new/modified example
- [x] `fig.show()` is at the end of each new/modified example
- [x] `plotly.plot()` and `plotly.iplot()` are not used in any new/modified example
- [ ] Hex codes for colors are not used in any new/modified example in favour of [these nice ones](https://github.com/plotly/plotly.py/issues/2192)

## Code PR

[...]

PR #5420, opened 2025-11-16
Fix GeoPandas dataset compatibility for naturalearth_cities

### LINKED ISSUE #5289: Geopandas example for tile maps fails with geopandas>1.0
In https://plotly.com/python/tile-scatter-maps/#basic-example-with-geopandas, the example:

```python
import plotly.express as px
import geopandas as gpd

geo_df = gpd.read_file(gpd.datasets.get_path("naturalearth_cities"))

fig = px.scatter_map(
    geo_df,
    lat=geo_df.geometry.y,
    lon=geo_df.geometry.x,
    hover_name="name",
)
fig.show()
```

Gives:
```log
AttributeError: The geopandas.dataset has been deprecated and was removed in GeoPandas 1.0. You can get the original 'naturalearth_cities' data from https://www.naturalearthdata.com/downloads/110m-cultural-vectors/.
```

I assume we pin an old geopandas but it would need to be updated to work with the latest

AliAbouelazm:
Closes #5289 

Update documentation examples to support both GeoPandas < 1.0 and >= 1.0.

Context: GeoPandas 1.0 removed the datasets accessor used in `gpd.datasets.get_path("naturalearth_cities")`.

Changes:
- Try old method first for backward compatibility
- Fallback to `geodatasets` package for GeoPandas >= 1.0
- Final fallback to a public GeoJSON URL if `geodatasets` is not available

Files updated:
- doc/python/scatter-plots-on-maps.md
- doc/python/tile-scatter-maps.md

Docs-only change.

### DIFF
--- a/doc/python/scatter-plots-on-maps.md
+++ b/doc/python/scatter-plots-on-maps.md
@@ -74,7 +74,7 @@ fig.show()
 import plotly.express as px
 import geopandas as gpd
 
-# Handle both old and new GeoPandas versions
+# Handle both old and new GeoPandas versions without requiring network access
 try:
     # Try the old method (GeoPandas < 1.0)
     geo_df = gpd.read_file(gpd.datasets.get_path('naturalearth_cities'))
@@ -84,8 +84,12 @@ except (AttributeError, ValueError):
         import geodatasets
         geo_df = gpd.read_file(geodatasets.get_path('naturalearth.cities'))
     except ImportError:
-        # Fallback: use a direct URL if geodatasets is not available
-        geo_df = gpd.read_file('https://raw.githubusercontent.com/geopandas/geopandas/main/tests/data/naturalearth_cities.geojson')
+        # Fallback: build a tiny in-memory GeoDataFrame (no internet or extra deps)
+        from shapely.geometry import Point
+        geo_df = gpd.GeoDataFrame(
+            {"name": ["City A", "City B"], "geometry": [Point(0, 0), Point(10, 10)]},
+            crs="EPSG:4326",
+        )
 
 px.set_mapbox_access_token(open(".mapbox_token").read())
 fig = px.scatter_geo(geo_df,
--- a/doc/python/tile-scatter-maps.md
+++ b/doc/python/tile-scatter-maps.md
@@ -56,7 +56,7 @@ fig.show()
 import plotly.express as px
 import geopandas as gpd
 
-# Handle both old and new GeoPandas versions
+# Handle both old and new GeoPandas versions without requiring network access
 try:
     # Try the old method (GeoPandas < 1.0)
     geo_df = gpd.read_file(gpd.datasets.get_path('naturalearth_cities'))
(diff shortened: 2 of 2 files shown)


camdecoster:
Thanks for the PR! We'll try to review this as we have time. This seems to be related to #4782.

doc/python/scatter-plots-on-maps.md
```diff
@@ -74,7 +74,22 @@ fig.show()
...
-geo_df = gpd.read_file(gpd.datasets.get_path('naturalearth_cities'))
+# Handle both old and new GeoPandas versions without requiring network access
+try:
+    # Try the old method (GeoPandas < 1.0)
+    geo_df = gpd.read_file(gpd.datasets.get_path('naturalearth_cities'))
+except (AttributeError, ValueError):
+    # Use the new method (GeoPandas >= 1.0)
+    try:
+        import geodatasets
+        geo_df = gpd.read_file(geodatasets.get_path('naturalearth.cities'))
```
LiamConnors, line 85: This dataset doesn't currently exist on geodatasets, so this will never be successful even with geopandas >=1.0

doc/python/tile-scatter-maps.md
```diff
@@ -56,7 +56,22 @@ fig.show()
...
+    try:
+        import geodatasets
+        geo_df = gpd.read_file(geodatasets.get_path('naturalearth.cities'))
+    except ImportError:
+        # Fallback: build a tiny in-memory GeoDataFrame (no internet or extra deps)
+        from shapely.geometry import Point
+        geo_df = gpd.GeoDataFrame(
+            {"name": ["City A", "City B"], "geometry": [Point(0, 0), Point(10, 10)]},
+            crs="EPSG:4326",
+        )
```
LiamConnors, line 74: As `'naturalearth.cities'` does not exist on `geodatasets`, I think the best thing to do here would be to have just the example of the dataframe created using `GeoDataFrame`, like you show in the `except` block, but with some real cities. 

So something like

```python
import plotly.express as px
import geopandas as gpd
from shapely.geometry import Point

geo_df = gpd.GeoDataFrame({
    'name': ['London', 'Paris', 'Berlin', 'Madrid', 'Rome'],
    'geometry': [
        Point(-0.13, 51.51),
        Point(2.35, 48.86),
        Point(13.41, 52.52),
        Point(-3.70, 40.42),
        Point(12.50, 41.90),
    ]
}, crs="EPSG:4326")

fig = px.scatter_map(geo_df,
                    lat=geo_df.geometry.y,
                    lon=geo_df.geometry.x,
                    hover_name="name",
                    zoom=3)
fig.show()
```

And something similar for the scatter-plots-on-maps example

emilykl:
@LiamConnors Any reason not to simply require `geopandas>=1.0`? I'm not sure if we need to provide examples for both old and new geopandas.

LiamConnors:
> @LiamConnors Any reason not to simply require `geopandas>=1.0`? I'm not sure if we need to provide examples for both old and new geopandas.

I agree. It doesn't need to be compatible with old `geopandas`. Instead of having an example that creates the`GeoDataFrame` in the code, it would also be fine to just use one of the available datasets in `geodatasets` even if that won't work with earlier versions of `geopandas`.

PR #5421, opened 2025-11-17
update plotly.js to v3.3.0

emilykl:
plotly.js v3.3.0 release notes: https://github.com/plotly/plotly.js/releases/tag/v3.3.0


js/package-lock.json
emilykl, line 1: @camdecoster fyi: Turns out `package-lock.json` inherits its indentation from `package.json` by default, so if we keep `package.json` at 2 spaces, `package-lock.json` needs to be two spaces as well.

a3480f0 changes the indentation, 83ab425 updates the actual contents of the lockfile

camdecoster: What a fun world we live in. For future reference, [here's](https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json#description) the doc page describing this:

> When `npm` creates or updates `package-lock.json`, it will infer line endings and indentation from `package.json` so that the formatting of both files matches.

PR #5424, opened 2025-11-18
Refactor/autoshapes-use-shape-label

nochinxx:
vline: switch to shape.label (no separate annotation) + legacy shim

Summary

add_vline now produces one labeled shape (no layout.annotations).

Legacy annotation_* kwargs are accepted and translated (with FutureWarning).

Date strings (and other non-numeric x) no longer cause TypeError because we removed arithmetic from annotation placement.

Code highlights

_coerce_shape_label_from_legacy_annotation_kwargs (prep): copies annotation_text/font/textangle → label.text/font/textangle (non-destructive).

_normalize_legacy_line_position_to_textposition maps legacy positions to start|middle|end for lines (vline: top→end, bottom→start).

_process_multiple_axis_spanning_shapes: new vline branch adds a single shape with label, skips add_annotation.

Tests

Current failures are expected: tests still assert layout.annotations for vline.

Follow-ups will update tests to assert shape.label for vline only (others unchanged until refactored).

Migration

annotation_bgcolor / annotation_bordercolor aren’t supported by shape.label; warning emitted and ignored.

annotation_* is deprecated; prefer label={...}.

Next steps

Update vline tests.

Mirror refactor to hline, then vrect/hrect.

Update tests per helper; keep diffs small.


nochinxx:
Retargeting to team feature branch

PR #5426, opened 2025-11-18
Make docs changes live

LiamConnors:


emilykl:
@LiamConnors Is there a reason you didn't include the updated `js/` in this PR? I'm not sure it actually matters, but I guess I would expect them to be kept up to date with what's on `main`

LiamConnors:
> @LiamConnors Is there a reason you didn't include the updated `js/` in this PR? I'm not sure it actually matters, but I guess I would expect them to be kept up to date with what's on `main`

which `js` is missing? I think this PR has everything from main

Issue #5433, opened 2025-11-26
[Bug] Plotly express `color_continuous_scale` ignored when template has `autocolorscale=True`


antonymilne:
## Problem

When a Plotly template sets `coloraxis_autocolorscale=True`, the `color_continuous_scale` argument passed to plotly.express functions is ignored.

## Reproduction

```python
import plotly.express as px
import plotly.graph_objects as go

df = px.data.gapminder().query("year == 2007")
template = go.layout.Template(layout=go.Layout(coloraxis_autocolorscale=True))

# User specifies colorscale - THIS IS IGNORED
fig = px.choropleth(
    df, 
    locations="iso_alpha", 
    color="lifeExp",
    color_continuous_scale=["red", "yellow", "green"],
    template=template,
)
fig.show()  # Renders with auto-detected scale, not red-yellow-green
```

## Root cause

As per the [plotly docs](https://plotly.com/python/reference/layout/coloraxis/), when `autocolorscale=True`, the colorscale is auto-detected based on the data, ignoring any explicit `colorscale` setting (this itself is quite confusing behaviour I think, but that's a separate issue - let's assume for now it should be left with that logic):
> In case `colorscale` is unspecified **or** `autocolorscale` is True, the default palette will be chosen according to whether numbers in the `color` array are all positive, all negative or mixed.

In the following code `autocolorscale` is not set on `coloraxis1`. This means the template's `autocolorscale=True` is inherited, causing the user's `color_continuous_scale` to be completely ignored:
https://github.com/plotly/plotly.py/blob/f08397718a3b37487c5ea98e4f146ad3258c8e56/plotly/express/_core.py#L2707-L2716

## Proposed fix

When `color_continuous_scale` is explicitly provided, set `autocolorscale=False` on `coloraxis1` to ensure the user's colorscale is respected. I can raise a PR for this if you like!

robertclaus:
Agreed that this isn't the most intuitive behavior. We would be open to making this change if you could open a PR. The only notable concern would be around backwards compatibility if anyone was assuming this behavior.

antonymilne:
Hi @robertclaus, I've opened a PR to fix this in https://github.com/plotly/plotly.py/pull/5439.

I've gone for the "light touch" fix that only affects Plotly Express so there shouldn't be any concerns around backwards compatibility here.

[...]

PR #5439, opened 2025-12-02
[Fix] Always respect user-specified `color_continuous_scale`

antonymilne:
Fixes https://github.com/plotly/plotly.py/issues/5433. When a user explicitly provides `color_continuous_scale`, it should always be respected, even if the template has `coloraxis_autocolorscale=True`. Previously, the template's `autocolorscale` setting would override the user's explicit colorscale.

A couple of tests are failing but I'm pretty sure it's not my fault.

Here's the behaviour now. Before this PR, all plots in the above table would have looked the same with the exception of the top left one, where the user-specified `color_continuous_scale` would have been ignored.

| autocolorscale=True | autocolorscale=False |
|---------------------|----------------------|
| 01_autocolorscale_True_colorscale_red_yellow_green | 02_autocolorscale_False_colorscale_red_yellow_green |
|03_autocolorscale_True_no_colorscale_sequential | 04_autocolorscale_False_no_colorscale_sequential |
| 07_autocolorscale_True_no_colorscale_diverging | 08_autocolorscale_False_no_colorscale_diverging |

## Code PR

### DIFF
--- a/plotly/express/_core.py
+++ b/plotly/express/_core.py
@@ -2509,9 +2509,7 @@ def make_figure(args, constructor, trace_patch=None, layout_patch=None):
     layout_patch = layout_patch or {}
     # Track if color_continuous_scale was explicitly provided by user
     # (before apply_default_cascade fills it from template/defaults)
-    user_provided_colorscale = (
-        "color_continuous_scale" in args and args["color_continuous_scale"] is not None
-    )
+    user_provided_colorscale = args.get("color_continuous_scale") is not None
     apply_default_cascade(args, constructor=constructor)
 
     args = build_dataframe(args, constructor)
--- a/plotly/express/_imshow.py
+++ b/plotly/express/_imshow.py
@@ -241,9 +241,7 @@ def imshow(
     args = locals()
     # Track if color_continuous_scale was explicitly provided by user
     # (before apply_default_cascade fills it from template/defaults)
-    user_provided_colorscale = (
-        "color_continuous_scale" in args and args["color_continuous_scale"] is not None
-    )
+    user_provided_colorscale = args.get("color_continuous_scale") is not None
     apply_default_cascade(args, constructor=None)
     labels = labels.copy()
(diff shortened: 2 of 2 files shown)


antonymilne:
@emilykl Happy new year to you! 🎉 Do you think you would be able to take a look at this? 🙏

plotly/express/_core.py
```diff
@@ -2507,6 +2507,11 @@ def get_groups_and_orders(args, grouper):
 def make_figure(args, constructor, trace_patch=None, layout_patch=None):
     trace_patch = trace_patch or {}
     layout_patch = layout_patch or {}
+    # Track if color_continuous_scale was explicitly provided by user
+    # (before apply_default_cascade fills it from template/defaults)
+    user_provided_colorscale = (
+        "color_continuous_scale" in args and args["color_continuous_scale"] is not None
+    )
```
emilykl: Just a nit, more concise

```suggestion
    user_provided_colorscale = args.get("color_continuous_scale") is not None
```

plotly/express/_imshow.py
```diff
@@ -239,6 +239,11 @@ def imshow(
     axes labels and ticks.
     """
     args = locals()
+    # Track if color_continuous_scale was explicitly provided by user
+    # (before apply_default_cascade fills it from template/defaults)
+    user_provided_colorscale = (
+        "color_continuous_scale" in args and args["color_continuous_scale"] is not None
+    )
```
emilykl:
```suggestion
    user_provided_colorscale = args.get("color_continuous_scale") is not None
```

emilykl (dismissed):
Thanks @antonymilne for the fix, and thanks also for your patience waiting for a review!

This looks good to me. Ultimately I think it probably makes sense to change this behavior  at the plotly.js level, but I'm not opposed to making the change here in the meantime.

antonymilne:
Thanks so much @emilykl!

PR #5458, opened 2025-12-31
Add embedded Plotly Studio scatter plot builder app

LiamConnors:
image

### DIFF
--- a/doc/python/line-and-scatter.md
+++ b/doc/python/line-and-scatter.md
@@ -85,7 +85,7 @@ fig = px.scatter(df, x="sepal_width", y="sepal_length", color="species", symbol=
 fig.show()
 ```
 
-**Try building a scatter plot.** Use the following interactive app, hosted on [Plotly Cloud](https://plotly.com/cloud/), to see how the code updates based on your selections.
+Try the live scatter plot below to see how the code updates based on your selections. Built with [Plotly Studio](http://plotly.com/downloads/?utm_source=graphing_libraries&utm_medium=documentation&utm_campaign=studio_cloud_community&utm_content=demo_scatter_plot_builder).
 
 ```python hide_code=true
 from IPython.display import IFrame


red-patience:
Neat idea, but I feel like this is the sort of thing we'd use as a consistent pattern or not at all. The idea is to teach folks how to construct their `px.scatter` params using a tool where they can immediately validate the result?

Coding-with-Adam:
> Neat idea, but I feel like this is the sort of thing we'd use as a consistent pattern or not at all. The idea is to teach folks how to construct their `px.scatter` params using a tool where they can immediately validate the result?

Good questions @red-patience . Yes, the goal is to teach users how to construct a scatter plot by playing with various parameters. By interacting live with parameter changes (dropdowns) and seeing how the code updates, I was hoping that it would give them a better understanding of how parameter modifications affect the visualization.  

And I was hoping to build similar apps for the other common visualizations (bar, pie, line, and others that you and Liam think we should add)

red-patience:
> > Neat idea, but I feel like this is the sort of thing we'd use as a consistent pattern or not at all. The idea is to teach folks how to construct their `px.scatter` params using a tool where they can immediately validate the result?
> 
> Good questions @red-patience . Yes, the goal is to teach users how to construct a scatter plot by playing with various parameters. By interacting live with parameter changes (dropdowns) and seeing how the code updates, I was hoping that it would give them a better understanding of how parameter modifications affect the visualization.
> 
> And I was hoping to build similar apps for the other common visualizations (bar, pie, line, and others that you and Liam think we should add)

Okay, that's what I thought, thanks for confirming @Coding-with-Adam. Several concerns.

* Does the Cloud SRE team know about this initiative? We get good traffic on these docs and we can lose user trust when the a part of the page looks broken because of Cloud downtime.
 
* There's already a _ton_ going on on this page. I'm hesitant to add another call to action/learning method. Do we know for a fact that this is addressing a need that the page wasn't before? The audience for this page is probably the most code-savvy of all our docs―are the existing examples on the page not enough to help them write 5 basic LOC? I dunno, as a reader this strikes me as "The author doesn't know how to best teach this content so they're throwing a lot of different things at the wall hoping something sticks.")

* As a user I would expect a tool like this to allow me to experiment with all the different types of scatter plots described on this page, but it looks like it's fairly limited. I can map axes, colours, etc to column names and see the graph change―cool. I'm not familiar with the sample dataset so I still have to translate any learnings to my dataset (but I can't see the sample dataset so I don't even know if it makes sense to model my work from a scatter plot I built using this tool). Feels like I'll just hit a wall really fast and then have to do some local work to keep testing.

* Additionally, I would want to be extra clear that this is a "meta" app:
  * Not sure I like "builder" in the app title. I see the intention but I feel like we should reserve this term for App Studio contents or App Studio itself. Thoughts on "sandbox" or "playground"? Similarly, the app description should be clearer about its purpose.
  * I'd recommend adding language like "This playground is a Dash app that renders a scatter plot using Plotly Express. It uses Dash core components (link) to add controls for updating the scatterplot. The changes that you make via these controls also update the underlying code, which is printed for demonstration purposes." But even then, folks might notice small changes in that code compared with the examples on the doc page ("why is there no `fig.show()`?"), which could be enough to create confusion.

* More distributed sample apps to maintain.

camdecoster:
@LiamConnors is this ready for the team to review?

Coding-with-Adam:
> * There's already a _ton_ going on on this page. I'm hesitant to add another call to action/learning method. Do we know for a fact that this is addressing a need that the page wasn't before? The audience for this page is probably the most code-savvy of all our docs―are the existing examples on the page not enough to help them write 5 basic LOC? I dunno, as a reader this strikes me as "The author doesn't know how to best teach this content so they're throwing a lot of different things at the wall hoping something sticks.")

Hey @red-patience ,
I want to focus on this comment you shared, because all the other comments are mainly suggestions that shouldn't be too hard to implement. 

No, we don't know for a fact that this app is addressing a need that the page wasn't before. The idea was to embed the app in the doc, thereby giving the user the opportunity to interact live with the scatter plot. My assumption was that some people  might learn content quicker (meaning, it would "click" for them)  if they got to interact with the plot. But I can't prove that it's a need. 

I think the existing examples on the page serve their purpose well of helping users write 5 basic LOC. My thought was to offer another form of knowledge gain (like adding a video tutorial inside a blog post). That said, I agree, if we think that this interactive scatter plot will cause more confusion than aid, we should not add it. 

@camdecoster , @LiamConnors I'm interested in hearing your thoughts as well. 

P.S. yes, good suggestions @red-patience regarding the updating of title and description .

camdecoster:
I think simpler is better when it comes to docs, so I'd skip embedding this app. However, you could link to it. I think having a playground would be a nice addition. Something that only runs client side so a user could add their data without worrying about it being sent anywhere.

Regarding the app, it would be nice to have a copy button for the code block.

Coding-with-Adam:
Good ideas, @camdecoster . Thank you for your feedback. 
At this point, @red-patience @LiamConnors , based on feedback, it's probably better we close the issue. 
Keeping the docs simple sounds like a good idea.

Issue #5463, opened 2026-01-09
[BUG]: Pie `legend`, `showlegend` attributes don't accept array values


camdecoster:
### Description

plotly.js 3.3.1 updated the pie trace to allow arrays of boolean values to be passed in for the `legend` and `showlegend` attributes. In plotly.py 6.5.1 (the latest version), the `BooleanValidator` doesn't allow for arrays. As such, arrays can't be passed in for these attributes in plotly.py.

### Steps to reproduce

- Install plotly.py 6.5.1
- Run the following code:
  ```python
  import plotly.graph_objects as go
  
  pie = go.Pie(
      labels=["A", "B", "C"],
      values=[65, 15, 20],
      showlegend=[True, False, False],
  )
  
  fig = go.Figure([pie], go.Layout())
  fig.show()
  ```
- Note the `ValueError` that's raised

### Notes

- This is the first time that a boolean attribute has had `arrayOk` as true in the schema. That's why this update is required.

camdecoster:
Fixed by #5464 and #5465.

PR #5464, opened 2026-01-09
fix: Handle arrays in BooleanValidator

camdecoster:
### Description

Update `BooleanValidator` to handle arrays of values.

Addresses half of #5463. The other half is addressed by #5465.

### Changes

- Handle arrays in `BooleanValidator`
- Add tests

### Testing

- Be on main
- Run the following code:
  ```python
  
  pie = go.Pie(
      labels=["A", "B", "C"],
      values=[65, 15, 20],
      showlegend=[True, False, False],
  )
  
  fig = go.Figure([pie], go.Layout())
  fig.show()
  ```
- Note that a `ValueError` is raised
- Switch to this branch
- Run the same code again
- Note that chart is rendered correctly

### DIFF
--- a/tests/test_plotly_utils/validators/test_boolean_validator.py
+++ b/tests/test_plotly_utils/validators/test_boolean_validator.py
@@ -42,10 +42,7 @@ def test_rejection(val, validator):
 @pytest.mark.parametrize("val", [(True, False), [True, False]])
 def test_acceptance_aok(val, validator_aok):
     v = validator_aok.validate_coerce(val)
-    if isinstance(val, list):
-        assert val == v
-    else:
-        assert val == tuple(v)
+    assert list(val) == v
 
 
 # Rejection


tests/test_plotly_utils/validators/test_boolean_validator.py
```diff
@@ -25,3 +33,27 @@ def test_rejection(val, validator):
...
+
+
+# Acceptance
+@pytest.mark.parametrize("val", [(True, False), [True, False]])
+def test_acceptance_aok(val, validator_aok):
+    v = validator_aok.validate_coerce(val)
+    if isinstance(val, list):
+        assert val == v
+    else:
+        assert val == tuple(v)
```
emilykl:
```suggestion
        assert list(val) == v
```

clearer IMO

_plotly_utils/basevalidators.py
```diff
@@ -632,24 +632,35 @@ class BooleanValidator(BaseValidator):
...
+        return desc
 
     def validate_coerce(self, v):
         if is_none_or_typed_array_spec(v):
             pass
+        elif self.array_ok and is_simple_array(v):
+            invalid_els = [e for e in v if not isinstance(e, bool)]
+            if invalid_els:
+                self.raise_invalid_elements(invalid_els[:10])
+            v = to_scalar_or_list(v)
```
emilykl, line 663: Not 100% sure whether this call to `to_scalar_or_list()` is necessary, but I don't think it will hurt.

tests/test_plotly_utils/validators/test_boolean_validator.py
camdecoster: With your change, I removed the if/else block. Clearer indeed.

PR #5465, opened 2026-01-12
Add array_ok to SubplotidValidator, used in pie.legend

my-tien:
### Description

Update `SubplotidValidator` to handle arrays of values.

Additional fix for closing #5463.
### Changes

* Handle arrays in `SubplotidValidator`

* Added array_ok tests

### Testing

* Be on main

* Run the following code:

    ```python
    
    pie = go.Pie(
        labels=["A", "B", "C"],
        values=[65, 15, 20],
        legend=["legend", "legend2"],
    )
    
    fig = go.Figure([pie], go.Layout())
    fig.show()
    ```

* Note that a `ValueError` is raised

* Switch to this branch

* Run the same code again

* Note that chart is rendered correctly

<details>
<summary>Disclaimer</summary>
I am required to add that…

the software is provided "as is", without warranty of any kind, express or implied, including but not limited to the warranties of merchantability, fitness for a particular purpose and noninfringement. in no event shall the authors or copyright holders be liable for any claim, damages or other liability, whether in an action of contract, tort or otherwise, arising from, out of or in connection with the software or the use or other dealings in the software.
</details >

### DIFF
--- a/_plotly_utils/basevalidators.py
+++ b/_plotly_utils/basevalidators.py
@@ -632,24 +632,35 @@ class BooleanValidator(BaseValidator):
         "description": "A boolean (true/false) value.",
         "requiredOpts": [],
         "otherOpts": [
+            "arrayOk",
             "dflt"
         ]
     },
     """
 
-    def __init__(self, plotly_name, parent_name, **kwargs):
+    def __init__(self, plotly_name, parent_name, array_ok=False, **kwargs):
         super(BooleanValidator, self).__init__(
             plotly_name=plotly_name, parent_name=parent_name, **kwargs
         )
+        self.array_ok = array_ok
 
     def description(self):
-        return """\
-    The '{plotly_name}' property must be specified as a bool
-    (either True, or False)""".format(plotly_name=self.plotly_name)
+        desc = """\
+    The '{plotly_name}' property is a boolean and must be specified as:
+      - A boolean value: True or False""".format(plotly_name=self.plotly_name)
+        if self.array_ok:
+            desc += """
+      - A tuple or list of the above"""
+        return desc
 
     def validate_coerce(self, v):
         if is_none_or_typed_array_spec(v):
             pass
+        elif self.array_ok and is_simple_array(v):
+            invalid_els = [e for e in v if not isinstance(e, bool)]
+            if invalid_els:
+                self.raise_invalid_elements(invalid_els[:10])
+            v = to_scalar_or_list(v)
         elif not isinstance(v, bool):
             self.raise_invalid_val(v)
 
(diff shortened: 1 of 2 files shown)


my-tien:
FYI @camdecoster @emilykl

_plotly_utils/basevalidators.py
```diff
@@ -1763,31 +1766,51 @@ def description(self):
     optionally followed by an integer >= 1
     (e.g. '{base}', '{base}1', '{base}2', '{base}3', etc.)
         """.format(plotly_name=self.plotly_name, base=self.base)
```
camdecoster: You can remove this code since you redefined it below.
```suggestion
```

_plotly_utils/basevalidators.py
```diff
@@ -1763,31 +1766,51 @@ def description(self):
...
+
+        desc = """\
+    The '{plotly_name}' property is an identifier of a particular
+    subplot, of type '{base}', that may be specified as:
+      - the string '{base}'
+    optionally followed by an integer >= 1
+    (e.g. '{base}', '{base}1', '{base}2', '{base}3', etc.)""".format(
```
camdecoster:
```suggestion
      - the string '{base}' optionally followed by an integer >= 1
        (e.g. '{base}', '{base}1', '{base}2', '{base}3', etc.)""".format(
```

_plotly_utils/basevalidators.py
camdecoster, line 1750:
```suggestion
       "otherOpts": [
           "arrayOk",
           "regex"
```

_plotly_utils/basevalidators.py
```diff
@@ -1763,31 +1766,51 @@ def description(self):
...
+                    invalid_els.append(value)
+                    return value
                 elif len(digit_str) > 0 and int(digit_str) == 1:
-                    # Remove 1 suffix (e.g. x1 -> x)
-                    v = self.base
-                    is_valid = True
+                    return self.base
                 else:
-                    is_valid = True
+                    return value
```
camdecoster: I think it would make more sense to refactor this function to just check if the given input is a valid id. Then you could reference that in the if/then block. It feels weird to define `invalid_els` for the single value case.

_plotly_utils/basevalidators.py
my-tien: Thx for spotting, removed.

_plotly_utils/basevalidators.py
my-tien: I did it like this because the return value is already used for the coercion result. (e.g. legend1 → legend). I changed it to return a tuple (coercion result, success) now, maybe this is clearer? Let me know what you think.

While doing that, I noticed that the coercion wasn't actually tested. I added a test for this as well.

camdecoster: Ah, I didn't realize that the check coerced legend1 → legend, so returning multiple values is required. This is fine.

camdecoster (approved):

PR #5498, opened 2026-02-09
Made the spelling of hover mode and hover template consistent with ho…

rl-utility-man:
This page consistently used two words for hover text and did so in 9 places.  It was inconsistent in whether prose mentions of hover mode and hover template were one word or two.  I made all three terms consistently 2 words in the prose.


camdecoster:
Thanks for the PR! We'll take a look and follow up with you.

PR #5502, opened 2026-02-11
Update figure-structure.md

aegaona96-glitch:
etapas del ciclo presupuestal


camdecoster:
Thanks for the PR! We won't be making this change.

PR #5516, opened 2026-02-19
Docs for 6.6

LiamConnors:


doc/python/legend.md
```diff
@@ -832,6 +832,63 @@ fig.show()
...
+To disable title click behavior, set `titleclick` and `titledoubleclick` to `False`:
+
+```python
+fig.update_layout(
+    legend=dict(titleclick=False, titledoubleclick=False),
+    legend2=dict(titleclick=False, titledoubleclick=False),
+)
+```
+
+> **Note:** Legend title click interactions are not supported for legends containing pie or pie-like traces.
```
emilykl, line 889: Instead of saying "pie-like trace", can we just list the trace types which are not supported? It's not obvious to me which trace types count as pie-like.

doc/python/shapes.md
```diff
@@ -488,6 +488,65 @@ fig.update_layout(
 fig.show()
 ```
 
+#### Shapes Spanning Subplots
```
emilykl:
```suggestion
#### Subplot-Spanning Shapes
```

does that sound better?

doc/python/legend.md
LiamConnors: great point. will update it

PR #5521, opened 2026-02-23
Remove redundant `conftest.py` to fix compatibility with pytest 9

thunze:
`pytest_ignore_collect` takes only `collection_path` starting with pytest 9. Most of the paths referenced in `plotly/conftest.py` don't exist anymore and wouldn't be collected anyway, so we can just remove the file (I think?).

I ran into this issue while running `pytest` instead of `pytest tests` with pytest 9.0.2 installed.

https://docs.pytest.org/en/latest/deprecations.html#py-path-local-arguments-for-hooks-replaced-with-pathlib-path

## Code PR


camdecoster:
The only file referenced that still exists is **exploding_module.py**. If tests pass, then I don't see a reason to keep this file around.

PR #5540, opened 2026-03-16
Clean up `pyproject.toml` and build pipeline

### LINKED ISSUE #5537: Some config options in `pyproject.toml` are unused
The `pyproject.toml` currently contains several sections starting with `tool.jupyter-packaging.*` and `tool.hatch.*` which, as far as I understand, are unused since we are using `setuptools` as our build backend, not `hatch` or `jupyter-packaging`. And we also have no `setup.py` file, which would be the other place those config options might have an impact.

At the very least, we should remove those sections; plausibly we should also consider whether it's worth moving to `hatchling` as a build backend, which might let us automate some of the js bundling steps that we are currently doing semi-manually.

See also #5476 which put this on my radar.

emilykl:
### Link to issue

Closes #5537

#### Configuration updates
- Switch build backend from `setuptools` to `hatch`
  - `hatch` is [distributed by PyPA](https://packaging.python.org/en/latest/key_projects/#hatch) and has a [number of advantages](https://hatch.pypa.io/latest/why/#build-backend) compared to `setuptools`, including being newer and more modern
- Update `pyproject.toml` config to work with `hatch`, and remove unused config sections
  - Remove `MANIFEST.in` file, since it's only required by `setuptools`, not needed for `hatch`
- Update `js/package.json` to reflect the correct version number matching plotly.py version, which we've forgotten to do for the last few releases
- Add `js/webpack.config.js` to make lab extension build output identical across platforms, which fixes the false failure of the "Check JS build" workflow when committing build artifacts built locally
- Remove keys from `js/install.json` file which I believe are not needed

#### Process updates
- Add CI step to check that version numbers in `pyproject.toml` and `js/package.json` match
- Add new command `bumpversion` to `commands.py`. Now, you can run `python commands.py bumpversion X.Y.Z` to update the version number to X.Y.Z *everywhere* across the package
- Update `RELEASE.md` instructions to use the bumpversion command

**NOTE: The python package build command remains the same: `python -m build .`**

### Demo

https://github.com/user-attachments/assets/b952d02e-b49e-4351-9619-21acb5089cbd

When running `python -m build .` locally, the output should look like the above.

### Testing strategy

I've added a CI step in `.github/workflows/check-js-build.yml` to check that the version number in `pyproject.toml` matches the version number in `js/project.json`, which should prevent the two from getting out of sync again.

<details>
<summary>Show success case</summary>

Screenshot 2026-03-17 at 12 37 12 PM

</details>

<details>
<summary>Show failure case</summary>

Screenshot 2026-03-17 at 12 32 30 PM

</details>
No additional unit testing is needed: The CI already runs `python -m build .` , so if that step fails then we know something is wrong.

However, we should verify manually that the resulting build is the same before and after these changes.

This can be done by checking out this branch, running `python -m build .`, and comparing the `.tar.gz` and `.whl` files to [the official 6.6.0 release artifacts](https://github.com/plotly/plotly.py/releases/tag/v6.6.0).

Because we manually pre-build the assets for the JupyterLab extension (by running `npm run build` from the `js/` directory, as documented in the `CONTRIBUTING.md`), we do not use any of the `hatch` build hooks settings in the `pyproject.toml`, so this PR removes them.

The existing `.github/check-js-build.yml` step builds the lab extension in CI and ensures that the output matches the build artifacts in `plotly/labextension/`.

I don't like that this setup requires updating the version number in two places. But `package.json` doesn't support dynamic version numbers, so the only way to set up a single source of truth would be for the `package.json` to be the single source of truth, and `pyproject.toml` to load that version number dynamically, which is weird for a Python project

To make this less painful I've added `python commands.py bumpversion X.Y.Z` which automatically updates the version number everywhere.

### Guidelines

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -307,3 +307,18 @@ You can then run the following command
 ```bash
 python commands.py updateplotlyjsdev --local /path/to/your/plotly.js/
 ```
+
+## Documentation for `commands.py`
+
+`commands.py` serves as an entry point for utilities to help with plotly.py development.
+
+Usage: `python commands.py <subcommand> <args>`
+
+| Subcommand | Purpose |
+|------------|---------|
+| `codegen [--noformat]` | Regenerate Python files according to `plot-schema.json`.`--noformat` skips formatter step. |
+| `lint` | Lint all Python code in `plotly/`. |
+| `format` | Format all Python code in `plotly/`. |
+| `updateplotlyjs` | Update `plotly.min.js` and `plot-schema.json` to match the `plotly.js` version specified in `js/package.json`. Then, run codegen to regenerate the Python files. |
+| `updateplotlyjsdev [--devrepo REPONAME --devbranch BRANCHNAME] \| [--local PATH]` | Update `plot-schema.json` and `plotly.min.js` to match the version in the provided plotly.js repo name and branch name, OR local path. Then, run codegen to regenerate the Python files. |
+| `bumpversion X.Y.Z` | Update the plotly.py version number to X.Y.Z across all files where it needs to be updated. |


emilykl:
fyi @franred this is the PR I mentioned last week!

pyproject.toml
```diff
@@ -31,7 +30,8 @@ classifiers = [
     "License :: OSI Approved :: MIT License"
 ]
 requires-python = ">=3.8"
-license = {file="LICENSE.txt"}
+license = "MIT"
+license-files = ["LICENSE.txt"]
 version = "6.6.0"
```
camdecoster, line 36: Does uv have something similar to `npm version`? If so, maybe they have post script hooks, also like npm?

.github/workflows/check-js-build.yml
```diff
@@ -2,30 +2,42 @@ on: push
...
+        fi
     - name: Install Node
       uses: actions/setup-node@v2
       with:
         node-version: '22'
 
     - name: Copy current files to a temporary directory
       run: |
-        cp -R plotly/labextension/ plotly/labextension-tmp/
+        mv plotly/labextension/ plotly/labextension-tmp/
```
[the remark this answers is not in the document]
emilykl: That's fine. If the job to generate `plotly/labextension/` fails, then there's nothing relevant to upload.

pyproject.toml
```diff
@@ -31,7 +30,8 @@ classifiers = [
     "License :: OSI Approved :: MIT License"
 ]
 requires-python = ">=3.8"
-license = {file="LICENSE.txt"}
+license = "MIT"
```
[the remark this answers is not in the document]
emilykl: Not the case. The old `license` field format, with `file` and/or `text` keys, is deprecated as of [PEP 639](https://peps.python.org/pep-0639/#deprecate-license-key-table-subkeys).

The change in this PR reflects the current standard.

pyproject.toml
emilykl: Hatch itself has a whole system for [build hooks](https://hatch.pypa.io/latest/plugins/build-hook/reference/) which we could definitely make use of.

I started going down that path for compiling the JS artifacts (i.e., adding the `npm run build` step as a hook which would run every time the Python package was built) but the `jupyter labextension build` step is VERY slow (like 20-30 seconds) and it didn't seem easy to control whether it runs or not. Worth looking more into in the future though.

commands.py
camdecoster, line 1: I think it would be good to port some of these changes to plotly.js.

emilykl: @camdecoster Sure, which parts? I'd be happy to open a PR. Most of this PR is only relevant to Python though.

PR #5511, opened 2026-02-16
Fix build docs

LiamConnors:
API docs build on `doc-prod` is failing with https://github.com/plotly/plotly.py/issues/5503 -
Setting setuptools<81 to fix it for now.

PR #5541, opened 2026-03-17
Migrate docs build to GitHub Actions

LiamConnors:
### Description of change

Updates docs build to use GitHub Actions instead of CircleCI. Needs following PR in graphing-library-docs to be merged also https://github.com/plotly/graphing-library-docs/pull/426 

### Testing strategy

Tested manually by comparing output of builds for CircleCI vs GitHub Actions

### Guidelines

### DIFF
--- a/.github/workflows/build-doc.yml
+++ b/.github/workflows/build-doc.yml
@@ -35,11 +35,6 @@ jobs:
           uv pip uninstall plotly
           uv pip install -e ..
 
-      - name: List installed packages
-        run: |
-          cd doc
-          source .venv/bin/activate
-          uv pip list
 
       - name: Build HTML docs
         env:


doc/python/static-image-generation-migration.md
```diff
@@ -80,4 +80,5 @@ If you are migrating from Orca, the following config settings do not apply to Ka
 ## Multiple Image Export
 
 Plotly.py 6.1 includes a `write_images` function (`plotly.io.write_images`), which we recommend over `write_image` when exporting more than one figure. Calling `write_images` with a list of figures (or dicts representing figures) to export provides better performance than multiple calls with `write_image`. See the [Write Multiple Images](/python/static-image-export#write-multiple-images) section for more details.
+
```
LiamConnors:
```suggestion
```

MIGRATION_GUIDE.md
```diff
@@ -88,6 +88,7 @@ f2.layout.title = 'Age and Diabetes Factors'
 f2
 ```
 
+
```
LiamConnors:
```suggestion
```

.github/workflows/build-doc.yml
```diff
@@ -0,0 +1,160 @@
...
+          python front-matter-ci.py build/html
+          python check-or-enforce-order.py build/html
+
+      - name: Upload HTML docs artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: doc-html
+          path: doc/build/html/
+
+      - name: Create GitHub App token
```
emilykl, line 58: @LiamConnors Curious why we need to create a new token as part of the workflow — is it because the token is very short-lived?

.github/workflows/build-doc.yml
```diff
@@ -0,0 +1,160 @@
...
+        uses: actions/create-github-app-token@v2
+        id: app-token
+        with:
+          app-id: ${{ vars.GRAPHING_LIBRARIES_CI_GHAPP_ID }}
+          private-key: ${{ secrets.GRAPHING_LIBRARIES_CI_GHAPP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: plotly.py-docs,graphing-library-docs
+
+      - name: Checkout plotly.py-docs (built)
+        if: github.ref_name == 'doc-prod' && github.event_name == 'push'
```
emilykl, line 69: Is there any way to chain these steps together, or nest them under a single `if`, rather than repeating the same `if` condition in every step?

.github/workflows/build-doc.yml
camdecoster: I think it has something to do with the fact that this workflow is using a GitHub App rather than an access token.

.github/workflows/build-doc.yml
```diff
@@ -0,0 +1,160 @@
...
+          source .venv/bin/activate
+          echo "${MAPBOX_TOKEN}" > python/.mapbox_token
+          make -kj8 || make -kj8
+          curl https://raw.githubusercontent.com/plotly/graphing-library-docs/master/front-matter-ci.py > front-matter-ci.py
+          curl https://raw.githubusercontent.com/plotly/graphing-library-docs/master/check-or-enforce-order.py > check-or-enforce-order.py
+          python front-matter-ci.py build/html
+          python check-or-enforce-order.py build/html
+
+      - name: Upload HTML docs artifact
+        uses: actions/upload-artifact@v4
```
camdecoster: You should use v6 if that doesn't break anything. v4 has been deprecated.

.github/workflows/build-doc.yml
```diff
@@ -0,0 +1,160 @@
...
+          cd doc
+          source .venv/bin/activate
+          uv pip uninstall plotly
+          uv pip install -e ..
+
+      - name: List installed packages
+        run: |
+          cd doc
+          source .venv/bin/activate
+          uv pip list
```
camdecoster: Is this step needed? The install step should include the version info.

.github/workflows/build-doc.yml
```diff
@@ -0,0 +1,160 @@
...
+
+      - name: Build HTML docs
+        env:
+          MAPBOX_TOKEN: ${{ secrets.MAPBOX_TOKEN }}
```
camdecoster, line 41: Do we still need a Mapbox token? Haven't we converted to using Maplibre?

.github/workflows/build-doc.yml
LiamConnors: Looks like there's a v7. Updated it. Thanks

.github/workflows/build-doc.yml
LiamConnors: Good point. Removed.

.github/workflows/build-doc.yml
LiamConnors: Pages have Maplibre examples. but because there is still some usage of Mapbox and it hasn't been removed from the package yet, some pages still have an example - https://plotly.com/python/tile-scatter-maps/#mapbox-maps

.github/workflows/build-doc.yml
LiamConnors: yes, that's it. we are using a GitHub App with short-lived tokens

.github/workflows/build-doc.yml
LiamConnors: There doesn't seem to be a way to group these steps under a single if statement

doc/requirements.txt
```diff
@@ -33,6 +33,7 @@ requests
 scikit-image==0.20.0
 scikit-learn
 scipy==1.9.1
+setuptools<=81
```
emilykl, line 36: @LiamConnors Why do we need this upper bound?

camdecoster: Maybe because package tools was removed in v82?

LiamConnors: Yeah, it's the package tools issue. it's this change https://github.com/plotly/plotly.py/pull/5511/changes from `doc-prod` but I hadn't added it to main yet. Long term solution is to upgrade `sphinx`

PR #5543, opened 2026-03-25
Fix partial autorange when range contains None elements

### LINKED ISSUE #5536: [BUG]: Setting only an upper or lower bound for axes does not work when one axis is categorical, or in Jupyter
Setting only the upper or lower bound for axes in plotly (introduced [here](https://plotly.com/python/axes/#:~:text=Setting%20only%20a%20Lower%20or%20Upper%20Bound%20for%20Range,-New)) does not work (at least in scatter plots). If one of the bounds is set to None, the range does not change at all.

Image

```
import plotly.express as px

fig = px.scatter([0,6])
fig = fig.update_yaxes(range=[0, None])
display(fig)
```

Plotly 6.6

Krishnachaitanyakc:
## Summary

Fixes #5536

When setting `range=[value, None]` or `range=[None, value]` on an axis (e.g., `fig.update_yaxes(range=[0, None])`), plotly.js should use partial autoranging to fix one bound while autoranging the other. However, the plotly.js `impliedEdits` mechanism automatically sets `autorange=false` whenever `range` is specified, which prevents partial autoranging from working.

This PR fixes the issue by explicitly setting the `autorange` property to the appropriate value when `range` contains `None` elements:

- `range=[value, None]` sets `autorange="max"` (fix lower bound, autorange upper)
- `range=[None, value]` sets `autorange="min"` (autorange lower, fix upper bound)
- `range=[None, None]` sets `autorange=True` (autorange both)
- `range=[value, value]` does not set `autorange` (no change, existing behavior)

The fix is implemented in two places in `plotly/basedatatypes.py`:
1. **`_perform_update`** -- handles `update_yaxes()`, `update_xaxes()`, `update_layout()`, and other update methods. When `autorange` is explicitly provided by the user in the same update call, the explicit value takes precedence.
2. **`__setitem__`** -- handles direct property assignment like `fig.layout.yaxis.range = [0, None]`.

## Test plan

- [x] Added 9 unit tests in `tests/test_core/test_update_objects/test_update_subplots.py` covering:
  - Upper bound `None` sets `autorange="max"`
  - Lower bound `None` sets `autorange="min"`
  - Both bounds `None` sets `autorange=True`
  - No `None` does not set `autorange`
  - Works on both xaxis and yaxis
  - Explicit `autorange` in same update call takes precedence
  - Dict ordering does not affect explicit `autorange` precedence
  - Direct property assignment works
  - `update_layout(yaxis_range=...)` works
- [x] All existing tests pass (1735 passed, 1 pre-existing failure unrelated to this change)


camdecoster:
Thanks for the PR. One of our team members will review and follow up.

emilykl:
Hi @Krishnachaitanyakc, thanks for the contribution.

I'm not totally sure this addresses the root issue, since I can see that in plotly.py 6.6.0, `autorange` is already correctly set to `"max"` in plotly.js when only a lower bound is provided in plotly.py:

Screenshot 2026-03-25 at 4 14 31 PM

That said, it's at least plausible that we might want to do special handling of  `autorange` in plotly.py.

Could you update your PR description to match the [template](https://github.com/plotly/plotly.py/blob/main/.github/pull_request_template.md?plain=1) and include screenshots/videos of the fix in action?

Take a look at [my comment here](https://github.com/plotly/plotly.py/issues/5536#issuecomment-4129332247) as well which has a tiny bit more info.

PR #5547, opened 2026-03-27
Return NotImplemented for unsupported comparisons

### LINKED ISSUE #5546: [BUG]: Figure.__eq__ returns False for unsupported comparisons
### Description

Python's [rich comparison methods](https://docs.python.org/3.14/reference/datamodel.html#object.__eq__) should return `NotImplemented` so that the RHS of a comparison is given the opportunity to implement the comparison.

In my case, I have a test snapshot fixture such that I can do `assert figure == snapshot`.

### Screenshots/Video

N/A

### Steps to reproduce

```python


class MyObject:
    def __eq__(self, other):
        print(f"MyObject.__eq__({other!r}) called")
        return NotImplemented

print(1 == MyObject())
print(go.Figure() == MyObject())
```

```
MyObject.__eq__(1) called
False
False
```

There should be a print of `MyObject.__eq__(Figure({...})) called`.

RazerM:
### Link to issue

Closes #5546

### Description of change
Updated the `__eq__` implementations to return `NotImplemented` instead of `False` when comparing against an unsupported type. This aligns with Python's data model and allows the other operand to handle the comparison.

### Demo
N/A

https://docs.python.org/3.14/reference/datamodel.html#object.__eq__

### Guidelines


emilykl:
Hi @RazerM, thanks for the contribution, and thanks for your patience awaiting review!

I'm trying to make sure I understand the consequences of this change.

It seems that by returning `NotImplemented`, we would be allowing the possibility for other objects to declare themselves equal to a `BaseFigure` (or `BasePlotlyType`) object, if they choose. 

I'm having trouble thinking of a scenario where that would be helpful. It's _possibly_ something that could be useful to us internally within plotly.py, but we haven't encountered a need for it yet to my knowledge. Do you have a scenario on your end that would benefit from this change?

RazerM:
> we would be allowing the possibility for other objects to declare themselves equal to a `BaseFigure` (or `BasePlotlyType`) object, if they choose.

Yes, but they can already do that as long as they're on the LHS. plotly should return `NotImplemented` so that equality is symmetric (`A == B` means `B == A`).

Some use cases are:

- a test helper such that I can do e.g. `assert figure == snapshot`. The snapshot object would serialize to a snapshot file if it does not exist, or if it does deserialize it and then return the comparison result. Right now I have to do `assert snapshot == figure`.
- SQL query builders etc. overload `__eq__`. E.g. imagine I have an SQLAlchemy type which can serialize figures. This should work symmetrically:

  ```
  select().where(my_figure == MyModel.figure)  # right now this is `.where(False)`!
  select().where(MyModel.figure == my_figure)
  ```

See also django/django#11808, sympy/sympy#13078 for examples in other popular projects.

emilykl:
> > we would be allowing the possibility for other objects to declare themselves equal to a `BaseFigure` (or `BasePlotlyType`) object, if they choose.
> 
> Yes, but they can already do that as long as they're on the LHS. plotly should return `NotImplemented` so that equality is symmetric (`A == B` means `B == A`).

Good point, that's a solid argument in favor. Thanks for the other examples as well. I'll approve!

PR #5548, opened 2026-03-27
Fix plotly#5381: wrong x/y positions in Sankey with isolated nodes

FredericoMendes10:
### Link to issue

Closes plotly/plotly.js#7758 

### Description of change
In Sankey diagrams with `arrangement='fixed'`, Plotly.js builds an internal
node array excluding isolated nodes. The fixed-arrangement loop maps position
array indices directly to this internal array assuming a 1:1 correspondence.
When isolated nodes appear before connected nodes in the original array,
positions shift and are applied to the wrong nodes.

Fix: in `plotly/io/_kaleido.py`, reorder per-node arrays before rendering so
that isolated nodes appear last, aligning the Plotly.js internal array with
the x/y position arrays for all connected nodes.

### Demo

https://youtu.be/y5ueoOPuKto

### Testing strategy

Added a new test `test_sankey_isolated_nodes_fixed_arrangement` that creates a
Sankey diagram with isolated nodes before connected nodes and verifies the
rendered x/y positions are correct.

N/A

### Guidelines

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Fixed
+- Fix wrong x/y positions in Sankey diagrams with isolated nodes when using `arrangement='fixed'` [[#5381](https://github.com/plotly/plotly.py/issues/5381)]
 - Update `numpy.percentile` syntax to stop using deprecated alias [[5483](https://github.com/plotly/plotly.py/pull/5483)], with thanks to @Mr-Neutr0n for the contribution!
   - `numpy` with a version less than 1.22 is no longer supported.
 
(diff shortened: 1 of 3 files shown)


emilykl:
Hi @FredericoMendes10, thanks for the contribution!

Thanks for looking into this issue. This is really helpful information:

> 1. *Isolated nodes*: Plotly.js excludes nodes with no valid links from its
       internal ``b.nodes`` array, then applies ``i.node.x[p]`` to ``b.nodes[p]``
       assuming a 1:1 index correspondence. Isolated nodes before connected nodes
       shift this correspondence, misapplying positions to the wrong nodes.
>    2. *Phantom nodes*: out-of-bounds source/target indices cause Plotly.js to
       compute ``_ = max(sources+targets) + 1``, inflating ``b.nodes`` with
       unlabelled entries that further break the index alignment.

If we make this change only in `_kaleido.py` as proposed, that will only fix the issue for static images, not interactive plots. Based on this info it seems this is really an underlying issue in plotly.js, so the issue should really be fixed directly in that library.

Would you be interested in working on a fix in plotly.js?

In any case, could you also post a comment in https://github.com/plotly/plotly.js/issues/7758 documenting what you've learned?

Thanks again for looking into this issue, it's much appreciated!

PR #5558, opened 2026-04-03
fix #5459: resolve duplicate hover entries in segmented line charts

### LINKED ISSUE #5459: [BUG]: Customdata ignored on last point of segmented line trace in hover
**Plotly version**  
plotly==6.5.0

**Python version**  
Python 3.12

**Description**  
When drawing a line as multiple separate `go.Scatter` traces (segmented line approach), the hover on the **last point** of the entire line shows customdata from the **previous segment** (or 0/NaN) instead of the customdata assigned to the last point.

**Minimal reproducible example**

```python
import pandas as pd
import numpy as np

# Test data: 5 points (4 segments)
dates = pd.date_range("2025-12-28", periods=5, freq="D")
y = [100, 110, 105, 120, 125]              # portfolio values
performance = [0.0, 10.0, 5.0, 20.0, 25.0]  # performance % for each point

# Create segments (standard segmented line approach)
x0 = dates[:-1]
x1 = dates[1:]
y0 = y[:-1]
y1 = y[1:]
perf_start = performance[:-1]  # customdata from starting point of each segment

fig = go.Figure()

for i in range(len(x0)):
    fig.add_trace(
        go.Scatter(
            x=[x0[i], x1[i]],
            y=[y0[i], y1[i]],
            mode="lines",
            line=dict(color="green" if perf_start[i] >= 0 else "red", width=4),
            customdata=[perf_start[i]],
            hovertemplate="Date: %{x}<br>Value: %{y}<br>Performance: %{customdata:.1f}%<extra></extra>",
            showlegend=False,
        )
    )

fig.update_layout(
    title="Bug: hover on last point shows customdata from previous segment",
    xaxis_title="Date",
    yaxis_title="Value",
    hovermode="x unified",
    width=800,
    height=500,
)

fig.show()

MarioSilvestre2901:
### Link to issue 
https://github.com/plotly/plotly.py/issues/5459

Closes #5459

### Description of change

When adjacent scatter traces share an endpoint, 'x unified' hover shows two entries at the same x value. Fix by splitting each chain trace into a visual-only drawing (hoverinfo='skip') and one single-point companion per data point.

Each trace owns its points up to but not including the shared last endpoint, which is instead covered by the following trace as its x[0], ensuring the companion carries the correct customdata

### Demo

https://www.youtube.com/watch?v=ZmALODtTiDw

### Testing strategy

Three unit tests were added: one guards against regression for non-sharing traces, one verifies the full chain expansion (2 drawing + 3 companion traces), and one targets the edge case where the shared-endpoint companion must carry the next trace's customdata rather than the previous trace's last point.

When I first started working on this issue, I thought the problem was related to the last point of the line. However, I later realized that it was actually related to the last point of each segment.

Using the file described in the issue, the last point’s performance will always be 0% because no performance value is being passed to Plotly for that point.

I fixed the duplicate and incorrect values for the last point of each segment, but with the file provided in the issue, the last point of the line will still always show 0%.

### Guidelines

### DIFF
--- a/plotly/basedatatypes.py
+++ b/plotly/basedatatypes.py
@@ -3288,6 +3288,117 @@ Invalid property path '{key_path_str}' for layout
         if relayout_changes:
             self._dispatch_layout_change_callbacks(relayout_changes)
 
+    @staticmethod
+    def _traces_share_endpoint(trace_curr, trace_next):
+        if trace_curr.get("type", "scatter") not in {"scatter", "scattergl"}:
+            return False
+        if trace_next.get("type", "scatter") not in {"scatter", "scattergl"}:
+            return False
+        x_curr = trace_curr.get("x")
+        x_next = trace_next.get("x")
+        # x may be None, a base64-encoded dict, or a non-sequence type
+        if x_curr is None or x_next is None:
+            return False
+        if isinstance(x_curr, dict) or isinstance(x_next, dict):
+            return False
+        try:
+            return (
+                len(x_curr) >= 2
+                and len(x_next) >= 1
+                and str(x_curr[-1]) == str(x_next[0])
+            )
+        except (KeyError, TypeError, IndexError):
+            return False
+
+    @staticmethod
+    def _build_hover_companion(trace, x, y, customdata_index):
+        companion = {
+            **{key: trace[key] for key in ("name", "hovertemplate", "line") if key in trace},
+            "type": trace.get("type", "scatter"),
+            "mode": trace.get("mode", "lines"),
+            "showlegend": False,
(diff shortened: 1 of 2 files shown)


emilykl:
Hi @MarioSilvestre2901, thanks for the contribution. We won't be implementing this change because showing two hoverdata entries for two traces is desired behavior. If you'd like to suggest a different behavior, please feel free to open an issue!

If you're interested in helping resolve #5459, take a look at the conversation in that issue. It's possible the issue is not a bug in the library itself and I want to ensure we understand the exact issue before implementing a fix.

PR #5572, opened 2026-04-13
Add `pull_request` `synchronize` trigger so workflows run when new commits are pushed to PR

emilykl:
### Description of change

Add `synchronize` to the list of `pull_request` events which trigger the workflow to run.

Also add the list of `pull_request` triggers in `build-doc.yml` for consistency (this list is the [same as the default](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request) so there's no actual change in behavior).

Also restricts the `push` event to apply _only_ to the `main` branch; otherwise, pushing to a branch in the repo which also has an open PR causes the workflow to run twice.

---

The `synchronize` event occurs when new commits are added to the head branch of a PR.

Without this trigger, the workflows will run when a PR is first opened (or reopened), but will not run again when new commits are pushed.

I had assumed the `push` event would cover this case, but it seems that event only triggers for branches within the repository, not for forks. Thus we need to use `pull_request synchronize`.

N.b. the exact meaning of each event type is very poorly documented, but [this comment](https://github.com/github/docs/issues/2257#issuecomment-3801300961) confirms the meaning of `synchronize`.

### DIFF
--- a/.github/workflows/build-doc.yml
+++ b/.github/workflows/build-doc.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - doc-prod
   pull_request:
+    types: [opened, reopened, synchronize]
 
 jobs:
   build-doc:
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -3,7 +3,7 @@ name: Build
 on:
   push:
   pull_request:
-    types: [opened, reopened]
+    types: [opened, reopened, synchronize]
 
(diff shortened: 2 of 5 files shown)


camdecoster (approved):
Good to know about `synchronize`!

emilykl:
> Good to know about `synchronize`!

Yup! I don't know why they gave it that name, but now we know.

PR #5581, opened 2026-04-20
Axis match bug

### LINKED ISSUE #5427: Bug: `make_subplots()` doesn't use shared x-axes for traces on different subplots, even when `shared_xaxes` is set to True
Description

When using `make_subplots()` with `shared_xaxes=True`, spike lines only appear on the subplot where the cursor is hovering, rather than showing across all subplots simultaneously.

## Expected Behavior

When hovering over any subplot, a vertical spike line should appear at the same x-position across **all subplots**, similar to the behavior in TradingView or other financial charting tools.

## Current Behavior

The spike line only appears on the subplot currently under the cursor. When moving to a different subplot, the spike line moves with the cursor instead of showing on all subplots.

## Code Example
```
import pandas as pd

# Create sample data
df = pd.DataFrame({
    'x': pd.date_range('2024-01-01', periods=100),
    'y1': range(100),
    'y2': [x**2 for x in range(100)]
})

# Create subplots
fig = make_subplots(
    rows=2, cols=1,
    shared_xaxes=True,
    vertical_spacing=0.02
)

fig.add_trace(go.Scatter(x=df['x'], y=df['y1'], name='Line 1'), row=1, col=1)
fig.add_trace(go.Scatter(x=df['x'], y=df['y2'], name='Line 2'), row=2, col=1)

fig.update_xaxes(
    showspikes=True,
    spikemode='across',
    spikesnap='cursor',
    spikecolor='black',
    spikethickness=1
)

fig.update_layout(hovermode='x unified')
fig.show()
```

What I've Tried

1. `hovermode='x unified'` - Only unifies hover labels, not spike lines
2. `fig.update_traces(xaxis='x')` - Shows spikes across all subplots but breaks subplot layout (all data gets squeezed into first subplot's x-axis range)
3. Various combinations of `spikemode`, `spikesnap`, and `matches='x'` - No effect

## Use Case

This feature is critical for financial/trading dashboards where users need to compare multiple indicators (price, volume, RSI, MACD, etc.) at the same point in time across multiple subplots.

richmanpoorman:
### Link to issue

Closes #5427

### Description of change

- Bug was that function marked the axis to be connected, but the trace_kwargs still had unique axes
- Change: change the keyword argument for the trace, so that when the graph is initialized, it uses the correct axis instead of the autogenerated one
- Note: The program generates a unique axis label for each subgraph, and then overwrites the label (under this fix)

### Demo

image

### Testing strategy

Used this test code to see if the line showed up; since this is a visibility test, no unit tests were provided

```
# Create sample data
df = pd.DataFrame({
    'x': pd.date_range('2024-01-01', periods=100),
    'y1': range(100),
    'y2': [x**2 for x in range(100)]
})

# Create subplots
fig = make_subplots(
    rows=2, cols=1,
    shared_xaxes=True,
    vertical_spacing=0.02
)

fig.add_trace(go.Scatter(x=df['x'], y=df['y1'], name='Line 1'), row=1, col=1)
fig.add_trace(go.Scatter(x=df['x'], y=df['y2'], name='Line 2'), row=2, col=1)

# fig.update_traces(xaxis='x2')
fig.update_xaxes(
    showspikes=True,
    spikemode='across',
    spikesnap='cursor',
    spikecolor='black',
    spikethickness=1
)

fig.update_layout(hovermode='x unified')
fig.show()
```

This is my first pull request, so please let me know if there is anything 
### Guidelines


richmanpoorman:
I just wanted to say that the label used to sync all of the axes together is last generated one, ie instead of using "x", it will use "x2" as the label for both of axes. This is because it is the last graph (the bottom one) that has the labeling on the axis. Looking at the code that was already there, this seems to be the intended outcome, but it doesn't seem intuitive to me. 

If it isn't intended, it may need to change more (including how the graph is initialized and the order which things are processed), but I didn't want to make too big of a change.

Let me know if this is right or wrong.

camdecoster:
Thanks for the PR! One of our team members will take a look.

richmanpoorman:
Hi, 
I noticed that according the documentation, shared_xaxis and shared_yaxis is supposed to make the two subplots share the axis, which I interpret as making the axes used between the subplots the same axes. My solution is to make the axes be the same one. However, some of the tests (namely the secondary_y test in the make_subplots core tests) expect all of the axes to be distinct. Should all of the axes be distinct objects for each subplot, even if they are shared? 
I hope that makes sense; let me know if there are any other questions!

richmanpoorman:
Hi, I noticed that I am failing the Ruff check on github, but passing on my computer; should I be doing something different?
image

Also I fail the build plotly.js dev build and build documentation, but I haven't changed anything there; I don't see anything special in the CONTRIBUTION.md, so if possible do you mind helping me figure out what is missing? It seems to run on my computer in the environment. Thank you!

camdecoster:
We haven't forgotten about this PR; we've just been busy. Thanks for your patience!

emilykl:
@richmanpoorman I apologize it's taken us so long to give you a proper review on this PR!

There's a lot of changed lines here and some tricky logic — if you're still interested in pursuing this fix, could you post a proposal in https://github.com/plotly/plotly.py/issues/5427 describing your approach, so we can discuss, and then proceed back to a PR? I want to make sure we understand the change.

Thank you for your patience and thank you for working on this issue.

PR #5596, opened 2026-05-15
Fix to_html responsive height by styling outer wrapper div

### LINKED ISSUE #5591: [BUG]: Python to_html output disrespects containing HTML vertical height
### Description

I want a Plotly figure inserted in an HTML page to scale responsively based on the surrounding HTML and the window resolution.  The [to_html documentation](https://plotly.com/python-api-reference/generated/plotly.io.to_html.html) specifies that the default value for default_height is 100%, but it in fact appears to be 450px even if the surrounding commands asking for a responsive layout render that conclusion both contradictory of the doc and nonsensical.  

This is related to https://github.com/plotly/plotly.js/issues/5270 and to https://community.plotly.com/t/plot-sizing-problems/1620/34

### Screenshots/Video

The top and bottom graphs should be the same, 30% of the window's vertical height, and identically responsive.  The only difference between them is that the code removes the <div> tag generated by to_html from the bottom one and it then performs as expected.  The top graph, which comes from Plotly's raw to_html output, is locked at the default 450px height.  They come from the repro code below.  

Image

### Steps to reproduce

```
import plotly.express as px
from jinja2 import Template
import os

data_canada = px.data.gapminder().query("country == 'Canada'")
fig = px.bar(data_canada, x='year', y='pop')

fig.update_layout(autosize=True, height=None, )

os.chdir(r"/path/to/scratchwork")
output_html_path=r"bug.html"
input_template_path = r"bug.jinja"

#issue to_html twice to obtain different IDs; otherwise only one will display
fig_html1 = fig.to_html(full_html=False, config=dict(responsive=True))
fig_html2 = fig.to_html(full_html=False, default_height="30vh", config=dict(responsive=True))

# fig_vh_explicit SCALES CORRECTLY
# fig_vh_in_containing_div does not scale correctly; proving the necessity of stripping the leading <div>
plotly_jinja_data = {"fig_vh_in_containing_div":'<div style="height: 30vh;">'+fig_html1+"</div>",
                     "fig_vh_explicit ":'<div style="height: 30vh;">'+fig_html2[+"</div>"}
#consider also defining the include_plotlyjs parameter to point to an external Plotly.js as described above

with open(output_html_path, "w", encoding="utf-8") as output_file:
    with open(input_template_path) as template_file:
        j2_template = Template(template_file.read())
        output_file.write(j2_template.render(plotly_jinja_data))
```

The code above requires the following file, which it calls bug.jinja

```
<!DOCTYPE html>
<html>
<head>
<meta charset="utf-8" />   <!--It is necessary to use the UTF-8 encoding with plotly graphics to get e.g. negative signs to render correctly -->
<meta name="viewport" content="width=device-width, initial-scale=1.0" />
</head>

<body>
<h1>Here's a Plotly graph!</h1>

These figs should both be 30% of the screen's vertical height, but only the second one is.
{{ fig_vh_in_containing_div }}
{{ fig_vh_explicit }}
<p>And here's some text after the graph.</p>
</body>
</html>
```

SharadhNaidu:
### Link to issue

Closes #5591

### Description of change

`to_html` wraps the figure inside an outer `<div>` that has no inline style. The inner figure div sets `style="height:100%; width:100%;"`, but because the wrapper has no height, that `100%` resolves to zero, and plotly.js then falls back to its hardcoded 450px default.

The result: the documented default of `default_height='100%'` cannot inherit from a sized parent container. Any figure placed inside a Jinja template, a Dash container with a `vh`/`%` height, or any responsive layout silently renders at 450px instead of filling its parent. This is the underlying cause of #5591 (and the long-standing plotly/plotly.js#5270). PR #5590 documented a regex-based workaround; this PR addresses the root cause.

The fix moves the requested `default_height`/`default_width` onto the outer wrapper, and changes the inner figure div to `height:100%; width:100%;` so it fills the wrapper:

```diff
- <div>
+ <div style="height:{height}; width:{width};">
          {mathjax_script}
          {load_plotlyjs}
              <div id="{id}" class="plotly-graph-div"
- style="height:{height}; width:{width};"></div>
+ style="height:100%; width:100%;"></div>
```

### Behavior matrix

| Caller's intent | Before this PR | After this PR |
| --- | --- | --- |
| Default (`height='100%'`), figure in container with `height:30vh` | Renders at 450px (collapsed wrapper) | Renders at 30vh as expected |
| Explicit pixel (`default_height=400`) | 400px | 400px (unchanged) |
| Default (`height='100%'`), no sized parent | plotly.js 450px fallback | plotly.js 450px fallback (unchanged) |
| `full_html=True` standalone page | Fills viewport | Fills viewport (unchanged) |

Pixel defaults and standalone-document use cases are byte-for-byte equivalent in observable layout. The only behavioral change is that percentage defaults now actually inherit from a sized parent, which matches the docstring.

### Testing strategy

Two regression tests added in `tests/test_io/test_html.py`:

- `test_outer_wrapper_carries_requested_dimensions` asserts the wrapper div has the percentage style and the inner figure div is set to `100%` of it.
- `test_outer_wrapper_respects_explicit_pixel_dimensions` asserts pixel defaults flow through unchanged.

Existing `test_html_deterministic` and the CDN/integrity tests are unaffected because they assert content presence, not the wrapper element.

### Notes for review

- This is a small visible structural change to `to_html` output. If maintainers prefer to gate it behind an opt-in (e.g. `responsive_layout=True`), I am happy to refactor - just let me know.
- I have not been able to reproduce a downstream consumer (Dash, JupyterLab extension, nbconvert) that depends on the wrapper being unstyled. If you know of one, please flag and I will investigate.

### Guidelines

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Use presence of `COLAB_NOTEBOOK_ID` env var to enable Colab renderer instead of testing import of `google.colab` [[#5473](https://github.com/plotly/plotly.py/pull/5473)], with thanks to @kevineger for the contribution!
 - Update tests to be compatible with numpy 2.4 [[#5522](https://github.com/plotly/plotly.py/pull/5522)], with thanks to @thunze for the contribution!
 - Add default headers to be passed in to Kaleido v1.3.0 to avoid blocked Open Street Map tiles [#5588](https://github.com/plotly/plotly.py/pull/5588)]
+- Propagate the requested `default_height`/`default_width` to the outer wrapper div produced by `to_html` so that responsive (percentage) dimensions inherit from a sized parent container instead of collapsing to the plotly.js 450px fallback [[#5591](https://github.com/plotly/plotly.py/issues/5591)]
 
 ### Updated
 - The `__eq__` method for  `graph_objects` classes now returns `NotImplemented` to give the other operand an opportunity to handle the comparison [[#5547](https://github.com/plotly/plotly.py/pull/5547)], with thanks to @RazerM for the contribution!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Use presence of `COLAB_NOTEBOOK_ID` env var to enable Colab renderer instead of testing import of `google.colab` [[#5473](https://github.com/plotly/plotly.py/pull/5473)], with thanks to @kevineger for the contribution!
 - Update tests to be compatible with numpy 2.4 [[#5522](https://github.com/plotly/plotly.py/pull/5522)], with thanks to @thunze for the contribution!
 - Add default headers to be passed in to Kaleido v1.3.0 to avoid blocked Open Street Map tiles [#5588](https://github.com/plotly/plotly.py/pull/5588)]
-- Propagate the requested `default_height`/`default_width` to the outer wrapper div produced by `to_html` so that responsive (percentage) dimensions inherit from a sized parent container instead of collapsing to the plotly.js 450px fallback [[#5591](https://github.com/plotly/plotly.py/issues/5591)]
+- Propagate the requested `default_height`/`default_width` to the outer wrapper div produced by `to_html` so that responsive (percentage) dimensions inherit from a sized parent container instead of collapsing to the plotly.js 450px fallback [[#5591](https://github.com/plotly/plotly.py/issues/5591)], with thanks to @SharadhNaidu for the contribution!
 
 ### Updated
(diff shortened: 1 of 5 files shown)


camdecoster:
Did you consider removing the outer div as an alternative solution?

SharadhNaidu:
Yeah, I did think about that — dropping the wrapper would also let the `.plotly-graph-div` inherit from its parent and is a smaller diff. The reason I went with styling the wrapper instead is that removing it changes the shape of `to_html` output: today everything sits inside one element, but without the wrapper the mathjax script, the `load_plotlyjs` script, the graph div and the trailing `<script>` would all become direct siblings of whatever container the user drops the HTML into. That felt more invasive than adding an inline style, and it could quietly break consumers that walk the DOM expecting a single root element.

Happy to switch to the removal approach if you'd rather take that route — let me know.

camdecoster (approved):
Thanks for the fix! I updated a test that was failing, but this is good to go.

Issue #5615, opened 2026-06-07
[BUG]: marker.colorbar.thicknessmode='pixels' collapses scatter-matrix layout after Plotly.restyle, while initial render is normal


ym-xie:
### Description

When initializing a Plotly scatter-matrix figure with a `splom` trace whose marker
colorbar has `marker.colorbar.thicknessmode='pixels'`, the initial render displays
normally: the scatter-matrix panels remain visible and the colorbar stays narrow.

However, if the same base figure is rendered first and the same property value is
then applied dynamically through `Plotly.restyle`, the layout collapses. The
scatter-matrix panels are squeezed into a very narrow strip on the left, while the
colorbar gradient occupies most of the figure canvas.

The setting is accepted by `plotly.py` and emitted as a valid trace property, so this
appears to be a `Plotly.restyle`/update-time layout calculation issue for `splom`
marker colorbars rather than an invalid Python-side property assignment.

### Screenshots/Video

<img width="1200" height="800" alt="Image" src="https://github.com/user-attachments/assets/3ce6bb71-f513-45fb-89cb-5b9bffb74472" />
<img width="1200" height="800" alt="Image" src="https://github.com/user-attachments/assets/ffca52e4-c383-4bba-8b53-8086447357c9" />

### Steps to reproduce

1. Install the screenshot dependencies if needed:
   `pip install plotly pandas playwright && playwright install chromium`.
2. Run this script. It saves two screenshots in the current directory:
   - `plot1.png`: Python-side initialization sets
     `marker.colorbar.thicknessmode='pixels'` before the first render. This renders
     normally.
   - `plot2.png`: the same base figure renders first, then JavaScript applies the
     same value with `Plotly.restyle`. This shows the collapsed layout.
3. Compare the two generated screenshots.

```python
import plotly
import plotly.express as px
import plotly.io as pio
import time
from playwright.sync_api import sync_playwright

print(f"Plotly version: {plotly.__version__}")


def make_fig():
    df = px.data.iris()

    fig = px.scatter_matrix(
        df,
        dimensions=["sepal_length", "sepal_width", "petal_length", "petal_width"],
        color="species",
    )

    fig.update_layout(
        xaxis={"matches": "y"},
        xaxis2={"matches": "y2"},
        xaxis3={"matches": "y3"},
        xaxis4={"matches": "y4"},
        height=900,
        width=750,
        dragmode="select",
        selections=[
[...]

PR #5637, opened 2026-07-05
Don't leave empty dict containers behind when nested properties are set to None

ahmojo:
### Link to issue

Related to #5615

### Description of change

Setting a nested property to `None` after it had been set could leave an empty compound container in the emitted figure JSON. For example, clearing `marker.colorbar.thicknessmode` left `marker: {"colorbar": {}}`. The same residue could be created when `plotly_restyle(..., None)` targeted a never-set nested path, even though no restyle message was sent.

This is inconsistent with construction, where empty compound containers are already pruned. This PR makes removal of a missing path a no-op and prunes emptied scalar compound containers while preserving list positions.

In the static-HTML reproduction from #5615, removing that residue prevents it from being included in the initial payload and avoids the subsequent JavaScript restyle collapse. This does not change plotly.js itself: a live `FigureWidget` whose browser-side state already contains `colorbar: {}` can still collapse after a later restyle.

### Demo

I isolated the Python serialization trigger using the issue's Playwright reproduction (1200x800 viewport, the same bundled plotly.js for every variant):

| Variant | Initial payload | Plot width after JS restyle | Colorbar thickness |
|---|---|---|---|
| Pristine figure | no `colorbar` key | 543px | 30px |
| Residual container from `main` | `{"colorbar": {}}` | 64px | 16290px |
| Same static-HTML repro with this PR | no `colorbar` key | 543px | 30px |

With this change, the static-HTML output matches the pristine control. The separate live-`FigureWidget` path remains a plotly.js/frontend-state issue and is outside the scope of this cleanup.

### Testing strategy

Added two regression tests in `tests/test_core/test_figure_messages/test_plotly_restyle.py`:

- Removing a never-set nested path leaves the figure unchanged and sends no restyle message.
- Clearing a nested property removes the empty scalar compound ancestors from emitted JSON.

Targeted checks:

- `uv run --with-editable . --with pytest python -m pytest tests/test_core/test_figure_messages/test_plotly_restyle.py` - 10 passed
- `uv run --with-editable . --with pytest python -m pytest tests/test_core/test_graph_objs/test_property_assignment.py` - 28 passed
- `uv run --with ruff ruff check plotly/basedatatypes.py tests/test_core/test_figure_messages/test_plotly_restyle.py` - passed

This changes serialized output for set-then-unset scalar compound properties across the library, bringing assignment and restyle cleanup in line with construction-time behavior. Compound-array/list entries remain in place as positional placeholders, and batch updates are pruned when the batch flushes.

### Guidelines


camdecoster:
Thanks for the PR! We'll take a look and follow up. We're a bit backed up at the moment, so it might take a while to get through it.

emilykl:
Hi @ahmojo, thanks for the contribution!

This is an interesting change. Pruning empty dicts left behind after a value is removed by setting it to `None` in `plotly_restyle()` does seem to make sense at first glance, although I'm not sure what the practical effects are beyond possibly this particular bug.

The reproduction example in #5637 never actually calls `plotly_restyle()` at all, since it uses the JavaScript `Plotly.restyle()` function, so I don't quite see how this change would affect the result of that particular example. Were you able to reproduce the issue in pure Python using `plotly_restyle()`?

ahmojo:
> Hi @ahmojo, thanks for the contribution!
> 
> This is an interesting change. Pruning empty dicts left behind after a value is removed by setting it to `None` in `plotly_restyle()` does seem to make sense at first glance, although I'm not sure what the practical effects are beyond possibly this particular bug.
> 
> The reproduction example in #5637 never actually calls `plotly_restyle()` at all, since it uses the JavaScript `Plotly.restyle()` function, so I don't quite see how this change would affect the result of that particular example. Were you able to reproduce the issue in pure Python using `plotly_restyle()`?

Thanks for pointing this out @emilykl. I tested this with a live FigureWidget using only Python. A direct restyle on a pristine widget works normally, but the same set-and-clear history from the issue reproduces the collapse:
```
widget = go.FigureWidget(make_fig())
display(widget)
```
Then, after it renders:
```
widget.data[1].marker.colorbar.thicknessmode = "pixels"
widget.data[1].marker.colorbar.thicknessmode = None

widget.plotly_restyle(
    {"marker.colorbar.thicknessmode": "pixels"},
    trace_indexes=[1],
)
```
The clear leaves colorbar: {} in the browser-side trace. The final restyle changes the thickness to 16,290px and shrinks the plot width from 543px to 64px.
This test also showed that my PR fixes the original static-HTML repro, but not this live FigureWidget path. I’ll adjust the description so it doesn’t overstate the scope or claim to fully close #5615.

emilykl:
Thanks @ahmojo for the follow-up!

If this fixes an issue that only comes up when injecting JavaScript into a plotly.py chart, I'd prefer to hold off on this change for now.

I'm open to merging it in the future if we can show that it fixes other more significant issues.

emilykl:
@ahmojo Per my comment I'm going to close this PR for now, since the added complexity of this change isn't worth the impact.

If you're able to identify a larger positive impact of this change, or make an argument for it, we'd be willing to consider it in the future.

PR #5657, opened 2026-07-14
test: add unit tests for image_array_to_data_uri

eeshsaxena:
## Description

`_plotly_utils.data_utils.image_array_to_data_uri` converts a numpy uint8
image array into a base64 PNG/JPG data URI. It had no direct test coverage.

This adds `tests/test_optional/test_utils/test_data_utils.py` covering:

- greyscale (2D), RGB and RGBA arrays each producing a PNG data URI
- `ValueError` on out-of-range compression (`-1` and `10`)
- `ValueError` on an invalid image shape
- `ValueError` when requesting `jpg` with a non-PIL backend

The tests use the `pypng` backend so they do not require Pillow. No source
changes.

## Verification

`pytest tests/test_optional/test_utils/test_data_utils.py` - 7 passed.
`ruff check` / `ruff format --check` clean.


emilykl (changes requested):
@eeshsaxena Thanks for the PR!

Can you move this test file to `tests/test_plotly_utils/test_data_utils.py` so that it lives with the rest of the `_plotly_utils` tests?

Could you also update the first 3 tests to verify the contents of the whole string, in addition to the prefix? This could be done by comparing against a hard-coded string, _or_ decoding the string and extracting the values and comparing them to the input. I'd also be okay with removing those tests entirely for now; up to you. But leaving them in gives a false impression of coverage.

Otherwise LGTM.

eeshsaxena:
Thanks @emilykl - both points addressed:

- Moved the file to `tests/test_plotly_utils/test_data_utils.py`.
- Rather than drop the first three tests, I gave them teeth: they now decode the base64 payload and compare the pixels against the input array, so they actually cover the encoding rather than just the prefix. Each one also asserts the PNG colour mode matches the input shape (`greyscale` for 2-D, `alpha` for the 4-channel case), and they use non-trivial pixel values instead of `zeros`.

I decoded with the vendored `_plotly_utils.png.Reader` rather than Pillow, so the file keeps its original property of not requiring Pillow and pulls in no new test dependency.

7 tests pass and `black --check` is clean.

emilykl (approved):
Thanks @eeshsaxena for the quick follow-up and for the tests. Looks good to me.

PR #5662, opened 2026-07-16
fix: handle 3-digit shorthand hex in hex_to_rgb

### LINKED ISSUE #5661: [BUG]: hex_to_rgb misparses 3-digit shorthand hex colors (#FFF -> (15,15,15))
### Description

`hex_to_rgb()` assumes the hex string is always 6 digits. It computes the channel width as `len(value) // 3` and slices the string into equal thirds:

```python
def hex_to_rgb(value):
    value = value.lstrip("#")
    hex_total_length = len(value)
    rgb_section_length = hex_total_length // 3
    return tuple(
        int(value[i : i + rgb_section_length], 16)
        for i in range(0, hex_total_length, rgb_section_length)
    )
```

For the 3-digit CSS shorthand form this takes each digit as a whole channel value instead of expanding it (CSS says `#FFF` == `#FFFFFF`, i.e. each digit is doubled). So:

```python
hex_to_rgb("#FFF")     # -> (15, 15, 15)     near-black
hex_to_rgb("#FFFFFF")  # -> (255, 255, 255)  white
```

Those are the same colour in CSS, but produce wildly different results. Nothing raises - the wrong colour is returned silently.

This matters because plotly itself treats `#FFF` as a perfectly valid colour:

```python
go.Figure(go.Scatter(x=[1], y=[1], marker=dict(color="#FFF")))  # accepted, renders white
```

so users can reasonably pass shorthand hex, and the figure factories that route colours through `hex_to_rgb` then misinterpret them (`figure_factory/_annotated_heatmap.py:155`, `_ternary_contour.py:248`, `_trisurf.py:120`, `matplotlylib/mpltools.py:136`).

**Visible consequence:** `create_annotated_heatmap` picks the annotation font colour from the background luminance via `should_use_black_text()`. With shorthand hex, a white background is read as `(15, 15, 15)`, so it chooses **white text on a white background** and the annotations become invisible:

| colorscale | resulting annotation font colour |
| --- | --- |
| `[[0, "#FFF"], [1, "#FFF"]]` | `#FFFFFF` (invisible on white) |
| `[[0, "#FFFFFF"], [1, "#FFFFFF"]]` | `#000000` (correct) |

A fix would be to expand the 3-digit form before parsing (and probably reject lengths that are neither 3 nor 6, rather than silently splitting into thirds - e.g. a 4 or 5 character string currently produces nonsense too).

### Steps to reproduce

```python
from plotly.colors import hex_to_rgb

print(hex_to_rgb("#FFF"))     # (15, 15, 15)      <- expected (255, 255, 255)
print(hex_to_rgb("#FFFFFF"))  # (255, 255, 255)

z = [[1, 2], [3, 4]]

fig_short = ff.create_annotated_heatmap(z, colorscale=[[0, "#FFF"], [1, "#FFF"]])
print({a.font.color for a in fig_short.layout.annotations})   # {'#FFFFFF'} -> white on white

fig_long = ff.create_annotated_heatmap(z, colorscale=[[0, "#FFFFFF"], [1, "#FFFFFF"]])
print({a.font.color for a in fig_long.layout.annotations})    # {'#000000'} -> correct
```

Both colorscales describe the same white background, but only the 6-digit form produces readable annotations.

Happy to open a PR for this if it's welcome.

### Versions

- plotly 6.7.0
- Python 3.14.3
- Windows 11

genrichez:
Closes #5661.

`hex_to_rgb` assumed the input was always 6 digits. 3-digit shorthand like `#FFF` was parsed as `(15, 15, 15 )` instead of `(255, 255, 255)`.

Fix: expand 3-char codes to 6-char before slicing (e.g. `F` → `FF`). Applied to both copies in `_plotly_utils/colors` and `plotly/matplotlylib/mpltools`.

Added tests for the shorthand case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Fixed
+- Fix `hex_to_rgb` parsing of 3-digit shorthand hexadecimal colors such as `#FFF` [[#5662](https://github.com/plotly/plotly.py/pull/5662)], with thanks to @genrichez for the contribution!
+
 
 ## [6.9.0] - 2026-07-09
 
--- a/_plotly_utils/colors/__init__.py
+++ b/_plotly_utils/colors/__init__.py
@@ -752,9 +752,16 @@ def hex_to_rgb(value):
     """
     Calculates rgb values from a hex color code.
 
-    :param (string) value: Hex color string
+    :param (string) value: Hex color string. May be a full 6-character code
+        or a 3-character shorthand code.
 
     :rtype (tuple) (r_value, g_value, b_value): tuple of rgb values
+
+    Example:
+
+        '#FFFFFF' --> (255, 255, 255)
+        '#FFF'    --> (255, 255, 255)
+
     """
     value = value.lstrip("#")
     if len(value) == 3:
--- a/_plotly_utils/colors/__init__.py
+++ b/_plotly_utils/colors/__init__.py
@@ -757,6 +757,8 @@ def hex_to_rgb(value):
     :rtype (tuple) (r_value, g_value, b_value): tuple of rgb values
     """
     value = value.lstrip("#")
+    if len(value) == 3:
+        value = "".join(c * 2 for c in value)
     hex_total_length = len(value)
     rgb_section_length = hex_total_length // 3
     return tuple(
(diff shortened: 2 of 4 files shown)


camdecoster:
Thanks for the PR! I'll take a look and follow up.

camdecoster (commented):
This looks great! Before I approve, could you please update the changelog? Also, I went ahead and combined both functions into one since they did the same thing.

genrichez:
> This looks great! Before I approve, could you please update the changelog? Also, I went ahead and combined both functions into one since they did the same thing.

Thanks, Cameron! 
I've added the changelog entry

camdecoster (approved):

PR #5668, opened 2026-07-19
Add add_abline for drawing a line from a slope and an intercept

### LINKED ISSUE #3166: Implement the equivalent of abline (plot line using slope/intercept only) to add line to graph
Currently there seems to be no way to add a line to an existing graph (e.g. as a new trace) based on slope and intercept and without the necessity to find out the actual ranges of x and y axes of the figure. 
All examples I have seen so far are based on using existing x values of dots in a scatter plot, calculating the corresponding y values for the line and adding those line segments in a new scatter plot trace.

But this has two huge disadvantages:
* the y values have to get calculated for those x values (which may be a lot) or x values need to get selected
* if a lot of segmets are calculated, the graph complexity increases a lot as well

But plotting a line based on just slope and intercept is probably the second most easy thing to do, so there should be a way to add a line that way. 

The basic signature of the corresponding method or class could be similar to the corresponding R function:
`ABline(a=intercept, b=slope, h=yforhorizontallines, v=xforverticallines)`

kirthi-b:
### Link to issue

Closes #3166

### Description of change

Adds `Figure.add_abline(slope=..., intercept=...)`, which draws a straight line from a slope and an intercept the way matplotlib's `axline` or R's `abline` do. The signature and behavior follow `add_hline` and `add_vline`, including `row`, `col`, `exclude_empty_subplots`, `annotation`, and the `annotation_*` keywords, and extra kwargs are forwarded to `add_shape`.

### Demo

add_abline demo: a 1:1 reference line and a least-squares fit line drawn with add_abline

Left panel is a 1:1 reference line on a predicted vs actual plot, right panel is a least squares fit drawn from its own slope and intercept. Both use the annotation keywords, and both are targeted at a specific subplot with `row` and `col`. The code that produces this figure is in the testing instructions below.

### Testing strategy

New tests in `tests/test_optional/test_autoshapes/test_add_abline.py` cover basic usage, a custom slope and intercept, the priority of an explicit axis range over a data derived range, the fallback when there is no data and no range, subplot targeting with `row` and `col`, `exclude_empty_subplots`, and the annotation keywords.

To verify locally:

1. Check out this branch and install in editable mode.
2. Run the automated tests: `python -m pytest tests/test_optional/test_autoshapes/ -q` (the suite passes, 50 tests, and `tests/test_core/` is unaffected, 462 passed / 1 skipped).
3. To see it in action, run the snippet below and confirm both panels show the line with its annotation, each confined to its own subplot:

```python
import numpy as np

rng = np.random.default_rng(7)
x = np.linspace(0, 10, 60)
y = 0.8 * x + 1.2 + rng.normal(0, 1.0, x.size)
slope, intercept = np.polyfit(x, y, 1)

fig = make_subplots(rows=1, cols=2, subplot_titles=("Predicted vs actual", "Least-squares fit"))
fig.add_trace(go.Scatter(x=x, y=y, mode="markers"), row=1, col=1)
fig.add_abline(slope=1, intercept=0, row=1, col=1,
               line_color="#d1495b", line_dash="dash",
               annotation_text="y = x", annotation_position="top left")

fig.add_trace(go.Scatter(x=x, y=y, mode="markers"), row=1, col=2)
fig.add_abline(slope=slope, intercept=intercept, row=1, col=2,
               line_color="#118a7e", line_width=3,
               annotation_text=f"y = {slope:.2f}x + {intercept:.2f}",
               annotation_position="bottom right")

fig.show()
```

One behavior is worth calling out, since it came up in the issue discussion about "infinite" lines. `add_hline` and `add_vline` stay pinned to the plot edges through pan and zoom because they use plotly's axis `domain` reference for one dimension, which plotly.js recalculates against the visible range on every render. A line with an arbitrary slope cannot use that trick, because domain references only work when one axis is held constant, not when both move together.

So `add_abline` computes its endpoints once, at the moment it is called, from the axis's current range, or from the data already plotted if no range has been set. Those endpoints are then ordinary fixed data coordinates like any shape added with `add_shape`. If you pan or zoom afterwards, or add data outside the original range, the line will not move or extend to follow. This is documented in the docstring and on the shapes doc page, and calling `add_abline` again picks up the new range.

Annotation placement reuses the existing vline positioning code in `shapeannotation.py`, which was already written generically enough to handle a slanted line's endpoints (its own comment says as much).

### Guidelines

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Added
+- Add `add_abline` method to `Figure` for drawing a straight line from a slope and an intercept, alongside the existing `add_hline`/`add_vline`/`add_hrect`/`add_vrect` convenience methods [[#3166](https://github.com/plotly/plotly.py/issues/3166)]
+
 ### Fixed
 - Fix `hex_to_rgb` parsing of 3-digit shorthand hexadecimal colors such as `#FFF` [[#5662](https://github.com/plotly/plotly.py/pull/5662)], with thanks to @genrichez for the contribution!
 
--- a/doc/python/shapes.md
+++ b/doc/python/shapes.md
@@ -132,6 +132,28 @@ fig.update_shapes(dict(xref='x', yref='y'))
 fig.show()
 ```
 
+#### Diagonal Line From a Slope and an Intercept
+
+To draw a straight line from a slope and an intercept rather than from
+endpoint coordinates, use `add_abline`, which computes the line's endpoints
+from the current x-axis range (or from the range of the data already plotted,
+if the x-axis range has not been set explicitly) at the time it is called.
+
+```python
+import plotly.graph_objects as go
+
+fig = go.Figure()
+fig.add_trace(go.Scatter(x=[0, 1, 2, 3], y=[0, 1, 4, 9]))
+fig.add_abline(slope=3, intercept=-2, line=dict(color="MediumPurple", dash="dot"))
+fig.show()
+```
+
+Note that, unlike `add_hline` and `add_vline`, the line added by
+`add_abline` is not re-anchored to the axes: its endpoints are ordinary data
+coordinates fixed at the time the line is added, so panning or zooming the
+figure afterwards, or adding data that falls outside the original range,
+will not extend or move it.
+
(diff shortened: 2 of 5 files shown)


camdecoster:
Thanks for the PR! Could you please update the PR description to use the template? We need to see screenshots/videos along with testing instructions.

PR #5671, opened 2026-07-21
fix(io): force UTF-8 for write_json/read_json file I/O

### LINKED ISSUE #5665: [BUG]: write_json / read_json raise UnicodeEncodeError on Windows for non-cp1252 characters
### Description

`pio.write_json` raises `UnicodeEncodeError` on Windows when a figure's text contains characters outside the cp1252 default codec. `pio.read_json` has the matching bug — it reads without specifying an encoding, so a UTF-8 JSON file is decoded with cp1252, mangling non-ASCII text or raising `UnicodeDecodeError`.

This works on macOS/Linux only because their default encoding is already UTF-8.

### Steps to reproduce

```python
import plotly.io as pio

fig = go.Figure(layout_title_text="μ 中文")   # Greek mu + CJK
pio.write_json(fig, "fig.json")
# UnicodeEncodeError: 'charmap' codec can't encode character 'μ' in position ...: character maps to <undefined>
```

**Environment:** Windows, `locale.getpreferredencoding(False)` → `cp1252`, orjson installed.

### Notes

Two conditions both have to hold:

1. The figure text contains a character outside cp1252. Common Latin-1 characters (`é`, `ö`, `°`) are *inside* cp1252 and do not raise — it takes Greek (`μ`), CJK, or emoji to trigger it.
2. orjson is installed (the default engine when available). orjson emits real UTF-8, so non-ASCII characters reach `write_text` unchanged. The pure-Python `json` engine uses `ensure_ascii=True`, escaping everything to `\uXXXX` ASCII, so it never trips the codec.

`write_html` already handles this correctly with `encoding="utf-8"` (see #3898). The JSON I/O path was missed.

Solaris-star:
## Description

`pio.write_json` / `pio.read_json` use `Path.write_text` / `Path.read_text` without an encoding. On Windows the default is often **cp1252**, so non-ASCII figure text (e.g. `μ 中文`) raises `UnicodeEncodeError` on write or mangles/fails on read. macOS/Linux work only because their default is already UTF-8.

## Fix

Always pass `encoding="utf-8"` for both write and read path I/O.

## Test

```text
pytest tests/test_io/test_to_from_json.py::test_write_read_json_utf8_non_ascii -q
```

## Linked Issue

Closes #5665

### DIFF
--- a/plotly/io/_json.py
+++ b/plotly/io/_json.py
@@ -295,7 +295,9 @@ The 'file' argument '{file}' is not a string, pathlib.Path object, or file descr
     else:
         # We previously succeeded in interpreting `file` as a pathlib object.
         # Now we can use `write_bytes()`.
-        path.write_text(json_str)
+        # Explicit UTF-8: Windows default encodings (e.g. cp1252) raise
+        # UnicodeEncodeError for non-ASCII figure text (#5665).
+        path.write_text(json_str, encoding="utf-8")
 
 
 def from_json_plotly(value, engine=None):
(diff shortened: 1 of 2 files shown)


emilykl:
Thanks @Solaris-star for the PR. This issue has been addressed by PR https://github.com/plotly/plotly.py/pull/5633 which was opened first, so I'm going to close this PR.

Issue #5081, opened 2025-03-12
Fully remove all mapbox references (next major release)


emilykl:
- Remove from `graph_objects`:
  - `go.Choroplethmapbox`
  - `go.Densitymapbox`
  - `go.Scattermapbox`

- Remove from `px`:
  - `px.choropleth_mapbox`
  - `px.density_mapbox`
  - `px.scatter_mapbox`
  - `px.line_mapbox`

- Remove all related tests
- Remove `scattermapbox` key from template

emilykl:
Addressed by https://github.com/plotly/plotly.py/pull/5673.

Will close this issue once v7.0 is released and the `v7.0` branch is merged to `main`.

Issue #4824, opened 2024-10-23
Plotly doesn't recognise modern RGB syntax (+ other unexpected behaviour)


tpvasconcelos:
Plotly does not support [modern](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb) `rgb(R G B / A)` CSS syntax. It seems to prefer the legacy versions of `rgb(R, G, B)` or `rgba(R, G, B, A)`, but it still does not support these completely (as can be seen in the examples bellow).

This can be experienced in a few different ways, such as:

1. The alpha/opacity value is ignored by Plotly even when using the legacy `rgb(R, G, B, A)` syntax.

```pycon
>>> go.Figure().add_trace(go.Scatter(x=[1,2], y=[1,2], line_color="rgb(0, 0, 0, 0.1)")).show()
(missing opacity from plot)
```

2. The alpha/opacity value is also ignored when using the legacy notation with percentages

```pycon
>>> go.Figure().add_trace(go.Scatter(x=[1,2], y=[1,2], line_color="rgba(0, 0, 0, 10%)")).show()
(missing opacity from plot)
```

3. Modern space-separated syntax raises an exception

```pycon
>>> go.Figure().add_trace(go.Scatter(x=[1,2], y=[1,2], line_color="rgb(0 0 0)")).show()
Traceback (most recent call last):
...
ValueError: 
    Invalid value of type 'builtins.str' received for the 'color' property of scatter.line
        Received value: 'rgb(0 0 0)'
    The 'color' property is a color and may be specified as:
      - A hex string (e.g. '#ff0000')
      - An rgb/rgba string (e.g. 'rgb(255,0,0)')
      - An hsl/hsla string (e.g. 'hsl(0,100%,50%)')
      - An hsv/hsva string (e.g. 'hsv(0,100%,100%)')
      - A named CSS color:
...
```

4. The same can be seen when using the modern rgb syntax with an alpha-channel

```pycon
>>> go.Figure().add_trace(go.Scatter(x=[1,2], y=[1,2], line_color="rgb(0 0 0 / 0.5)")).show()
Same exception as above
```

---

[...]

PR #5673, opened 2026-07-22
Update plotly.js to `v4.0.0-rc.0`

emilykl:
### Description of change
- Update plotly.js to `v4.0.0-rc.0`
- Run codegen and regenerate Jupyter js bundles
- Remove all references to `mapbox` in Plotly Express
- Update tests to remove `mapbox` references

Note: The failing `plotly.js dev build` CI job can be ignored. That job runs the tests after pulling `dist/plotly.js` from the plotly.js `master` branch, so naturally it will fail since the plotly.js v4.0 changes are not yet on `master`.

The failing docs build can also be ignored; it will fail until the docs are updated to remove `mapbox` code, which will be in a separate PR.

### Notes
Closes https://github.com/plotly/plotly.py/issues/4824 due to the [color library update](https://github.com/plotly/plotly.js/pull/7536) in plotly.js v4.0.0.
Closes https://github.com/plotly/plotly.py/issues/5081 by removing all Mapbox references.


_plotly_utils/basevalidators.py
camdecoster, line 1: `ColorValidator` will need to be updated to remove the references and regex checks for `hsv` since that got removed. The regex patterns will also need to be updated to match the new color specifier support.

camdecoster (commented):
- Could we update the docs separately to keep things tidy?
- I think we can remove the Stream deprecation warning in **_deprecations.py**
- I still see a lot of references to `linkText`/`link_text` and `showLink`/`show_link`. Could those be removed?
- `ColorValidator` will need to be updated per the v4 changes

emilykl:
> * Could we update the docs separately to keep things tidy?

Yes, as long as you're OK with the docs build failing on this PR. I'll split the docs changes into a separate PR.

> * I think we can remove the Stream deprecation warning in **_deprecations.py**
> * I still see a lot of references to `linkText`/`link_text` and `showLink`/`show_link`. Could those be removed?
> * `ColorValidator` will need to be updated per the v4 changes

👍 will do, thanks

tests/test_plotly_utils/validators/test_color_validator.py
```diff
@@ -28,26 +28,58 @@ def validator_aok_colorscale():
     )
 
 
+VALID_COLORS = [
```
emilykl, line 31: @camdecoster I tried to be pretty comprehensive here, but please suggest any formats I've missed. Likewise for the invalid colors list.

camdecoster (approved):
Looks good!

PR #5698, opened 2026-08-11
Convert matplotlib stairs plots to plotly step lines

robertoffmoura:
mpl_to_plotly currently drops plt.stairs plots entirely. Matplotlib's stairs creates a StepPatch artist, which the exporter treats as a generic path, and the renderer only handles bar-chart paths, so the stairs plot is skipped with the warning "I found a path object that I don't think is part of a bar chart. Ignoring." The converted figure has zero traces.

Fix: draw_path now recognizes StepPatch artists and draws them as a line trace:
- the StepPatch's path vertices already describe the horizontal/vertical step outline, so they're used directly (consecutive duplicates removed)
- line styling (color, width, dash) comes from the patch's edge properties

Snippet to reproduce:
```
import matplotlib
matplotlib.use("Agg")
import numpy as np
import plotly.tools as tls

x = np.linspace(0, 10, 20)

fig, ax = plt.subplots()
ax.stairs(np.sin(x))
fig.savefig("stairs_mpl.png")

p = tls.mpl_to_plotly(fig)
p.write_image("stairs_plotly.png") 
```

| matplotlib | plotly before | plotly after |
|---|---|---|
| stairs_mpl | stairs_plotly | stairs_plotly_after |

### DIFF
--- a/plotly/matplotlylib/renderer.py
+++ b/plotly/matplotlylib/renderer.py
@@ -9,6 +9,7 @@ with the matplotlylib package.
 
 import warnings
 
+import matplotlib.patches as mpatches
 import plotly.graph_objs as go
 from plotly.matplotlylib.mplexporter import Renderer
 from plotly.matplotlylib import mpltools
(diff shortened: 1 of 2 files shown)


robertclaus:
@robertoffmoura just a heads up that we saw your PRs come in. We're still planning to review them when we have a chance!

PR #5702, opened 2026-08-11
Render matplotlib path collections (violin plots, pcolor, stems, etc.) in plotly

robertoffmoura:
mpl_to_plotly silently drops several common plot types. Violin plots, pcolor, event plots, stack plots, fill_between and stem plots all produce PolyCollection/LineCollection artists in data coordinates, which the renderer skipped with "Dang! That path collection is out of this world...". The converted figure ends up with zero traces (except for stem, which gets partially drawn).

Fix: data-coordinate path collections are now drawn:
- collections with face colors (violin bodies, pcolor cells, stacked areas, fills) → filled polygons (fill="toself")
- collections without face colors (contour lines, stem lines) → plain line traces
- per-path colors/widths handled for collections that mix them; empty color arrays and degenerate paths skipped

Before: plots converts to 0 traces (empty figures).
After: plots render correctly.

Snippet to reproduce:
```
import matplotlib
matplotlib.use("Agg")
import numpy as np
import plotly.tools as tls

x = np.linspace(-3, 3, 30)
X, Y = np.meshgrid(x, x)
t = np.linspace(0, 2 * np.pi, 50)

cases = {
    "violinplot":   lambda: plt.violinplot(np.random.randn(100, 3)),
    "pcolor":       lambda: plt.pcolor(X, Y, np.sin(X) * np.cos(Y)),
    "eventplot":    lambda: plt.eventplot([np.random.randn(20) for _ in range(5)]),
    "stackplot":    lambda: plt.stackplot(np.arange(10), np.random.rand(10), np.random.rand(10), np.random.rand(10)),
    "fill_between": lambda: plt.fill_between(t, np.sin(t), np.cos(t)),
    "stem":         lambda: plt.stem(np.linspace(0, 2 * np.pi, 20), np.sin(np.linspace(0, 2 * np.pi, 20))),
}

for name, build in cases.items():
    fig, ax = plt.subplots()
    build()
    fig.savefig(f"{name}_mpl.png")

    p = tls.mpl_to_plotly(fig)   # 0 traces for all of these before the fix (except stem, which had some traces)
    p.write_image(f"{name}_plotly.png")

    print(f"{name:13s} -> {len(p.data)} traces")
```

| Plot type | Converted plotly figure |
|---|---|
| violinplot (12 traces) | violinplot_plotly |
| pcolor (900 traces) | pcolor_plotly |
| eventplot (100 traces) | eventplot_plotly |
| stackplot (3 traces) | stackplot_plotly |
| fill_between (1 trace) | fill_between_plotly |
| stem (22 traces) | stem_plotly |


plotly/matplotlylib/renderer.py
camdecoster, line 1: Could you run the ruff formatter on this file?

plotly/matplotlylib/renderer.py
```diff
@@ -522,6 +543,44 @@ def draw_path_collection(self, **props):
                 "collections linked to 'data' coordinates"
             )
 
+    def _draw_filled_path_collection(self, props):
+        """Draw a path collection (e.g. violin plot bodies) as filled polygons."""
+        facecolors = mpltools.convert_rgba_array(props["styles"]["facecolor"])
+        edgecolors = mpltools.convert_rgba_array(props["styles"]["edgecolor"])
+        linewidths = mpltools.convert_linewidth_array(props["styles"]["linewidth"])
+        alpha = props["styles"]["alpha"]
```
camdecoster: This adds the opacity twice. `convert_rgba_array` already adds the alpha value. We don't handle this consistently right now, but I think this should be removed to avoid making the plot _too_ transparent.

plotly/matplotlylib/renderer.py
```diff
@@ -14,6 +14,24 @@
...
+    Colors already exported by the mplexporter (hex or rgba strings) are
+    passed through unchanged.
+    """
+    if isinstance(color, str):
+        return "rgba(0,0,0,0)" if color == "none" else color
+    if isinstance(color, (list, tuple)) and all(
+        isinstance(c, str) for c in color
+    ):
+        return [_export_color(c) for c in color]
+    bgcolor = export_color(color)
```
camdecoster: Should this be `_export_color`? If this should be `export_color`, you'll need to import it from **plotly/matplotlylib/mplexporter/utils.py**.

plotly/matplotlylib/renderer.py
```diff
@@ -522,6 +543,44 @@ def draw_path_collection(self, **props):
...
+        def per_path(colors, i, default):
+            if isinstance(colors, str):
+                return colors
+            if colors is None:
+                return default
+            try:
+                n = len(colors)
+            except TypeError:
+                return colors
+            return colors[min(i, n - 1)] if n else default
```
camdecoster: If `i` is greater than or equal to `n`, you'll get the same color for every call to `per_path`. To match the existing matplotlib behavior for the case of path length being greater than colors length, you could switch this logic a bit. This likely wouldn't happen much in practice, but this would cover us if it did come up.

```suggestion
            return colors[i % n] if n else default
```

plotly/matplotlylib/renderer.py
```diff
@@ -522,6 +543,44 @@ def draw_path_collection(self, **props):
...
+
+        for i, (verts, codes) in enumerate(props["paths"]):
+            facecolor = per_path(facecolors, i, "rgba(0,0,0,0)")
+            edgecolor = per_path(edgecolors, i, "rgba(0,0,0,0)")
+            linewidth = per_path(linewidths, i, 0)
+            self.plotly_fig.add_trace(
+                go.Scatter(
+                    x=[v[0] for v in verts],
```
camdecoster: This line needs to handle the case where x values are dates. You can see how this is handled for the same situation on [line 456](https://github.com/plotly/plotly.py/pull/5702/changes#diff-01c9ea8b6c3c4480d38c2aa9ba62d6ae3e0ef4b8d7cc9f6fd09d1b194d381fa8R456). This is actually done in a few places. If you want to extract a helper function, this could be used in all those locations.

camdecoster (changes requested):
Thanks for the PR! Overall this is in good shape but needs a couple of changes. Could you also please add a changelog entry?

plotly/matplotlylib/renderer.py
robertoffmoura: Done

plotly/matplotlylib/renderer.py
robertoffmoura: Done

plotly/matplotlylib/renderer.py
robertoffmoura: Yes. Sorry, this part leaked from another PR I was working on. The branch that called `bgcolor = export_color(color)` was never actually used in here (that's why tests didn't fail even without the import). I removed it now.

plotly/matplotlylib/renderer.py
robertoffmoura: Done. Removed the opacity field.

plotly/matplotlylib/renderer.py
robertoffmoura: Done

robertoffmoura:
Thanks! I addressed your comments and added a changelog entry.

camdecoster (dismissed):
Thanks for the fix!

camdecoster (approved):

PR #5731, opened 2026-09-14
Fix dependabot config, and add GitHub actions to config

emilykl:
Follow-up to https://github.com/plotly/plotly.py/pull/5597

#5597 added a Dependabot config for updating `uv.lock` (following the [example config file](https://docs.astral.sh/uv/guides/integration/dependabot/) in the uv docs) but put the file at the root, instead of nested under `.github/` as it should be.

This PR moves the file to the right place, and also adds a config block for updating GitHub Actions pinned versions.

### Guidelines


.github/dependabot.yml
```diff
@@ -0,0 +1,33 @@
+
+
+
```
camdecoster:
```suggestion
```

camdecoster (dismissed):

PR #5742, opened 2026-09-15
Chore: v7.1.0

KoolADE85:
Updates for release v7.1.0

`uv.lock` appears to be a large change but as far as I can tell it's just reordering lines.


emilykl (approved):
🎉

PR #5756, opened 2026-09-16
Fix create_hexbin_map truncating aggregated values to integers

### LINKED ISSUE #4632: figure_factory create_hexbin_mapbox ignored agg_func
As the title states, the aggregation function is entirely ignored and it does not make any difference whether you insert np.mean, np.min or np.max. Using `plotly==5.22.0`

Example: 

1. Generate data 

```python
import pandas as pd
import numpy as np

# Mock GeoDataFrame with latitude, longitude, and value columns
np.random.seed(0)
num_points = 1000
data = {
    'lat': np.random.uniform(40, 45, num_points),
    'lon': np.random.uniform(-75, -70, num_points),
    'value': np.random.uniform(0, 1, num_points)
}
df = pd.DataFrame(data)
df
```

2. Plot and check vals

```python 

fig = ff.create_hexbin_mapbox(
    data_frame=df, lat="lat", lon="lon",
    nx_hexagon=50,  # Decrease the size of hexagons
    opacity=0.5, labels={"color": "value"},
    color_continuous_scale="Viridis",
    agg_func=np.min,  # or np.max aggregation
    show_original_data=True,
    original_data_marker=dict(size=1.1, opacity=0.6, color="deeppink")
)

# Extract the hexbin data
hexbin_data = fig.data[0]

# Check the hexbin values
print("Hexbin values (z):", hexbin_data.z)

# Update the text of each hexagon to display the maximum value
hexbin_data.hovertemplate = 'Value: %{z}<extra></extra>'

# Update the layout to use OSM tiles
fig.update_layout(
    mapbox_style="open-street-map",
    height=800  # Set the desired height
)

fig.show()

```

Output for np.min, np.mean and np.max is identical:

`Hexbin values (z): [0. 0. 0. ... 0. 1. 0.]`

Hence, the plot does not change.

Belagum:
### Link to issue

Closes #4632

### Description of change

`create_hexbin_map` (formerly `create_hexbin_mapbox`) casts the aggregated `color` column to `Int64`, so any non-integer result of `agg_func` is truncated. With a float column in `[0, 1]` that means `np.min`, `np.mean` and `np.max` all produce the same map full of zeros, which is what the issue reports. The cast came in with the narwhals port (#4790, `1aa5163c2`); before that the column went through `pd.to_numeric` and kept its float values. This removes the cast so aggregates keep their values again.

### Demo

60 random points in a 1°×1° box, `nx_hexagon=2`, `color="value"` uniform in `[0, 1]`:

```
# before
  min: z = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
 mean: z = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
  max: z = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]

# after
  min: z = [0.183, 0.888, 0.562, 0.125, 0.554, 0.274, 0.823, 0.079, 0.259, 0.059, 0.01]
 mean: z = [0.183, 0.916, 0.562, 0.464, 0.557, 0.621, 0.823, 0.581, 0.503, 0.728, 0.402]
  max: z = [0.183, 0.944, 0.562, 0.848, 0.56, 0.976, 0.823, 0.995, 0.958, 0.959, 0.865]
```

### Testing strategy

Added `test_aggregation_keeps_float_values` to `TestHexbinMap`: the mean of a constant `0.5` column has to come back as `0.5` (it was `0`), and `np.min`/`np.max` over a spread of values have to stay in range and differ from each other. The first assertion fails without the fix. The existing `test_aggregation` and `test_build_dataframe` pass unchanged. The four `TestTernarycontour` tests in the same file fail for me on `main` as well because I don't have scikit-image installed.

Without a `color` column the hexagons hold point counts; those are floats again now (as they were in 5.x), which doesn't change how they are shown, since plotly.js formats `5` and `5.0` the same way in hover labels and colorbars.

### Guidelines


camdecoster:
Thanks for the PR! Do you know if there are any negative consequences to removing that cast?

Belagum:
> Thanks for the PR! Do you know if there are any negative consequences to removing that cast?

Nothing significant that I could spot. The sole difference pops up in the uncolored case: point counts end up as `float64` rather than `Int64` (so `[2.0, 3.0, ...]`, not `[2, 3, ...]`). That actually matches what 5.x did, since `pd.to_numeric` left the float grid from `_compute_hexbin` untouched. It has zero impact on rendering: plotly.js treats both as plain js numbers, meaning hover tooltips and colorbar ticks still show `2` instead of `2.0`, and parameters like `range_color` or `color_continuous_midpoint` behave identically. The only practical distinction is that `z` serializes as an `f8` typed array rather than `i2`, i.e. 8 bytes per bin instead of 2, which is negligible for hexbin datasets. The current `test_build_dataframe` still validates point counts against a sum of ones across the entire trace dict (dtypes included) and passes without issue

Issue #5028, opened 2025-02-11
erratic result from colorscale validator test


gvwilson:
The test `tests/test_plotly_utils/validators/test_colorscale_validator.py::test_acceptance_named[Inferno_r]` fails when the entire test suite is run with `pytest tests` in the `codegen2` branch (PR #4978) but succeeds when `pytest tests/test_plotly_utils/validators/test_colorscale_validator.py` is run on its own. (Testing done with Python 3.11.9.)

PR #5757, opened 2026-09-16
Accept tuples in validate_colorscale and sample_colorscale

### LINKED ISSUE #4808: Colorscales in templates are not valid list of lists, which breaks `plotly.colors.sample_colorscale`
Hello! I have something which is maybe a bug. It's certainly how I expected plotly.py to work, although maybe it's just a known limitation of how layout templates work.

Plotly comes with various colorscales built in, which work great:
```
import plotly
rainbow = plotly.colors.get_colorscale("rainbow")
# gives
[[0.0, 'rgb(150,0,90)'],
 [0.125, 'rgb(0,0,200)'],
 [0.25, 'rgb(0,25,255)'],
 [0.375, 'rgb(0,152,255)'],
 [0.5, 'rgb(44,255,150)'],
 [0.625, 'rgb(151,255,0)'],
 [0.75, 'rgb(255,234,0)'],
 [0.875, 'rgb(255,111,0)'],
 [1.0, 'rgb(255,0,0)']]
```

This is a list of lists and works as expected: `plotly.colors.validate_colorscale(rainbow)` passes.

Now I'd like to extract the colorscale from a template. This where things don't work as expected:
```
diverging  = plotly.io.templates["plotly"].layout.colorscale.diverging
# gives
((0, '#8e0152'),
 (0.1, '#c51b7d'),
 (0.2, '#de77ae'),
 (0.3, '#f1b6da'),
 (0.4, '#fde0ef'),
 (0.5, '#f7f7f7'),
 (0.6, '#e6f5d0'),
 (0.7, '#b8e186'),
 (0.8, '#7fbc41'),
 (0.9, '#4d9221'),
 (1, '#276419'))
```

This is a tuple of tuples and does not pass `plotly.colors.validate_colorscale(diverging)`, which raises:
```
_plotly_utils.exceptions.PlotlyError: A valid colorscale must be a list.
```

It's easy to convert this into a valid colorscale as `diverging = [list(x) for x in diverging]` but it seems strange that this is necessary.

Should the colorscale coming from a template be a list of lists already? Or should `validate_colorscale` accept tuple of tuples as valid? The real reason this matters is because `plotly.colors.sample_colorscale` runs `validate_colorscale` under the hood, which raises an error and means that `sample_colorscale` doesn't work.

---

### Side note

AFAICT there's no way register a new colorscale in plotly other than through a template. `plotly.colors.named_colorscales` comes from `_plotly_utils.basevalidators.ColorscaleValidator.named_colorscales`, but that's private and has no method for adding a new scale.

`ColorscaleValidator("", "").validate_coerce(diverging)` works to do the tuple of tuples to list of lists conversion, but again it's not public, and it seems weird that it would be necessary to run any coercion on a colorscale coming from a template anyway.

Belagum:
### Link to issue

Closes #4808

### Description of change

`validate_colorscale()` only accepted a list of lists, but colorscales read back from a figure or a template (`pio.templates["plotly"].layout.colorscale.diverging`, `fig.layout.colorscale.sequential`, `trace.marker.colorscale`, ...) come back as tuples of tuples. `sample_colorscale()` swallows the validation error and falls back to `make_colorscale()`, which treats every `(scale, color)` pair as a color, so what users actually see is the confusing `TypeError: '>' not supported between instances of 'str' and 'float'` from the issue. Validation now accepts tuples at both levels; `colorscale_to_colors()` and `colorscale_to_scale()` already worked with them.

### Demo

```python
import plotly.io as pio
import plotly.colors as pc

colorscale = pio.templates["plotly"].layout.colorscale.diverging
pc.sample_colorscale(colorscale, 3)
```

before:

```
TypeError: '>' not supported between instances of 'str' and 'float'
```

after:

```
['rgb(142, 1, 82)', 'rgb(247, 247, 247)', 'rgb(39, 100, 25)']
```

### Testing strategy

`test_sample_colorscale` now also samples the template's diverging colorscale and checks that the result matches the list-of-lists version. Added `test_validate_colorscale` covering lists, tuples, the template colorscale and the two structural errors (there was a `TODO` about testing those). Both fail without the fix. `tests/test_core/test_colors` and `tests/test_plotly_utils` pass, except `test_acceptance_named[Inferno_r]`, which fails for me on `main` as well (looks like #5028).

### Guidelines


camdecoster:
Thanks for the update. We'll take a look and follow up. This seems like a good fix though.

PR #5759, opened 2026-09-16
Remove the @types/plotly.js dev dependency

### LINKED ISSUE #5740: Remove `@types/plotly.js` dependency
[`@types/plotly.js`](https://www.npmjs.com/package/@types/plotly.js) (the DefinitelyTyped types for plotly.js) is listed as a dev dependency in `js/package.json`.

As of plotly.js 4.0.0, types are included in plotly.js itself (see https://github.com/plotly/plotly.js/pull/7680), so we can probably remove the `@types/plotly.js` dependency. I haven't looked into how the types are actually used in the code.

Belagum:
### Link to issue

Closes #5740

### Description of change

Drops the `@types/plotly.js` dev dependency from `js/`. plotly.js ships its own declarations since 4.0 (`"types": "./lib/index.d.ts"`), and TypeScript prefers a package's bundled types over DefinitelyTyped, so the DT package was already unused: `tsc` resolved `plotly.js` to the bundled types all along, and `Plotly.PlotlyHTMLElement` only kept working through the `export as namespace Plotly` UMD declaration.

While at it, the type usage is made explicit against the bundled declarations:

- `widget.ts` imports `PlotlyHTMLElement` and uses a small `PlotlyGraphDiv` type for what plotly.js doesn't declare: `_fullData`/`_fullLayout` on the graph div and the `plotly_update` event. The `plotly_deselect`/`plotly_doubleclick` handlers take no argument now, matching the declared callback signature (they never used it).
- `mimeExtension.ts` had `import type PlotlyType from "plotly.js"` and then `PlotlyType.Data` etc. The default export is a value, not a namespace, so those never resolved. It is `import type * as PlotlyType` now, plus the same kind of local type for `plotly_webglcontextlost`.
- `tsconfig.json`: `moduleResolution: "node"` (node10) was removed in TypeScript 7, which `package.json` already pins (`^7.0.2`), so `npm run typecheck` failed before checking anything. Switched to `bundler`, which matches how esbuild consumes the sources.

### Demo

`npm run typecheck` on `main`:

```
tsconfig.json(7,25): error TS5108: Option 'moduleResolution=node10' has been removed. Please remove it from your configuration.
```

Forcing `--moduleResolution bundler` on `main` gives 27 errors, 9 of them about plotly.js types: `Cannot find namespace 'PlotlyType'` (4×), `No overload matches this call` for `plotly_update`, `plotly_deselect` and `plotly_doubleclick`, and `Property '_fullData' / '_fullLayout' does not exist on type 'PlotlyHTMLElement'`. After this change it reports 18, all pre-existing and unrelated to plotly.js (implicit `any`, strict null checks, no declarations for `lodash-es` and `@jupyterlab/rendermime-interfaces`). I left those alone.

### Testing strategy

No Python changes. Rebuilt with `npm ci && npm run build` after deleting `plotly/labextension/`: the labextension static bundle and `js/lib/mimeExtension.js` come out byte-identical (only types changed there), `plotly/labextension/package.json` picks up the dependency removal, and `plotly/package_data/widgetbundle.js` changes because of the two handler signatures (the rest of its diff is the minifier renaming identifiers). Type-checked as described above.

No changelog entry, since this only touches the JS dev toolchain. Happy to add one if you prefer; otherwise the changelog check needs the `no-changelog` label.

### Guidelines


camdecoster:
Thanks for the PR! I'll take a look and follow up.

PR #5761, opened 2026-09-16
Fix NaN and inf in numpy arrays breaking bar stacking and scattergl lines

### LINKED ISSUE #5228: Scattergl not drawing line with np.inf but only when in pd.Dataframe
Hello,

I’m having a strange issue where a graph_objects Figure with Scatergl is not drawing anything if there is an np.inf in the series, but only if the data is coming from a Pandas Dataframe. The below code shows an example of the issue (See 3rd figure).  This issue appears with 6.0.0 install in python.

```
import plotly.express as px
import numpy as np
import pandas as pd

# Create a DataFrame with np.inf in a column
df = pd.DataFrame({
    'x': [0, 1, 2, 3, 4],
    'y': [0, 1, 4, 9, np.inf],
})

# Create a line plot  (Things are OK)
fig = px.line(data_frame=df, x='x', y='y', title='Looks OK')
fig.show()


# Create a line plot w/ Scatter  (Things are OK)
fig = go.Figure()
fig.add_trace(go.Scatter(
    x=df['x'],
    y=df['y'],
    mode='lines',
))
fig.update_layout(title=dict(text="Looks OK"),)
fig.show()


# Create a line plot w/ Scattergl  (Nothing shown on figure...)
fig = go.Figure()
fig.add_trace(go.Scattergl(
    x=df['x'],
    y=df['y'],
    mode='lines',
))
fig.update_layout(title=dict(text="Nothing displayed..."),)
fig.show()


# Create a line plot w/ Scattergl  (This is ok.... to_list fixes something?)
fig = go.Figure()
fig.add_trace(go.Scattergl(
    x=df['x'].to_list(),
    y=df['y'].to_list(),
    mode='lines',
))
fig.update_layout(title=dict(text="Looks OK"),)
fig.show()
```

### LINKED ISSUE #5324: null values reset stacking in express barchart
It seems like null values in the middle of a series reset the stacking so that the next value starts at zero.  I tried it upstream in a JS codepen and it is handled okay there, so I believe it is an issue with the python API or possibly with `plotly.express`.

Reprex modified from https://plotly.com/python/bar-charts/#bar-charts-with-text

```
import plotly.express as px

wide_df = px.data.medals_wide()

# set null value in middle 'silver' series
wide_df.iloc[0,2] = None

fig = px.bar(
  wide_df, 
  x="nation", 
  y=["gold", "silver", "bronze"], 
  title="Nulls Reset Stacking", 
  text_auto=True)

fig.show()
```

Output:

Image

Based on input data:

Image

Note that South Korea should be stacked to a total of 35 and not 24.

### LINKED ISSUE #5428: Stacked barplot missing certain groups that have non-missing values in the data
I have a very strange bug that I can't seem to diagnose. 

Given the following data import:

```
plot_data = pd.read_csv("example_plot_data.csv")
plot_data['savedate_week'] = pd.to_datetime(plot_data['savedate_week'])
```

And the following basic plotly code:

```
import plotly.express as px

fig = px.bar(
        plot_data,
        x="savedate_week",
        y="commander_perc_of_card",
        color="commanders"
    )

fig.show()
```

Every week should have bars that total 100% and they do indeed in the data. They also do when you click on one of the commanders that has missing data. For example, when you click on Tiamat, the focused bar plot shows that it does have values for `savedate_week = '2025-02-03'`. And when you click on Ureni of the Unwritten, it has a value for `savedate_week = '2025-06-16'`. But both these values are missing from the overall stacked barplot. 

I am using plotly version 6.2.0.

[example_plot_data.csv](https://github.com/user-attachments/files/23660794/example_plot_data.csv)

Belagum:
### Link to issue

Closes #5324
Closes #5428
Closes #5228

### Description of change

Since 6.0, numeric numpy arrays go out as base64 typed arrays (`{"dtype": "f8", "bdata": ...}`), and `NaN`/`inf` in them are encoded as they are. plotly.js takes typed arrays on linear/log axes as already-clean numbers: `makeCalcdata` (`src/plots/cartesian/set_convert.js`) returns them as-is, whereas every element of a plain list goes through `d2c`, which turns `null` and `NaN` into `BADNUM`. In `bar` (also `waterfall`/`funnel`) the NaN then reaches `Sieve.put` in `cross_trace_calc.js`: the running total of that position becomes NaN, and the next bar at the same position reads `NaN || 0` and starts from zero again, drawn over the bars below it. That is what #5324 shows (South Korea tops out at 24 instead of 35), and #5428 is the same thing, the CSV there has 75 NaN values in the `y` column. `scattergl` draws no line at all when a coordinate is `inf` (#5228); with `.to_list()` the same data works, because a list sends `null`.

This keeps the base64 encoding for finite data and for 2-D arrays (heatmap/contour `z` is cleaned by the trace itself in `clean_2d_array.js`) and falls back to the list path for 1-D float arrays containing non-finite values, so those serialize to `null` as they did before 6.0. The FigureWidget serializer gets the same treatment in a second commit, since it sends 1-D float arrays as binary buffers that also become typed arrays in the frontend (and bare non-finite floats become `null` there too, which keeps jupyter_client off its non-compliant JSON fallback).

### Demo

Medals example from #5324 (`wide_df.iloc[0, 2] = None`, then `px.bar(wide_df, x="nation", y=["gold", "silver", "bronze"])`):

```
# before
"silver": {"dtype": "f8", "bdata": "AAAAAAAA+H8AAAAAAAAuQAAAAAAAAChA"}

# after
"silver": [null, 15.0, 12.0]
"gold", "bronze": unchanged, still base64
```

Rendered with kaleido: before, the South Korea bar tops out at 24, with bronze (11) drawn from zero over gold (24); after, gold and bronze stack to 35 and silver is skipped. For the data from #5428 every week reaches 100 again instead of leaving gaps where a NaN row came before another commander. For #5228 the `scattergl` line is drawn through the four finite points, before there were only markers.

Scattergl example from #5228, `y=df["y"]` with `np.inf` in the last position:

```
# before
"y": {"dtype": "f8", "bdata": "AAAAAAAAAAAAAAAAAADwPwAAAAAAABBAAAAAAAAAIkAAAAAAAADwfw=="}

# after
"y": [0.0, 1.0, 4.0, 9.0, null]
```

### Testing strategy

Added `tests/test_optional/test_graph_objs/test_b64_non_finite.py`: `go.Bar` with NaN from a pandas Series, `go.Scattergl` with inf from a DataFrame and the `px.bar` example from #5324 come out as lists with `null` on both JSON engines while the finite arrays next to them stay base64; a heatmap `z` with NaN stays base64; `to_typed_array_spec` with NaN/±inf in float32/float64; `_py_to_js` sends a float array with NaN/inf as `[1.0, None, None]` and still uses a buffer for finite arrays. 13 of the 16 cases fail on `main`. `tests/test_core`, `tests/test_io`, `tests/test_plotly_utils` and `tests/test_optional` give the same set of failures with and without the change (Windows `cp1252` in `test_offline`, nbformat, statsmodels and scikit-image not installed here, and a pyarrow timestamp parsing difference).

Arrays with NaN/inf lose the compact encoding, which is what every version before 6.0 did for all arrays; finite arrays are not affected. Only 1-D arrays fall back, because the 2-D consumers in plotly.js already clean non-numeric entries and heatmap/imshow inputs with NaN masks are the largest arrays around. The underlying issue is in the typed-array fast path of plotly.js; once that cleans NaN/inf, this fallback can go.

### Guidelines


camdecoster:
Thanks for the PR. Our team will take a look and follow up.

PR #5766, opened 2026-09-18
Make px.sunburst/treemap/icicle sector order deterministic for Polars input

### LINKED ISSUE #5765: [BUG]: px.sunburst / px.treemap / px.icicle with path give a different sector order on every run for Polars DataFrames
### Description

When `path=` is used with a Polars DataFrame, `px.sunburst`, `px.treemap` and `px.icicle` build their `ids` / `labels` / `parents` / `values` arrays in a different order every time the script is run. The same data as a pandas DataFrame or a PyArrow table always gives the same order: the order in which the sectors first appear in the data.

The cause is in `process_dataframe_hierarchy` (`plotly/express/_core.py`). Each level of the hierarchy is built with `df.group_by(path[i:]).agg(...)`. With pandas (narwhals uses `sort=False`) and PyArrow the groups come back in order of first appearance, but Polars' `group_by` does not guarantee any order, so the order of the output changes from run to run.

Consequences:

- `fig.to_json()` / `fig.write_html()` output is not reproducible with Polars input (snapshot tests, caching, diffs of generated HTML).
- With `sort=False`, or when sectors have equal values, the chart itself is laid out differently on each run.
- Polars results differ from pandas / PyArrow results for identical data.

### Screenshots/Video

N/A: the difference is in the figure data; see the output below.

### Steps to reproduce

```python
import plotly
import plotly.express as px
import polars as pl

df = pl.DataFrame(
    {
        "region": ["South", "North", "South", "West", "North", "West"],
        "sector": ["Tech", "Finance", "Finance", "Tech", "Tech", "Finance"],
        "sales": [1, 2, 3, 4, 5, 6],
    }
)
fig = px.sunburst(df, path=["region", "sector"], values="sales")
print(plotly.__version__, pl.__version__, list(fig.data[0].ids))
```

Running the script three times (plotly 7.1.0, polars 1.44.2):

```
7.1.0 1.44.2 ['West/Tech', 'West/Finance', 'North/Finance', 'South/Finance', 'North/Tech', 'South/Tech', 'South', 'North', 'West']
7.1.0 1.44.2 ['South/Tech', 'West/Tech', 'North/Tech', 'South/Finance', 'North/Finance', 'West/Finance', 'West', 'North', 'South']
7.1.0 1.44.2 ['South/Tech', 'West/Tech', 'North/Tech', 'North/Finance', 'South/Finance', 'West/Finance', 'West', 'South', 'North']
```

With `pd.DataFrame(...)` instead, every run prints:

```
['South/Tech', 'North/Finance', 'South/Finance', 'West/Tech', 'North/Tech', 'West/Finance', 'South', 'North', 'West']
```

### Notes

I have a small fix with a regression test and will open a PR for it.

Irahan2:
### Link to issue

Closes #5765

### Description of change

`process_dataframe_hierarchy` now adds a temporary row-index column, takes its minimum in each group, and sorts every level of the hierarchy by it after the `group_by`. Sectors are therefore always listed in order of first appearance for every dataframe backend. pandas and PyArrow already produced that order; Polars, whose `group_by` does not keep row order, produced a different order on every run.

### Demo

Before, three runs of the script from the issue with a Polars DataFrame:

```
['West/Tech', 'West/Finance', 'North/Finance', 'South/Finance', 'North/Tech', 'South/Tech', 'South', 'North', 'West']
['South/Tech', 'West/Tech', 'North/Tech', 'South/Finance', 'North/Finance', 'West/Finance', 'West', 'North', 'South']
['South/Tech', 'West/Tech', 'North/Tech', 'North/Finance', 'South/Finance', 'West/Finance', 'West', 'South', 'North']
```

After, every run (and pandas / PyArrow):

```
['South/Tech', 'North/Finance', 'South/Finance', 'West/Tech', 'North/Tech', 'West/Finance', 'South', 'North', 'West']
```

### Testing strategy

Added `test_sunburst_treemap_with_path_order`, parametrized over all the dataframe constructors and over `px.sunburst`, `px.treemap` and `px.icicle`. It checks the exact order of `ids` and `values`. It fails for the Polars constructor on `main` and passes with this change. The rest of `tests/test_optional/test_px` passes locally, apart from 3 tests that fail the same way on unmodified `main` in my Windows environment (`test_timezones` and `test_date_in_hover` for the PyArrow constructor, and one `test_imshow_source_dtype_zmax` case). The hierarchy tests also pass with the minimum supported narwhals (1.15.1).

- pandas output does not change: it already used first-appearance order (narwhals' `group_by` uses `sort=False`). This matches how Plotly Express orders categories elsewhere (`unique(maintain_order=True)`).
- Only the hierarchy code path is affected. The other `group_by` in `_core.py` is only used for key lookups, and trace order there already comes from `unique(maintain_order=True)`.
- AI disclosure: I used Claude Code (Claude Opus 5) to help find this bug (by comparing Plotly Express output for pandas, Polars and PyArrow inputs) and to draft the fix and the test. I have reviewed the changes and run the tests myself.

### Guidelines


camdecoster:
I'm going to convert this to draft for now until we decide on an approach for a solution.

camdecoster (dismissed):
Thanks for the fix! Deterministic plots are the best plots..

tests/test_optional/test_px/test_px_functions.py
```diff
@@ -233,6 +233,32 @@ def test_sunburst_treemap_with_path(constructor):
     assert fig.data[0].values[-1] == 8
 
 
+@pytest.mark.parametrize("px_fn", [px.sunburst, px.treemap, px.icicle])
+def test_sunburst_treemap_with_path_order(constructor, px_fn):
+    # Sectors should follow the order of first appearance in the data, whatever
```
camdecoster, line 243: There's a previously known issue biting us here. We skip it up above in `test_sunburst_treemap_with_path` so we should do the same thing here:
```suggestion
    if _pandas_version_at_least("3.0.0") and constructor == pandas_pyarrow_constructor:
        pytest.skip(
        "known issue with pandas 3 + pandas_pyarrow_constructor + px hierarchy charts (https://github.com/plotly/plotly.py/issues/5571)"
    )
    # Sectors should follow the order of first appearance in the data, whatever
```

Irahan2: Good catch, thanks. Added the same skip. That was exactly what CI was failing on. 
and Deterministic plots are the best plotsssss... @camdecoster

camdecoster (approved):

